PR #18, opened 2026-02-24
fix: auto-exclude react-server packages from RSC optimizeDeps

FredKSchott:
## Summary

Packages with a `react-server` export condition (e.g. `swr`) get pre-bundled with a stripped-down entry in the RSC environment, breaking client imports (like `useSWR`) before the RSC plugin's `"use client"` transform can strip them.

At startup, scan the project's direct dependencies for packages declaring a `react-server` export condition and automatically add them to `environments.rsc.optimizeDeps.exclude`.

Uses `createRequire` for resolution so it works across all package managers.

Ref: https://github.com/vitejs/vite-plugin-react/issues/736, https://github.com/vitejs/vite-plugin-react/issues/775

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -351,6 +351,108 @@ function getPackageName(id: string): string | null {
   return rest.split("/")[0] || null;
 }
 
+/**
+ * Find installed packages that declare a `react-server` export condition
+ * and should be excluded from RSC `optimizeDeps`.
+ *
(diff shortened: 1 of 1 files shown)


hi-ogawa:
Just released `@vitejs/plugin-rsc` v0.5.21 with https://github.com/vitejs/vite-plugin-react/pull/1119. I'd appreciate feedback. Thanks! :pray:

PR #27, opened 2026-02-24
fix running on windows

niicojs:
fix ERR_UNSUPPORTED_ESM_URL_SCHEME when running on windows


andlewis:
I think you might have missed the one on line 2910 as well

schemburkar:
> I think you might have missed the one on line 2910 as well

works without it on some paths.. but agree, needed in general

PR #30, opened 2026-02-24
Fix binary response handling in Pages Router prod server

linares222:
Fixes a Pages Router prod-server response handling bug where `response.text()` could corrupt binary payloads.

### Changes
- Switch Pages Router prod response reads from `response.text()` to `Buffer.from(await response.arrayBuffer())`
- Add `/api/binary` fixture route returning known bytes
- Add regression test asserting exact byte preservation for `/api/binary`

### Why
API routes can return raw bytes (`res.end(Buffer)`). Decoding as text before writing can mangle non-UTF8 bytes.


southpolesteve:
Thank you!

PR #57, opened 2026-02-25
fix: export all google font

### LINKED ISSUE #72: `next/font/google` shim missing named exports for fonts like Figtree, DM_Sans, etc.
## Environment

- **vinext**: 0.0.8 (also reproduced on 0.0.5)
- **Next.js**: 16.1.6
- **Platform**: Cloudflare Workers + `vinext dev`

## Reproduce

1. In any component (server or client): `import { Figtree } from 'next/font/google'`
2. `vinext build` → fails with something like `"Figtree" is not exported by ".../shims/font-google.ts"`

`Inter` in a Server Component layout works fine because the shim does export it (or handles it as default).

## Root cause

The font shim ([source](https://github.com/cloudflare/vinext/blob/4d0a00864446d482e573710e06936ad0cc3e1eb2/packages/vinext/src/shims/font-google.ts)) does not export `Figtree` (or other Google Fonts like `DM_Sans`, `Poppins`, etc.) as named exports. Only a subset of fonts appear to be shimmed.

This affects any app using non-default Google Fonts via `next/font/google` named imports.

## Workaround

Replace the `next/font/google` import with a CSS `@import` from Google Fonts CDN:

```css
/* In your CSS module */
@import url('https://fonts.googleapis.com/css2?family=Figtree:wght@700&display=swap');
```

Then use `font-family: 'Figtree', system-ui, sans-serif` directly in CSS instead of the JS font object's `className`.

## Suggested fix

Either generate named exports for all Google Fonts (matching Next.js behavior), or document the CSS `@import` workaround as the recommended approach for fonts beyond the currently shimmed set.

Verified on 0.0.8 — the font shim source still lacks these named exports.

### LINKED ISSUE #190: Some google fonts may have issues [vite] Internal server error: (0 , __vite_ssr_import_2__.Space_Grotesk) is not a function
Was having an issue with this specific font somehow it wasn't either loaded from the cdn or perhaps it wasn't registered in the same way, I can't really tell it was declared the same way as the other fonts

```
const spaceGrotesk = Space_Grotesk({  // <-- it was having an issue with this specific font
  subsets: ["latin"],
  variable: "--font-heading",
}); 

const inter = Inter({
  subsets: ["latin"],
  variable: "--font-sans",
});

const jetbrainsMono = JetBrains_Mono({
  subsets: ["latin"],
  variable: "--font-mono",
});
```

after removing it the website loaded, I would like to help because i am sure if one font has an issue it means other fonts probably will at some point, but don't know what to look for exactly

hyoban:
Otherwise, users will encounter import errors.

Fixes #72
Fixes #190


andrebrito16:
Fixes #72

PR #79, opened 2026-02-25
fix(navigation): cache getServerSnapshot fallback in useSearchParams

gembleman:
useSyncExternalStore compares snapshots with Object.is. The previous getServerSnapshot returned `new URLSearchParams()` on every call when no server context was available, producing a new object reference each time and triggering an infinite re-render loop in React.

Fix: introduce a module-level `_cachedServerSearchParams` that is allocated once and reused, matching the same referential-stability pattern already used by `getSearchParamsSnapshot` for the client snapshot.


FredKSchott:
LGTM! Thanks @gembleman

PR #84, opened 2026-02-26
fix: make vinext bin resilient in workspace installs

ruskzane:
### What
When installing the monorepo with pnpm, fixture/example workspaces that depend on `vinext` can emit warnings because the package `bin` points at `dist/cli.js` before the build runs.

This change adds a small top-level `cli.js` wrapper (checked into the repo and included in npm files) and updates `bin` to point to it. The wrapper forwards to `dist/cli.js` when present, and otherwise prints a clear instruction to run the build.

### Why
- Avoids ENOENT warnings during workspace installs
- Keeps published package behavior the same (prepack still builds dist)

### Notes
- `prepare` still runs `tsc` so dist exists after install.

### DIFF
--- a/packages/vinext/cli.js
+++ b/packages/vinext/cli.js
@@ -1,0 +1,20 @@
+#!/usr/bin/env node
+import fs from "node:fs";
+import path from "node:path";
+import url from "node:url";
+
+const __dirname = path.dirname(url.fileURLToPath(import.meta.url));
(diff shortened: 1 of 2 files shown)


FredKSchott:
This is a nice quality of life improvement for maintainers but we're laser focused right now on stability. Closing this for now but would love to revisit once things settle in a few weeks/months.

PR #90, opened 2026-02-26
fix: detect src/app App Router

toddLiao469469:
## Summary
   - Align `hasAppDir` with Next.js behavior by checking both `app/` and `src/app/`.
   - Fix `vinext init` misclassification for projects that use `src/app`.
   - Add a regression test in `tests/init.test.ts` to ensure `src/app` projects are treated as App Router projects and install App Router dependencies.

   ## Why
   Next.js supports both `app/` and `src/app/`; matching this behavior improves compatibility and avoids incorrect setup during `vinext init`.

   ## Testing
   - `pnpm exec vitest run tests/init.test.ts`


FredKSchott:
LGTM! Did a separate review to confirm that everywhere else is updated to correct check. Thanks @toddLiao469469

PR #128, opened 2026-02-26
fix(shims): copy request headers into mutable Headers in headersContextFromRequest

harrisrobin:
## Summary

In Cloudflare Workers, `Request.headers` is immutable per the Fetch API spec. `headersContextFromRequest` was storing the original `request.headers` reference directly in `HeadersContext`, so when `applyMiddlewareRequestHeaders` later called `ctx.headers.set()` (to propagate middleware-modified headers into server component context), Workers threw:

```
TypeError: Can't modify immutable headers.
```

This is triggered on every request when middleware returns `NextResponse.next({ request: { headers } })` — a common pattern for passing data like `x-request-url` from middleware to server components.

## Fix

Wrap with `new Headers(request.headers)`, which creates an independent mutable copy while preserving all header values.

```diff
- headers: request.headers,
+ headers: new Headers(request.headers),
```

## Test plan

- [x] New test asserting the context holds a mutable copy, not the original `Request.headers` reference
- [x] All 2026 existing tests pass
- [x] Lint: 0 warnings/errors (`oxlint`)
- [x] Typecheck: clean (`tsgo --noEmit`)

/bonk

### DIFF
--- a/packages/vinext/src/shims/headers.ts
+++ b/packages/vinext/src/shims/headers.ts
@@ -190,7 +190,10 @@ export function headersContextFromRequest(request: Request): HeadersContext {
     }
   }
   return {
-    headers: request.headers,
+    // Copy into a mutable Headers instance. In Cloudflare Workers the original
+    // Request.headers is immutable; applyMiddlewareRequestHeaders() needs to
+    // call .set() on this object after middleware runs.
+    headers: new Headers(request.headers),
     cookies,
   };
 }
--- a/tests/shims.test.ts
+++ b/tests/shims.test.ts
@@ -266,6 +266,29 @@ describe("next/headers shim", () => {
     expect(ctx.headers.get("cookie")).toBe("a=1; b=2");
   });
 
+  it("headersContextFromRequest returns mutable headers (not the immutable Request.headers)", async () => {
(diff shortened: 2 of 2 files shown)


solracnyc:
Can confirm this bug in production. We hit the exact same `TypeError: Can't modify immutable headers` on Cloudflare Workers with vinext 0.0.15 + React 19.2.4.

**Environment:**
- vinext 0.0.15 (also reproduced on 0.0.16)
- React 19.2.4 / react-dom 19.2.4
- Clerk middleware (`clerkMiddleware` from `@clerk/nextjs/server`) — triggers `NextResponse.next({ request: { headers } })` on every request
- Cloudflare Workers (not Pages)

**Impact:** Every route returned HTTP 500. The worker tail showed `TypeError: Can't modify immutable headers` on all requests.

**Our workaround:** We cloned the request with mutable headers in our worker entry point before passing to vinext's app-router handler:

```ts
// worker/index.ts
const requestWithMutableHeaders = new Request(request, {
  headers: new Headers(request.headers),
  cf: (request as Request & { cf?: unknown }).cf,
} as RequestInit);

return handler.fetch(requestWithMutableHeaders);
```

This is essentially the same fix as this PR but applied at the app level instead of inside vinext. Would be great to get this merged upstream so we can remove our workaround. Thanks @harrisrobin for filing this!

PR #133, opened 2026-02-26
fix: log errors from failed error reporting calls

heyitsmohdd:
What This PR Does

Found a bug where errors were being silently swallowed. If something went wrong while trying to report an error, that error just disappeared into the void. Not great for debugging.

Fixed 3 spots in the app dev server:

1. Server actions - when they fail and we try to report the error, we now log it if that reporting itself fails

2. Route handlers - same thing

3. Async components with loading boundaries - these are handled correctly but now we also log errors in development so you can see what's happening

No breaking changes. Just adds some error logging so we can actually debug when things go wrong.

Tests: All pass - 1400+ tests, typecheck, lint all good.


heyitsmohdd:
Hey Fred, that explanation about why the catch block was noisy makes sense now. Glad the other changes helped

Issue #144, opened 2026-02-26
Error : Did you mean to import "next/font/local/index.js"?


technewwings:
trying to port my current payload cms + nextjs 16 project to vinext.
```
2:01:52 AM [vite] Internal server error: Directory import '/Users/play-area/electrofit/node_modules/next/font/local' is not supported resolving ES modules imported from /Users/play-area/electrofit/node_modules/geist/dist/mono.js
Did you mean to import "next/font/local/index.js"?
      at finalizeResolution (node:internal/modules/esm/resolve:262:11)
      at moduleResolve (node:internal/modules/esm/resolve:859:10)
      at defaultResolve (node:internal/modules/esm/resolve:983:11)
      at nextResolve (node:internal/modules/esm/hooks:748:28)
      at o (file:///Users/play-area/electrofit/node_modules/@tailwindcss/node/dist/esm-cache.loader.mjs:1:69)
      at nextResolve (node:internal/modules/esm/hooks:748:28)
      at Hooks.resolve (node:internal/modules/esm/hooks:240:30)
      at MessagePort.handleMessage (node:internal/modules/esm/worker:199:24)
      at [nodejs.internal.kHybridDispatch] (node:internal/event_target:827:20)
      at MessagePort.<anonymous> (node:internal/per_context/messageport:23:28)
```

PR #157, opened 2026-02-27
fix: resolve directory imports for next/font shims

hoangnv170752:
1/ Summary
- Fix directory imports for next/font/local and next/font/google when imported from third-party packages like geist
- Add resolveId hook handling for imports without .js extension that Vite's ESM resolution rejects
2/ Problem
- Libraries like geist import next/font/local as a directory:
`import localFont from "next/font/local"
`
- This fails in Vite's ESM mode with:
```
Directory import '/node_modules/next/font/local' is not supported resolving ES modules
Did you mean to import "next/font/local/index.js"?
```

The existing resolve.alias only works for user code, not for dependencies in node_modules.
3/ Solution
- Extend the resolveId hook to catch directory-style imports (without .js extension) and resolve them through the shim map, similar to the existing handling for .js extension imports.

### DIFF
--- a/packages/vinext/src/entries/app-browser-entry.ts
+++ b/packages/vinext/src/entries/app-browser-entry.ts
@@ -1,6 +1,7 @@
 import { resolveRuntimeEntryModule } from "./runtime-entry-module.js";
 import type { VinextLinkPrefetchRoute } from "../client/vinext-next-data.js";
 import type { AppRoute } from "../routing/app-router.js";
+import type { RouteManifest } from "../routing/app-route-graph.js";
 
 /**
  * Generate the virtual browser entry module.
@@ -9,7 +10,10 @@ import type { AppRoute } from "../routing/app-router.js";
  * embedded RSC payload and handles client-side navigation by re-fetching
  * RSC streams.
  */
-export function generateBrowserEntry(routes: readonly AppRoute[] = []): string {
+export function generateBrowserEntry(
+  routes: readonly AppRoute[] = [],
+  routeManifest: RouteManifest | null = null,
+): string {
   const entryPath = resolveRuntimeEntryModule("app-browser-entry");
   const prefetchRoutes: VinextLinkPrefetchRoute[] = routes
     .filter((route) => isLinkPrefetchRoute(route))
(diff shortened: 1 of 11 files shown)


southpolesteve (changes requested):
## Review: Request Changes

I've read the full diff, the existing `resolveId` logic, the `resolve.alias` setup, and the linked issue #144. I agree with the ask-bonk review's concerns and want to add a more fundamental one.

### Blocking: This fix likely doesn't address the actual bug

The error in issue #144 comes from **Node's native ESM resolver**, not from Vite's module graph:

```
at finalizeResolution (node:internal/modules/esm/resolve:262:11)
at moduleResolve (node:internal/modules/esm/resolve:859:10)
at defaultResolve (node:internal/modules/esm/resolve:983:11)
at nextResolve (node:internal/modules/esm/hooks:748:28)
at o (file:///.../node_modules/@tailwindcss/node/dist/esm-cache.loader.mjs:1:69)
```

The `@tailwindcss/node` ESM loader hook is resolving `next/font/local` through Node's module system, completely outside Vite's plugin pipeline. Neither `resolve.alias` nor `resolveId` hooks fire for imports resolved by Node's native ESM loader.

This means:
1. `resolve.alias` (line 2010) already handles `"next/font/local"` for imports that go through Vite's module graph. The shim map has an exact entry for it (line 1828).
2. The new `resolveId` block is redundant with `resolve.alias` for the same set of imports.
3. For the actual bug (Node's ESM resolver in the Tailwind loader), neither approach works.

**Before this can be merged, the author needs to demonstrate that this fix actually resolves the error from issue #144.** A reproduction repo or test showing the fix working would be sufficient.

The real fix for this issue likely involves one of:
- Adding `"next"` to `ssr.noExternal` so Vite processes `next/*` imports through its pipeline instead of deferring to Node
- Configuring `ssr.resolve.conditions` or `ssr.resolve.alias` to handle the resolution
- Investigating why the Tailwind CSS ESM loader is involved and whether that's the actual root cause

### Non-blocking: Code duplication (agree with ask-bonk)

If this block does end up being needed, it should be merged with the existing `.js` handler (lines 2190-2197) as ask-bonk suggested:

```ts
if (cleanId.startsWith("next/")) {
  const baseId = cleanId.endsWith(".js") ? cleanId.slice(0, -3) : cleanId;
  if (nextShimMap[baseId]) {
    const shimPath = nextShimMap[baseId];
    return shimPath.endsWith(".js") ? shimPath : shimPath + ".js";
  }
}
```

### Non-blocking: Misleading comment (agree with ask-bonk)

The comment says "Handle directory imports without extension" but the condition `cleanId.startsWith("next/") && !cleanId.endsWith(".js")` matches every `next/*` import that doesn't end in `.js`, including `"next/link"`, `"next/navigation"`, etc. These aren't directory imports. The comment should accurately describe what the code does.

### Non-blocking: No tests

There are no tests for this change. At minimum, a unit test should verify the `resolveId` handler returns the correct shim path for `"next/font/local"` and `"next/font/google"` bare imports.

### Summary

The core issue is that I'm not convinced this change fixes the reported bug. The error originates in Node's ESM resolver (via Tailwind's loader hook), which is outside Vite's plugin system. Please provide evidence that this fix resolves the issue, or investigate the actual resolution path that's failing.

james-elicx:
Hi @hoangnv170752, would you be able to try with the latest version of Vinext and see if this is still an issue please? If it is, would you be able to include a reproduction?

PR #163, opened 2026-02-27
fix: add scope constraint to Bonk agent prompts

### LINKED ISSUE #161: bonk: improve review prompt
When invoked in a comment, Bonk can occasionally reply to the wrong issue or PR number. This is a problem and needs to be fixed.

Example: https://github.com/cloudflare/vinext/pull/45

We should review the viguy prompt and ensure that the agent is better steered towards ONLY commenting on the issue or PR it was invoked on.

When updating the prompt, ensure you follow anthropic best practice practices for prompt engineering and LLM guardrails.

## research

Look at previous workflow runs where Bonk ran, and use that to cross reference issues and PRs. Look at the last 3-4 days of Bonk responses. 

You should also reference a related issue at https://github.com/cloudflare/workerd/pull/6150 - however this applies to new pull request automation and not manual comment-driven automation.

elithrar:
Bonk can occasionally comment on or interact with the wrong issue/PR when invoked via `/bonk`. Fixes #161.

The viguy prompt had no explicit constraint scoping the agent to the triggering issue/PR — and step 3 ("Search for related issues and PRs") actively encouraged exploring other issues without guardrails against interacting with them. This mirrors the same problem fixed in cloudflare/workerd#6150 for auto-triggered PR reviews.

- add "Scope constraint" section to `viguy.md` as the highest-priority rule, placed early in the prompt for primacy. constrains all `gh` write commands to the invoking issue/PR, with explicit instructions to only link (not interact with) related issues found during research.
- update step 3 of "Before starting work" to mark related-issue search as read-only with bold warning
- add one-line scope constraint to `reviewer.md` for the reviewer subagent
- increase `fetch-depth` from 1 to 30 in both `bonk.yml` and `bigbonk.yml` — gives the agent meaningful git history for diffs, reducing the need to search elsewhere for context (same change as workerd#6150)

Verification is manual — trigger `/bonk` on a PR and confirm it only references the correct PR number.

### DIFF
--- a/.opencode/agents/reviewer.md
+++ b/.opencode/agents/reviewer.md
@@ -20,7 +20,7 @@ permission:
 
 You are a senior code reviewer for **vinext** — a Vite plugin that reimplements the Next.js API surface with Cloudflare Workers as the primary deployment target.
 
-**Scope constraint:** You are reviewing one specific PR. Confirm the PR number matches what triggered you before posting any review. Do not comment on or interact with other PRs or issues.
+**Scope constraint:** You are reviewing one specific PR. The `$PR_NUMBER` environment variable contains the PR you were invoked on — use it as the source of truth, not numbers mentioned in comments or descriptions. Before posting any review or comment, verify the target matches `$PR_NUMBER`. Do not interact with any other PR or issue.
 
 ## Review standards
 
--- a/.opencode/agents/viguy.md
+++ b/.opencode/agents/viguy.md
@@ -13,7 +13,8 @@ You are a senior engineer specializing in Vite internals, Next.js internals, Typ
 
 You have been invoked on a specific GitHub issue or PR. **All your actions must target only that issue or PR.**
 
-- Before running any `gh` command that writes (comment, review, close, merge, create), confirm you are targeting the correct issue/PR number from your invocation context.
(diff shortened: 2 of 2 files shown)


.github/workflows/bonk.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          fetch-depth: 1
+          fetch-depth: 30
```
elithrar:
```suggestion
          fetch-depth: 30 # 30 is a balance in providing enough context for the agent vs. pulling down the entire repo history. It is not a scientific number. 27 or 31 or 36 would likely work as well.
```

.github/workflows/bigbonk.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          fetch-depth: 1
+          fetch-depth: 30
```
elithrar:
```suggestion
          fetch-depth: 30 # 30 is a balance in providing enough context for the agent vs. pulling down the entire repo history. It is not a scientific number. 27 or 31 or 36 would likely work as well.
```

PR #176, opened 2026-02-27
fix: bound fetch cache key body serialization

dknecht:
### Motivation

- The fetch cache key generator previously serialized and fully buffered many `BodyInit` types (streams, blobs, form data) with no size limit, allowing an attacker to cause unbounded memory growth during cached `fetch()` calls.
- The goal is to prevent OOM scenarios while preserving caching behavior for normal-sized bodies and ensuring the original request body remains usable for the real network fetch.

### Description

- Add a configurable hard limit `MAX_CACHE_KEY_BODY_BYTES` (1 MiB) and a `BodyTooLargeForCacheKeyError` to guard cache-key body serialization and abort unsafe buffering.
- Rework stream handling to use `ReadableStream.prototype.tee()` so one branch is consumed for cache-key hashing while the other branch is forwarded to the real fetch without reconstructing a full in-memory buffer.
- For bodies that exceed the size limit (Blob, `Uint8Array`, streamed chunks, `FormData` files), short-circuit cache-key generation and fall back to passing the request through to `originalFetch` (i.e., bypass caching) to avoid buffering; `_ogBody` restoration remains supported for safe cases.
- Update tests to reflect preserved stream forwarding behavior and add checks that oversized `Blob` and `ReadableStream` bodies bypass the cache and still perform the network fetch.

### Testing

- Ran unit tests with `pnpm test tests/fetch-cache.test.ts`, which passed (`54 tests` all green). 
- Ran type checking with `pnpm run typecheck`, which completed successfully with no type errors.
- New/updated tests exercise stream preservation and oversized-body fallback behavior and passed as part of the test run.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69a1f69e2e14832396f748a282248f53)

### DIFF
--- a/packages/vinext/src/shims/fetch-cache.ts
+++ b/packages/vinext/src/shims/fetch-cache.ts
@@ -37,7 +37,7 @@ import { AsyncLocalStorage } from "node:async_hooks";
 const HEADER_BLOCKLIST = ["traceparent", "tracestate"];
 
 // Cache key version — bump when changing the key format to bust stale entries
-const CACHE_KEY_PREFIX = "v1";
+const CACHE_KEY_PREFIX = "v2";
 const MAX_CACHE_KEY_BODY_BYTES = 1024 * 1024; // 1 MiB
 
 class BodyTooLargeForCacheKeyError extends Error {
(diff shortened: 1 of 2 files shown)


threepointone (approved):
This feels good to merge.

PR #196, opened 2026-02-28
fix: extract webpack resolve.alias from next.config plugins (#177)

### LINKED ISSUE #177: Issues with `next-intl` library
### The issue

Vinext doesn't seem to work with `next-intl`.
I created a brand new NextJS project using `create-next-app`, added the default `next-intl` configuration and default `vinext` setup.
I get the following error when I run the app:

Image

Running the exact same setup with default nextjs works fine.

### Package versions:

| Package  | Version |
| ----------- | ----------- |
| vinext | 0.0.15 |
| next | 16.1.6 |
| react |  19.2.4 |
| next-intl | 4.8.3 |

### Possible cause

When I start the vinext dev server, I get the following warning:

`[vinext] next.config option "webpack" is not yet supported and will be ignored`

I assume this is because of the NextJS config injected by the next-intl: `export default withNextIntl(nextConfig);`

SeolJaeHyeok:
## Summary

  - next-intl's createNextIntlPlugin / withNextIntl registers "next-intl/config" → user's i18n/request.ts via webpack resolve.alias
  - Since vinext ignores webpack config, this alias was never applied — next-intl/config resolved to a stub that throws "Couldn't find next-intl config file"
  - Fix: probe config.webpack() with the real project root to extract string-valued aliases, then merge them into Vite's resolve.alias (vinext's own shims
  always take precedence)
  - Suppresses the spurious "webpack not yet supported" warning when alias extraction succeeds

 Fixes #177

### DIFF
--- a/.agents/skills/migrate-to-vinext/references/compatibility.md
+++ b/.agents/skills/migrate-to-vinext/references/compatibility.md
@@ -91,7 +91,7 @@ Tested and working:
 - next-themes
 - nuqs
 - next-view-transitions
-- next-intl
+- next-intl (partial: `createNextIntlPlugin` requires manual webpack alias workaround, see #202)
 - better-auth
 - @vercel/analytics
 - tailwindcss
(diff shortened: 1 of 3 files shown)


southpolesteve:
Thanks for the PR @SeolJaeHyeok! We're looking at this now.

PR #217, opened 2026-03-01
fix(rewrites): serve static files from public/ when rewrite target is a .html path

### LINKED ISSUE #199: Add support for rewrites
It seems like vinext are not working with rewrites properly:

https://nextjs.org/docs/app/api-reference/config/next-config-js/rewrites

Our config:

```
import type { NextConfig } from "next";

const nextConfig: NextConfig = {
  devIndicators: false,
  output: "standalone",
  reactCompiler: true,
  reactStrictMode: true,
  turbopack: {
    rules: {
      "*.css": {
        loaders: ["@tailwindcss/webpack"],
      },
    },
  },
  typedRoutes: true,
  typescript: {
    ignoreBuildErrors: true,
  },
  async rewrites() {
    // These pages are static HTML pages that handle auth flows.
    // We can hide the .html suffix when being accessed.
    return [
      {
        source: "/auth/no-access",
        destination: "/auth/no-access.html",
      },
      {
        source: "/auth/automatic-sign-in",
        destination: "/auth/automatic-sign-in.html",
      },
      {
        source: "/auth/signed-out",
        destination: "/auth/signed-out.html",
      },
    ];
  },
};

export default nextConfig;

```

yunus25jmi1:
Fixes #199.

## Problem

When `next.config` rewrites() map a clean URL to a static `.html` file in `public/` (a common pattern for serving pre-built static pages like `/auth/no-access` → `/auth/no-access.html`), vinext was returning 404 because no Next.js route matches `.html` paths.

## Root Cause

After `afterFiles` rewrites resolve a path to `*.html`, the routing logic in all three server paths would find no matching app/pages route for the `.html` URL and return 404, without ever checking the filesystem's `public/` directory.

## Fix

Three server paths updated:

1. **Pages Router dev** (`packages/vinext/src/index.ts`): After afterFiles/fallback rewrites, before the final `handler()` call, check if the resolved URL points to a file in `public/` and serve it directly.

2. **App Router dev** (`packages/vinext/src/server/app-dev-server.ts`): The generated RSC entry now checks `public/` for the rewritten pathname before rendering the 404 page. Added a new optional `root` parameter to `generateRscEntry()` so the public directory path is embedded in the generated virtual module.

3. **Production server** (`packages/vinext/src/server/prod-server.ts`): After afterFiles rewrites and after fallback rewrites produce a path with a file extension, `tryServeStatic()` is called against the built `clientDir` (which contains `public/` files) before passing to SSR.

Also added a `staticMimeType()` helper in `index.ts` to ensure correct `Content-Type` headers.

## Tests

- Integration tests in `app-router.test.ts` and `pages-router.test.ts`: `GET /static-html-page` (rewritten → `/static-html-page.html`) returns 200 with correct HTML content and `text/html` Content-Type.
- Unit test in `app-router.test.ts`: `generateRscEntry()` embeds the `public/` path in the generated code.
- Fixture `public/static-html-page.html` files added to both `app-basic` and `pages-basic` fixtures; `next.config` rewrites updated.

## What async rewrites() flat array does

Per Next.js semantics, when `rewrites()` returns a flat array, all rules go into `afterFiles`. vinext already handles this correctly via `resolveNextConfig()`. This PR fixes the serving of the rewritten destination when it's a static file.


james-elicx:
For what it's worth, you don't need to keep tagging us
/bigbonk review

yunus25jmi1:
Merge the PR to main. The PR is in a good state.

yunus25jmi1:
Addressed all bonk review items:

1. **Hoisted `resolvedPublicDir`** in `index.ts` — computed once before both afterFiles and fallback blocks instead of redundantly in each
2. **Replaced bare `catch {}`** in `app-dev-server.ts` — both afterFiles and fallback static file checks now log non-ENOENT errors
3. **Added `wasm: "application/wasm"`** to `mime.ts` — relevant for Cloudflare Workers apps
4. **Documented leading-slash assumption** — added comments in both `app-dev-server.ts` and `index.ts` explaining that `"." + pathname` relies on rewrite destinations starting with `/`, and that the traversal guard catches malformed paths regardless
5. **Merged upstream/main** — resolved conflicts, kept both `bodySizeLimit` (from #338) and `root` parameter

Re: making `root` required — kept it optional for forward compatibility as bonk noted is also fine. The warning + null propagation cleanly disables the feature when missing.

yunus25jmi1:
Addressed latest bonk review items:

1. **Hoisted `__publicDir` to module scope** in generated RSC entry — replaces repeated `JSON.stringify(publicDir)` inline literals with a single `const __publicDir` at module scope (next to `__mimeTypes`, `__configRewrites`, etc.)
2. **Added fallback rewrite static file tests** — new test in both `app-router.test.ts` and `pages-router.test.ts` that exercises the fallback rewrite path targeting a `.html` file in `public/`. Added corresponding fixture config entries and `public/fallback-page.html` files.
3. **Updated unit test** — `generateRscEntry` unit test now verifies `__publicDir` is hoisted to module scope in the generated code.

Re `root` required vs optional: kept optional for now since bonk said "current approach is also fine for forward compatibility" and it makes the function more defensive for future callers.

All 213 app-router tests and 130/132 pages-router tests pass (2 pre-existing failures unrelated to this PR: "hydration proxy script" and "client bundle chunks").

yunus25jmi1:
Addressed the Bonk review item:

- **`path.join` → `path.resolve` for `publicDir`** in `app-dev-server.ts` — now consistent with `index.ts`.  
  `path.resolve` guarantees a fully normalized absolute path, making the traversal guard comparison more robust. The unit test has been updated accordingly.

Re: the 10 positional parameters — acknowledged as technical debt, but not addressed in this PR as Bonk noted.

All 213 app-router tests pass. Typecheck clean.

yunus25jmi1:
Hi @southpolesteve @james-elicx @elithrar - this PR has been open for about a month now. The CI is passing and the merge conflicts have been resolved. Could we get a final review? This fixes issue #199 - serving static files from public/ when rewrite target is a `.html` path.

james-elicx:
Hi, please could you allow us a little bit of time to check on PRs before tagging in the future :)

packages/vinext/src/index.ts
```diff
@@ -2894,7 +2963,11 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
...
 
-              await handler(req, res, resolvedUrl, mwStatus);
+              try {
+                await handler(req, res, resolvedUrl, mwStatus);
+              } catch (e) {
+                next(e);
+              }
             } catch (e) {
               next(e);
             }
```
james-elicx: Reading this diff, the error was already being caught and passed to next as it's running inside a try/catch

packages/vinext/src/index.ts
```diff
@@ -2877,6 +2920,32 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
                     await proxyExternalRewriteNode(req, res, fallbackRewrite);
                     return;
                   }
+                  // Check if fallback targets a static file in public/
```
james-elicx: this code looks almost identical to the other block above - potential to re-use / share impls instead of duplicating

packages/vinext/src/entries/app-rsc-entry.ts
```diff
@@ -1893,6 +1906,29 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
...
+      const __afterExtname = __nodePath.extname(cleanPathname);
+      if (__afterExtname && __publicDir !== null) {
+        // "." + cleanPathname works because rewrite destinations always start with "/";
+        // the traversal guard below catches any malformed path regardless.
+        const __afterPublicFile = __nodePath.resolve(__publicDir, "." + cleanPathname);
+        if (__afterPublicFile.startsWith(__publicDir + __nodePath.sep)) {
+          try {
+            const __afterStat = __nodeFs.statSync(__afterPublicFile);
+            if (__afterStat.isFile()) {
+              const __afterContent = __nodeFs.readFileSync(__afterPublicFile);
```
james-elicx: I would like to avoid adding statSync and readFileSync in entries if possible as they may introduce issues if the codepath is hit on Workers.

packages/vinext/src/entries/app-rsc-entry.ts
```diff
@@ -1893,6 +1906,29 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
         return proxyExternalRequest(request, __afterRewritten);
       }
       cleanPathname = __afterRewritten;
+      // If the rewritten path has a file extension, it may point to a static
+      // file in public/. Serve it directly before route matching.
```
james-elicx: How does this differ to the implementation we have for after middleware runs? Is this custom implementation needed? Can we re-use the existing one instead?

https://github.com/cloudflare/vinext/blob/5f6bccbd039bbbbbf7e0d516f0737080cef5ce15/packages/vinext/src/entries/app-rsc-entry.ts#L1672-L1682

packages/vinext/src/server/mime.ts
james-elicx, line 1: happy for this to be split into a separate pr if you want to get this mime types refactor in without waiting for this pr to go through

packages/vinext/src/server/prod-server.ts
```diff
@@ -501,7 +500,7 @@ async function tryServeStatic(
   if (!resolved) return false;
 
   const ext = path.extname(resolved.path);
-  const ct = CONTENT_TYPES[ext] ?? "application/octet-stream";
+  const ct = mimeType(ext.startsWith(".") ? ext.slice(1) : ext);
```
james-elicx: the startswith check should probably be built into the utility

james-elicx (changes requested):

wojtekmaj (approved):
I can confirm this fixes my issue, see https://github.com/cloudflare/vinext/issues/199#issuecomment-4395837999

yunus25jmi1:
@james-elicx @wojtekmaj, I've thoroughly reviewed the latest changes in this PR. The implementation now correctly handles middleware headers, includes robust traversal guards, and maintains parity across all server environments. The core issue #199 is definitively resolved and verified by the new integration tests. Please take a look at the final cleanup and consider for merge.

PR #247, opened 2026-03-04
fix: set correct NODE_ENV for build

### LINKED ISSUE #246: Set correct NODE_ENV for build
Currently, NODE_ENV may not be set to production during the build, which can lead to incorrect output.

hyoban:
Fixes #246

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -1698,11 +1698,15 @@ hydrate();
           }
         }
         // Align NODE_ENV with Next.js semantics: build -> production, serve -> development.
-        const resolvedNodeEnv = mode === "test"
-          ? "test"
-          : env?.command === "build"
-          ? "production"
-          : "development";
+        // Next.js unconditionally forces NODE_ENV during build/dev, so we do the same.
+        let resolvedNodeEnv: string;
+        if (mode === "test") {
+          resolvedNodeEnv = "test";
+        } else if (env?.command === "build") {
+          resolvedNodeEnv = "production";
+        } else {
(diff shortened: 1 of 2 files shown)


southpolesteve (commented):
Addressed all review feedback and added an additional fix:

**Review feedback addressed:**
1. **Ternary readability** - Replaced nested ternary with if/else block for the NODE_ENV resolution logic.
2. **beforeEach/afterEach scope** - Kept file-scoped (necessary because other test blocks like `optimizeDeps` and `treeshake` also call `mainPlugin.config()`, which now mutates `process.env.NODE_ENV`). Added a comment explaining why.
3. **Test coverage** - Added three new test cases:
   - `command: "serve"` verifies `"development"`
   - `command: "build"` without explicit `mode` verifies `"production"`
   - User-defined `config.define` with `process.env.NODE_ENV` verifies the guard skips injection
4. **Guard condition** - Left as-is; the `typeof` check defensively handles the `null` case.

**Additional fix (from my own review):**
- The `nextConfig.env` loop (which injects env vars from `next.config.js`) could silently override the NODE_ENV define we just set. If a user had `env: { NODE_ENV: "development" }` in their next.config, it would overwrite `"production"` during build. Added a `if (key === "NODE_ENV") continue` guard, matching Next.js behavior (Next.js also ignores NODE_ENV in the env config key).

Also extracted a `setupTmpProject()` helper in the test file to reduce boilerplate across the NODE_ENV test cases.

All checks pass locally: typecheck clean, lint 0 warnings, 37/37 build-optimization tests pass, 1041/1041 core tests pass.

PR #250, opened 2026-03-04
fix(rsc): emit actionable hint when non-plain object crosses server-client boundary

### LINKED ISSUE #237: Only plain objects, and a few built-ins, can be passed to Client Components from Server Components
Hello, i'm trying to use vinext to replace next 16.1.6, my current next project works without giving me the error, but as soon as i use vinext this issue pops up 

```js
Only plain objects, and a few built-ins, can be passed to Client Components from Server Components. Classes or null prototypes are not supported.
  [..., Module, Null, ..., ..., [], 1]
        ^^^^^^
    at resolveErrorDev (C:\Users\specy\Desktop\progetti\discerns\client\node_modules\.vite\deps_ssr\react-server-dom-webpack_client__edge.js:4022:148)
    at processFullBinaryRow (C:\Users\specy\Desktop\progetti\discerns\client\node_modules\.vite\deps_ssr\react-server-dom-webpack_client__edge.js:4813:29)
    at progress (C:\Users\specy\Desktop\progetti\discerns\client\node_modules\.vite\deps_ssr\react-server-dom-webpack_client__edge.js:5009:20)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

I'm aware that i cannot pass Class instances etc to the server-client boundary, but i don't seem to be passing any (as i recall at least). I cannot debug further as the application is quite large and the error does not give any context of what file might be causing it.

Using next 16.1.6 i don't have this error

SeolJaeHyeok:
Fixes #237 

### Problem

  When a Server Component passes a non-serializable value (class instance, ESM module namespace object, `Date`, `Map`, or `Object.create(null)`) as a prop to a Client Component, React's RSC serializer throws:

  > Only plain objects, and a few built-ins, can be passed to Client Components
  from Server Components. Classes or null prototypes are not supported.
    [..., Module, Null, ..., ..., [], 1]
          ^^^^^^

  The error surfaces in the browser with no indication of **which component or prop** triggered it, making it nearly impossible to debug in large applications. Users migrating from plain Next.js are especially affected because the same code silently works there (see root cause below).

  ---

  ### Root Cause

  This is a fundamental difference between how Vite and webpack handle ES modules:

  | | Next.js (webpack) | vinext (Vite/ESM) |
  |---|---|---|
  | `import * as X from './mod'` | `{ __esModule: true, ...exports }` — **plain object** | True ESM Module Namespace Object — **non-plain** |
  | RSC serializable? | ✅ Yes (passes `isPlainObject` check) | ❌ No (internal `[[Module]]` slot) |

  When users pass `import * as utils` as a prop to a Client Component, webpack wraps it as a plain `__esModule` object that React's RSC serializer accepts. Vite produces a real ESM Module Namespace Object with a null-like prototype that React correctly rejects. **vinext's behavior is actually more spec-correct** — Next.js/webpack was accidentally permitting invalid semantics.

  ---

  ### Solutions Considered

  #### ✅ Solution 1 — User-land fix (always applicable)

  Convert non-serializable values to plain objects before passing them across
  the RSC boundary:

  ```ts
  // ❌ Fails: class instance
  const service = new UserService();
  return <UserCard service={service} />;

  // ✅ Works: extract plain data
  const user = await service.getUser(1);
  return <UserCard user={{ id: user.id, name: user.name }} />;

  // ❌ Fails: module namespace object
  import * as config from './config';
  return <AppInfo config={config} />;

  // ✅ Works: spread into a plain object
  return <AppInfo config={{ ...config }} />;

  // ❌ Fails: Date instance
  return <Card createdAt={new Date()} />;

  // ✅ Works: serialize to string
  return <Card createdAt={new Date().toISOString()} />;
```

  #### ✅ Solution 2 — Actionable dev-mode hint in rscOnError (this PR)

  Enhanced rscOnError in app-dev-server.ts to detect this specific error
  in dev mode and print a structured, actionable message to the server console:

```
[vinext] RSC serialization error: a non-plain object was passed from a Server Component to a Client Component.

Common causes:
  * Passing a module namespace (import * as X) directly as a prop.
    Unlike Next.js (webpack), Vite produces real ESM module namespace objects
    which are not serializable. Fix: pass individual values instead,
    e.g. <Comp value={module.value} />
  * Passing a class instance (new Foo()) as a prop.
    Fix: convert to a plain object, e.g. { id: foo.id, name: foo.name }
  * Passing a Date, Map, or Set. Use .toISOString(), [...map.entries()], etc.
  * Passing Object.create(null). Use { ...obj } to restore a prototype.
```

  No behavior change in production — the existing digest hash logic is unchanged.

  #### ❌ Solution 3 — Auto-convert Module Namespace Objects in vinext (not applied)

  An alternative approach was considered: intercept props before
  renderToReadableStream and automatically convert ESM Module Namespace
  Objects to plain objects via { ...moduleNS } or
  Object.fromEntries(Object.entries(moduleNS)).

  This was rejected for three reasons:

  1. Semantic corruption. Silently flattening a module namespace drops
  important distinctions — live bindings become static snapshots, and any
  non-enumerable exports or symbols are lost without warning. It would make
  vinext behave like webpack rather than being correct.
  2. Tree-shaking breakage. Auto-converting import * as X to a plain
  object and passing it to a Client Component means all exports of that
  module would be bundled into the client, defeating the RSC boundary's
  purpose of keeping server-only code out of the client bundle.
  3. Hides the real bug. The user's code has a genuine error — passing
  non-serializable data across an RSC boundary. Silently fixing it
  server-side masks the problem instead of helping the user understand and
  fix their code. A clear error message is the correct resolution.

  ---
### Changes

  - packages/vinext/src/server/app-dev-server.ts: added non-plain object detection block in rscOnError (dev-only, no production impact)

  ---
###  Testing

Minimal reproduction:

```ts
  // app/components/UserCard.tsx
  'use client';
  export function UserCard({ service }: { service: any }) {
    return <div />;
  }

  // app/page.tsx  (Server Component)
  import { UserCard } from './components/UserCard';
  class UserService { constructor(public baseUrl: string) {} }

    // Triggers the error — class instance across RSC boundary
    return <UserCard service={new UserService('https://api.example.com')} />;
  }
  ```

### DIFF
--- a/packages/vinext/src/server/app-dev-server.ts
+++ b/packages/vinext/src/server/app-dev-server.ts
@@ -321,21 +321,22 @@ function rscOnError(error) {
     error.message.includes("Only plain objects, and a few built-ins, can be passed to Client Components")
   ) {
     console.error(
-      "[vinext] RSC serialization error: a non-plain object was passed from a Server Component to a Client Component.\n" +
-      "\n" +
-      "Common causes:\n" +
-      "  * Passing a module namespace (import * as X) directly as a prop.\n" +
-      "    Unlike Next.js (webpack), Vite produces real ESM module namespace objects\n" +
-      "    which are not serializable. Fix: pass individual values instead,\n" +
-      "    e.g. <Comp value={module.value} />\n" +
-      "  * Passing a class instance (new Foo()) as a prop.\n" +
-      "    Fix: convert to a plain object, e.g. { id: foo.id, name: foo.name }\n" +
-      "  * Passing a Date, Map, or Set. Use .toISOString(), [...map.entries()], etc.\n" +
-      "  * Passing Object.create(null). Use { ...obj } to restore a prototype.\n" +
-      "\n" +
+      "[vinext] RSC serialization error: a non-plain object was passed from a Server Component to a Client Component.\\n" +
+      "\\n" +
+      "Common causes:\\n" +
+      "  * Passing a module namespace (import * as X) directly as a prop.\\n" +
+      "    Unlike Next.js (webpack), Vite produces real ESM module namespace objects\\n" +
+      "    which are not serializable. Fix: pass individual values instead,\\n" +
+      "    e.g. <Comp value={module.value} />\\n" +
+      "  * Passing a class instance (new Foo()) as a prop.\\n" +
+      "    Fix: convert to a plain object, e.g. { id: foo.id, name: foo.name }\\n" +
+      "  * Passing a Date, Map, or Set. Use .toISOString(), [...map.entries()], etc.\\n" +
+      "  * Passing Object.create(null). Use { ...obj } to restore a prototype.\\n" +
+      "\\n" +
       "Original error:",
       error.message,
     );
+    return undefined;
   }
 
   // In production, generate a digest hash for non-navigation errors
--- a/tests/app-router.test.ts
+++ b/tests/app-router.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeAll, afterAll } from "vitest";
+import { describe, it, expect, beforeAll, afterAll, vi } from "vitest";
 import { createBuilder, type ViteDevServer } from "vite";
 import path from "node:path";
 import fs from "node:fs";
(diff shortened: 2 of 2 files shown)


southpolesteve (commented):
Pushed fixes to the branch (7ea4fce):

**Bug fix: unterminated string literals in generated code**

The `\n` escape sequences in the `console.error` hint strings were inside a template literal, so they resolved to actual newlines in the generated JavaScript output. This produced unterminated string literals that broke production builds (`Unterminated string constant` from Rollup). Fixed by double-escaping to `\\n` so the output code gets valid `\n` escape sequences.

**Review feedback addressed:**
1. Added explicit `return undefined` after the hint `console.error` block (ask-bonk item 1)
2. Added runtime tests that extract `rscOnError` from the generated code via `new Function()` and verify three behaviors (ask-bonk item 2):
   - Digest passthrough for navigation errors (redirect/notFound)
   - `console.error` hint fires and returns `undefined` for RSC serialization errors
   - Returns `undefined` without logging for generic dev errors

**Dev/prod parity check:** `rscOnError` only exists in the generated RSC entry (from `app-dev-server.ts`). The production server delegates to the built RSC entry, so the fix is inherited. The `NODE_ENV !== "production"` guard correctly suppresses the hint in production builds. No parity issues.

Typecheck, lint, and all `rscOnError` tests pass (8/8).

PR #252, opened 2026-03-04
fix: apps with both app + pages router returning 404s

james-elicx:
If an app has both an app directory and a pages directory, all its app directory routes return 404s.

Regression test is achieved through simply having a pages directory route live in the app directory fixture. Without this patch, you then get 404s.


james-elicx:
Was not expecting to see 2 middleware header failures pop up from those changes... Interesting.

PR #257, opened 2026-03-05
fix(ssr): preload client reference modules before first SSR render

### LINKED ISSUE #256: fix(ssr): first request after server start always returns 500
## Bug Description

The **first HTTP request** after starting the vinext production server (`vinext start`) **always returns a 500 Internal Server Error**. All subsequent requests succeed with 200. This is 100% reproducible on every server restart.

## Root Cause

During production SSR, client reference modules (`"use client"` components) are loaded lazily via async `import()` through the `setRequireModule({ load })` mechanism provided by `@vitejs/plugin-rsc`. On the very first request:

1. `handleSsr` calls `createFromReadableStream` to create the React element tree from the RSC stream
2. `renderToReadableStream` attempts to render the HTML shell synchronously
3. When React encounters a client component reference, it calls `__vite_rsc_client_require__(id)` which returns a **Promise** (the async `import()` has not completed yet)
4. Since there is no `<Suspense>` boundary wrapping the root shell, React SSR cannot render a fallback — it rejects with `undefined`
5. The `onError` callback receives `undefined`, and `renderToReadableStream` rejects
6. The production server catches this as `"Server error: undefined"` and returns 500

On subsequent requests, the modules are already cached by the `memoize()` wrapper in `setRequireModule`, so `__vite_rsc_client_require__` resolves synchronously from cache and SSR succeeds.

## Steps to Reproduce

1. Build a vinext app with any `"use client"` components: `vinext build`
2. Start the production server: `vinext start`
3. Send a request immediately: `curl http://localhost:3120/`
4. **Result:** 500 Internal Server Error
5. Send the same request again: `curl http://localhost:3120/`
6. **Result:** 200 OK

This reproduces on every restart cycle — `systemctl restart` / `vinext start` / any cold start.

## Environment

- vinext: v0.0.20
- @vitejs/plugin-rsc: v0.5.21
- Node.js: v22.x
- OS: Linux (Ubuntu 24.04)

## Proposed Fix

Eagerly preload all client reference modules at the start of `handleSsr`, before `renderToReadableStream` runs. The `virtual:vite-rsc/client-references` module (provided by `@vitejs/plugin-rsc`) exposes the map of all client component IDs to their async import functions. By calling `__vite_rsc_client_require__` for each ID before rendering, the memoize cache is warmed and all subsequent `require` calls during SSR resolve synchronously.

The fix is minimal (16 lines, single file) and has no impact on subsequent requests since `memoize()` returns cached values immediately.

PR incoming: #TBD

gagipro:
## Summary

Fixes #256 — the first HTTP request after `vinext start` always returns 500.

## Problem

On cold start, client reference modules (`"use client"` components) are loaded lazily via
async `import()` through `@vitejs/plugin-rsc`'s `setRequireModule({ load })`. When `handleSsr`
runs `renderToReadableStream` on the first request, React SSR tries to render the HTML shell
but the client component imports haven't resolved yet. With no `<Suspense>` boundary wrapping
the root, React cannot render a fallback and rejects with `undefined`, causing a 500.

All subsequent requests work because `memoize()` in `setRequireModule` caches resolved modules.

## Fix

Eagerly preload all client reference modules at the start of `handleSsr`, before
`renderToReadableStream` runs:

```typescript
import * as _clientRefs from "virtual:vite-rsc/client-references";

export async function handleSsr(rscStream, navContext, fontData) {
  // Preload client refs so first SSR render can resolve them synchronously
  if (_clientRefs?.default && globalThis.__vite_rsc_client_require__) {
    await Promise.all(
      Object.keys(_clientRefs.default).map((id) =>
        globalThis.__vite_rsc_client_require__(id).catch?.(() => {})
      )
    );
  }
  // ... rest of handleSsr unchanged
}
```

The `virtual:vite-rsc/client-references` module (provided by `@vitejs/plugin-rsc`) exposes
the map of all client component IDs. The `__vite_rsc_client_require__` global is already set
by `initialize()` before `handleSsr` is called.

## Scope

- **1 file changed**, 16 insertions
- `packages/vinext/src/server/app-dev-server.ts` — inside `generateSsrEntry()` template
- No changes to `@vitejs/plugin-rsc` or any other package

## Testing

Verified on a production vinext app with 11 client component references:
- **Before fix:** first request after restart → 500 (100% reproducible)
- **After fix:** 5/5 restart cycles → 200 on first request, all routes (/, /login, /dashboards, /datasources, /dashboards/[id]) return 200

## Notes

- The preload runs once per request but the `memoize()` cache means only the first call per ID
  does actual work. The overhead on subsequent requests is negligible (synchronous Map lookups).
- The defensive `?.default` and `?.catch` guards ensure the fix is safe if `@vitejs/plugin-rsc`
  changes its export shape in the future.
- This fix is in the generated SSR entry (template literal in `generateSsrEntry()`), so it
  affects both dev and production SSR.

### DIFF
--- a/packages/vinext/src/entries/app-ssr-entry.ts
+++ b/packages/vinext/src/entries/app-ssr-entry.ts
@@ -12,6 +12,9 @@ import { setNavigationContext, ServerInsertedHTMLContext } from "next/navigation
 import { runWithNavigationContext as _runWithNavCtx } from "vinext/navigation-state";
 import { safeJsonStringify } from "vinext/html";
 import { createElement as _ssrCE } from "react";
+import * as _clientRefs from "virtual:vite-rsc/client-references";
+
+let _clientRefsPreloaded = false;
 
 /**
  * Collect all chunks from a ReadableStream into an array of text strings.
@@ -150,6 +153,29 @@ function createRscEmbedTransform(embedStream) {
  *   and the data needs to be passed to SSR since they're separate module instances.
  */
 export async function handleSsr(rscStream, navContext, fontData) {
+  // Eagerly preload all client reference modules before SSR rendering.
+  // On the first request after server start, client component modules are
+  // loaded lazily via async import(). Without this preload, React's
+  // renderToReadableStream rejects because the shell can't resolve client
+  // components synchronously (there is no Suspense boundary wrapping the
+  // root). The memoized require cache ensures this is only async on the
+  // very first call; subsequent requests resolve from cache immediately.
+  // See: https://github.com/cloudflare/vinext/issues/256
+  // _clientRefs.default is the default export from the virtual:vite-rsc/client-references
+  // namespace import — a map of client component IDs to their async import functions.
+  if (!_clientRefsPreloaded && _clientRefs.default && globalThis.__vite_rsc_client_require__) {
+    await Promise.all(
+      Object.keys(_clientRefs.default).map((id) =>
+        globalThis.__vite_rsc_client_require__(id).catch((err) => {
+          if (process.env.NODE_ENV !== "production") {
+            console.warn("[vinext] failed to preload client ref:", id, err);
+          }
+        })
+      )
+    );
+    _clientRefsPreloaded = true;
+  }
+
   // Wrap in a navigation ALS scope for per-request isolation in the SSR
   // environment. The SSR environment has separate module instances from RSC,
   // so it needs its own ALS scope.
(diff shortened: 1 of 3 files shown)


gagipro:
Thanks for the review @hi-ogawa — good points.

You're right that this fix only covers the `"use client"` references that vinext itself auto-wraps (via `@vitejs/plugin-rsc`). It doesn't help if someone manually writes `React.lazy(() => import(...))` inside their own client component — those lazy imports would still race on the first SSR.

A couple of questions:

1. **Do you have a link to the React SSR + lazy component issues you mentioned?** Would be helpful to check whether it's the same root cause or a separate React-level bug.
2. **Do you have a concrete repro for the manually-written `React.lazy` edge case in a vinext app?** If so, I can add a test with an artificially slow lazy component as you suggested. Otherwise, would you be open to tracking that as a follow-up issue?

This PR targets the common case (every `"use client"` component → 500 on first request, 100% repro rate). Happy to extend the fix or open a separate issue for the manual `React.lazy` edge case.

PR #262, opened 2026-03-05
feat(font): replace generated font catalog with Vite import-rewriting transform

### LINKED ISSUE #200: Explore Vite transform for next/font/google imports (eliminate generated font catalog)
## Context

PR #57 solved the immediate problem of missing named exports by code-generating all ~1,923 Google Font exports from the Google Fonts API. This works and tree-shakes correctly, but it requires maintaining a generated file that needs periodic re-runs as Google adds new fonts.

## How Next.js Does It

Next.js handles `next/font/google` entirely at compile time via an SWC transform:

1. SWC detects `import { Inter } from 'next/font/google'`
2. Rewrites `const inter = Inter({...})` into a CSS import with query params
3. A webpack loader downloads fonts and generates `@font-face` CSS at build time

The `next/font/google` module in Next.js only contains `declare function` type stubs. There's zero runtime JavaScript. The font system is a compile-time illusion.

## Proposed Approach

We could implement a similar pattern using a Vite `transform` hook:

1. Detect `import { X } from 'next/font/google'` in the importing file
2. Rewrite it to `import { createFontLoader } from 'vinext/shims/font-google-base'; const X = createFontLoader("X Name");`
3. The transform derives the font family name from the export name (same logic the Proxy already uses)

### Benefits

- No generated file to maintain (no need to re-run a script when Google adds fonts)
- Works for any font automatically, including ones Google hasn't added yet
- Zero bundle overhead (only fonts actually used get `createFontLoader` calls)
- Closer to how Next.js handles it architecturally

### Considerations

- Need to handle edge cases: re-exports, dynamic imports, barrel files
- IDE autocomplete would still need the `.d.ts` declarations file (could keep generating that separately)
- Must work correctly in all three Vite environments (RSC, SSR, client)

## Current State

The generated catalog from PR #57 works well and is the right solution for now. This issue tracks exploring the transform approach as a future improvement.

yunus25jmi1:
## Summary

Closes #200

Instead of a 1928-line generated font catalog (`font-google.generated.ts`) that re-exports every Google Font as a `createFontLoader` call, the `vinext:google-fonts` Vite plugin now rewrites `next/font/google` import statements at compile time.

### Before

```ts
// font-google.generated.ts (1928 lines, ~70KB)
export const Inter: FontLoader = createFontLoader('Inter');
export const Roboto: FontLoader = createFontLoader('Roboto');
// ... 1926 more lines
```

### After

The plugin rewrites imports at compile time across all Vite environments (RSC, SSR, client):

```ts
// User writes:
import { Inter, Roboto_Mono } from 'next/font/google'

// Plugin transforms to:
import { createFontLoader as __vinext_clf } from 'next/font/google';
const Inter = __vinext_clf('Inter');
const Roboto_Mono = __vinext_clf('Roboto Mono');
```

Self-hosted font fetching during production builds continues to work as before.

## Changes

- **Delete** `packages/vinext/src/shims/font-google.generated.ts` (1928 lines / ~70 KB)
- Update `font-google.ts` to export `createFontLoader` instead of re-exporting the catalog
- Add `createFontLoader` export to `next-shims-font-google.generated.d.ts` for IDE autocomplete
- Rewrite `vinext:google-fonts` Vite plugin to perform import-time transforms in dev + build
- Fix Proxy in `font-google-base.ts` to handle underscore-style names (`Roboto_Mono` → `'Roboto Mono'`) for users who skip the Vite plugin
- Update `generate-google-fonts.js` to only generate `.d.ts` (no longer generates `.ts` catalog)
- Update tests to use `createFontLoader` API instead of named exports

## Testing

- All 54 `font-google.test.ts` tests pass
- All 7 font/google tests in `shims.test.ts` pass
- All 9 `entry-templates.test.ts` tests pass


packages/vinext/src/index.ts
```diff
@@ -2882,109 +2889,170 @@ hydrate();
...
+          // Parse individual import specifiers
+          const rawSpecifiers = importMatch[1].split(",").map((s) => s.trim()).filter(Boolean);
 
-          const s = new MagicString(code);
-          let hasChanges = false;
+          // Utility exports that are NOT font names — keep as regular imports
+          const utilityExports = new Set([
+            "buildGoogleFontsUrl", "getSSRFontLinks", "getSSRFontStyles",
+            "getSSRFontPreloads", "createFontLoader",
+          ]);
```
james-elicx: would be worth thinking about if there's a way for us to ensure this is kept in-sync. would also be good to add a comment about it as well.

yunus25jmi1:
@james-elicx re-review the PR. Conflicts fixed.

PR #273, opened 2026-03-05
fix(config): pass correct Next.js phase to function-form configs + instrumentation test coverage

NathanDrake2406:
## Summary

- **fix(config):** `unwrapConfig()` hardcoded `"phase-development-server"` when calling function-form Next.js configs, even during production builds. Apps that branch on `phase` in `next.config.ts` (e.g. `(phase) => phase === "phase-production-build" ? prodConfig : devConfig`) received the wrong config during `vinext build`. Now maps Vite's `env.command` to the correct Next.js phase (`"build"` → `"phase-production-build"`, `"serve"` → `"phase-development-server"`).

- **test(instrumentation):** Added 11 tests for `server/instrumentation.ts` which previously had zero coverage. Covers `findInstrumentationFile` (file discovery priority, src/ fallback, null when absent), `runInstrumentation` (register lifecycle, onRequestError storage, error isolation), and `reportRequestError` (handler forwarding, no-op without handler, handler error isolation).

- **fix(test):** Fixed 6 pre-existing failures in `static-export.test.ts` App Router section. The test used a local `startFixtureServer` that loaded the fixture's `vite.config.ts` via `configFile`, which caused RSC plugin timing issues on Node 25. Switched to the shared `startFixtureServerHelper` from `helpers.ts` (uses `configFile: false` with explicit plugins), matching the pattern used by `app-router.test.ts`.

## Test plan

- [x] `npx vitest run tests/next-config.test.ts` — 3 tests verifying phase propagation (production build phase, dev default, object-form ignores phase)
- [x] `npx vitest run tests/instrumentation.test.ts` — 11 tests covering file discovery, register/onRequestError lifecycle, error isolation
- [x] `npx vitest run tests/static-export.test.ts` — 14 tests now all pass (was 6 failures)
- [x] `npx vitest run` — full suite passes (2182 passed, 0 failed)
- [x] `pnpm run typecheck` — clean


NathanDrake2406:
@southpolesteve pls have a look :)

PR #277, opened 2026-03-06
fix: stop stripping credential headers from rewrite proxy requests

### LINKED ISSUE #264: Bug: Rewrite proxy strips cookie/authorization headers, breaking backend auth
## Description

`proxyExternalRequest` (in `config-matchers.ts`) and `__proxyExternalRequest` (inlined in `app-dev-server.ts`) unconditionally strip `cookie`, `authorization`, `x-api-key`, and `proxy-authorization` headers from proxied rewrite requests. This breaks authentication for users who use `rewrites` to proxy to their own backend services.

## Root Cause

The proxy functions contain:

```ts
headers.delete("cookie");
headers.delete("authorization");
headers.delete("x-api-key");
headers.delete("proxy-authorization");
```

This was added as a security measure to prevent credential leakage to third-party origins. However, the most common use case for rewrites is proxying to the user's own backend, where these credentials are required.

## Reproduction

```ts
// next.config.ts
const nextConfig: NextConfig = {
  rewrites: () => [
    {
      source: "/api/:path*",
      destination: `${process.env.BACKEND_URL}/api/:path*`,
    },
  ],
};
```

Any authenticated request to `/api/...` will fail because the `authorization` header and cookies are stripped before reaching the backend.

## Expected Behavior

Next.js itself does **not** strip these headers for rewrite proxies — it forwards all headers as-is to the destination. vinext should match this behavior for compatibility.

## Suggested Fix

Remove the blanket credential stripping. If security hardening is desired for third-party origins, consider:

1. Matching Next.js behavior by default (forward all headers)
2. Optionally allowing users to configure header stripping per-rewrite rule

## Environment

- vinext: 0.0.21

17hz:
Fixes #264

## Problem

`proxyExternalRequest` and `__proxyExternalRequest` unconditionally strip `cookie`, `authorization`, `x-api-key`, and `proxy-authorization` headers from all proxied rewrite requests. This breaks authentication for the most common use case — proxying to the user's own backend.

## Context

Next.js does not strip these headers for rewrites. It forwards all request headers as-is to the rewrite destination. vinext's blanket stripping is a deviation from Next.js behavior that breaks real-world usage.

## Fix

Remove the credential-stripping logic from both:

- `config-matchers.ts` — `proxyExternalRequest` (used by prod server)
- `app-dev-server.ts` — `__proxyExternalRequest` (inlined in dev server virtual module)

Internal `x-middleware-*` headers are still stripped as before.

### DIFF
--- a/tests/app-router.test.ts
+++ b/tests/app-router.test.ts
@@ -2910,7 +2910,7 @@ describe("App Router external rewrite proxy credential stripping", () => {
     await new Promise<void>((resolve) => mockServer?.close(() => resolve()));
   });
 
-  it("strips credential headers from proxied requests to external rewrite targets", async () => {
+  it("forwards credential headers through proxied requests to external rewrite targets", async () => {
     mockResponseMode = "plain";
     capturedHeaders = null;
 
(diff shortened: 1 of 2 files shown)


james-elicx:
cc @threepointone - i believe this was related to the security advisory changes

PR #293, opened 2026-03-06
fix: stub node:async_hooks in client builds to prevent Rollup errors

Divkix:
## Summary

- Shim files import `AsyncLocalStorage` from `node:async_hooks`. When resolved into the **client** (browser) build, Vite externalizes `node:async_hooks` to `__vite-browser-external` — an empty stub with no named exports. This causes **Rollup named-export failures**.
- Server environments (rsc, ssr) use real `node:async_hooks` natively via `nodejs_compat`. Only the client environment needs a stub.
- Adds a browser-safe `AsyncLocalStorage` stub at `packages/vinext/src/shims/async-hooks-browser.ts` with no-op methods (`run()` passes through, `getStore()` returns `undefined` — matching existing `_als.getStore() ?? _fallbackState` patterns in shims).
- Extends the `resolveId` filter to match `node:async_hooks` and redirect to the stub **only** when `this.environment.name === "client"`.

## Reproduction

1. Invalidate build cache (e.g., change chunk graph or delete `.vite/`)
2. Run `bun run build` on an App Router project with shims that import `AsyncLocalStorage`
3. Build fails with Rollup error: `"AsyncLocalStorage" is not exported by "__vite-browser-external"`

## Fix

Two changes:
1. **New file**: `packages/vinext/src/shims/async-hooks-browser.ts` — browser-safe no-op `AsyncLocalStorage` class
2. **Modified**: `packages/vinext/src/index.ts` — extended `resolveId` hook to redirect `node:async_hooks` → stub in client environment only

## Test plan

- [x] Clean build succeeds without Rollup named-export errors
- [x] The browser stub's `run()` correctly passes through the callback return value
- [x] Server environments (rsc, ssr) still use real `node:async_hooks` — no behavioral change
- [x] `biome check` passes on all changed files

### DIFF
--- a/tests/shims.test.ts
+++ b/tests/shims.test.ts
@@ -8034,16 +8034,16 @@ describe("node:async_hooks resolveId", () => {
     return typeof hook === "function" ? hook : hook?.handler;
   }
 
-  function getConfigPlugin() {
-    const vinext = require("../packages/vinext/src/index.js").default;
+  async function getConfigPlugin() {
+    const vinext = (await import("../packages/vinext/src/index.js")).default;
     const plugins = vinext() as any[];
     const plugin = plugins.find((p: any) => p.name === "vinext:config");
     if (!plugin) throw new Error("vinext:config plugin not found");
     return plugin;
   }
 
-  it("resolves node:async_hooks to stub in client environment", () => {
-    const plugin = getConfigPlugin();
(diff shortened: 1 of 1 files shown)


james-elicx:
It feels wrong that this would be being imported into a browser bundle in the first place... Did you look into what was pulling it in and if there were ways to stop that from doing so?

PR #303, opened 2026-03-06
fix: rebuild reqCtx for rewrites after middleware

james-elicx:
All three rewrite types — `beforeFiles`, `afterFiles`, and `fallback` — run after middleware per the Next.js execution order:

```
headers → redirects → Middleware → beforeFiles → filesystem → afterFiles → fallback
```

This applies to both App Router and Pages Router (confirmed in the [Next.js 14 docs](https://nextjs.org/docs/14/pages/building-your-application/routing/middleware#matching-paths)).

## What was wrong

`has`/`missing` conditions on rewrite rules were evaluated against the pre-middleware request context, so middleware-injected cookies/headers were invisible to the matcher.

Additionally, `beforeFiles` rewrites were running *before* middleware in `app-dev-server.ts` — a pre-existing ordering bug, now fixed.

## What changed

- **`app-dev-server.ts`**: Moved `beforeFiles` block to after middleware execution. Added `__buildPostMwRequestContext()` helper that reads the ALS-stored headers context (post-middleware) and converts `Map<string,string>` → `Record<string,string>` for cookie lookups. All three rewrite types now use this post-middleware context.
- **`prod-server.ts`**: `beforeFiles` now uses `postMwReqCtx` (was incorrectly using the pre-middleware snapshot). `headers` and `redirects` correctly retain the pre-middleware context.
- **`deploy.ts`**: Already correct for `afterFiles`/`fallback`; `beforeFiles` also switched to `postMwReqCtx`.
- **`shims/headers.ts`**: Exported `getHeadersContext()` for use in the post-middleware context helper.

## Tests added

Regression tests covering all three rewrite types across both routers:

| Rewrite type | Router | Server under test |
|---|---|---|
| `afterFiles` | Pages Router | dev + prod (`prod-server.ts`) |
| `beforeFiles` | App Router | dev (`app-dev-server.ts`) |
| `beforeFiles` | Pages Router | prod (`prod-server.ts`) |
| `fallback` | App Router | dev (`app-dev-server.ts`) |

Each test verifies: without `?mw-auth` the rule does not match (404); with `?mw-auth` middleware injects the cookie and the rule matches (200, rewrites to `/about`).

### DIFF
--- a/packages/vinext/src/server/prod-server.ts
+++ b/packages/vinext/src/server/prod-server.ts
@@ -766,9 +766,9 @@ async function startPagesRouterServer(options: PagesRouterServerOptions) {
       });
 
       // Build request context for has/missing condition matching.
-      // Pages Router: headers, redirects, and beforeFiles all run before
-      // middleware, so they use this pre-middleware snapshot. afterFiles
-      // and fallback rewrites use postMwReqCtx (built after middleware).
+      // headers and redirects run before middleware and use this pre-middleware
+      // snapshot. beforeFiles, afterFiles, and fallback all run after middleware
+      // per the Next.js execution order, so they use postMwReqCtx below.
       const reqCtx: RequestContext = requestContextFromRequest(webRequest);
 
       // ── 4. Run middleware ─────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


packages/vinext/src/server/app-dev-server.ts
```diff
@@ -1361,23 +1383,8 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
...
-        setNavigationContext(null);
-        return __proxyExternalRequest(request, __rewritten);
-      }
-      pathname = __rewritten;
-    }
-  }
-
   const isRscRequest = pathname.endsWith(".rsc") || request.headers.get("accept")?.includes("text/x-component");
-  let cleanPathname = pathname.replace(/\\.rsc$/, "");
+  let cleanPathname = pathname.replace(/.rsc$/, "");
```
james-elicx:
```suggestion
  let cleanPathname = pathname.replace(/\\.rsc$/, "");
```

PR #349, opened 2026-03-08
test: reproduce issue #346 — renderHTTPAccessFallbackPage drops params from layouts

### LINKED ISSUE #346: renderHTTPAccessFallbackPage does not pass params to layout components
`renderHTTPAccessFallbackPage` does not pass `params` to layout components

---

When vinext renders HTTP access fallback pages (404, 403, 401), layout components are rendered without the `params` prop, causing layouts that destructure route params to crash with:

```
Cannot destructure property 'lang' of '(intermediate value)' as it is undefined.
```

## Root cause

In the bundled output, `renderHTTPAccessFallbackPage` creates layout elements with only `{ children }` and no `params`:

```js
// renderHTTPAccessFallbackPage path — no params passed
const LayoutComponent = layouts[i3]?.default;
if (LayoutComponent) {
  element = createElement(LayoutComponent, { children: element });
}
```

The normal render path correctly passes params:

```js
// normal render path — params passed correctly
const layoutProps = { children: element, params: makeThenableParams(params) };
```

This affects both the RSC and non-RSC branches within `renderHTTPAccessFallbackPage`.

## Reproduction

1. Create a Next.js app with a dynamic `[lang]` route segment
2. In `app/[lang]/layout.tsx`, destructure `lang` from `params`:

```tsx
export default async function RootLayout({
  params,
  children,
}: {
  params: Promise<{ lang: string }>;
  children: React.ReactNode;
}) {
  const { lang } = await params;
  // use lang...
}
```

3. Navigate to any URL that triggers a 404 (e.g. `/en/nonexistent-page`)
4. The layout crashes because `params` is `undefined`

## Expected behavior

`renderHTTPAccessFallbackPage` should pass `params: makeThenableParams(params)` to layout components, matching the normal render path.

## Workaround

Add nullish coalescing when destructuring params in layouts:

```ts
const { lang } = (await params) ?? {};
```

## Environment

- vinext: `^0.0.22`
- next: `15.5.2`

james-elicx:
## Summary

Adds a failing test that reproduces #346: when a 404 is triggered under a dynamic `[lang]` route segment, `renderHTTPAccessFallbackPage` passes no `params` to layout components. This causes the scoped `not-found.tsx` boundary to be skipped entirely and the root not-found to render instead (or a crash if the layout doesn't guard against `undefined` params).

## Failing tests

- `notFound() under dynamic [lang] layout passes params to layout (issue #346)` — expects scoped "Not Found (params-layout)", gets root "404 - Page Not Found"
- `RSC request: notFound() under dynamic [lang] layout passes params (issue #346)` — same via `.rsc` endpoint

## What was added

- **Fixture** `tests/fixtures/app-basic/app/nextjs-compat/not-found-params-layout/[lang]/`
  - `layout.tsx` — destructures `lang` from `params` (the affected component type)
  - `page.tsx` — simple page under the dynamic segment
  - `not-found.tsx` — scoped fallback ("Not Found (params-layout)")
- **Tests** in `tests/nextjs-compat/not-found.test.ts` — sanity check (passes) + two failing regression tests

No fix is included; this PR exists to track the reproduction.

Closes #346

### DIFF
--- a/packages/vinext/src/server/app-dev-server.ts
+++ b/packages/vinext/src/server/app-dev-server.ts
@@ -504,11 +504,12 @@ async function renderHTTPAccessFallbackPage(route, statusCode, isRscRequest, req
     // client-side tree reconciliation.
     const _treePositions = route?.layoutTreePositions;
     const _routeSegs = route?.routeSegments || [];
-    const _fallbackParams = opts?.params || {};
+    const _fallbackParams = opts?.matchedParams ?? route?.params ?? {};
+    const _asyncFallbackParams = makeThenableParams(_fallbackParams);
     for (let i = layouts.length - 1; i >= 0; i--) {
       const LayoutComponent = layouts[i]?.default;
       if (LayoutComponent) {
-        element = createElement(LayoutComponent, { children: element });
+        element = createElement(LayoutComponent, { children: element, params: _asyncFallbackParams });
         const _tp = _treePositions ? _treePositions[i] : 0;
         const _cs = __resolveChildSegments(_routeSegs, _tp, _fallbackParams);
         element = createElement(LayoutSegmentProvider, { childSegments: _cs }, element);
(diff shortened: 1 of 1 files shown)


james-elicx:
closing in favour of https://github.com/cloudflare/vinext/pull/348

PR #373, opened 2026-03-09
perf: load instrumentation without blocking isolate initialization

james-elicx:


james-elicx:
/bigbonk so what's the verdict of the review for this pr?

PR #384, opened 2026-03-09
fix: move optional dependencies to peerDependencies

hyoban:
We are currently installing these dependencies on the user side.


hyoban:
Version is updated for `react-server-dom-webpack`. Since these two dependencies are indeed optional for the pages router, I don't plan to change them.

PR #399, opened 2026-03-10
feat: support generateSitemaps() for paginated sitemaps

### LINKED ISSUE #397: Feature Request: Support for `generateSitemaps()`
### Description

  vinext currently supports basic `sitemap.ts` metadata file routes (single `<urlset>` per route segment), but does not support Next.js's [`generateSitemaps()`](https://nextjs.org/docs/app/api-reference/functions/generate-sitemaps) API for
  paginated sitemaps.

  ### Current Behavior

  - `sitemap.ts` default export is called with **no arguments** (`await metaFn()`)
  - Each `sitemap.ts` maps to a single fixed URL (e.g., `/photos/sitemap.xml`)
  - `sitemapToXml()` only produces `<urlset>` — no `<sitemapindex>` serialization exists
  - The `generateSitemaps()` named export is ignored if present

  ### Expected Behavior

  Per the [Next.js API](https://nextjs.org/docs/app/api-reference/functions/generate-sitemaps):

  1. If a `sitemap.ts` exports `generateSitemaps()`, call it to get a list of `{ id }` objects
  2. For each `id`, serve the sitemap at `/{segment}/sitemap/{id}.xml` by calling `sitemap({ id })`
  3. Serve a `<sitemapindex>` at `/{segment}/sitemap.xml` listing all paginated sitemap URLs — this requires a new serializer alongside `sitemapToXml()` that produces `<sitemapindex>` with `<sitemap><loc>` entries

  This is required for collections that exceed the 50,000 URL limit per sitemap file (per the [sitemaps protocol](https://www.sitemaps.org/protocol.html)).

NathanDrake2406:
## Summary

- Adds support for Next.js's [`generateSitemaps()`](https://nextjs.org/docs/app/api-reference/functions/generate-sitemaps) API for paginated sitemaps
- When a `sitemap.ts` exports `generateSitemaps()`, individual sitemaps are served at `/{segment}/sitemap/{id}.xml`
- The base URL (`/{segment}/sitemap.xml`) is no longer served when `generateSitemaps` exists (matches Next.js behavior)
- Invalid IDs return 404 after validation against the `generateSitemaps()` result

## How it works

The runtime metadata route handler in the generated RSC entry now checks if a sitemap module exports `generateSitemaps`. If so:

1. **URL matching**: Instead of matching the exact `servedUrl` (e.g. `/products/sitemap.xml`), it matches the pattern `/{prefix}/sitemap/{id}.xml`
2. **ID validation**: Calls `generateSitemaps()` and checks the requested ID exists in the returned array
3. **Rendering**: Calls the default export with `{ id: rawId }` and serializes with `sitemapToXml()`

Each paginated sitemap is a standalone `<urlset>` document. Next.js does not generate a `<sitemapindex>` — the issue description was incorrect about that part.

**Note on the issue**: #397 requested `<sitemapindex>` support, but Next.js doesn't implement that. Users who want a sitemap index can create a separate root `sitemap.ts` that manually lists the child sitemap URLs.

Closes #397

## Test plan

- [x] `GET /products/sitemap/0.xml` → 200, XML with batch-0 entries only
- [x] `GET /products/sitemap/1.xml` → 200, XML with batch-1 entries only
- [x] `GET /products/sitemap/99.xml` → 404 (invalid ID)
- [x] `GET /products/sitemap.xml` → 404 (base URL not served with generateSitemaps)
- [x] `scanMetadataFiles` discovers nested `products/sitemap.ts`
- [x] Existing sitemap tests still pass (root `/sitemap.xml` without generateSitemaps)
- [x] All 214 app-router tests pass
- [x] Entry template snapshots updated
- [x] CI: format, lint, typecheck, vitest, playwright

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -1532,7 +1532,11 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
     // generateSitemaps() support — paginated sitemaps at /{prefix}/sitemap/{id}.xml
     // When a sitemap module exports generateSitemaps, the base URL (e.g. /products/sitemap.xml)
     // is no longer served. Instead, individual sitemaps are served at /products/sitemap/{id}.xml.
-    if (metaRoute.type === "sitemap" && metaRoute.isDynamic && typeof metaRoute.module.generateSitemaps === "function") {
+    if (
+      metaRoute.type === "sitemap" &&
+      metaRoute.isDynamic &&
+      typeof metaRoute.module.generateSitemaps === "function"
+    ) {
       const sitemapPrefix = metaRoute.servedUrl.slice(0, -4); // strip ".xml"
       // Match exactly /{prefix}/{id}.xml — one segment only (no slashes in id)
       if (cleanPathname.startsWith(sitemapPrefix + "/") && cleanPathname.endsWith(".xml")) {
@@ -1541,7 +1545,9 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
         const sitemaps = await metaRoute.module.generateSitemaps();
         const matched = sitemaps.find(function(s) { return String(s.id) === rawId; });
         if (!matched) return new Response("Not Found", { status: 404 });
-        // Pass the original typed id from generateSitemaps() so numeric IDs stay numeric
+        // Pass the original typed id from generateSitemaps() so numeric IDs stay numeric.
+        // TODO: wrap with makeThenableParams-style Promise when upgrading to Next.js 16
+        // full-Promise param semantics (id becomes Promise<string> in v16).
         const result = await metaRoute.module.default({ id: matched.id });
         if (result instanceof Response) return result;
         return new Response(sitemapToXml(result), {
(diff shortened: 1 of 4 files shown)


NathanDrake2406:
**Open question for @southpolesteve @james-elicx:**

Currently this PR matches Next.js behavior exactly — when `generateSitemaps()` exists, the base URL (e.g. `/products/sitemap.xml`) returns 404 and only `/products/sitemap/{id}.xml` URLs are served.

Would it be worth also serving a `<sitemapindex>` at the base URL? This would list all paginated sitemap URLs so search engines can auto-discover them:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
  <sitemap><loc>https://example.com/products/sitemap/0.xml</loc></sitemap>
  <sitemap><loc>https://example.com/products/sitemap/1.xml</loc></sitemap>
</sitemapindex>
```

**Pros:** Search engines auto-discover all child sitemaps without users manually maintaining a root sitemap. The URL would otherwise 404, so it's additive and non-breaking.

**Cons:** Behavioral divergence from Next.js. Users porting Next.js apps would get an extra response they didn't have before.

Happy to add it in this PR or as a follow-up if you think it's worth the divergence.

PR #404, opened 2026-03-10
perf: eliminate page probe double-execution of page components

james-elicx:
The RSC entry had a "page probe" that called PageComponent() outside React's render cycle before every RSC render, solely to detect redirect()/notFound()/forbidden()/unauthorized() throws. The probe's result was discarded — only thrown errors mattered. Then the actual RSC render called the same component again, executing all data fetching and computation a second time.

For async server components (the common case — pages that fetch data), this meant every fetch, database query, and API call ran twice per request. On Cloudflare Workers, where there's no persistent process cache to deduplicate fetches, this doubled CPU time for data-heavy pages.

Investigation of Next.js source confirmed they do NOT use a probe. They handle redirect/notFound entirely during rendering: errors inside Suspense boundaries are caught by React error boundaries (rendered in-stream with status 200), while errors that escape all boundaries are caught by an outer try/catch before any bytes are flushed.

Replace the page probe with three strategies based on page type:

1. Routes WITH loading.tsx (Suspense): stream the RSC response directly. Errors inside the Suspense boundary flow through the RSC stream as error references for client-side handling, matching Next.js behavior.

2. SYNC pages (no loading.tsx): a lightweight sync-only probe catches redirect/notFound thrown synchronously by the page function. This is practically free — sync server components don't do data fetching. Errors from async children inside inline <Suspense> boundaries flow through the stream as before.

3. ASYNC pages (no loading.tsx): buffer the entire RSC stream, then check onError for redirect/notFound. Components execute only once during the actual RSC render. Latency characteristics are identical because without Suspense, nothing can stream until all async work completes anyway.

The layout probe is preserved unchanged — it's needed for correct parent-boundary selection when a layout throws notFound() (knowing which layout threw determines which not-found.tsx boundary to render).

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -2177,46 +2177,6 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
     if (_layoutProbeResult instanceof Response) return _layoutProbeResult;
   }
 
-  // Pre-render the page component to catch redirect()/notFound() thrown synchronously.
-  // Server Components are just functions — we can call PageComponent directly to detect
-  // these special throws before starting the RSC stream.
-  //
-  // For routes with a loading.tsx Suspense boundary, we skip awaiting async components.
-  // The Suspense boundary + rscOnError will handle redirect/notFound thrown during
-  // streaming, and blocking here would defeat streaming (the slow component's delay
-  // would be hit before the RSC stream even starts).
-  //
-  // Because this calls the component outside React's render cycle, hooks like use()
-  // trigger "Invalid hook call" console.error in dev. The module-level ALS patch
-  // suppresses the warning only within this request's execution context.
-  const _hasLoadingBoundary = !!(route.loading && route.loading.default);
-  const _pageProbeResult = await _suppressHookWarningAls.run(true, async () => {
-    try {
-      const testResult = PageComponent({ params });
-      // If it's a promise (async component), only await if there's no loading boundary.
-      // With a loading boundary, the Suspense streaming pipeline handles async resolution
-      // and any redirect/notFound errors via rscOnError.
-      if (testResult && typeof testResult === "object" && typeof testResult.then === "function") {
-        if (!_hasLoadingBoundary) {
-          await testResult;
-        } else {
-          // Suppress unhandled promise rejection — with a loading boundary,
-          // redirect/notFound errors are handled by rscOnError during streaming.
-          testResult.catch(() => {});
-        }
-      }
-    } catch (preRenderErr) {
-      const specialResponse = await handleRenderError(preRenderErr);
-      if (specialResponse) return specialResponse;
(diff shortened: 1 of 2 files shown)


packages/vinext/src/entries/app-rsc-entry.ts
```diff
@@ -2177,63 +2177,109 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
...
+  //    is discarded; only the thrown error matters.
+  //
+  // 3. ASYNC pages (no loading.tsx): buffer the entire RSC stream to catch
+  //    redirect/notFound thrown after await (e.g. fetch then notFound).
+  //    This is the key optimization — the old "page probe" awaited the full
+  //    async component, causing every data fetch to run twice. Buffering the
+  //    stream instead means components execute only once. Latency is the same
+  //    because without Suspense, nothing can stream until all async work is done.
+  const _hasLoadingBoundary = !!(route.loading && route.loading.default);
+  const _isAsyncPage = PageComponent.constructor?.name === "AsyncFunction";
```
james-elicx, line 2235: Checking `constructor.name === 'AsyncFunction'` is fragile. A regular function that returns a Promise (e.g. an HOC wrapper or `function Page() { return fetch().then(...) }`) will be misclassified as sync. The old probe caught these via `typeof testResult.then === 'function'`. Because the new sync probe drops the return value, a Promise rejection carrying `redirect()` or `notFound()` will become an unhandled rejection and will never be intercepted for non-Suspense routes, causing full-page loads to receive a 200 with an embedded error reference instead of the correct redirect/404 response. Consider also probing the return type, or falling back to stream buffering for any page that yields a thenable.

packages/vinext/src/entries/app-rsc-entry.ts
```diff
@@ -2177,63 +2177,109 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
...
+    // (during RSC rendering), eliminating the double data fetching that
+    // the old page probe caused.
+    const _rscReader = rscStream.getReader();
+    const _chunks = [];
+    while (true) {
+      const { value, done } = await _rscReader.read();
+      if (value) _chunks.push(value);
+      if (done) break;
+    }
+
```
james-elicx, line 2269: This loop has no try/catch around `await _rscReader.read()`. If React fatally errors the stream (a scenario that bypasses `onError` and tears down the readable stream), `read()` will reject and the unhandled exception will crash the request. Wrap the read loop in a try/catch so you can propagate or log the fatal error cleanly.

james-elicx (commented):
This is a well-structured optimization that eliminates the double-execution penalty for async server components. Two correctness concerns: (1) the `AsyncFunction` constructor-name check misses regular functions that return Promises, so special-error throws from those pages bypass interception for non-Suspense routes; (2) the RSC-stream buffering loop has no error handling around `reader.read()`, which can reject if React fatally errors the stream.

PR #419, opened 2026-03-10
fix(isr): namespace cache keys by buildId

### LINKED ISSUE #415: isr: namespace cache keys by buildId or deploy version
## Problem

ISR cache keys are currently formatted as `<router>:<pathname>` (e.g., `pages:/about`). When a new version of the app is deployed, the old cached entries may be served because the cache key hasn't changed. This can cause stale content from a previous build to be served to users.

## Proposed Solution

Include `buildId` in the ISR cache key format: `<router>:<buildId>:<pathname>`.

The `buildId` is already available at runtime via `process.env.__VINEXT_BUILD_ID` (injected by the Vite define plugin). The change is to pass it through to `isrCacheKey()` and include it in the generated key.

This ensures that each deployment gets its own cache namespace, preventing stale cache reuse across deploys.

## Acceptance Criteria

- [ ] `isrCacheKey()` accepts optional `buildId` parameter
- [ ] When `buildId` is provided, key format is `<router>:<buildId>:<pathname>`
- [ ] Without `buildId`, format is unchanged (backward compat)
- [ ] Long paths with `buildId` still hash correctly when exceeding length limit
- [ ] All callers updated: `dev-server.ts`, `pages-server-entry.ts`
- [ ] Tests cover all cases

Divkix:
## Summary
- Include `buildId` in ISR cache key generation to prevent stale cache reuse across deploys
- Key format changes from `<router>:<pathname>` to `<router>:<buildId>:<pathname>` when buildId is available
- Backward compatible: without buildId, key format is unchanged
- Updated all callers: `dev-server.ts`, `pages-server-entry.ts`

Closes #415

## Test plan
- [ ] Unit tests for `isrCacheKey()` with and without buildId
- [ ] Long path hashing still works with buildId prefix
- [ ] Backward compatibility: no buildId = old format
- [ ] Typecheck, lint, format pass

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -341,7 +341,7 @@ function __triggerBackgroundRegeneration(key, renderFn, ctx) {
 // even after adding the build ID and suffix. FNV-1a 64 is used for the hash (two
 // 32-bit rounds) to give a ~64-bit output with negligible collision probability for
 // realistic pathname lengths.
-// Keep in sync with isrCacheKey() in server/isr-cache.ts.
+// Keep prefix construction and hashing logic in sync with isrCacheKey() in server/isr-cache.ts.
 function __isrFnv1a64(s) {
   // h1 uses the standard FNV-1a 32-bit offset basis (0x811c9dc5).
   let h1 = 0x811c9dc5;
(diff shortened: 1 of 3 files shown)


james-elicx:
Funnily enough I was about to change this in my other pr as part of fixing ISR 😅

PR #436, opened 2026-03-11
perf: startup and cache micro-optimizations

Divkix:
Closes #437

## Summary

- **Cache `hasMdxFiles()` result** per root directory — avoids redundant recursive filesystem walks when `config()` fires per Vite environment (RSC/SSR/Client = 3+ times)
- **Cache `resolvePostcssStringPlugins()` result** per project root — skips repeated `existsSync` checks across 17 PostCSS config file candidates on each `config()` invocation
- **Replace byte-by-byte base64 encode/decode** in KV cache handler with `Buffer` APIs for significantly faster serialization
- **Pass pre-normalized URL to `nodeToWebRequest()`** in App Router prod server — eliminates redundant `normalizePath` call in the RSC handler for pathological URLs

## Test plan

- [x] 9 new tests in `tests/startup-cache.test.ts` — MDX scan caching (5) and PostCSS config caching (4)
- [x] 3 new edge-case tests in `tests/kv-cache-handler.test.ts` — 1 MiB buffer, null bytes, all 256 byte values
- [x] 5 new tests in `tests/node-to-web-request.test.ts` — `urlOverride` parameter behavior
- [x] All 33 kv-cache-handler tests pass
- [x] Typecheck (`tsgo --noEmit`) clean
- [x] Lint (`oxlint`) clean
- [x] Format (`oxfmt --check`) clean

PR #450, opened 2026-03-11
refactor: phase 1/2 unify per-request ALS into a shared request context

### LINKED ISSUE #451: Phase 1/2: Consolidate per-request ALS scopes and land initial Pages Router parity
## Problem

Every vinext request path historically relied on multiple nested `AsyncLocalStorage.run()` calls from different shim modules. That adds measurable overhead on Workers and makes request-scoped state harder to reason about.

## Solution

Consolidate request-scoped state into a single `UnifiedRequestContext` backed by one ALS instance. Individual shims read from the unified store when inside the unified scope and fall back to their standalone ALS outside it for compatibility.

### Phase 1 (this issue / PR #450)

- App Router request handling + ISR background regeneration
- Shared unified context infrastructure across `headers`, `navigation-state`, `cache`, `cache-runtime`, `fetch-cache`, and `request-context`
- Pages Router dev SSR now runs inside the unified scope
- Generated Pages Router server entry now runs inside the unified scope
- `router-state.ts` and `head-state.ts` read from unified context when inside the scope
- Pages Router ISR background regeneration is wrapped in unified context for parity
- Dual-path fallbacks remain in place for compatibility and staged rollout safety

### Deferred to Phase 2 (#478)

`#478` is an umbrella follow-up issue and may span more than one PR.

The first follow-up PR should focus on:

- Pages Router integration/parity tests for concurrent router/head isolation in dev and prod
- Final validation of the Pages Router production path, which currently delegates through the unified `renderPage()` entry
- Documenting the decision for per-call ALS scopes: `cacheContextStorage` and `_unstableCacheAls`
- ALS architecture documentation

Fallback cleanup is conditional after that validation work. Standalone fallback ALS code should only be removed where it is clearly unnecessary; otherwise it should be kept with rationale.

## Related

- PR: #450
- Phase 2/2 umbrella follow-up: #478

Divkix:
## Summary

- Introduces a single `UnifiedRequestContext` backed by one `AsyncLocalStorage` instance
- Replaces the 5-6 nested ALS scopes used by App Router requests with one unified per-request scope
- Lands the initial Pages Router rollout needed for parity: dev SSR, generated Pages server entry, unified-aware router/head shims, and Pages ISR background regeneration

**Depends on:** #432, #433, #434, #435, #436, #442

Closes #451
Follow-up tracked in #478

## Included In This PR

| File / area | Change |
|------|--------|
| `shims/unified-request-context.ts` | **NEW** unified ALS module |
| `shims/headers.ts` | Dual-path `_getState()`, unified-aware wrapper semantics |
| `shims/navigation-state.ts` | Dual-path `_getState()`, unified-aware wrapper semantics |
| `shims/cache.ts` | Unified-aware request-scoped cache state |
| `shims/cache-runtime.ts` | Unified-aware private cache state |
| `shims/fetch-cache.ts` | Unified-aware tag collection + `ensureFetchPatch()` |
| `shims/request-context.ts` | Unified-aware execution context access |
| `shims/router-state.ts` | Pages Router SSR state reads from unified scope when present |
| `shims/head-state.ts` | Pages Router head state reads from unified scope when present |
| `server/dev-server.ts` | Pages Router dev SSR wrapped in unified scope |
| `entries/pages-server-entry.ts` | Generated Pages Router server entry wrapped in unified scope |
| `entries/app-rsc-entry.ts` | App Router handler + ISR regen wrapped in unified scope |
| `tests/unified-request-context.test.ts` | New unit coverage for unified ALS behavior |

## Deferred To #478

`#478` is an umbrella follow-up issue and may span more than one PR.

The first follow-up PR should focus on:

- Pages Router integration parity tests for concurrent router/head isolation in dev and prod
- Validation of the production Pages Router path, which currently delegates through unified `renderPage()`
- Documentation of the per-call ALS decision for `cacheContextStorage` and `_unstableCacheAls`
- ALS architecture documentation

Fallback cleanup remains conditional and should only happen if those tests/validation show the standalone paths are truly unnecessary.

## What Is Intentionally Still Separate

- `cacheContextStorage` in `cache-runtime.ts`: per-cache-call ALS, not per-request
- `_unstableCacheAls` in `cache.ts`: per-call scope for `unstable_cache()`
- Standalone fallback ALS paths in shims: kept for compatibility and staged rollout safety until the phase-2 validation work is complete

## Test Plan

- [x] Unified request-context unit tests
- [x] Existing shim/request-context coverage
- [x] App Router + ISR + fetch-cache + features coverage
- [x] Routing + nextjs-compat coverage
- [x] Typecheck, lint, format
- [x] CI full suite + Playwright E2E

### DIFF
--- a/packages/vinext/src/entries/pages-server-entry.ts
+++ b/packages/vinext/src/entries/pages-server-entry.ts
@@ -267,8 +267,9 @@ import { runWithPrivateCache } from "vinext/cache-runtime";
 import { ensureFetchPatch, runWithFetchCache } from "vinext/fetch-cache";
 import { runWithRequestContext as _runWithUnifiedCtx, createRequestContext as _createUnifiedCtx } from "vinext/unified-request-context";
 import "vinext/router-state";
+import { runWithServerInsertedHTMLState } from "vinext/navigation-state";
 import { runWithHeadState } from "vinext/head-state";
-import { runWithI18nState } from "vinext/i18n-state";
+import "vinext/i18n-state";
 import { setI18nContext } from "vinext/i18n-context";
 import { safeJsonStringify } from "vinext/html";
 import { decode as decodeQueryString } from "node:querystring";
@@ -368,9 +369,13 @@ async function renderIsrPassToStringAsync(element) {
   // The cache-fill render is a second render pass for the same request.
   // Reset render-scoped state so it cannot leak from the streamed response
   // render or affect async work that is still draining from that stream.
-  return await runWithHeadState(() =>
-    _runWithCacheState(() =>
-      runWithPrivateCache(() => runWithFetchCache(async () => renderToStringAsync(element))),
+  // Keep request identity state (pathname/query/locale/executionContext)
+  // intact: this second pass still belongs to the same request.
+  return await runWithServerInsertedHTMLState(() =>
+    runWithHeadState(() =>
+      _runWithCacheState(() =>
+        runWithPrivateCache(() => runWithFetchCache(async () => renderToStringAsync(element))),
+      ),
     ),
   );
 }
@@ -735,10 +740,9 @@ async function _renderPage(request, url, manifest) {
   const __uCtx = _createUnifiedCtx({
     executionContext: _getRequestExecutionContext(),
   });
-  return runWithI18nState(() =>
-    _runWithUnifiedCtx(__uCtx, async () => {
-      ensureFetchPatch();
-  try {
+  return _runWithUnifiedCtx(__uCtx, async () => {
+    ensureFetchPatch();
+    try {
     if (typeof setSSRContext === "function") {
       setSSRContext({
         pathname: patternToNextFormat(route.pattern),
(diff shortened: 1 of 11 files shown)


Divkix:
This PR’s goal is to collapse the per-request ALS layers on the main request/render paths, not to remove every standalone store immediately.

We still need the fallback path today because there are real call-sites that run outside a unified request scope. The main ones are non-request/build flows like static export, plus the public shim helpers that can still be called outside a wrapped request context. So removing the standalone ALS/fallbacks in this PR would be a behavior change, not just cleanup.

On `ensureFetchPatch()`: we could hide that behind a higher-level helper, but I kept it explicit because fetch patching is process-global setup, while `runWithRequestContext()` is only about request-local state. Keeping them separate avoids coupling the unified context module to fetch-cache, and the current request entry points are few enough that the explicit call is manageable.

For coverage: the main request render entry points are wrapped now (App Router prod/dev, Pages render, and the App Router revalidation path), so the consolidated ALS is what server-side request handling uses in normal operation. It’s not yet true that literally every server-side path in the repo is inside the unified scope though, which is why the fallback still exists.

If we want to go further, I think that’s a reasonable follow-up: wrap the remaining non-unified entry points and then remove fallback stores once there are no supported callers left outside the unified scope.

Divkix:
## Scope Update

Phase 1 originally stated "App Router only" but review feedback identified that Pages Router ISR regeneration was non-functional without unified context. Added minimal ISR fix to ensure parity:

- Pages Router ISR regeneration now wrapped in unified context (commit b8c2c12)
- Ensures patched fetch and ALS context during background regen
- Full Pages Router consolidation still deferred to Phase 2 (#478)

This fix is surgical (only touches the ISR regeneration callback, ~15 lines) and doesn't overlap with Phase 2 work.

PR #482, opened 2026-03-12
fix: add next shim js variants

hyoban:
It looks like `resolveId` does not work for `optimizeDeps`


james-elicx:
Thanks, I was going to do this today so you've saved me some time!

PR #490, opened 2026-03-12
docs: add pre-commit checklist to AGENTS.md

NathanDrake2406:
@james-elicx , I've wasted so much time updating this dumb ass snapshot and failing CI due to formatting, let's just add this so the bot runs it automatically every time, it doesn't hurt.


NathanDrake2406:
> We could add a precommit hook for fmt / lint / entry-template snapshots if you want as well?

I didn't know that you could add a precommit hook in git lol. All good I just set it up

PR #493, opened 2026-03-12
fix: register reportRequestError with ctx.waitUntil on Workers

NathanDrake2406:
## Summary

Follow-up to #492. On Cloudflare Workers, the isolate can terminate after the Response is sent, killing in-flight error reports (e.g. Sentry HTTP requests) that were fire-and-forget.

- `reportRequestError()` now self-registers with `ctx.waitUntil()` via the ALS-backed `getRequestExecutionContext()`, matching the existing pattern used by `KVCacheHandler`, ISR regeneration, and fetch cache revalidation
- On Node.js (dev or `vinext start`), `getRequestExecutionContext()` returns `null` — fire-and-forget remains correct since the process doesn't die
- Since `reportRequestError` never rejects (internal try/catch), the redundant `.catch()` at all 5 call sites across both routers is removed

## Test plan

- [x] New test: `registers the report promise with ctx.waitUntil on Workers` — verifies `waitUntil` is called with the report promise inside `runWithExecutionContext`
- [x] New test: `does not call waitUntil when no execution context is available` — verifies no-op on Node.js
- [x] All 13 instrumentation tests pass
- [x] `pages-router.test.ts` passes (166 tests)
- [x] `entry-templates.test.ts` passes (12 tests)
- [x] `api-handler.test.ts` passes (40 tests)
- [ ] CI: full Vitest + Playwright E2E suite

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -535,9 +535,7 @@ function rscOnError(error, requestInfo, errorContext) {
       error instanceof Error ? error : new Error(String(error)),
       requestInfo,
       errorContext,
-    ).catch((reportErr) => {
-      console.error("[vinext] Failed to report render error:", reportErr);
-    });
+    );
   }
 
   // In production, generate a digest hash for non-navigation errors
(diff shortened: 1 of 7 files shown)


NathanDrake2406:
@james-elicx Any big fat tasks that requires lots of tokens? Mine reset tomorrow I've got so much left :(

PR #497, opened 2026-03-12
fix(route-handlers): align Allow and default-export behavior with Next.js

Dayifour:
# Summary
Align route handler behavior with Next.js for HTTP method handling.

Keep implicit OPTIONS auto-response (204) and sort its Allow header deterministically.

Stop sending Allow on 405 Method Not Allowed responses (Next.js parity).

Reject default export in route handlers (only named HTTP method exports are valid).

Add a dedicated invalid fixture and update integration/e2e tests accordingly.

Refresh ON-3 tracking note and generated entry snapshots.

Testing

`# Unit/Integration tests
pnpm test tests/entry-templates.test.ts
pnpm test app-router.test.ts -t "errors when a route handler exports a default function"`

`# E2E tests (Windows/PowerShell syntax)
$env:PLAYWRIGHT_PROJECT='app-router'; pnpm run test:e2e tests/e2e/app-router/api-routes.`

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -60,11 +60,9 @@ function collectRouteHandlerMethods(handler) {
   return methods;
 }
 
-function buildRouteHandlerAllowHeader(exportedMethods, includeImplicitOptions) {
+function buildRouteHandlerAllowHeader(exportedMethods) {
   const allow = new Set(exportedMethods);
-  if (includeImplicitOptions) {
-    allow.add("OPTIONS");
-  }
+  allow.add("OPTIONS");
   return Array.from(allow).sort().join(", ");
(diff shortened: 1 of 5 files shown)


NathanDrake2406:
> On a side note, I’m curious... how did you (or your agent) run those comparisons against the Next.js source so quickly? The table is super helpful and I'd love to use that workflow for future PRs to ensure I'm aligned from the start.


Use Context7. It's pretty much free

Dayifour:
Thanks @NathanDrake2406 @james-elicx for the tips! Context7 and OpenCode sound awesome, I'll definitely check them out to better align my future PRs.

Just pushed the fixes to address the bot's feedback: I've switched the default export handling to a 405 + dev warning to match Next.js. I've also updated the tests to match.

Let me know if it looks better now!

PR #498, opened 2026-03-12
fix: dedup stale background refetches in fetch cache

NathanDrake2406:
## Summary

- Every concurrent request hitting a stale fetch cache entry was independently calling `originalFetch()`, creating a thundering herd on popular endpoints
- Added a per-cache-key dedup map (`pendingRefetches`) shared across RSC/SSR environments via `Symbol.for()`, matching the existing ISR regeneration dedup pattern in `isr-cache.ts`
- Added a 60s timeout safety net that force-cleans dedup entries when upstream fetches hang, with identity checks in both `.finally()` and `setTimeout` to prevent slot-ownership races
- Improved error log to include URL and truncated cache key for production observability

## Test plan

- [x] Concurrent stale hits trigger only one background refetch (deferred promise gate ensures deterministic timing)
- [x] Completed refetches allow new ones (lifecycle test)
- [x] Failed refetches clean up dedup entry, allowing retry (error-path test)
- [x] Hung fetches are force-cleaned after 60s timeout (`vi.useFakeTimers`)
- [x] Late-settling hung fetch does not evict a replacement refetch's slot (race test)
- [x] `_resetPendingRefetches()` called in `beforeEach` for test isolation
- [ ] CI: full Vitest suite + Playwright E2E

### DIFF
--- a/packages/vinext/src/shims/fetch-cache.ts
+++ b/packages/vinext/src/shims/fetch-cache.ts
@@ -667,6 +667,10 @@ function createPatchedFetch(): typeof globalThis.fetch {
             if (pendingRefetches.get(cacheKey) === refetchPromise) {
               pendingRefetches.delete(cacheKey);
             }
+          const timeoutId = setTimeout(() => {
+            if (pendingRefetches.get(cacheKey) === refetchPromise) {
+              pendingRefetches.delete(cacheKey);
+            }
           }, DEDUP_TIMEOUT_MS);
 
           getRequestExecutionContext()?.waitUntil(refetchPromise);
--- a/packages/vinext/src/shims/fetch-cache.ts
+++ b/packages/vinext/src/shims/fetch-cache.ts
@@ -610,6 +610,10 @@ function createPatchedFetch(): typeof globalThis.fetch {
           const cleanInit = stripNextFromInit(init);
           const refetchPromise = originalFetch(input, cleanInit)
             .then(async (freshResp) => {
+              // Only cache successful responses — a transient 500 must not
+              // overwrite previously-good cached data.
+              if (!freshResp.ok) return;
+
               const freshBody = await freshResp.text();
               const freshHeaders: Record<string, string> = {};
               freshResp.headers.forEach((v, k) => {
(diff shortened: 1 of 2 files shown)


packages/vinext/src/shims/fetch-cache.ts
```diff
@@ -581,44 +604,78 @@ function createPatchedFetch(): typeof globalThis.fetch {
...
-              freshHeaders[k] = v;
+        // Background refetch — deduped so only one in-flight refetch runs
+        // per cache key, preventing thundering herd on popular endpoints.
+        if (!pendingRefetches.has(cacheKey)) {
+          const cleanInit = stripNextFromInit(init);
+          const refetchPromise = originalFetch(input, cleanInit)
+            .then(async (freshResp) => {
+              // Only cache successful responses — a transient 500 must not
+              // overwrite previously-good cached data.
+              if (!freshResp.ok) return;
```
james-elicx:
```suggestion
              if (freshResp.status !== 200) return;
```

i think we should tighten this a bit more

james-elicx: (similarly for the other path where it checks status)

packages/vinext/src/shims/fetch-cache.ts
```diff
@@ -581,44 +604,78 @@ function createPatchedFetch(): typeof globalThis.fetch {
...
-              freshHeaders[k] = v;
+        // Background refetch — deduped so only one in-flight refetch runs
+        // per cache key, preventing thundering herd on popular endpoints.
+        if (!pendingRefetches.has(cacheKey)) {
+          const cleanInit = stripNextFromInit(init);
+          const refetchPromise = originalFetch(input, cleanInit)
+            .then(async (freshResp) => {
+              // Only cache 200 responses — a transient error or unexpected
+              // status must not overwrite previously-good cached data.
+              if (freshResp.status !== 200) return;
```
NathanDrake2406, line 615: Good call @james-elicx — tightened both paths in b4840c0. Background revalidation guard and cache-miss path now use `status === 200` instead of `response.ok`.

Also noticed that `isr-cache.ts:69` has the same cross-environment sharing problem — `pendingRegenerations` is a plain module-level `Map` instead of using `Symbol.for()` on `globalThis`. RSC and SSR get separate copies, so dedup doesn't work across environments. Will tackle that in a separate PR.

james-elicx (approved):

PR #503, opened 2026-03-12
fix: MemoryCacheHandler revalidate:0 creates immediately-stale entries

NathanDrake2406:
## Summary

- `MemoryCacheHandler.set` had two code paths for computing `revalidateAt`. The `ctx` path guarded with `revalidate > 0`, but the `data` path only checked `typeof data.revalidate === "number"` — letting `revalidate: 0` set `revalidateAt = Date.now()`, creating immediately-stale entries.
- This was inconsistent with `KVCacheHandler` which already had the `> 0` guard on both paths, meaning the two handlers produced different staleness for the same input (Memory → immediately stale, KV → no TTL).
- Adds the missing `> 0` guard to `MemoryCacheHandler.set` to align both code paths and both cache handler implementations.

## Test plan

- [x] Added test: `data.revalidate:0` does not create immediately-stale entries
- [x] Added test: `data.revalidate:0` behaves consistently with `ctx.revalidate:0`
- [x] Existing ISR cache tests pass (stale entries, fresh entries, tag invalidation)
- [x] Full `features.test.ts` suite passes (252 tests)
- [x] All cache-related tests pass (`shims`, `fetch-cache`, `isr-cache`, `kv-cache-handler` — 1063 tests)
- [ ] CI: Format, Lint, Typecheck, Vitest, Playwright E2E

### DIFF
--- a/packages/vinext/src/cloudflare/kv-cache-handler.ts
+++ b/packages/vinext/src/cloudflare/kv-cache-handler.ts
@@ -238,22 +238,24 @@ export class KVCacheHandler implements CacheHandler {
     }
     const tags = [...tagSet];
 
-    // Determine revalidation time
-    let revalidateAt: number | null = null;
+    // Resolve effective revalidate — data overrides ctx.
+    // revalidate: 0 means "don't cache", so skip storage entirely.
+    let effectiveRevalidate: number | undefined;
     if (ctx) {
       const revalidate = (ctx as any).cacheControl?.revalidate ?? (ctx as any).revalidate;
-      if (typeof revalidate === "number" && revalidate > 0) {
-        revalidateAt = Date.now() + revalidate * 1000;
+      if (typeof revalidate === "number") {
+        effectiveRevalidate = revalidate;
       }
     }
-    if (
-      data &&
-      "revalidate" in data &&
-      typeof data.revalidate === "number" &&
-      data.revalidate > 0
-    ) {
-      revalidateAt = Date.now() + data.revalidate * 1000;
+    if (data && "revalidate" in data && typeof data.revalidate === "number") {
+      effectiveRevalidate = data.revalidate;
     }
+    if (effectiveRevalidate === 0) return Promise.resolve();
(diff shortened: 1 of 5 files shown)


packages/vinext/src/cloudflare/kv-cache-handler.ts
```diff
@@ -238,22 +238,24 @@ export class KVCacheHandler implements CacheHandler {
...
-      data.revalidate > 0
-    ) {
-      revalidateAt = Date.now() + data.revalidate * 1000;
+    if (data && "revalidate" in data && typeof data.revalidate === "number") {
+      effectiveRevalidate = data.revalidate;
     }
+    if (effectiveRevalidate === 0) return Promise.resolve();
+
+    const revalidateAt =
+      typeof effectiveRevalidate === "number" && effectiveRevalidate > 0
```
james-elicx, line 256:
```suggestion
      typeof effectiveRevalidate === "number"
```

nit: already asserted it can't be 0

packages/vinext/src/shims/cache.ts
```diff
@@ -192,17 +192,24 @@ export class MemoryCacheHandler implements CacheHandler {
...
       }
     }
     if (data && "revalidate" in data && typeof data.revalidate === "number") {
-      revalidateAt = Date.now() + data.revalidate * 1000;
+      effectiveRevalidate = data.revalidate;
     }
+    if (effectiveRevalidate === 0) return;
+
+    const revalidateAt =
+      typeof effectiveRevalidate === "number" && effectiveRevalidate > 0
```
james-elicx, line 210:
```suggestion
      typeof effectiveRevalidate === "number"
```
ditto

packages/vinext/src/cloudflare/kv-cache-handler.ts
james-elicx: oh wait is this supposed to catch negative numbers? that feels like an invalid input....

PR #527, opened 2026-03-13
fix: `use server` closure variable collision with local declarations

james-elicx:
This was observed in the wild with Payload CMS.

### Fix

Add a `"pre"`-enforced Vite plugin (`vinext:fix-use-server-closure-collision`) that runs before `plugin-rsc` sees the file. For any `"use server"` function whose body declares a variable that shadows an outer-scope name, it renames the inner declaration **and all its usages within that function body** to `__local_<name>`. The outer binding is untouched.

After the rename, `periscopic` no longer sees `cookies` referenced inside the action, so it is not added to `bindVars` at all. Only genuinely-closed-over variables (like `config`, which is not redeclared inside the action) remain.

### Tests

- `tests/use-server-closure-collision.test.ts` — unit tests that exercise the plugin transform directly:
  - Reproduces the bug by calling `transformHoistInlineDirective` on raw source and confirming the duplicate declaration
  - Verifies the fix eliminates the collision and produces correct output
  - Regression guard for the non-colliding case

### DIFF
--- a/tests/use-server-closure-collision.test.ts
+++ b/tests/use-server-closure-collision.test.ts
@@ -12,11 +12,11 @@
  * and parses correctly.
  */
 
-import { describe, it, expect } from "vitest";
-import { parseAst } from "vite";
+import { describe, it, expect } from "vite-plus/test";
+import { parseAst } from "vite-plus";
 import { transformHoistInlineDirective } from "@vitejs/plugin-rsc/transforms";
 import vinext from "../packages/vinext/src/index.js";
-import type { Plugin } from "vite";
+import type { Plugin } from "vite-plus";
 
 // ---------------------------------------------------------------------------
 // Helpers


james-elicx:
> 1k lines approved first pass 😮 Do you do the same loop locally before bonk? Like plan -> implement -> review -> re-implement -> review ... til the LLM approve locally?

With these kinds of pieces of work, I'm very involved and treat it more like pair programming where I frequently guide what the agent is doing, and will often interrupt the agent if it veers off-course from where I think it should be going.

So I get very heavily involved in getting these bigger pieces of work in the direction I want them to go. Then I review it myself and also have the agent review the diff as well, while continuously testing it with my reproduction.

Depending on what it is, I will often use the agent panel in Zed where I'm able to edit the code at the same time as the agent. It's usually small things or addressing comments that I give the agent a bit more freedom on. I generally avoid going full hands-off for things that aren't small.

PR #543, opened 2026-03-15
fix: use waitUntil in after() for Cloudflare Workers

NathanDrake2406:
## Summary

- `after()` was scheduling tasks as fire-and-forget microtasks via `Promise.resolve().then(task)`, which are silently dropped when Cloudflare Workers terminate after response completion
- Now uses `getRequestExecutionContext().waitUntil(promise)` when an execution context exists, falling back to fire-and-forget for Node.js dev server
- Added `throwIfInsideCacheScope("after()")` guard matching Next.js behavior (same pattern as `connection()`, `headers()`, `cookies()`)
- Kept `after()` synchronous (`function after(): void`) with static imports to ensure `waitUntil` is registered immediately, not on a deferred microtask

## Test plan

- [x] Test: `after()` calls `waitUntil` on execution context when one exists
- [x] Test: `after()` falls back to fire-and-forget when no execution context
- [x] Test: `after()` throws inside `"use cache"` scope
- [x] All 688 tests pass
- [x] Typecheck and lint clean


NathanDrake2406:
@james-elicx i don't get this bug :(

NathanDrake2406:
Btw could you pls explain more about the half baked static rendering implementation and what's ur approach going to be? 
I wanna learn 😃 my claude is gonna take a beating tonight

NathanDrake2406:
HAHAHAHA so it is absolutely random his name is NATHAN DRAKE? OMFG

PR #553, opened 2026-03-15
feat: production prerender pipeline

james-elicx:
## Summary

- Adds \`prerenderPages()\` and \`prerenderApp()\` in \`build/prerender.ts\` — render all static/ISR routes using **production bundles only**, no \`ViteDevServer\` dependency at prerender time
- \`runPrerender()\` in \`build/run-prerender.ts\` consolidates route scanning, progress reporting, and manifest writing for both App and Pages Router (hybrid projects included); the old \`runPrerenderWithDevServer()\` shim is removed
- \`static-export.ts\` becomes a thin adapter — all rendering logic lives in the prerender layer
- Prerendered HTML/RSC output for non-export builds goes to \`dist/server/prerendered-routes/\` instead of \`dist/client/\`. On Cloudflare Workers, \`not_found_handling: "none"\` means every request hits the worker first so files in \`dist/client/\` are never auto-served for page requests — they were uploaded but inert. Keeping prerendered output out of \`dist/client/\` also prevents a future issue where KV pre-population would cause ISR routes to be served as stale static files forever. \`output: 'export'\` builds are unaffected (still write to \`dist/client/\`)
- \`prod-server.ts\` now serves prerendered \`.html\` files for extension-less paths (e.g. \`/about\` → \`about.html\`)
- \`NoOpCacheHandler\` added to \`shims/cache.ts\` to skip wasteful ISR writes during prerender; \`CacheHandler\` is save/restored around prerender calls for isolation
- Build report integrates \`PrerenderResult\` — routes confirmed static by speculative render are upgraded from \`unknown\` → \`static\` in the output
- Hybrid projects (both \`app/\` and \`pages/\`): after the App Router multi-env build, a separate SSR-only build produces the Pages Router bundle so the prerender phase finds both bundles
- Adds \`rscOutDir\` / \`ssrOutDir\` / \`clientOutDir\` / \`disableAppRouter\` options to \`VinextOptions\` so tests can build multiple fixtures in parallel into isolated output directories
- Miscellaneous cleanups: extract \`loadBundle()\`, rename \`isExplicitlyDynamic\` → \`isConfiguredDynamic\`, \`SetCtx\` interface, \`CacheResultWrapper\`, fix single-row table corner, unify \`onProgress\` to single call site, remove unused \`appDir\` from \`PrerenderAppOptions\`
- Support for pre-rendering workers builds via Wrangler servers.

## Tests

- 36 new tests in \`tests/prerender.test.ts\` covering \`prerenderPages\`, \`prerenderApp\`, hybrid, and export wiring
- All suites pass: **265 app-router + 172 pages-router + 36 prerender + 14 static-export = 487 tests**
- Typecheck, lint, and fmt all clean

## Future tasks
- Serving pre-rendered responses from memory cache.
- Populating remote cache during deployment.
- Configurable concurrency for pre-render pipeline.
- Dedupe RSC requests during pre-render and extract RSC payload from HTML response.
- Improve progress bar tracking for the pre-render phase to be a better reflection of the work that is happening.
- Expose a better public API for manual external control over pre-rendering and static exports.
- Improve static analysis detection classification for routes when not going through the pre-render pipeline.
- Lightweight worker and auto-config options for static exports.
- Wire up new static export fixture to actual tests and deployments once static export functionality is finished.
- Provide some information in the route report about what was pre-rendered.
- Turn on pre-rendering + KV population for the deployed example apps.
- Improve pre-rendering test suites.
- Performance improvements for resolveParentParams lookups.
- Switch unstable_dev to unstable_startWorker and add more informational logs about the server being started.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,29 +15,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint:
-    name: Lint
+  check:
+    name: Check
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: voidzero-dev/setup-vp@v1
-      - run: vp run lint
-
-  format:
-    name: Format
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: voidzero-dev/setup-vp@v1
-      - run: vp run fmt:check
-
-  typecheck:
-    name: Typecheck
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: voidzero-dev/setup-vp@v1
-      - run: vp run typecheck
+      - name: Build plugin (needed for benchmark workspace type resolution)
+        run: vp run build
+      - run: vp run check
 
   test:
     name: Vitest
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,8 +21,8 @@ pnpm test                                        # Vitest — full suite (~2 min
 pnpm test tests/routing.test.ts                  # Run a single test file (~seconds)
 pnpm test tests/shims.test.ts tests/link.test.ts # Run specific files
 pnpm run test:e2e                                # Playwright E2E tests (all projects, use PLAYWRIGHT_PROJECT=<name> to target one)
-pnpm run typecheck                               # TypeScript via tsgo (fast)
-pnpm run lint                                    # oxlint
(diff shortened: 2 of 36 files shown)


james-elicx:
Built this whole solution around the Node.js build output, so it's a bit tricky getting it to play well with the Workers output as that will need running in a workerd context. Will let the agents try and fix and hopefully have something mergeable tomorrow evening. :(

james-elicx:
/bigbonk review the cloudflare workers prerendering setup in this pr

PR #603, opened 2026-03-20
fix: include query string in route handler ISR cache key

southpolesteve:
## Summary

- Route handler ISR cache keys now include the query string, so `/api/data?user=alice` and `/api/data?user=bob` produce distinct cache entries.
- Applied to both the cache read (HIT) and cache write (MISS) paths.

## Details

Previously, `__isrRouteKey(cleanPathname)` used only the pathname for the cache key. Route handlers that return different responses based on `searchParams` would cache the first response and serve it for all subsequent requests regardless of query parameters.

The fix extracts the query string from `request.url` and appends it to the pathname before computing the cache key. Only GET/HEAD requests are affected (POST/PUT/etc. are never ISR-cached).

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -2256,8 +2256,15 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
     }
 
     // Include query string in route handler ISR cache key so different
-    // query parameters produce distinct cache entries.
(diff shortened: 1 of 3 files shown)


southpolesteve:
/bigbonk Addressed both review items: (1) sorted query params to normalize key order, (2) updated assertion and regenerated 6 snapshots. All tests pass locally. Please re-review.

PR #613, opened 2026-03-20
fix: detect dynamic API usage in route handlers to prevent stale ISR cache hits

southpolesteve:
## Summary

Route handlers that access `request.headers` or `request.nextUrl.searchParams` are dynamic and should not be served from the ISR cache. This PR adds runtime dynamic detection matching Next.js behavior.

## Changes

1. **`__proxyRouteRequest`**: Wraps the `Request` passed to route handlers in a Proxy. Accessing `.headers` or `.nextUrl.searchParams` calls `markDynamicUsage()`. Methods are bound to the original target to preserve Web API internal slots (`.json()`, `.text()`, etc.).

2. **`__dynamicRouteHandlers`**: A module-level `Set` that remembers route patterns whose handlers used dynamic APIs. The ISR cache read condition checks this set and skips the cache for known-dynamic handlers.

3. **`dynamicUsageCount`**: Replaces the boolean `dynamicUsageDetected` flag with a counter, enabling future counter-based bracketing of handler execution.

## Why not query-string-in-cache-key?

[James pointed out](https://github.com/cloudflare/vinext/pull/603#pullrequestreview-2903462834) that including query strings in the cache key diverges from Next.js, which uses pathname-only keys ([source](https://github.com/vercel/next.js/blob/9f181bd11af5f532c529a6f168fe40505f0a4d75/packages/next/src/build/templates/app-route.ts#L181-L187)). In Next.js, handlers that read searchParams are detected as dynamic and never cached at all. This PR matches that behavior.

## Proxy granularity

| Access | Dynamic? | Reason |
|---|---|---|
| `request.method` | No | Same for all ISR requests |
| `request.url` (string) | No | Can't track downstream usage |
| `request.nextUrl.pathname` | No | Pathname is the route |
| `request.nextUrl.searchParams` | **Yes** | Varies per request |
| `request.headers` | **Yes** | Varies per request |

The one gap (same as Next.js): `new URL(request.url).searchParams` bypasses the Proxy.

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -2385,7 +2385,7 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
       const previousHeadersPhase = setHeadersAccessPhase("route-handler");
       const __proxiedRequest = __proxyRouteRequest(request, markDynamicUsage);
       try {
-        const response = await handlerFn(__proxiedRequest, { params });
+        const response = await handlerFn(request, { params });
         const dynamicUsedInHandler = consumeDynamicUsage();
         const handlerSetCacheControl = response.headers.has("cache-control");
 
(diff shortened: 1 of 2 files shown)


southpolesteve:
Superseded by v3 branch with minimal diff (no counter rename).

PR #620, opened 2026-03-21
fix: prevent useActionState state becoming undefined when redirect() is called (#589)

### LINKED ISSUE #589: `useActionState` receives `undefined` state when Server Action calls `redirect()`
## Bug Description

When a Server Action used with `useActionState` calls `redirect()` (from `next/navigation`), the `state` returned by `useActionState` becomes `undefined` instead of retaining the previous state. This causes a `TypeError` on the next render if the component accesses any property on `state`.

## Steps to Reproduce

1. Create a Server Action that calls `redirect()` on success:
```ts
// app/actions.ts
'use server';
import { redirect } from 'next/navigation';
import type { ActionResult } from './types';

export async function createPost(
  _prev: ActionResult,
  formData: FormData,
): Promise<ActionResult> {
  // ... create post logic ...
  redirect(`/posts/${postId}`); // <-- triggers the bug
}
```

2. Use the action with `useActionState` in a Client Component:
```tsx
'use client';
import { useActionState } from 'react';
import { createPost } from './actions';

const initial = { success: false, error: '' };

export function PostForm() {
  const [state, formAction, pending] = useActionState(createPost, initial);

  // TypeError: Cannot read properties of undefined (reading 'success')
  // state is `undefined` after redirect() is called
  return (
    <form action={formAction}>
      {!state.success && state.error && <p>{state.error}</p>}
      <button type="submit">Submit</button>
    </form>
  );
}
```

3. Submit the form. The Server Action runs, calls `redirect()`, and the component receives `undefined` as the new state, crashing with:
```
TypeError: Cannot read properties of undefined (reading 'success')
```

## Expected Behavior

When `redirect()` is called inside a Server Action, the browser should navigate to the target URL. The `state` value from `useActionState` should either retain the previous state or remain in a defined shape — it should never become `undefined`.

This is the standard Next.js behavior: `redirect()` throws a special internal error that Next.js catches to perform the navigation, and `useActionState` is not involved in that process.

## Actual Behavior

`state` becomes `undefined` after `redirect()` is called in the Server Action, causing a `TypeError` on the next render.

## Workaround

Avoid calling `redirect()` inside Server Actions used with `useActionState`. Instead, return a success payload from the action and navigate client-side using `router.push()` in a `useEffect`:

```tsx
useEffect(() => {
  if (state.success && state.data?.postId) {
    router.push(`/posts/${state.data.postId}`);
  }
}, [state, router]);
```

## Environment

- vinext: 0.0.31
- - Next.js App Router (Server Actions)
- - - Cloudflare Workers (via `@cloudflare/vite-plugin`)

yunus25jmi1:
## Description

This PR fixes issue #589 where `useActionState` receives `undefined` state when a Server Action calls `redirect()` from `next/navigation`, causing a `TypeError` on the next render.

---

## Problem

When a server action used with `useActionState` calls `redirect()`:

1. The server catches the redirect error and returns redirect headers
2. **Problem:** Server was sending **empty response body**
3. Client's RSC navigation expects a valid RSC payload but receives empty body
4. Navigation fails silently, leaving `useActionState` with `undefined` state
5. TypeError on next render when accessing state properties

**Example:**

```tsx
'use client';
import { useActionState } from 'react';
import { redirect } from 'next/navigation';

async function createPost(_prev: ActionResult, formData: FormData): Promise<ActionResult> {
  // ... create post logic ...
  redirect(`/posts/${postId}`); // <-- triggers the bug
}

export function PostForm() {
  const [state, formAction, pending] = useActionState(createPost, { success: false, error: '' });

  // TypeError: Cannot read properties of undefined (reading 'success')
  return (
    <form action={formAction}>
      {!state.success && state.error && <p>{state.error}</p>}
      <button type="submit">Submit</button>
    </form>
  );
}
```

---

## Root Cause

Server action redirects were sending **empty response body** with only headers:

```typescript
// Before: empty body
return new Response("", { status: 200, headers: redirectHeaders });
```

RSC navigation requires a valid RSC payload to parse and render. Empty body causes navigation to fail.

---

## Solution

**Pre-render the redirect target's RSC payload on the server** and include it in the response body. This matches Next.js behavior where the redirect response includes the target page's RSC payload for soft SPA navigation.

### Server-side (app-rsc-entry.ts)

```typescript
// After: render redirect target's RSC payload
const redirectMatch = matchRoute(redirectPathname);
if (redirectMatch) {
  const redirectElement = buildPageElement(redirectRoute, redirectParams, ...);
  const rscStream = renderToReadableStream({ root: redirectElement, ... });
  return new Response(rscStream, { status: 200, headers: redirectHeaders });
}
```

### Client-side (app-browser-entry.ts)

```typescript
// Check if response has RSC payload
const hasRscPayload = contentType.includes('text/x-component') && fetchResponse.body;

if (hasRscPayload) {
  // Soft RSC navigation (SPA-like)
  const result = await createFromFetch(Promise.resolve(fetchResponse), ...);
  getReactRoot().render(result.root);
} else {
  // Fallback: hard redirect for empty body
  window.location.assign/replace(actionRedirect);
}
```

---

## Benefits

1. ✅ **Fixes issue #589** - useActionState no longer receives undefined state
2. ✅ **Preserves SPA navigation** - Soft RSC navigation instead of hard page load
3. ✅ **Matches Next.js parity** - Pre-renders redirect target like Next.js does
4. ✅ **Backward compatible** - Falls back to hard redirect if payload unavailable

---

### Core Fix

- **`packages/vinext/src/entries/app-rsc-entry.ts`**: Pre-render redirect target's RSC payload
- **`packages/vinext/src/server/app-browser-entry.ts`**: Use RSC navigation when payload available, fallback to hard redirect

### Test Coverage

- **`tests/fixtures/app-basic/app/actions/actions.ts`**: Added `redirectWithActionState` test action
- **`tests/fixtures/app-basic/app/action-state-redirect/page.tsx`**: Added test page
- **`tests/e2e/app-router/server-actions.spec.ts`**: Added E2E test for issue #589 with waitForHydration

---

## Testing

- ✅ Shims test: 730 tests passed
- ✅ Entry templates snapshot updated
- ✅ E2E test added for the fix

---

## Related Issues

- Fixes #589

---

Signed-off-by: Md Yunus <admin@yunuscollege.eu.org>

### DIFF
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -141,7 +141,7 @@ function registerServerActionCallback(): void {
           return undefined;
         }
       } catch {
-        // Fall through to client-side navigation if URL parsing fails.
+        // Fall through to hard redirect below if URL parsing fails.
       }
 
       // Use hard redirect for all action redirects because vinext's server


packages/vinext/src/server/app-browser-entry.ts
```diff
@@ -143,17 +144,16 @@ function registerServerActionCallback(): void {
...
+      // requires a valid RSC payload. This is a known parity gap with Next.js,
+      // which pre-renders the redirect target's RSC payload.
       const redirectType = fetchResponse.headers.get("x-action-redirect-type") ?? "replace";
       if (redirectType === "push") {
-        window.history.pushState(null, "", actionRedirect);
+        window.location.assign(actionRedirect);
       } else {
-        window.history.replaceState(null, "", actionRedirect);
+        window.location.replace(actionRedirect);
       }
```
james-elicx, line 156: I just want to make sure I understand this - are we saying that we're essentially replacing one bug with another?

Can we fix the underlying issue instead?

yunus25jmi1: yes, i will look into it.

PR #645, opened 2026-03-22
feat: seed memory cache from pre-rendered routes

### LINKED ISSUE #561: Serve pre-rendered responses from memory cache
When an app has pre-rendered routes, those should be used to stub the memory cache (the default ISR option).

NathanDrake2406:
## Summary

- Reads `vinext-prerender.json` at production server startup and populates the `MemoryCacheHandler` with pre-rendered HTML/RSC files from `dist/server/prerendered-routes/`
- First request to any pre-rendered page is now a cache HIT instead of a full re-render
- Adds `buildId`, `router`, and `trailingSlash` to the prerender manifest so the seeding function can construct matching cache keys and locate files correctly

## Changes

| File | Change |
|------|--------|
| `src/server/seed-cache.ts` | New `seedMemoryCacheFromPrerender()` function |
| `src/build/prerender.ts` | Add `router` to `PrerenderRouteResult`, add `buildId`/`trailingSlash` to manifest |
| `src/build/run-prerender.ts` | Pass `buildId`/`trailingSlash` to `writePrerenderIndex` |
| `src/server/prod-server.ts` | Call seeding at App Router production server startup |
| `tests/seed-cache.test.ts` | 11 unit tests |

## Scope

App Router only for now. Pages Router seeding requires the prerender phase to also write `pageData` JSON files (not currently done), so it's left for a follow-up.

Closes #561

## Test plan

- [x] 11 unit tests covering: ISR routes, static routes, dynamic routes, index route, trailingSlash layout, skipped/errored routes, missing manifest, missing HTML files, missing RSC files, multiple routes
- [x] Existing prerender tests (46) and ISR cache tests (42) pass unchanged
- [x] Type-aware lint clean

### DIFF
--- a/packages/vinext/src/build/prerender.ts
+++ b/packages/vinext/src/build/prerender.ts
@@ -51,6 +51,8 @@ export type PrerenderRouteResult =
        * Omitted for non-dynamic routes where pattern === path.
        */
       path?: string;
+      /** Which router produced this route. Used by cache seeding. */
+      router: "app" | "pages";
     }
   | {
       route: string;
@@ -586,6 +588,7 @@ export async function prerenderPages({
             status: "rendered",
             outputFiles,
             revalidate,
+            router: "pages",
             ...(urlPath !== route.pattern ? { path: urlPath } : {}),
           };
         } catch (e) {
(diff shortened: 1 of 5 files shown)


james-elicx:
Only had a brief look to see if this handled Workers as well - do you have any thoughts on how this could be adapt to support that environment as well? Admittedly there would be gotcha's we need to make sure people are aware of if they use the memory cache with Workers, but it would be good to stub that with the prerendered routes rather than just SSR.

Issue #623, opened 2026-03-21
RFC: Per-request store API


james-elicx:
This issue is a **request for feedback on the public API** for our per-request store API. The idea was originally proposed in #608.

Internally, the store would interact with a map that's inside a request-scoped `AsyncLocalStorage`.

I've been doing some thinking this morning about the public API that we'd be exposing, as it would be nice have it well-defined from the get-go to avoid potentially introducing  breaking changes in the future.

## Store Interface

My thinking is that we would want to expose full control for interacting with the underlying store, which also providing some higher-level abstractions to reduce boilerplate.

The store itself would expose the following:

```ts
// checks the underlying map.
// does type-narrowing so that get(key) calls in if-statements have the non-undefined type.
has(key: string): boolean;
// returns value from underlying map. narrowed when combined with has(key).
get(key: string): unknown | undefined;
// inserts value to underlying map and returns the value back.
set(key: string, value: unknown): unknown;
// checks if the map already has a value and returns it, otherwise sets the value and returns it.
getOrSet(key: string, factory: () => unknown): unknown;
```

Methods accepting a callback may possibly need async variants as well, or the other way around with async as the default and sync as the variant, or neither if we find a nice way to make both work in one function.

## Store Consumption

I had been thinking through a few different ways to approach this, and the following is what I arrived at for a nice balance of type-safety and control.

It would be great to get thoughts from others, or suggestions if you would like to see something else!

**Function that returns a type-safe interface to the underlying map**

```ts
import { createRequestStore } from 'vinext/server';

// safe to be called in global scope as it doesn't access the inner store
const store = createRequestStore<{ prisma: PrismaClient }>();

const getPrisma = () =>
  store.getOrSet("prisma", () => {
    const pool = new Pool();
    return new PrismaClient(pool);
  });

const getPrisma = () => {
  if (store.has("prisma")) {
    return store.get("prisma"); // type-narrowed by has(...)
  }

  const pool = new Pool();
  const client = new PrismaClient(pool);

  return store.set("prisma", client);
};
```

The alternative approach that was on my mind would be a Proxy to the underlying map where you pass the type stored to each function call, but that would sacrifice some of the type-safety.

**Abstraction on top of the store for callbacks**

Another option that would be interesting to explore is a `cacheForRequest` function, that takes a callback and lazily initialises the value in the request store on first access. Similar to how React's `cache` function works, but not limited to server components.

```ts
import { cacheForRequest } from 'vinext/server';

[...]

Issue #537, opened 2026-03-14
Alternating request failures (success, fail, success...) when using Drizzle + Postgres lazily evaluated in vinext / Cloudflare Workers


sindhukhrisna:
## Description

I am deploying a Next.js application to Cloudflare Workers (using vinext). I am using Drizzle ORM with the postgres driver connecting to a Supabase database.

Because Cloudflare Workers do not allow asynchronous I/O (like establishing a database connection) in the global scope, the standard Drizzle initialization fails during the build step. To bypass this, I implemented a lazy initialization pattern using a JavaScript Proxy.

While this successfully builds on both Vercel and Cloudflare Workers, runtime behavior on Cloudflare is inconsistent. Specifically, requests fail exactly every other time:

* 1st Request: Success
* 2nd Request: Fail
* 3rd Request: Success
* 4th Request: Fail

(This happens regardless of whether I use Hyperdrive or a DB pooler URL).

---

### Attempt 1: Standard Drizzle Setup (Fails during build)

If I use the standard initialization, the worker fails to build/deploy due to global scope I/O restrictions.

```ts
import * as schema from "./schemas/schema-main";
import * as schema2 from "./schemas/schema-secondary";
import { drizzle } from 'drizzle-orm/postgres-js'
import { env } from "cloudflare:workers";
import postgres from 'postgres'

const poolDb = postgres(
  env.HYPERDRIVE.connectionString ?? "", {
  prepare: false,
});

export const db = drizzle({
  client: poolDb,
  schema: schema,
})
```

### Build Error (Attempt 1)

```
16:48:27.291 ✘ [ERROR] A request to the Cloudflare API (/accounts/<ACCOUNT_ID>/workers/scripts/<PROJECT_NAME>/versions) failed.
16:48:27.291  Uncaught Error: Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and generating random values are not allowed within global scope. To fix this error, perform this operation within a handler. https://developers.cloudflare.com/workers/runtime-apis/handlers/
16:48:27.291    at null.<anonymous> (index.js:36463:41)
16:48:27.291   [code: 10021]
```

---

### Attempt 2: Lazy Evaluation using Proxy (Builds successfully, but runtime alternates failing)

To fix the build error, I used a Proxy to lazily instantiate the database connection only when a query is actually executed.

```ts
import * as schema from "./schemas/schema-main";
import { drizzle } from 'drizzle-orm/postgres-js'
[...]

PR #646, opened 2026-03-22
feat: implement cacheForRequest() per-request factory cache

JamesbbBriz:
## Summary

Implements `cacheForRequest()` from RFC #623 — a per-request factory cache that lazily initializes values on first call and returns the cached result for subsequent calls within the same request.

## API

```ts
import { cacheForRequest } from "vinext/cache-for-request";

const getPrisma = cacheForRequest(() => {
  const pool = new Pool({ connectionString: env.HYPERDRIVE.connectionString });
  return new PrismaClient({ adapter: new PrismaPg(pool) });
});

// In any server code:
const prisma = getPrisma(); // first call creates, subsequent calls reuse
```

**Key design decisions:**
- Factory function reference = cache key (no string keys, no cross-module collision)
- Async factories cache the Promise itself (concurrent-safe: two `await getDb()` share one invocation)
- Outside request scope: factory runs every time, no caching (safe for tests/build-time)

## Changes

| File | Change |
|------|--------|
| `shims/cache-for-request.ts` | New — `cacheForRequest()` implementation (17 lines of logic) |
| `shims/unified-request-context.ts` | Add `requestCache: WeakMap<Function, unknown>` field + shallow-clone comment sync |
| `package.json` | Add `"./cache-for-request"` export |

## Blast radius

Zero. New file + one new field with `new WeakMap()` default (~0.1μs per request). No existing behavior changes.

## Test plan

Tested with Prisma v7 + Hyperdrive on Workers. Happy to add unit tests — let me know the preferred test file location.

Addresses #623 · Relates to #537

### DIFF
--- a/tests/cache-for-request.test.ts
+++ b/tests/cache-for-request.test.ts
@@ -23,7 +23,7 @@ describe("cacheForRequest", () => {
     const get = cacheForRequest(factory);
 
     const ctx = createRequestContext();
-    runWithRequestContext(ctx, () => {
+    void runWithRequestContext(ctx, () => {
       const a = get();
       const b = get();
       expect(a).toBe(b);
@@ -38,7 +38,7 @@ describe("cacheForRequest", () => {
     const getB = cacheForRequest(factoryB);
 
     const ctx = createRequestContext();
-    runWithRequestContext(ctx, () => {
+    void runWithRequestContext(ctx, () => {
       expect(getA()).toBe("a");
       expect(getB()).toBe("b");
       expect(factoryA).toHaveBeenCalledTimes(1);


packages/vinext/package.json
```diff
@@ -21,6 +21,10 @@
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js"
     },
+    "./cache-for-request": {
+      "types": "./dist/shims/cache-for-request.d.ts",
+      "import": "./dist/shims/cache-for-request.js"
+    },
```
james-elicx: can this be made available through `vinext/cache`?

PR #651, opened 2026-03-22
feat: add optional dev-time `next typegen`

pacexy:
## Summary
- run automatically during dev server startup when Next.js is installed
- rerun typegen when app/pages route files are added or removed, with a small debounce and serialized execution
- add tests covering startup, opt-out via , route-file watch reruns, and missing-Next fallback behavior

## Notes
- this PR is intended as a temporary stopgap to improve developer experience
- Vinext startup still succeeds when `next` is not installed; typegen is skipped in that case
- the long-term goal is still native Vinext type generation rather than depending on `next typegen`

## Testing
- vp check
- vp test tests/tsconfig-paths-vite8.test.ts tests/typegen.test.ts

### DIFF
--- a/tests/typegen.test.ts
+++ b/tests/typegen.test.ts
@@ -120,7 +120,7 @@ describe("next typegen controller", () => {
 });
 
 describe("vinext dev typegen integration", () => {
-  it("invokes the project-local Next.js typegen command on dev server startup by default", async () => {
+  it("runs the project-local `next typegen` command on dev server startup by default", async () => {
     const root = createTempProject("vinext-typegen-start-");
     let server: ViteDevServer | null = null;
     try {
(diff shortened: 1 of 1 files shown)


packages/vinext/src/typegen.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+export function resolveNextTypegenBin(root: string): string | null {
+  try {
+    const projectRequire = createRequire(path.join(root, "package.json"));
+    return projectRequire.resolve("next/dist/bin/next");
+  } catch {
+    return null;
+  }
+}
```
james-elicx, line 36: We should avoid have having a hard dependency/requirement on Next.js for Vinext to be usable

pacexy: That makes sense. I agree Vinext itself should not depend on Next.js to be usable.

In this PR, my intent was only to add an optional dev-time integration: if `next` is installed in the project already, Vinext will opportunistically run `next typegen`; if it is not installed, startup still succeeds and we skip it.

For now, this is mainly intended as a pragmatic way to improve the developer experience in the short term. We can still follow up later with a native Vinext typegen implementation so this no longer depends on Next.js at all.

PR #670, opened 2026-03-23
feat: populate Workers KV with pre-rendered pages at deploy time

### LINKED ISSUE #562: Populating remote cache during deployment
If an application has configured a remote cache handler like Workers KV, and has pre-rendered assets, those should automatically populate KV using Wrangler at deploy-time.

For KV, see the OpenNext implementation.

NathanDrake2406:
## Summary

Closes #562. **Depends on #653** — requires `buildId`, `trailingSlash`, and `router` fields in the prerender manifest. Should be merged after #653.

- **New module** `cloudflare/populate-kv.ts` — reads prerender manifest + HTML/RSC files, constructs cache entries matching the runtime format exactly, uploads via Cloudflare REST bulk API
- **Deploy integration** — automatic step 6d between prerender asset copy and wrangler deploy when `CLOUDFLARE_API_TOKEN` + `VINEXT_CACHE` KV namespace + prerender manifest all exist. `--no-populate-kv` to opt out.
- **TPR refactor** — replaces TPR's inline `uploadToKV` (80 lines) with shared helpers, fixing three bugs: wrong key format (`cache:<path>` → `cache:app:<buildId>:<path>:html`), wrong TTL (10x revalidate → fixed 30-day), missing tags (empty `[]` → full hierarchy for `revalidatePath()`)

### Key format parity

Deploy-time keys match the runtime `__isrCacheKey` in `app-rsc-entry.ts` exactly:
```
[appPrefix:]cache:app:<buildId>:<pathname>:<suffix>
```

Two entries per route (`:html` + `:rsc`) with full tag hierarchy (`_N_T_/layout`, intermediate layouts, leaf page tag) so `revalidatePath()` works on seeded entries.

### Design decisions

- **App Router only** — Pages Router uses different keys (`pages:...`) and value shape (`kind: "PAGES"`). Routes without `router: "app"` are skipped.
- **Batching** — by count (10,000 max) and bytes (95 MB with 5 MB headroom from 100 MB API limit)
- **Fixed 30-day KV TTL** — matches `KVCacheHandler.set()` default
- **No `--app-prefix` flag** — removed after review; it would write prefixed keys without wiring the prefix into the generated runtime, creating silently unreadable entries. `appPrefix` remains on the programmatic `PopulateKVOptions` for advanced users who configure both sides.

## Test plan

- [x] 40 unit tests covering key parity, tag parity, entry serialization, bulk batching, end-to-end flow, router filtering, graceful skips
- [x] `vp check` passes (0 warnings, 0 errors)
- [ ] Manual: `CLOUDFLARE_API_TOKEN=... vinext deploy --prerender-all` on a test project with ISR routes
- [ ] Verify KV entries via `wrangler kv key list --binding VINEXT_CACHE`
- [ ] First request shows `X-Vinext-Cache: HIT`

### DIFF
--- a/packages/vinext/src/deploy.ts
+++ b/packages/vinext/src/deploy.ts
@@ -60,8 +60,6 @@ export interface DeployOptions {
   tprWindow?: number;
   /** Disable automatic KV population with pre-rendered pages */
   noPopulateKv?: boolean;
-  /** KVCacheHandler appPrefix for KV key construction */
-  appPrefix?: string;
 }
 
 // ─── CLI arg parsing (uses Node.js util.parseArgs) ──────────────────────────
@@ -80,7 +78,6 @@ const deployArgOptions = {
   "tpr-limit": { type: "string" },
   "tpr-window": { type: "string" },
   "no-populate-kv": { type: "boolean", default: false },
-  "app-prefix": { type: "string" },
 } as const;
 
 export function parseDeployArgs(args: string[]) {
@@ -109,7 +106,6 @@ export function parseDeployArgs(args: string[]) {
     tprLimit: parseIntArg("tpr-limit", values["tpr-limit"]),
     tprWindow: parseIntArg("tpr-window", values["tpr-window"]),
     noPopulateKv: values["no-populate-kv"],
-    appPrefix: values["app-prefix"]?.trim() || undefined,
   };
 }
 
--- a/packages/vinext/src/deploy.ts
+++ b/packages/vinext/src/deploy.ts
@@ -27,7 +27,8 @@ import {
   findInNodeModules as _findInNodeModules,
 } from "./utils/project.js";
 import { getReactUpgradeDeps } from "./init.js";
-import { runTPR } from "./cloudflare/tpr.js";
+import { parseWranglerConfig, resolveAccountId, runTPR } from "./cloudflare/tpr.js";
+import { populateKV } from "./cloudflare/populate-kv.js";
 import { runPrerender } from "./build/run-prerender.js";
 import { loadDotenv } from "./config/dotenv.js";
 import { loadNextConfig, resolveNextConfig } from "./config/next-config.js";
@@ -57,6 +58,10 @@ export interface DeployOptions {
   tprLimit?: number;
   /** TPR: analytics lookback window in hours (default: 24) */
   tprWindow?: number;
+  /** Disable automatic KV population with pre-rendered pages */
+  noPopulateKv?: boolean;
+  /** KVCacheHandler appPrefix for KV key construction */
+  appPrefix?: string;
 }
 
 // ─── CLI arg parsing (uses Node.js util.parseArgs) ──────────────────────────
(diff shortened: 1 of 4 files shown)


packages/vinext/src/deploy.ts
```diff
@@ -1356,6 +1361,34 @@ export async function deploy(options: DeployOptions): Promise<void> {
...
+        if (kvResult.skipped) {
+          console.log(`\n  KV populate: Skipped (${kvResult.skipped})`);
+        } else if (kvResult.entriesUploaded > 0) {
+          console.log(
+            `\n  KV populate: ${kvResult.routesProcessed} routes → ${kvResult.entriesUploaded} entries (${(kvResult.durationMs / 1000).toFixed(1)}s)`,
+          );
+        }
+      }
+    }
+  }
```
james-elicx, line 1390: I think we need to better think about the API for this before implementing it.

There shouldn't be an implicit assumption that KV is the cache handler people use.

I have been thinking about this over the weekend - we need to ensure the solution we arrive at will scale to any cache handler rather than being KV-specific.

I'm not still not 100% on what that should look like. Some things I was thinking about at the weekend were having a static property on the cache handler itself that can define a method for build/deploy-time bulk-population, and then provide the handler using the next.config.js cacheHandler option.

That could work, but I think it might have an issue with bundling dependencies that we don't want bundled if it were to use wrangler for the population logic for example.

Lets continue a conversation about how this can be implemented in a way that scales to any cache handler on the issue before iterating on impl

james-elicx (changes requested):

packages/vinext/src/deploy.ts
NathanDrake2406: Sorry for zoomin! I'll refactor it out of deploy and turn this into a KV adapter instead. Do you think the foundation part is good enough tho?

NathanDrake2406: Yeah i think a config is best here. I'll get back to u w the details. Need to nerd out a bit

james-elicx: Lets discuss on the issue what good looks like for a long-term solution 👍

NathanDrake2406: > Lets discuss on the issue what good looks like for a long-term solution 👍

instead of a static method, what about a named export? It feels cleaner and more modular, less coupled.
I can refactor this PR to use the seeding core and the KV part to a KV adapter

james-elicx: As the decision we make will have to be supported for a long time ideally without needing breaking changes, it would probably be good for us to come up with a few different options that we can weigh up pros/cons for and hopefully get some engagement from others in the community on, with a caveat that whatever approach we take should still be fully compatible with the cache handler Next.js config option.

NathanDrake2406: Yeah I think you should open an issue and pin it, maybe a discussion?

james-elicx: There's a placeholder issue for remote cache population - we can use that one perhaps?

NathanDrake2406: @southpolesteve. Would appreciate your expertise on this!

james-elicx: I've posted some initial thoughts on the GitHub issue.

james-elicx:
I'm going to close this for now based on the above conversation. I've started the discussion about options the API could potentially look like on the issue for this feature - it would be good to get feedback / engagement if you're interested in taking a look. Lets try and ensure new features are well-defined before using AI to generate PRs for them.

Issue #639, opened 2026-03-22
App Router client navigation double-flashes Suspense fallbacks and janky back-button scroll restoration


NathanDrake2406:
## Summary

Client-side navigation in vinext App Router visibly commits partial destination UI (Suspense fallbacks) instead of keeping the old content on screen until the new content is ready. This produces a double-flash on list/detail transitions and janky scroll restoration on back/forward navigation.

This does not happen in the equivalent app running on Next.js.

## Reproduction

Any vinext App Router app with async server components inside Suspense boundaries will show this. Minimal pattern:

**Page (server component):**

```tsx
// app/page.tsx
import { Suspense } from "react";
import { FilterToggle } from "./FilterToggle";

async function ItemList({ filter }: { filter: string }) {
  // Simulate DB query — any async server component triggers it
  await new Promise((r) => setTimeout(r, 300));
  const items = filter
    ? [{ id: 1, name: `Filtered: ${filter}` }]
    : [{ id: 1, name: "Item A" }, { id: 2, name: "Item B" }, { id: 3, name: "Item C" }];
  return <ul>{items.map((i) => <li key={i.id}>{i.name}</li>)}</ul>;
}

export default async function Page({ searchParams }: { searchParams: Promise<{ filter?: string }> }) {
  const { filter = "" } = await searchParams;
  return (
    <div>
      <h1>{filter ? `Filtered: ${filter}` : "All items"}</h1>
      <FilterToggle current={filter} />
      <Suspense fallback={<div style={{ height: 400 }} />}>
        <ItemList filter={filter} />
      </Suspense>
    </div>
  );
}
```

**Client component:**

```tsx
// app/FilterToggle.tsx
"use client";
import { useRouter } from "next/navigation";

export function FilterToggle({ current }: { current: string }) {
  const router = useRouter();
  return (
    <button onClick={() => router.push(current ? "/" : "/?filter=active")}>
      {current ? "Clear filter" : "Apply filter"}
    </button>
  );
}
```

### Steps — double flash

1. `vinext dev`, open the page
2. Click "Apply filter" — navigates to `/?filter=active`
3. Click "Clear filter" — navigates back to `/`
4. **Observe:** the heading updates to "All items" while the Suspense fallback (empty space) is visible, then the list pops in afterward — two visible stages instead of one clean swap

### Steps — back button scroll jank

1. Add a detail route (`app/item/[id]/page.tsx`) with an async server component
2. Scroll down on the list page
[...]

PR #690, opened 2026-03-26
fix: cross-route client navigation hangs in Firefox (#652)

### LINKED ISSUE #652: Cross-route client navigation hangs in Firefox (startTransition never commits)
## Description

Cross-route client-side navigation (e.g. `/top` to `/tv-shows`) hangs indefinitely in Firefox. The old page stays visible and the new page never renders. Same-route navigations (e.g. changing searchParams on `/top`) work fine in all browsers.

This is Firefox-specific and only reproduces on production builds served via `wrangler dev` (workerd). Chrome works fine for all navigation types.

## Root cause

React's `startTransition` never commits when the **entire component tree** is replaced during a cross-route navigation in Firefox. The transition render starts but React never calls the commit phase, so `useLayoutEffect` callbacks never fire and the navigation promise never resolves.

This is **not a React bug**. Next.js avoids this entirely because their segment-level cache architecture only swaps changed route segments inside `startTransition` -- parent layouts stay mounted, so the transition is always an incremental update. vinext replaces the full RSC tree on every cross-route navigation, which is a much larger update that React's transition scheduler apparently cannot finalize in Firefox.

Investigation trace (with debug logging in the browser entry):

1. Link click handler fires correctly
2. `navigateImpl` is called
3. `navigateRsc` fetches the RSC response (200 OK, valid payload)
4. Response is buffered and parsed successfully
5. `renderNavigationPayload` calls `startTransition(() => setState({...}))`
6. **Hangs here** -- React never commits the transition, even after 30+ seconds

No React errors are reported via `onUncaughtError` or `onRecoverableError` callbacks on the root.

## Proper fix

Implement segment-level caching (a `CacheNode` tree that mirrors route segments). On navigation, diff the old and new route trees and only swap segments that changed. This would make `startTransition` updates incremental (matching Next.js behavior) and eliminate the Firefox hang.

See analysis in #639 for details on how Next.js handles this via `layout-router.tsx` + `segment-cache/navigation.ts`.

## Workaround

PR #643 works around this by detecting cross-route vs same-route navigations and only using `startTransition` for same-route navigations. Cross-route navigations use synchronous state updates, which commit immediately in all browsers:

```ts
const isSameRoute = url.pathname === window.location.pathname;
// ...
renderNavigationPayload(payload, snapshot, commitEffect, isSameRoute);
```

This means cross-route navigations lose the "keep old UI visible during loading" behavior, but the page navigates. Same-route navigations (filter changes) still benefit from transitions.

## Reproduction

1. Build any App Router app with multiple routes that have Suspense boundaries
2. Serve via `wrangler dev`
3. Open in Firefox
4. Click a link that navigates to a different route (different pathname)
5. The page stays on the old route indefinitely

## Environment

- Firefox (tested on latest)
- Production build on workerd via `wrangler dev`
- Does NOT reproduce in Chrome
- Does NOT reproduce on Vite dev server (Node.js)
- Does NOT affect Next.js (segment-level updates avoid the problem)

Divkix:
## Summary

Fixes #652 — `startTransition` never commits in Firefox when the entire RSC component tree is replaced during cross-route navigation. The old page stays visible indefinitely in production builds on workerd.

**Root cause**: vinext replaces the full RSC tree on every cross-route navigation inside `startTransition`. The React transition scheduler cannot finalize this in Firefox. The previous `flushSync` approach also caused the Suspense double-flash from #639.

**Fix**: Two-phase navigation commit with same-route detection:
- **Same-route navigations** (search param changes): use `startTransition` for smooth incremental updates
- **Cross-route navigations** (different pathname): use synchronous updates that bypass the Firefox scheduler issue
- **URL/history commit deferred** to `useLayoutEffect` so hooks see the pending URL during transitions

### Key changes

- **`navigation.ts`**: `ClientNavigationState` on `Symbol.for` global (survives multiple module instances), render snapshot context for hook consistency during transitions, RSC response snapshot/restore for visited cache, unified `navigateClientSide()` entry point, history suppression helpers
- **`app-browser-entry.ts`**: `BrowserRoot` component with `useState`-managed tree, `NavigationCommitSignal` that defers URL commit to `useLayoutEffect`, visited response cache with LRU eviction (50 entries, 5min TTL — LRU via delete+re-insert promotion on hit, evict oldest-inserted when over capacity), navigation ID counter for stale navigation bailout, manual scroll restoration
- **`link.tsx`**: removed duplicated `isHashOnlyChange`/`scrollToHash`, delegates to `navigateClientSide`
- **`form.tsx`**: App Router GET forms delegate to `navigateClientSide`
- **`global.d.ts`**: extended `__VINEXT_RSC_NAVIGATE__` signature with `navigationKind` and `historyUpdateMode`

### Edge cases handled

- Rapid clicking (navigation ID counter prevents stale commits)
- Back/forward instant replay from visited response cache
- Server action redirects
- Hash-only changes (no RSC fetch)
- External URLs (falls through to `location.assign`)
- HMR (clears all navigation caches)
- Multiple Vite module instances (`Symbol.for` global state)

## Test plan

- [x] `vp check` passes (only pre-existing benchmark errors remain)
- [x] `tests/shims.test.ts` — 742 tests pass
- [x] `tests/form.test.ts` — updated assertions for new navigate signature, all pass
- [x] `tests/prefetch-cache.test.ts` — compatible with new `PrefetchCacheEntry` type, all pass
- [x] `tests/link.test.ts` — all pass
- [x] `tests/routing.test.ts` — all pass
- [x] `tests/app-router.test.ts` — all pass
- [x] `tests/entry-templates.test.ts` — all pass
- [x] New E2E tests in `tests/e2e/app-router/navigation-regressions.spec.ts` covering:
  - Cross-route navigation completes without hanging
  - Same-route navigation (search param change)
  - Back/forward navigation
  - Rapid same-route navigation settles correctly
  - Cross-route then same-route navigation
  - `usePathname`/`useSearchParams`/`useParams` hook sync
  - Provider page dynamic param navigation
  - Round-trip SPA state preservation
- [x] CI: full Vitest suite
- [x] CI: Playwright E2E (chromium, firefox, webkit)


packages/vinext/src/server/app-browser-entry.ts
```diff
@@ -204,21 +606,77 @@ async function main(): Promise<void> {
...
       const url = new URL(href, window.location.origin);
       const rscUrl = toRscUrl(url.pathname + url.search);
+      // Use startTransition for same-route navigations (searchParam changes)
+      // so React keeps the old UI visible during the transition. For cross-route
+      // navigations (different pathname), use synchronous updates — React's
+      // startTransition hangs in Firefox when replacing the entire tree.
+      // NB: During rapid navigations, window.location.pathname may not reflect
+      // the previous navigation's URL yet (URL commit is deferred). This could
+      // cause misclassification (synchronous instead of startTransition or vice
+      // versa), resulting in slightly less smooth transitions but correct behavior.
```
james-elicx, line 623: i think it would be good to see if we can potentially find a better fix for this as a follow-up...

packages/vinext/src/shims/navigation.ts
```diff
@@ -272,127 +282,465 @@ export function getPrefetchedUrls(): Set<string> {
...
+type ClientNavigationGlobal = typeof globalThis & {
+  [_CLIENT_NAV_STATE_KEY]?: ClientNavigationState;
+};
+
+function getClientNavigationState(): ClientNavigationState | null {
+  if (isServer) return null;
+
+  const globalState = window as ClientNavigationGlobal;
+  if (!globalState[_CLIENT_NAV_STATE_KEY]) {
+    globalState[_CLIENT_NAV_STATE_KEY] = {
```
james-elicx, line 481: nit: `??=`

tests/e2e/app-router/navigation-regressions.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+// this directory (navigation.spec.ts, server-client-only.spec.ts, after.spec.ts,
+// etc.). Port 4174 is the vite preview default for the app-basic fixture.
+const BASE = "http://localhost:4174";
+
+async function waitForHydration(page: import("@playwright/test").Page) {
+  await expect(async () => {
+    const ready = await page.evaluate(() => !!(window as any).__VINEXT_RSC_ROOT__);
+    expect(ready).toBe(true);
+  }).toPass({ timeout: 10_000 });
+}
```
james-elicx, line 13: should probably be pulled out into a shared helper - i think a few tests do something like this

james-elicx (approved):
Tried it out on a complex app as well and it looks like it's working well.

PR #721, opened 2026-03-30
test: regression tests for inherited intercepting route slot deduplication

Debbl:
## Summary

The fix for this bug (inherited parallel slots causing the same intercepting page to be counted multiple times during validation) has already been merged into main. This PR retains the regression tests and the `intercepting-routes` playground example so the behavior stays covered going forward.

## What's included

- `tests/intercepting-routes-build.test.ts` — production build regression test: verifies a route structure with an inherited `@modal` slot and an intercepting `(.)photo/[id]` route builds successfully
- `tests/routing.test.ts` — unit regression test: verifies that inherited intercepting slots sharing the same target pattern are correctly allowed (not rejected as duplicates)
- `examples/app-router-playground/app/intercepting-routes/` — playground demo showing the intercepting routes pattern with a modal slot

## Background

When a parallel slot is inherited by child routes, the same intercepting page can appear on multiple discovered routes. The old validation flattened all intercepted target patterns and treated repeated entries from inherited slots as separate route definitions, incorrectly throwing:

`You cannot have two routes that resolve to the same path`

The fix (de-duplicating by target pattern before validation) landed in main separately. These tests guard against regressions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ coverage/
 
 # Worktrees
 .worktrees/
+.claude
 
 # Next.js reference clone (for searching source/tests)
 .nextjs-ref/
@@ -41,3 +42,5 @@ result
 result-*
 .direnv/
 .corepack/
+
+.vinext


packages/vinext/src/routing/app-router.ts
```diff
@@ -173,13 +173,18 @@ export async function appRouter(
...
+    for (const slot of route.parallelSlots) {
+      for (const intercept of slot.interceptingRoutes) {
+        // Inherited slots can surface the same intercepting page on multiple
+        // child routes. De-dupe by page path so we only validate distinct
+        // physical intercept definitions while still rejecting real conflicts.
+        uniqueInterceptTargetPatterns.set(intercept.pagePath, intercept.targetPattern);
+      }
+    }
+  }
+  validateRoutePatterns(Array.from(uniqueInterceptTargetPatterns.values()));
```
james-elicx: The old approach could have been wrapped in a set to achieve the same result I imagine.

I wonder if it could potentially be possible to have two conflicting pagePath values and therefore one overwrites the others targetPattern 🤔

Debbl: @james-elicx I changed it a bit. Mind taking another look?

james-elicx: I think it could potentially be possible for two interceptions to target the same route pattern

What I meant when I wrote this was that you could potentially achieve the same desired outcome with:
```ts
    [...new Set(routes.flatMap((route) =>
      route.parallelSlots.flatMap((slot) =>
        slot.interceptingRoutes.map((intercept) => intercept.targetPattern),
      ),
    ))]
```
just deduping the array we originally created

what do you think?

Debbl: I updated the PR to follow your suggestion.

After discussing it a bit more with AI, I think the `targetPattern` dedupe is probably fine for this PR. What I still want to validate is the behavior when there are multiple intercepting routes targeting the same route, for example:

```
app/feed/page.tsx
app/feed/@modal/(..)photo/[id]/page.tsx

app/profile/page.tsx
app/profile/@modal/(..)photo/[id]/page.tsx

app/photo/[id]/page.tsx
```

I’m going to put together a small demo around this case to compare the behavior in Next.js and vinext, just to make sure we’re matching the intended semantics.

james-elicx: No worries, I'll mark as draft for the moment while you're looking at that

Debbl: Ok, I opened [#730](https://github.com/cloudflare/vinext/issues/730) to track this issue separately.

james-elicx: The Set dedupe was merging in another PR so I'm going to merge this as additional regression tests.

PR #731, opened 2026-03-31
[codex] Fix wrapped use server exports

### LINKED ISSUE #722: Incompatible with next-safe-action
I am trying to use https://github.com/next-safe-action/next-safe-action with Vinext but I am getting this error

```
12:21:32 AM [vite] Internal server error: unsupported non async function
  Plugin: rsc:use-server
  File: /Users/[USER]/projects/vinext-nextjs/src/test.action.ts:5:0
  1  |  "use server";
  2  |  import { actionClient } from "./lib/safe-action";
  3  |  export const testAction = actionClient.action(async () => {
     |  ^
  4  |          return { message: "Hello, world!" };
  5  |  });
```

# Steps to reproduce
1. Clone https://github.com/LubomirGeorgiev/vinext-next-safe-action
2. Run `pnpm dev:vinext`

southpolesteve:
## Summary

Fix file-level `"use server"` exports that wrap an inline async action, such as `next-safe-action`'s `actionClient.action(async () => ...)` pattern.

## Root Cause

`@vitejs/plugin-rsc` validates file-level `"use server"` exports syntactically and rejects wrapped exports whose initializer is a call expression instead of a direct async function expression. That causes vinext to fail with `unsupported non async function` for valid Next.js-compatible patterns.

## Changes

- add a pre-transform plugin that rewrites wrapped file-level `"use server"` exports into local bindings plus `export { ... }`
- register that plugin ahead of `@vitejs/plugin-rsc`
- add a regression test covering the `next-safe-action` shape and the upstream strict proxy transform behavior

## Validation

- `pnpm test tests/use-server-wrapped-exports.test.ts tests/use-server-closure-collision.test.ts`
- `pnpm run check`
- built local `vinext`, linked it into the reporter's repro app from issue #722, and confirmed `GET /` changed from `500 unsupported non async function` to `200`

Closes #722.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -64,6 +64,7 @@ import { asyncHooksStubPlugin } from "./plugins/async-hooks-stub.js";
 import { clientReferenceDedupPlugin } from "./plugins/client-reference-dedup.js";
 import { createInstrumentationClientTransformPlugin } from "./plugins/instrumentation-client.js";
 import { createOptimizeImportsPlugin } from "./plugins/optimize-imports.js";
+import { fixUseServerWrappedExportsPlugin } from "./plugins/fix-use-server-wrapped-exports.js";
 import { fixUseServerClosureCollisionPlugin } from "./plugins/fix-use-server-closure-collision.js";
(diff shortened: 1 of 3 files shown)


james-elicx:
Fixed upstream.

PR #736, opened 2026-04-01
fix(tpr): use correct KV cache key format matching runtime

NathanDrake2406:
## Summary

- TPR's `uploadToKV` wrote KV keys as `cache:<pathname>` but the runtime `KVCacheHandler.get()` reads `cache:app:<buildId>:<pathname>:html` — every TPR-seeded entry was a dead key (guaranteed cache miss)
- Extracted `buildTprKVPairs()` as an exported pure function that delegates key construction to the shared `isrCacheKey()` from `server/isr-cache.ts`
- Reads `buildId` from the prerender manifest (`vinext-prerender.json`) at upload time so keys include the build identifier

### Before
```
cache:/blog/hello          ← TPR writes this
cache:app:abc123:/blog/hello:html  ← runtime expects this
```
Never matches → every TPR page is a cold miss.

### After
```
cache:app:abc123:/blog/hello:html  ← TPR writes this
cache:app:abc123:/blog/hello:html  ← runtime expects this
```

## Test plan

- [x] 7 unit tests covering key format parity, header-based revalidate, TTL clamping (min 60s, max 30d), no-buildId backward compat, KVCacheEntry serialization
- [x] 216 existing deploy tests pass
- [x] 42 ISR cache tests pass
- [x] `vp check` clean

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -3229,24 +3229,30 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
     },
     // Write BUILD_ID to dist/server/ so post-build tools (TPR, seed-cache) can
     // read the build identifier without depending on the prerender manifest.
-    {
-      name: "vinext:build-id",
-      apply: "build",
-      enforce: "post",
-      writeBundle: {
-        sequential: true,
-        order: "post",
-        handler(options) {
-          const envName = this.environment?.name;
-          if (envName !== "rsc") return;
-
-          const outDir = options.dir;
-          if (!outDir) return;
-
-          fs.writeFileSync(path.join(outDir, "BUILD_ID"), nextConfig!.buildId);
+    // Uses closeBundle (not writeBundle) with a one-time write guard so the file
+    // is written exactly once per build regardless of how many environments are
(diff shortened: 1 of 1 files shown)


packages/vinext/src/cloudflare/tpr.ts
```diff
@@ -854,13 +874,24 @@ export async function runTPR(options: TPROptions): Promise<TPRResult> {
...
 
   // ── 10. Upload to KV ──────────────────────────────────────────
+  // Read buildId from prerender manifest so KV keys match the runtime format.
+  let buildId: string | undefined;
+  try {
+    const manifestPath = path.join(root, "dist", "server", "vinext-prerender.json");
+    const raw = fs.readFileSync(manifestPath, "utf-8");
+    buildId = JSON.parse(raw).buildId;
+  } catch {
+    // Best-effort — proceed without buildId
```
james-elicx: TPR doesn't use the actual prerender pipeline at the moment so this would never realistically find a build ID, and you wouldn't do TPR with --prerender-all at the same time

james-elicx: we should probably always write the build id to its own file regardless of prerender or not like next.js does (`BUILD_ID` i believe) and use that

PR #753, opened 2026-04-02
feat: encode interception context in App Router payload IDs and caches

NathanDrake2406:
## Summary

Part of #726 (PR 4). This adds the interception-context identity that PR 2c intentionally deferred. Intercepted App Router navigations now carry their source route through the payload IDs, prefetch cache, visited-response cache, and history state, so `/feed -> /photos/42` modal navigations stop aliasing direct `/photos/42` page loads.

The practical effect is that intercepted and direct renders of the same visible pathname no longer stomp each other. Repeated feed/gallery modal navigations reuse the right cached payload, back/forward restores the right source context, and the ordinary visited-response cache still works for non-intercepted navigations.

### What changed

- **Payload metadata and IDs** — App Router payloads now include `__interceptionContext` when an intercepted render is active, and route/page IDs are encoded with that context instead of keying only by pathname
- **Browser router state** — `app-browser-entry` and `app-browser-state` now carry committed interception context alongside `elements`, `routeId`, and `rootLayoutTreePath`, and write it into history state so traversal can recover the original source route
- **Prefetch and visited-response caches** — prefetch keys and visited-response keys are now interception-aware, so `/photos/42.rsc` from `/feed` and `/gallery` do not alias each other. Direct-route payloads still omit interception metadata, so visited-response storage preserves the request-side cache partition when the payload does not provide one
- **Client requests** — `next/link` prefetch and App Router navigations send `X-Vinext-Interception-Context` when the current render is inside an intercepted flow, and the server reads that header when building intercepted payloads
- **Fixtures and tests** — adds a `gallery` source route plus targeted integration/E2E coverage for direct-vs-intercepted cache separation, repeated intercepted navigations, refresh behavior, and interception-aware prefetch entries

### What this does NOT do

Server action / refresh parity for intercepted routes is still deferred. This PR intentionally leaves action re-renders on the direct-route path and keeps the scope to payload encoding + browser-side cache/history identity.

## Test plan

- [x] `vp test run tests/app-elements.test.ts tests/app-browser-entry.test.ts tests/prefetch-cache.test.ts`
- [x] `vp test run tests/app-router.test.ts -t "renders intercepted photo modal on RSC navigation from feed"`
- [x] `vp run test:e2e -- tests/e2e/app-router/advanced.spec.ts --project app-router`
- [x] `git commit` pre-commit hook (`vp check --fix`)

### DIFF
--- a/packages/vinext/src/server/app-browser-state.ts
+++ b/packages/vinext/src/server/app-browser-state.ts
@@ -4,6 +4,7 @@ import type { ClientNavigationRenderSnapshot } from "../shims/navigation.js";
 
 export type AppRouterState = {
   elements: AppElements;
+  interceptionContext: string | null;
   renderId: number;
   navigationSnapshot: ClientNavigationRenderSnapshot;
   rootLayoutTreePath: string | null;
@@ -12,6 +13,7 @@ export type AppRouterState = {
 
 export type AppRouterAction = {
   elements: AppElements;
+  interceptionContext: string | null;
   navigationSnapshot: ClientNavigationRenderSnapshot;
   renderId: number;
   rootLayoutTreePath: string | null;
@@ -21,6 +23,7 @@ export type AppRouterAction = {
 
 export type PendingNavigationCommit = {
   action: AppRouterAction;
+  interceptionContext: string | null;
   rootLayoutTreePath: string | null;
   routeId: string;
 };
@@ -37,6 +40,7 @@ export function routerReducer(state: AppRouterState, action: AppRouterAction): A
     case "navigate":
       return {
         elements: mergeElements(state.elements, action.elements, action.type === "traverse"),
+        interceptionContext: action.interceptionContext,
         navigationSnapshot: action.navigationSnapshot,
         renderId: action.renderId,
         rootLayoutTreePath: action.rootLayoutTreePath,
@@ -45,6 +49,7 @@ export function routerReducer(state: AppRouterState, action: AppRouterAction): A
     case "replace":
       return {
         elements: action.elements,
+        interceptionContext: action.interceptionContext,
         navigationSnapshot: action.navigationSnapshot,
         renderId: action.renderId,
         rootLayoutTreePath: action.rootLayoutTreePath,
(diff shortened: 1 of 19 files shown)


packages/vinext/src/server/app-browser-state.ts
```diff
@@ -101,12 +106,15 @@ export async function createPendingNavigationCommit(options: {
...
       navigationSnapshot: options.navigationSnapshot,
       renderId: options.renderId,
       rootLayoutTreePath: metadata.rootLayoutTreePath,
       routeId: metadata.routeId,
       type: options.type,
     },
+    // Convenience aliases — always equal action.interceptionContext / action.rootLayoutTreePath / action.routeId.
+    interceptionContext: metadata.interceptionContext,
     rootLayoutTreePath: metadata.rootLayoutTreePath,
     routeId: metadata.routeId,
```
james-elicx, line 119: these feel unnecessary to have if they're just duplicates of what's in action.

tests/e2e/app-router/advanced.spec.ts
```diff
@@ -2,6 +2,12 @@ import { test, expect } from "@playwright/test";
 
 const BASE = "http://localhost:4174";
 
+async function waitForAppRouterHydration(page: import("@playwright/test").Page) {
+  await page.waitForFunction(() => typeof window.__VINEXT_RSC_NAVIGATE__ === "function", null, {
+    timeout: 10_000,
+  });
+}
```
james-elicx, line 9: i believe a helper already exists for this that could be reused

packages/vinext/src/shims/navigation.ts
```diff
@@ -274,6 +277,26 @@ export function toRscUrl(href: string): string {
...
+export function getCurrentInterceptionContext(): string | null {
+  if (isServer) {
+    return null;
+  }
+
+  return (
+    readHistoryStateInterceptionContext(window.history.state) ??
+    stripBasePath(window.location.pathname, __basePath)
+  );
+}
```
james-elicx, line 298: hmm this reads that it will always return an interception context on the client-side even if there might not be one

PR #767, opened 2026-04-03
feat: static/dynamic layout detection for skip-header optimization

### LINKED ISSUE #756: Static/dynamic detection
## Problem

Pls refer to https://github.com/cloudflare/vinext/issues/726 ￼ for context

Phase 1 layout persistence works by always rendering all entries on the server and merging them on the client.

The planned Phase 3 skip-header optimization (`X-Vinext-Router-Skip`) needs to know which layout entries are safe to skip re-rendering. That requires a way to classify persisted layout entries as either:

- static: safe to reuse from client cache
- dynamic: must be re-rendered, or later handled with explicit stale-time semantics

Without this classification, skip-header work has no reliable foundation. This can be based off main I think since it's orthogonal

NathanDrake2406:
## Summary

Part of #726 (Phase 3, PR 1). Adds a three-layer classification system that determines whether each layout in an App Router route tree is static or dynamic. This metadata is the foundation for PR 7's skip-header optimization, which avoids re-rendering static layouts on RSC navigations. Wiring is done in PR 768.

> **Stacked on #755** (PR 5). Review only the last 6 commits.

### Three-layer classification

```
Segment config: force-dynamic?           → DYNAMIC (final)
Segment config: force-static?            → STATIC (final)
Module graph: no dynamic shim imports?   → STATIC (proven, skip probe)
Module graph: has dynamic shim imports?  → run probe at request time
Probe: dynamicUsageDetected = true?      → DYNAMIC (observed)
Probe: dynamicUsageDetected = false?     → STATIC (observed)
```

**Layer 1 — Segment config** (`classifyLayoutSegmentConfig` in `report.ts`). Reads `export const dynamic` / `export const revalidate` from layout source files. Unlike page classification, positive `revalidate` values return null (ISR is a page concept) — only the extremes (`0` → dynamic, `Infinity` → static) are decisive.

**Layer 2 — Module graph** (`classifyLayoutByModuleGraph` in `layout-classification.ts`). BFS traversal of each layout's dependency tree via a `ModuleInfoProvider` abstraction over Vite's `this.getModuleInfo`. If no transitive dynamic shim import (headers, cache, server) is found, the layout is provably static without any runtime cost.

**Layer 3 — Runtime probe** (extended `probeAppPageLayouts` in `app-page-execution.ts`). For layouts that import dynamic shims but may not call them on every request, the probe runs the layout function with isolated dynamic scope tracking. The `runWithIsolatedDynamicScope` callback is injected from the RSC entry, keeping the execution module decoupled from ALS internals.

### Wire format

`__layoutFlags` metadata key in the RSC payload: `Record<string, "s" | "d">` mapping layout IDs to compact flags. Follows the established pattern of `__route`, `__rootLayout`, `__interceptionContext`.

On client-side navigation, the router merges new flags with existing ones (navigate) or replaces them entirely (replace), building up cumulative knowledge of which layouts are safe to skip.

### Type design

`LayoutClassificationOptions` groups the three classification fields (`getLayoutId`, `runWithIsolatedDynamicScope`, `buildTimeClassifications`) into a single optional object. This makes the all-or-nothing invariant type-safe — you either provide the full classification context or nothing.

`LayoutFlags` is canonically defined in `app-elements.ts` and re-exported from `app-page-execution.ts`.

`readAppElementsMetadata` accepts `Record<string, unknown>` since the RSC payload carries heterogeneous values. `parseLayoutFlags` validates at the wire boundary using a type predicate guard.

### New files

| File | Purpose |
|------|---------|
| `packages/vinext/src/build/layout-classification.ts` | Module graph traversal + combined classification |
| `tests/layout-classification.test.ts` | 13 tests for module graph + combined classification |

## Test plan

- [x] `tests/build-report.test.ts` — 7 tests for `classifyLayoutSegmentConfig`
- [x] `tests/layout-classification.test.ts` — 13 tests for module graph traversal + combined classification
- [x] `tests/app-page-execution.test.ts` — 5 tests for per-layout probe classification (dynamic detection, build-time skip, backward compat, error default)
- [x] `tests/app-page-probe.test.ts` — 2 tests for layout flags propagation + special error handling with classification
- [x] `tests/app-elements.test.ts` — 2 tests for `__layoutFlags` metadata parsing + backward compat
- [x] `tests/app-browser-entry.test.ts` — 2 tests for layoutFlags merge (navigate) and replace semantics
- [x] `tests/app-router.test.ts` — 276 integration tests pass with zero regressions
- [x] `vp check` — format, lint, type checks all clean

```bash
vp test run tests/build-report.test.ts -t "classifyLayoutSegmentConfig"
vp test run tests/layout-classification.test.ts
vp test run tests/app-page-execution.test.ts tests/app-page-probe.test.ts
vp test run tests/app-elements.test.ts tests/app-browser-entry.test.ts
vp test run tests/app-router.test.ts
vp check
```

Refs #726
Closes #756

### DIFF
--- a/packages/vinext/src/server/app-page-execution.ts
+++ b/packages/vinext/src/server/app-page-execution.ts
@@ -29,7 +29,7 @@ export type ProbeAppPageLayoutsResult = {
 };
 
 export type LayoutClassificationOptions = {
-  /** Build-time classifications from segment config or module graph. */
+  /** Build-time classifications from segment config or module graph, keyed by layout index. */
   buildTimeClassifications?: ReadonlyMap<number, "static" | "dynamic"> | null;
   /** Maps layout index to its layout ID (e.g. "layout:/blog"). */
   getLayoutId: (layoutIndex: number) => string;
--- a/packages/vinext/src/server/app-page-execution.ts
+++ b/packages/vinext/src/server/app-page-execution.ts
@@ -1,3 +1,7 @@
+import type { LayoutFlags } from "./app-elements.js";
+
+export type { LayoutFlags };
+
 export type AppPageSpecialError =
   | { kind: "redirect"; location: string; statusCode: number }
   | { kind: "http-access-fallback"; statusCode: number };
@@ -19,11 +23,18 @@ export type BuildAppPageSpecialErrorResponseOptions = {
   specialError: AppPageSpecialError;
 };
 
-export type LayoutFlags = Readonly<Record<string, "s" | "d">>;
-
 export type ProbeAppPageLayoutsResult = {
   response: Response | null;
   layoutFlags: LayoutFlags;
+};
+
+export type LayoutClassificationOptions = {
+  /** Build-time classifications from segment config or module graph. */
+  buildTimeClassifications?: ReadonlyMap<number, "static" | "dynamic"> | null;
+  /** Maps layout index to its layout ID (e.g. "layout:/blog"). */
+  getLayoutId: (layoutIndex: number) => string;
+  /** Runs a function with isolated dynamic usage tracking per layout. */
+  runWithIsolatedDynamicScope: <T>(fn: () => T) => Promise<{ result: T; dynamicDetected: boolean }>;
 };
 
 export type ProbeAppPageLayoutsOptions = {
@@ -31,15 +42,8 @@ export type ProbeAppPageLayoutsOptions = {
   onLayoutError: (error: unknown, layoutIndex: number) => Promise<Response | null>;
   probeLayoutAt: (layoutIndex: number) => unknown;
   runWithSuppressedHookWarning<T>(probe: () => Promise<T>): Promise<T>;
-
-  /** Build-time classifications from segment config or module graph. */
-  buildTimeClassifications?: ReadonlyMap<number, "static" | "dynamic"> | null;
-  /** Maps layout index to its layout ID (e.g. "layout:/blog"). */
-  getLayoutId?: (layoutIndex: number) => string;
-  /** Runs a function with isolated dynamic usage tracking per layout. */
-  runWithIsolatedDynamicScope?: <T>(
-    fn: () => T,
-  ) => Promise<{ result: T; dynamicDetected: boolean }>;
+  /** When provided, enables per-layout static/dynamic classification. */
+  classification?: LayoutClassificationOptions | null;
 };
 
 export type ProbeAppPageComponentOptions = {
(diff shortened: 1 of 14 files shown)


NathanDrake2406:
Thanks for the thorough pass. Walking through each item:

### Addressed

**#1 (`layout-classification.ts:46`, BFS `shift()`):** done in 1b851172. Swapped to a head pointer (`let head = 0; queue[head++]`). The 13 existing tests in `tests/layout-classification.test.ts` still pass.

**#7 (`tests/app-page-execution.test.ts`, build-time + special-error gap):** done in 86042390. New test `"returns special error response when build-time classified layout throws during error probe"`. Two build-time-static layouts, `probeLayoutAt(1)` throws, `onLayoutError` resolves to a 404, and we assert: the special-error response propagates, the build-time flag for layout 1 is recorded before the short-circuit, and layout 0 is never reached. The mock also installs a `runWithIsolatedDynamicScope` that throws if called, which proves the build-time branch never falls through to Layer 3.

### Skipped intentionally

**#2 (`app-page-execution.ts:129`, truthiness comment):** the existing comment on the next line already explains the intent (\"Build-time classified... skip dynamic isolation, but still probe for special errors\"). The fall-through reads naturally from the type. A second comment about how \`undefined\` happens to evaluate falsy felt like commenting on commenting.

**#3 (`app-page-execution.ts:141`, debug log for swallowed probe errors):** conservative dynamic fallback on probe failure is the intended production behavior. Adding a debug log without a known failure mode would be defensive observability with no actionable signal. Happy to revisit if we see a real pattern of probe-failure noise in the wild.

### Acknowledged, no action needed

**#4 (`app-elements.ts:109`, `isLayoutFlagsRecord` edge cases):** ack. The wire format only emits plain objects, but the guard handles the theoretical Date/RegExp/Map case correctly anyway.

**#5 (`app-browser-state.ts:100`, \`d → s\` requires server resend):** the invariant you named already holds. \`probeAppPageLayouts\` in \`packages/vinext/src/server/app-page-execution.ts\` iterates \`layoutCount - 1 → 0\` and writes \`layoutFlags[cls.getLayoutId(layoutIndex)]\` for every layout in the current tree on every request. The server always re-emits flags for the entire current tree, so a \`\"d\" → \"s\"\` flip is reflected on the next navigation that includes the layout.

**#6 (`app-page-render.ts:121`, unused `preRenderResult.layoutFlags`):** ack. Consumption lands in PR #768 (\`pr-7-skip-header\`), the next branch in this stack.

packages/vinext/src/server/app-page-execution.ts
```diff
@@ -98,24 +118,64 @@ export async function buildAppPageSpecialErrorResponse(
...
+        // Layer 3: probe with isolated dynamic scope to detect per-layout
+        // dynamic API usage (headers(), cookies(), connection(), etc.)
+        try {
+          const { dynamicDetected } = await cls.runWithIsolatedDynamicScope(() =>
+            options.probeLayoutAt(layoutIndex),
+          );
+          layoutFlags[cls.getLayoutId(layoutIndex)] = dynamicDetected ? "d" : "s";
+        } catch (error) {
+          // Probe failed — conservatively treat as dynamic.
+          layoutFlags[cls.getLayoutId(layoutIndex)] = "d";
```
james-elicx, line 148: i wonder if it's worth having a separate flag for failures that we interpret as dynamic but still track as being errored

james-elicx:
sorry about the delay

PR #795, opened 2026-04-08
feat: csp support

hyoban:
Currently, inline scripts are being blocked by CSP.

```
self.__VINEXT_RSC_PARAMS__={}
self.__VINEXT_RSC_NAV__=...
self.__VINEXT_RSC_CHUNKS__.push(...)
self.__VINEXT_RSC_DONE__=true
```


hyoban:
Resolved. Updating the lock file to fix the CI.

hyoban:
Re: “nonce on `<link rel="stylesheet">` is a spec mismatch and diverges from Next.js behavior” — this does not match current Next.js behavior: the local canary reference still adds `nonce` to Pages Router stylesheet/preload links in [`packages/next/src/pages/_document.tsx`](https://github.com/vercel/next.js/blob/canary/packages/next/src/pages/_document.tsx), and [`test/e2e/app-dir/next-dynamic-csp-nonce/next-dynamic-csp-nonce.test.ts`](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/next-dynamic-csp-nonce/next-dynamic-csp-nonce.test.ts) explicitly asserts nonce-bearing dynamic preload links, so removing `nonce` from `link` tags here would introduce a parity regression rather than fix one.

hyoban:
resolved

PR #809, opened 2026-04-09
fix: apply middleware request-header overrides to App Route request objects

southpolesteve:
## Summary

App Route handlers now receive a request object rebuilt from middleware request-header overrides, so `request.headers` and `headers()` agree after middleware mutations.

## Details

Middleware request-header overrides were already applied to the `next/headers` ALS context, so `headers()` reflected the middleware-modified request state. But App Route handlers still received a tracked `NextRequest` built from the original request object, so `request.headers` exposed the pre-middleware Authorization/Cookie headers.

This change:
- preserves a clone of the middleware response headers before internal `x-middleware-*` headers are stripped for response safety
- threads that preserved header set into the typed App Route execution module
- rebuilds the underlying `Request` with `buildRequestHeadersFromMiddlewareResponse()` before constructing the tracked `NextRequest`

## Tests

Adds an App Route fixture and integration test proving that after middleware deletes credential headers and injects `x-from-middleware`, both:
- `request.headers`
- `headers()`

see the same middleware-modified values.

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -1188,7 +1188,7 @@ export default async function handler(request, ctx) {
     // Per-request container for middleware state. Passed into
     // _handleRequest which fills in .headers and .status;
     // avoids module-level variables that race on Workers.
-    const _mwCtx = { headers: null, status: null };
+    const _mwCtx = { headers: null, requestHeaders: null, status: null };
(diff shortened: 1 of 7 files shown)


southpolesteve:
/bigbonk Rebased onto latest main and resolved the conflict cleanly. Focused integration coverage still passes locally for all three middleware override cases: App page, App route, and mixed Pages fallback. Please re-review.

PR #817, opened 2026-04-10
fix: apply middleware request-header overrides before App->Pages fallback

southpolesteve:
## Summary

In mixed App+Pages projects, the App Router production fallback now rebuilds the request from middleware request-header overrides before delegating to Pages SSR.

## Details

Middleware request-header overrides were already applied to:
- App page rendering
- App Route handlers

But the App Router production fallback still delegated to `renderPage()` with the original `Request` object. That meant a Pages route reached via App->Pages fallback could still see the original `Authorization` / `Cookie` headers in `getServerSideProps`, even when middleware had deleted them and injected trusted replacement headers.

This change:
- rebuilds the fallback request with `buildRequestHeadersFromMiddlewareResponse()` using the preserved middleware request-header payload
- passes that rebuilt request into `renderPage()`
- keeps passing the preserved middleware header payload into the existing `middlewareHeaders` parameter

## Tests

Adds a production-only regression test verifying that `/pages-header-override-delete` in a mixed App+Pages project sees:
- `authorization: null`
- `cookie: null`
- `x-from-middleware: hello-from-middleware`

before `getServerSideProps` runs.

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -31,6 +31,10 @@ import { isProxyFile } from "../server/middleware.js";
 // resolve these at code-generation time and embed them as absolute paths.
 const configMatchersPath = resolveEntryPath("../config/config-matchers.js", import.meta.url);
 const requestPipelinePath = resolveEntryPath("../server/request-pipeline.js", import.meta.url);
+const middlewareRequestHeadersPath = resolveEntryPath(
+  "../server/middleware-request-headers.js",
+  import.meta.url,
+);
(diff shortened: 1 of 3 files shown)


southpolesteve:
Closing as duplicate / mixed branch. The actual  code fix was already merged in PR #813. This branch was based on pre-merge history, so GitHub still showed the same fallback fix against the old merge base, and it also picked up unrelated  ISR test work. Restarting  from current  on a clean branch.

southpolesteve:
Follow-up context: the actual #619 code fix was already merged in PR #813. This branch was based on pre-merge history, so GitHub still showed the same fallback fix against the old merge base, and it also picked up unrelated #614 ISR test work. Closing this PR as a duplicate / mixed branch and restarting #614 cleanly from current main.

PR #818, opened 2026-04-10
fix: strip Set-Cookie from App Route ISR cache entries

southpolesteve:
## Summary

App Route ISR cache values no longer persist `Set-Cookie` headers.

## Details

Cached App Route responses are shared across requests. Persisting `Set-Cookie` inside the ISR cache would replay per-request cookies to later visitors on HIT/STALE responses.

This change updates `buildAppRouteCacheValue()` to keep dropping:
- `set-cookie`
- `x-vinext-cache`
- `cache-control`

## Tests

Adds a dedicated cacheable route handler fixture (`/api/static-set-cookie`) with `revalidate = 1` that sets `Set-Cookie` directly on the response.

Regression coverage verifies:
- first request: `MISS` and `Set-Cookie` present
- second request: `HIT` and `Set-Cookie` absent
- body is still cached correctly across the HIT

### DIFF
--- a/packages/vinext/src/server/app-route-handler-response.ts
+++ b/packages/vinext/src/server/app-route-handler-response.ts
@@ -89,13 +89,12 @@ export async function buildAppRouteCacheValue(response: Response): Promise<Cache
   const headers: CachedRouteValue["headers"] = {};
 
   response.headers.forEach((value, key) => {
+    // Never persist Set-Cookie into shared ISR cache entries. Cached route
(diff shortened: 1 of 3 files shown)


james-elicx:
I think this might be a deviation from Next.js, albeit arguably a correct one.

https://github.com/vercel/next.js/blob/canary/packages/next/src/build/templates/app-route.ts#L369-L398

southpolesteve:
Yeah good catch @james-elicx. I'm honestly torn about how to proceed here.

Issue #472, opened 2026-03-11
feat: support assetPrefix in next.config


elydelva:
## Summary

`assetPrefix` from `next.config` is currently not read or used anywhere in vinext. Setting it is silently ignored.

## Current behavior

In production, all static assets (fonts, JS chunks, etc.) are always served from:

```
{basePath}/assets/[file]-[hash].[ext]
```

The URL is generated by Vite's asset pipeline with no way to inject a CDN prefix.

## Expected behavior

When `assetPrefix` is set in `next.config`:

```ts
// next.config.ts
const nextConfig: NextConfig = {
  assetPrefix: 'https://assets.example.com',
};
```

All asset URLs should be prefixed accordingly:

```
https://assets.example.com/assets/Font-[hash].woff2
https://assets.example.com/_next/static/chunks/...
```

## Why this matters

When deploying to Cloudflare Workers with a gateway pattern (a hub Worker routing requests to multiple service Workers), every request — including static assets like fonts and images — goes through the gateway Worker and incurs an invocation.

With `assetPrefix` support, static assets could be served from a dedicated subdomain (e.g. `assets-app.example.com`) backed by a Cloudflare Workers Assets-only deployment with `run_worker_first: false`. This would serve fonts, CSS, and JS bundles **directly from Cloudflare's edge with zero Worker invocations**.

## Investigation

Verified by reading `dist/config/next-config.js` — `resolveNextConfig` never extracts `assetPrefix`. The `vinext:local-fonts` Vite plugin (index.js lines 3153–3219) rewrites font paths to Vite static asset imports, but Vite's `base` (driven by `basePath`) is the only configurable prefix — there is no `assetPrefix` injection at any stage.

## Suggested approach

In `resolveNextConfig`, extract `assetPrefix` and pass it through to the Vite plugin. The plugin would then configure Vite's [`experimental.renderBuiltUrl`](https://vitejs.dev/guide/build#advanced-base-options) to prefix resolved asset URLs:

```ts
// vite.config equivalent
build: {
  rollupOptions: {},
},
experimental: {
  renderBuiltUrl(filename) {
    return assetPrefix + '/' + filename;
  }
}
```

This would cover JS/CSS chunks, font files, and images processed through Vite's asset pipeline.

PR #819, opened 2026-04-10
fix: emit served URLs for next/font/google self-hosted assets

Shorebirdmgmt:
## Summary

`fetchAndCacheFont()` in `packages/vinext/src/plugins/fonts.ts` downloads Google Fonts `.woff2` files into `<root>/.vinext/fonts/<family>-<urlHash>/` and rewrites the cached `@font-face` CSS with `css.split(fontUrl).join(path.join(fontDir, filename))` — an absolute dev-machine filesystem path. That CSS is then embedded verbatim as `_selfHostedCSS` in the server bundle, and every downstream consumer reads from the same leaked string: the injected `<style data-vinext-fonts>` block's `@font-face { src: url(...) }`, the HTML body's `<link rel="preload">` tags, and the HTTP `Link:` response header.

Production requests on workerd produce header and body preload entries like `</home/<user>/<project>/.vinext/fonts/geist-<hash>/geist-<hash>.woff2>` — the browser follows the absolute `href` to `<origin>/home/<user>/...`, workerd returns 404 (the cached files are never copied into `dist/client/` either), and the console fills with `downloadable font: download failed` and `preloaded with link preload was not used` warnings on every page view. Because preload is high-priority for fonts, the broken request contends with real critical-path traffic and Cloudflare's 103 Early Hints path would emit these broken entries before the HTML even starts streaming.

Any app using stock `next/font/google` self-hosted mode is affected — no user-facing config can work around it (`#472` tracks `assetPrefix` support but that's orthogonal, and the leak happens regardless of whether `assetPrefix` is set).

## Root cause

Three downstream font-preload emitters all read from the same in-memory array populated by `collectFontPreloadsFromCSS()` in `shims/font-google-base.ts`, which extracts `url(...)` references from `_selfHostedCSS` via regex:

1. The injected `<style data-vinext-fonts>` block's `@font-face { src: url(...) }` (via `ssrFontStyles` in `shims/font-google-base.ts`).
2. The HTML body's `<link rel="preload">` tags emitted from `server/app-ssr-entry.ts:renderFontHtml()` via `fontData.preloads[*].href`.
3. The HTTP `Link:` response header, built by `buildAppPageFontLinkHeader()` in `server/app-page-execution.ts` and set on the response in `server/app-page-response.ts:242`.

All three read from the same source of truth, so a single fix at the CSS level propagates to every emitter.

The upstream source — the cached CSS — is written by `fetchAndCacheFont()`:

```ts
// packages/vinext/src/plugins/fonts.ts
for (const [fontUrl, filename] of urls) {
  const filePath = path.join(fontDir, filename);
  // ... download + write to disk ...
  // Rewrite CSS to use absolute path (Vite will resolve /@fs/ for dev, or asset for build)
  css = css.split(fontUrl).join(filePath);
}
```

The "Vite will resolve /@fs/ for dev, or asset for build" comment describes the intended behaviour but is not what actually happens: the CSS is embedded as a **JavaScript string literal** in the bundle, and Vite's asset pipeline operates on CSS files and `import`/`new URL(...)` references — not on string literals inside JS. The filesystem path gets baked into `_selfHostedCSS` verbatim, and nothing downstream rewrites it.

Separately, `fetchAndCacheFont` leaves the downloaded `.woff2` files in `<root>/.vinext/fonts/` and nothing ever copies them into `dist/client/` — so even a correctly-rewritten URL would 404 in production without a companion copy step.

## Fix

Two changes in `packages/vinext/src/plugins/fonts.ts`, both inside the existing `vinext:google-fonts` plugin:

1. **`_rewriteCachedFontCssToServedUrls()`** — a small helper (3 lines of real logic plus regression-detection JSDoc) that replaces the absolute `cacheDir` prefix in a cached CSS string with the served URL namespace `/assets/_vinext_fonts`. Called from `injectSelfHostedCss()` right before the CSS string is `JSON.stringify`'d into the bundle, so every downstream consumer sees the rewritten URL.

2. **`writeBundle` hook** (client environment only) — recursively copies every `.woff2`/`.woff`/`.ttf`/`.otf`/`.eot` file out of `<root>/.vinext/fonts/` into `<clientOutDir>/assets/_vinext_fonts/`, preserving the `<family>-<hash>/` subdirectory structure. The existing `_headers` rule for `/assets/*` already covers the new namespace, and `StaticFileCache` picks the copied files up via its recursive walk of `dist/client/`, so `Content-Type: font/woff2`, `Cache-Control: public, max-age=31536000, immutable`, and an automatic content-hashed ETag all flow through without any server-side changes.

No config surface is added — this does not implement `#472` (`assetPrefix` / `basePath` support), it just stops the absolute filesystem path from leaking into the preload emitters regardless of whether `assetPrefix` is set.

## Tests

Two new regression tests, both verified to fail on `main` and pass with the fix:

- **Unit** — `tests/font-google.test.ts`: five cases in a new `_rewriteCachedFontCssToServedUrls` describe block drive the helper directly with (a) a realistic cached CSS containing multiple `url(...)` references, (b) multi-occurrence replacement where the same path appears multiple times in a single block, (c) a cache directory containing regex metacharacters (`/tmp/build (1)/...`) to prove split/join is safer than a constructed regex, (d) CSS that never references the cache directory (no-op), and (e) an empty cacheDir defensive guard so a split on `""` doesn't insert the URL namespace between every character.
- **Integration** — `tests/app-router.test.ts`, in a new `App Router Production server self-hosted next/font/google headers` describe block. Builds the existing `tests/fixtures/font-google-multiple` fixture (stock `Geist` + `Geist_Mono` via `next/font/google`) with a mocked fetch that stands in for the Google Fonts CDN. The mock returns CSS with real `https://fonts.gstatic.com/...` URLs so `fetchAndCacheFont`'s regex actually exercises the path-rewriting code that was the bug source — returning CSS with already-relative URLs would sidestep the failure mode. Starts `startProdServer()` on a random port and asserts on the raw HTTP response: (a) `Link:` header contains the served URL and neither the absolute fixture path nor `.vinext/fonts` appears anywhere in it, (b) body `<link rel="preload">` tags match `/assets/_vinext_fonts/...`, (c) the injected `<style data-vinext-fonts>` block's `@font-face src: url()` also uses the served URL, and (d) the copied font file serves 200 with `content-type: font/woff2` and an `immutable` cache-control.

A separate fixture is used instead of extending `app-basic` because `app-basic` is shared by every integration test in `tests/app-router.test.ts` — adding `next/font/google` to its root layout would force a real Google Fonts network fetch into every test run in the file. The mocked-fetch approach keeps the test hermetic.

## Verified

With the fix applied:

| Check | Result |
|---|---|
| `pnpm test:unit` | **2853 / 2853** pass |
| `pnpm test:integration` | **1262 / 1262** pass (2 pre-existing skips) |
| `pnpm run check` (lint + format + types) | **0** warnings, **0** errors |
| `pnpm test:e2e` — `app-router` project | **322 / 322** pass (3 flaky on retry, 9 pre-existing skips) |
| `pnpm test:e2e` — `cloudflare-workers` project (Miniflare) | **38 / 38** pass (1 flaky on retry) |

Both new tests were also re-run with the fix reverted on a clean tree to confirm they deterministically catch the regression — the integration suite printed the leaked path verbatim in the `<style data-vinext-fonts>` assertion, e.g.

```
src: url(/home/<user>/.../.vinext/fonts/geist-4db05770f54f/geist-2fba11a4.woff2) format('woff2');
```

and the font-file `200 OK` assertion failed because the cached `.woff2` files were never copied into `dist/client/`.

End-to-end runtime verification on `workerd` (Miniflare, via `wrangler dev`) against `examples/app-router-cloudflare` with a `next/font/google` layout import:

**Before**

```
$ curl -sI http://localhost:4999/
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Link: </home/<user>/<project>/.vinext/fonts/geist-4db05770f54f/geist-8e42e564.woff2>; rel=preload; as=font; type=font/woff2; crossorigin, </home/<user>/<project>/.vinext/fonts/geist-4db05770f54f/geist-bd9fc9d8.woff2>; rel=preload; as=font; type=font/woff2; crossorigin, </home/<user>/<project>/.vinext/fonts/geist-4db05770f54f/geist-79e7a99c.woff2>; rel=preload; as=font; type=font/woff2; crossorigin
```

```
$ curl -s http://localhost:4999/ | grep -oE '<link[^>]*preload[^>]*woff2[^>]*>'
<link rel="preload" href="/home/<user>/<project>/.vinext/fonts/geist-4db05770f54f/geist-8e42e564.woff2" as="font" type="font/woff2" crossorigin />
<link rel="preload" href="/home/<user>/<project>/.vinext/fonts/geist-4db05770f54f/geist-bd9fc9d8.woff2" as="font" type="font/woff2" crossorigin />
<link rel="preload" href="/home/<user>/<project>/.vinext/fonts/geist-4db05770f54f/geist-79e7a99c.woff2" as="font" type="font/woff2" crossorigin />
```

```
$ curl -sI 'http://localhost:4999/home/<user>/<project>/.vinext/fonts/geist-4db05770f54f/geist-8e42e564.woff2'
HTTP/1.1 404 Not Found
```

**After**

```
$ curl -sI http://localhost:4999/
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Link: </assets/_vinext_fonts/geist-4db05770f54f/geist-8e42e564.woff2>; rel=preload; as=font; type=font/woff2; crossorigin, </assets/_vinext_fonts/geist-4db05770f54f/geist-bd9fc9d8.woff2>; rel=preload; as=font; type=font/woff2; crossorigin, </assets/_vinext_fonts/geist-4db05770f54f/geist-79e7a99c.woff2>; rel=preload; as=font; type=font/woff2; crossorigin
```

```
$ curl -s http://localhost:4999/ | grep -oE '<link[^>]*preload[^>]*woff2[^>]*>'
<link rel="preload" href="/assets/_vinext_fonts/geist-4db05770f54f/geist-8e42e564.woff2" as="font" type="font/woff2" crossorigin />
<link rel="preload" href="/assets/_vinext_fonts/geist-4db05770f54f/geist-bd9fc9d8.woff2" as="font" type="font/woff2" crossorigin />
<link rel="preload" href="/assets/_vinext_fonts/geist-4db05770f54f/geist-79e7a99c.woff2" as="font" type="font/woff2" crossorigin />
```

```
$ curl -sI http://localhost:4999/assets/_vinext_fonts/geist-4db05770f54f/geist-8e42e564.woff2
HTTP/1.1 200 OK
Content-Type: font/woff2
Cache-Control: public, max-age=31536000, immutable
ETag: "99db0b97ae3a4012f8b03153490d9518"
CF-Cache-Status: HIT
```

The `writeBundle` copy hook landed the cached files in `dist/client/assets/_vinext_fonts/geist-4db05770f54f/` and the existing `_headers` rule for `/assets/*` applies an immutable cache-control header at the edge — workerd returned `CF-Cache-Status: HIT` on the second request, identical to every other hashed asset.

## References

- #472 — `assetPrefix` / `basePath` support in `next.config` (orthogonal: this PR does not implement `assetPrefix`, it just stops the absolute filesystem path from leaking into the preload emitters regardless of whether `assetPrefix` is set)
- #812 — parallel `applyMiddlewareRequestHeaders` sealed-snapshot invalidation fix that followed the same report-and-fixture pattern

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -77,6 +77,7 @@ import {
   createLocalFontsPlugin,
   _findBalancedObject,
   _findCallEnd,
+  _rewriteCachedFontCssToServedUrls,
 } from "./plugins/fonts.js";
 import { hasWranglerConfig, formatMissingCloudflarePluginError } from "./deploy.js";
 import { computeLazyChunks } from "./utils/lazy-chunks.js";
@@ -4205,6 +4206,6 @@ export { hasMdxFiles as _hasMdxFiles };
 export { _mdxScanCache };
 export { scanPublicFileRoutes as _scanPublicFileRoutes };
 export { parseStaticObjectLiteral as _parseStaticObjectLiteral };
-export { _findBalancedObject, _findCallEnd };
+export { _findBalancedObject, _findCallEnd, _rewriteCachedFontCssToServedUrls };
 export { stripServerExports as _stripServerExports };
 export { asyncHooksStubPlugin as _asyncHooksStubPlugin };
--- a/packages/vinext/src/plugins/fonts.ts
+++ b/packages/vinext/src/plugins/fonts.ts
@@ -80,21 +80,37 @@ export const VINEXT_FONT_URL_NAMESPACE = "_vinext_fonts";
  * (e.g. `/home/user/project/.vinext/fonts/geist-<hash>/geist-<hash>.woff2`)
  * and any production request fetches `<origin>/home/user/...` → 404.
  *
+ * `assetsDir` must match whatever Vite has resolved for
+ * `build.assetsDir` on the client environment — otherwise the embedded
+ * CSS URLs and the files emitted by the `writeBundle` hook would diverge
+ * and a user who customizes `build.assetsDir` (e.g. to `"static"`) would
+ * see 404s on every preload. The call site in `injectSelfHostedCss`
+ * passes the resolved value through from plugin state. The default is
+ * kept only so the exported helper can be driven directly from unit
+ * tests without synthesizing a full plugin context.
+ *
  * Uses split/join rather than regex because `cacheDir` is an absolute
  * filesystem path that may contain regex metacharacters on unusual
  * filesystems.
  */
-export function _rewriteCachedFontCssToServedUrls(css: string, cacheDir: string): string {
+export function _rewriteCachedFontCssToServedUrls(
+  css: string,
+  cacheDir: string,
+  assetsDir: string = DEFAULT_ASSETS_DIR,
+): string {
   if (!cacheDir || !css.includes(cacheDir)) return css;
-  return css.split(cacheDir).join(`/${DEFAULT_ASSETS_DIR}/${VINEXT_FONT_URL_NAMESPACE}`);
+  const prefix = assetsDir || DEFAULT_ASSETS_DIR;
+  return css.split(cacheDir).join(`/${prefix}/${VINEXT_FONT_URL_NAMESPACE}`);
 }
 
 /**
- * Default Vite `build.assetsDir` — mirrors Vite's own default. Used to
- * build the served URL prefix in `_rewriteCachedFontCssToServedUrls` so
- * the constant does not depend on plugin state. The matching `writeBundle`
- * hook reads the real `envConfig.build.assetsDir` from Vite and will
- * fall back to this value when it is unset.
+ * Default Vite `build.assetsDir` — mirrors Vite's own default. Used as
+ * the fallback for the `assetsDir` parameter of
+ * `_rewriteCachedFontCssToServedUrls` so the exported helper can be unit
+ * tested without synthesizing plugin state. Production call sites thread
+ * the real `envConfig.build.assetsDir` resolved by Vite through so that
+ * the embedded CSS URLs always match the directory the `writeBundle`
+ * hook copies the font files into.
  */
 const DEFAULT_ASSETS_DIR = "assets";
 
@@ -610,6 +626,14 @@ export function createGoogleFontsPlugin(fontGoogleShimPath: string, shimsDir: st
         if (!code.includes("next/font/google")) return null;
         if (id.startsWith(shimsDir)) return null;
 
+        // Read the resolved `build.assetsDir` from the current Vite
+        // environment so it can be closed over by the inner
+        // `injectSelfHostedCss` helper (a plain function declaration
+        // where `this` is untyped). Captured at the top of the hook so
+        // a single handler invocation always threads one consistent
+        // value through every font-loader call site it rewrites.
+        const transformAssetsDir = this.environment?.config?.build?.assetsDir ?? DEFAULT_ASSETS_DIR;
+
         const s = new MagicString(code);
         let hasChanges = false;
         let proxyImportCounter = 0;
@@ -771,7 +795,20 @@ export function createGoogleFontsPlugin(fontGoogleShimPath: string, shimsDir: st
           // the browser can actually resolve. The plugin's writeBundle hook
           // copies the referenced font files to the matching location under
           // the client output directory so the URLs serve 200s, not 404s.
-          const servedCSS = _rewriteCachedFontCssToServedUrls(localCSS, cacheDir);
+          //
+          // `transformAssetsDir` is captured at the top of the outer
+          // transform handler (where `this.environment` is bound by
+          // Rollup to the plugin context) and closed over here. This
+          // keeps the embedded URL prefix in lockstep with the directory
+          // the writeBundle hook copies files into, so a user who
+          // customizes `build.assetsDir` (e.g. to `"static"`) sees both
+          // the CSS and the copy target move together — otherwise the
+          // rewritten URLs would 404 in production.
+          const servedCSS = _rewriteCachedFontCssToServedUrls(
+            localCSS,
+            cacheDir,
+            transformAssetsDir,
+          );
 
           // Inject _selfHostedCSS into the options object
           const escapedCSS = JSON.stringify(servedCSS);
@@ -879,12 +916,27 @@ export function createGoogleFontsPlugin(fontGoogleShimPath: string, shimsDir: st
       handler(outputOptions: { dir?: string }) {
         // Only copy on the client build — the server/SSR environments
         // don't serve static assets.
+        //
+        // Optional chaining on `this.environment` matches the convention
+        // used by the other build-time plugins in `src/index.ts` (the
+        // `vinext:precompress` and `vinext:cloudflare-build` plugins both
+        // guard on `this.environment?.name !== "client"`). Vite 6+ always
+        // populates `this.environment` inside writeBundle, but keeping
+        // the guard makes the hook safely no-op if the code is ever
+        // executed in a context where Rollup invokes it without a bound
+        // environment (e.g. a thin unit test harness that invokes the
+        // hook directly). Concretely: under normal Vite builds this
+        // always resolves, the early-return is never taken.
         if (this.environment?.name !== "client") return;
         if (!cacheDir || !fs.existsSync(cacheDir)) return;
         const outDir = outputOptions.dir;
         if (!outDir) return;
 
-        const assetsDir = this.environment?.config?.build?.assetsDir ?? DEFAULT_ASSETS_DIR;
+        // Read the resolved `build.assetsDir` from the same environment
+        // that the transform-time rewrite read it from, so the embedded
+        // URL prefix and the physical copy location cannot diverge even
+        // if a user customizes `build.assetsDir`.
+        const assetsDir = this.environment.config?.build?.assetsDir ?? DEFAULT_ASSETS_DIR;
         const targetRoot = path.join(outDir, assetsDir, VINEXT_FONT_URL_NAMESPACE);
 
         // Recursive copy of every cached font file. Skip the companion
--- a/packages/vinext/src/plugins/fonts.ts
+++ b/packages/vinext/src/plugins/fonts.ts
@@ -45,6 +45,59 @@ export const GOOGLE_FONT_UTILITY_EXPORTS = new Set([
   "createFontLoader",
 ]);
 
+/**
+ * Served URL prefix for self-hosted Google Font files.
+ *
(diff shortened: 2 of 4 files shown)


Shorebirdmgmt:
Thanks for the review. Pushed 93af737 addressing all three points.

### 1. `build.assetsDir` divergence — fixed

Threaded the resolved `assetsDir` through from plugin state. `_rewriteCachedFontCssToServedUrls` now takes an optional third parameter (defaulting to `DEFAULT_ASSETS_DIR` so the exported helper still drives cleanly from unit tests without a plugin context). At the call site in `injectSelfHostedCss`, I read `this.environment?.config?.build?.assetsDir ?? DEFAULT_ASSETS_DIR` once at the top of the outer transform `handler` and close over it — doing this in the inner `injectSelfHostedCss` declaration directly triggered `TS2683: 'this' implicitly has type 'any'` because function declarations shadow `this`. The `writeBundle` hook already read the real value, so both call sites now consume the same source and cannot diverge by construction.

**End-to-end verification on Miniflare** with `examples/app-router-cloudflare` and `build: { assetsDir: "static" }`:

```
$ find dist/client -name '*.woff2'
dist/client/static/_vinext_fonts/geist-4db05770f54f/geist-8e42e564.woff2
dist/client/static/_vinext_fonts/geist-4db05770f54f/geist-bd9fc9d8.woff2
dist/client/static/_vinext_fonts/geist-4db05770f54f/geist-79e7a99c.woff2

$ curl -sI http://localhost:4997/ | grep -i '^link:'
Link: </static/_vinext_fonts/geist-4db05770f54f/geist-8e42e564.woff2>; rel=preload; as=font; type=font/woff2; crossorigin, </static/_vinext_fonts/geist-4db05770f54f/geist-bd9fc9d8.woff2>; rel=preload; as=font; type=font/woff2; crossorigin, </static/_vinext_fonts/geist-4db05770f54f/geist-79e7a99c.woff2>; rel=preload; as=font; type=font/woff2; crossorigin

$ curl -sI http://localhost:4997/static/_vinext_fonts/geist-4db05770f54f/geist-8e42e564.woff2
HTTP/1.1 200 OK
Content-Type: font/woff2
Cache-Control: public, max-age=31536000, immutable
ETag: "60f7c478fdf66f978f14ec8a5d2abfff"
CF-Cache-Status: HIT
```

Two new unit test cases cover the custom `assetsDir` path and the empty-string fallback (so the URL can never acquire a `//` segment if a misconfigured environment passes `""`).

**One caveat worth flagging separately**: while investigating I noticed `server/prod-server.ts` hardcodes `pathname.startsWith("/assets/")` at lines 995, 1286, and 1551 as the static-file fast path. That's a pre-existing limitation affecting every custom `build.assetsDir` user, not just fonts — on the Node.js production server a custom `assetsDir` currently 404s for all hashed assets, not just for the rewritten font URLs. Fixing that is out of scope for this PR since it predates the font leak and touches an unrelated codepath, but happy to file a follow-up issue if useful. On Cloudflare Workers (the primary deployment target) the asset binding serves the entire `dist/client/` tree recursively, so the custom-`assetsDir` case works end-to-end today, as the Miniflare capture above shows.

### 2. Optional chaining on `this.environment` — documented

Left the optional chaining in place and added a comment explaining: it matches the existing convention in `vinext:precompress` and `vinext:cloudflare-build` (both in `src/index.ts`), `this.environment` is always populated during a real Vite build, the guard is there so the hook safely no-ops if invoked outside a bound environment (e.g. from a thin unit-test harness), and the early-return path is unreachable during a normal build. The same comment clarifies why the `writeBundle` hook reads `this.environment.config.build.assetsDir` at hook time rather than via a closure from `configResolved` — Vite's top-level `configResolved` receives the global config, but per-environment `build.assetsDir` may differ, so reading from `this.environment.config` at hook time is the only way to get the value Vite actually used for the current environment.

### 3. Fetch mock `String(input)` — hardened

Replaced with a small `resolveFetchUrl` helper that handles `string`, `URL`, and `Request` inputs explicitly before falling back to `String()`. As you noted, the build plugin always passes string URLs today so this is theoretical, but it prevents a future vinext internals change or test helper from silently making the mock fall through to a real network fetch for Google Fonts.

### Verification matrix (re-run with follow-up)

| Check | Result |
|---|---|
| `pnpm test:unit` | **2855 / 2855** pass (+2 new `assetsDir` cases) |
| `pnpm test:integration` | **1262 / 1262** pass (2 pre-existing skips) |
| `pnpm run check` | **0** warnings, **0** errors |
| `pnpm test:e2e` — `app-router` | **322 / 322** pass (2 flaky on retry, 9 skipped) |
| `pnpm test:e2e` — `cloudflare-workers` (Miniflare) | **38 / 38** pass |

Shorebirdmgmt:
Both non-blocking nits addressed in a5c2b72 to close the review trail cleanly.

### 1. Stale `fetchAndCacheFont` comment

Replaced the old `Vite will resolve /@fs/ for dev, or asset for build` comment — that was the incorrect assumption that seeded this bug in the first place — with an explicit description of the handoff: `fetchAndCacheFont` writes absolute paths as the on-disk intermediate form, `_rewriteCachedFontCssToServedUrls()` rewrites the prefix at embed time, and the filesystem path must never reach the bundle, the injected `<style data-vinext-fonts>` block, the body `<link rel="preload">` tags, or the HTTP `Link:` header. New comment ends with an explicit "do not resurrect that assumption" warning for future maintainers.

### 2. `fetch(input, init)` signature preserved in the test mock

The mock signature is now `async (input: unknown, init?: RequestInit) =>` and the fallback forwards `init` to `originalFetch(input as RequestInfo, init)`. Consistent with the `resolveFetchUrl` hardening from the previous round — future-proofing the mock against a vinext internals change or test helper that relies on headers/method in the fallback path.

Both changes are surgical (comment-only + 2-line signature fix). Verified:

- `pnpm run check` — 0 warnings, 0 errors, all 919 files formatted
- `pnpm test:unit -t "_rewriteCachedFontCssToServedUrls"` — 7/7 pass (unchanged)
- `pnpm test:integration -t "self-hosted next/font/google"` — 4/4 pass (unchanged)

Thanks for the thorough review — approval is unchanged by these nits, landing them just to keep the codebase tidy and the intent explicit for the next person who touches this pipeline.

PR #822, opened 2026-04-11
fix: always mark pages that receive searchParams as dynamic

southpolesteve:
## Summary

Pages that read `searchParams` are now always marked dynamic, even when the first request has an empty query string.

## Details

Previously, `markDynamicUsage()` was only called when the URL contained query parameters (`hasSearchParams`). A request to `/page` (no query) would render with `searchParams = {}`, not be marked dynamic, and get ISR-cached. A later request to `/page?filter=violet` would then receive the cached empty-query response.

The fix removes the `hasSearchParams` guard so `markDynamicUsage()` is called unconditionally whenever `searchParams` is provided to the page component. This matches Next.js behavior where accessing the `searchParams` prop signals dynamic rendering regardless of whether the URL has query parameters.

## Tests

Adds a production integration test with a dedicated fixture page (`/search-params-page`) that reads `searchParams`:
- first request with no query: not cached, renders `filter=none`
- second request with `?filter=violet`: not cached, renders `filter=violet`

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -1035,15 +1035,15 @@ async function buildPageElements(route, params, routePath, opts, searchParams, i
     // when the query string is empty -- pages that do "await searchParams" need
     // it to be a thenable rather than undefined.
     pageProps.searchParams = makeThenableParams(spObj);
-    // If the URL has query parameters, mark the page as dynamic.
-    // In Next.js, only accessing the searchParams prop signals dynamic usage,
-    // but a Proxy-based approach doesn't work here because React's RSC debug
-    // serializer accesses properties on all props (e.g. $$typeof check in
-    // isClientReference), triggering the Proxy even when user code doesn't
-    // read searchParams. Checking for non-empty query params is a safe
-    // approximation: pages with query params in the URL are almost always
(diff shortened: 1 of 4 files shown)


packages/vinext/src/entries/app-rsc-entry.ts
```diff
@@ -1035,15 +1035,15 @@ async function buildPageElements(route, params, routePath, opts, searchParams, i
...
-    // dynamic, and this avoids false positives from React internals.
-    if (hasSearchParams) markDynamicUsage();
+    // Mark the render as dynamic whenever searchParams is provided. In Next.js,
+    // accessing the searchParams prop via a Proxy triggers dynamic rendering.
+    // We cannot use a Proxy here because React's RSC debug serializer accesses
+    // properties on all props ($$typeof, etc.), causing false positives.
+    // Instead, we unconditionally mark dynamic: any page that receives
+    // searchParams is request-dependent, even when the first request has an
+    // empty query string. Without this, the empty-query render would be cached
+    // and replayed to later requests with different query parameters.
```
james-elicx, line 1045: Would it be possible to filter out those internals in a proxy instead? If so, we could use the proxy approach and align behaviour with Next.js.

This change as-is would mark every route we give search params to as dynamic, regardles of whether they use search params or not.

PR #827, opened 2026-04-13
chore: remove 'use server' collision workaround plugin

james-elicx:
fixed upstream and version with fix is already a dep


james-elicx:
/bigbonk re-review - i removed that test file

PR #835, opened 2026-04-13
fix: return error for client hook used without directive

### LINKED ISSUE #834: Some hooks silently become undefined without an error when 'use client' is missing
## Description

When 'use client' is removed from a file, certain hooks that depend on it silently resolve to undefined — with no warning or error message from vinext.

## Steps to Reproduce
1. Create a component that uses a hook requiring 'use client' (e.g. useState, useRouter)
2. Remove the 'use client' directive at the top of the file
3. Observe that the hook returns undefined silently

## Expected Behavior
vinext should throw an error (similar to Next.js) indicating that 'use client' is required for the given hook to work.

## Actual Behavior
No error or warning is shown. The hook silently returns undefined, making the root cause difficult to diagnose.

## Screenshots
Next.js:
Image

vinext:
Image

## Additional Context
Next.js surfaces this mistake as an explicit error. vinext currently provides no equivalent feedback, which significantly degrades the developer experience when debugging missing 'use client' directives.

james-elicx:
fixes #834

**Summary**
- Hooks like usePathname() from next/navigation now throw a clear error when called from a Server Component without the "use client" directive, matching Next.js behavior
- Previously these hooks silently returned fallback values, making the root cause extremely difficult to diagnose
- Uses the same react-server export condition pattern as Next.js: the RSC Vite environment resolves next/navigation to a server-safe variant that provides error-throwing stubs for client-only hooks

**Approach**
A new navigation.react-server.ts shim re-exports server-safe APIs (redirect, notFound, setNavigationContext, etc.) from the canonical navigation.ts and adds error-throwing stubs for client-only hooks. The resolveId hook routes next/navigation to this variant in the RSC environment and to the full shim elsewhere.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -354,16 +354,18 @@ const _shimsDir = path.resolve(__dirname, "shims") + "/";
 const _fontGoogleShimPath = resolveShimModulePath(_shimsDir, "font-google");
 
 /**
- * Shims that have a `.react-server.ts` variant for the RSC environment.
- * Maps every import specifier that should be intercepted → the base shim name
- * (without extension). In the RSC env the resolveId hook appends
- * `.react-server`; in other envs it resolves to the base name.
+ * Shims with a `.react-server.ts` variant for the RSC environment.
+ * Maps import specifier → base shim name. In the RSC env, resolveId
+ * appends `.react-server`; in other envs it resolves to the base.
+ *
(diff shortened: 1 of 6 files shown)


Bananamilk452:
does this PR works only for next/navigation hooks? or all the other client only hooks?

james-elicx:
Ah sorry @Bananamilk452, I read useRouter in your issue and asked the agent to do next/navigation stuff. I'll sort out React hooks in https://github.com/cloudflare/vinext/pull/844

PR #847, opened 2026-04-15
fix(plugins): resolve optimized imports past local barrels

### LINKED ISSUE #845: App Router build can still reach antd barrel exports despite default optimizePackageImports, causing `[plugin rsc:use-client] unsupported ExportAllDeclaration`
### Summary

`vinext build` for an App Router project can still fail on `antd` barrel exports even though `antd` is in vinext's default `optimizePackageImports` set.

In my case, the build failed inside `@vitejs/plugin-rsc` with:

```txt
[plugin rsc:use-client] .../node_modules/antd/es/button/index.js
Error: unsupported ExportAllDeclaration
```

After tracing both vinext and `@vitejs/plugin-rsc` source, this looks like a vinext compatibility gap rather than just a raw upstream `plugin-rsc` bug:

- vinext already claims support for `experimental.optimizePackageImports`
- vinext's default optimized package list already includes `antd`
- vinext already introduced `vinext:optimize-imports` specifically to avoid RSC-unfriendly barrel import paths
- but in this case the import optimization did not fully shield the dependency chain, and `plugin-rsc` still reached a client-boundary barrel module containing `export *`

So the actionable vinext issue seems to be:

> the current optimize-imports path is not sufficient to prevent some `antd` subpath barrels from leaking into the App Router RSC pipeline.

---

### Environment

- Framework mode: App Router
- Build phase: production build (`vinext build`)
- Package manager: npm
- React: 19.x
- vinext: current release at the time of testing
- Vite: 8.x
- `@vitejs/plugin-rsc`: installed through vinext App Router setup
- UI library: `antd`

---

### Expected behavior

`vinext build` should succeed for a normal App Router project importing components from `antd`, especially since:

- `antd` is listed in vinext's default optimized package set
- vinext documents `experimental.optimizePackageImports` as supported in RSC/SSR environments
- this class of barrel-import problem was already addressed by `vinext:optimize-imports`

At minimum, vinext should rewrite the import chain deeply enough that `plugin-rsc` does not end up analyzing a `"use client"` barrel module with `ExportAllDeclaration`.

---

### Actual behavior

The App Router build reached `antd/es/button/index.js`, and `@vitejs/plugin-rsc` failed with:

```txt
Error: unsupported ExportAllDeclaration
```

That failure is thrown by `plugin-rsc`'s `"use client"` transform when it sees `ExportAllDeclaration` while generating client-reference metadata.

---

### Why I believe this is actionable in vinext

I understand the immediate throw site is upstream in `@vitejs/plugin-rsc`, but the reason I am filing this in vinext is:

1. vinext explicitly advertises support for `experimental.optimizePackageImports`
2. vinext includes `antd` in the default optimized package set
3. vinext already has a dedicated `vinext:optimize-imports` plugin intended to avoid exactly this class of RSC/barrel problems
4. therefore, if a standard `antd` usage path still leaks a barrel like `antd/es/button/index.js` into `plugin-rsc`, that is still a vinext compatibility gap even if the final throw comes from upstream

In other words, I am **not** claiming vinext is the original source of the exception text. I am claiming vinext currently does not fully shield this dependency graph even though it appears intended to.

---

### Root-cause analysis

After reading the vinext, Next.js, and `@vitejs/plugin-rsc` sources, my current understanding is:

#### 1. `plugin-rsc` does not support `ExportAllDeclaration` in the `"use client"` client-reference pipeline

The hard failure is in `@vitejs/plugin-rsc`:

- `packages/plugin-rsc/src/plugin.ts`
- `packages/plugin-rsc/src/transforms/proxy-export.ts`
- `packages/plugin-rsc/src/transforms/utils.ts`
- `packages/plugin-rsc/src/transforms/wrap-export.ts`

The relevant behavior is effectively:

```ts
if (node.type === 'ExportAllDeclaration') {
  throw new Error('unsupported ExportAllDeclaration')
}
```

So if `plugin-rsc` reaches a `"use client"` barrel module containing `export *`, the build fails.

#### 2. vinext does try to mitigate this

vinext already has:

- `vinext:optimize-imports`
- a default optimized package list that includes `antd`
- documentation stating barrel imports are rewritten to direct sub-module imports in RSC/SSR environments

This suggests vinext is already aware of exactly this category of problem.

#### 3. The likely gap is that the rewrite is not deep enough

My understanding is that vinext rewrites top-level source imports such as:

```ts
import { Button } from 'antd'
```

into a more specific subpath import.

However, in this failure mode that still appears to leave a later subpath barrel in the graph, e.g. a path resolving to something like:

```txt
antd/es/button/index.js
```

If that subpath is itself a `"use client"` barrel with `export *`, then `plugin-rsc` still sees it and throws.

So the problem is not that optimizePackageImports is disabled for `antd`.
The problem seems to be that the current vinext optimization path does not fully eliminate all problematic barrel layers before `plugin-rsc` analyzes the module.

---

### Related vinext issues / prior work

This report seems closely related to existing vinext work in this area:

- #100
- #137
- #138

My understanding is:

- #100 reported barrel-import-related RSC failures
- #137 proposed a vinext-side optimization approach
- #138 implemented `vinext:optimize-imports`

This new report appears to be a remaining gap / edge case in the same problem family rather than a completely unrelated bug.

If useful, I can also open a companion upstream issue in `vitejs/vite-plugin-react`, because the immediate unsupported AST node is clearly there as well.

---

### Repro characteristics

I do not currently have a minimized public reproduction repo prepared yet, but the failure pattern is straightforward:

1. Start with a vinext App Router project
2. Use `antd`
3. Build with `vinext build`
4. The RSC pipeline eventually reaches an `antd` client-boundary barrel module
5. Build fails with `unsupported ExportAllDeclaration`

If a minimal reproduction is required, I can prepare one.

---

### What I think would fix it

Any of these would probably address the issue on the vinext side:

1. Make `vinext:optimize-imports` flatten `antd` imports more aggressively so the graph does not stop at subpath barrels like `antd/es/button`
2. Add additional package-specific handling for `antd` barrel chains in App Router RSC builds
3. Ensure the rewritten path lands on the concrete implementation module rather than a secondary barrel
4. Otherwise document this as a known gap and recommend a temporary workaround until upstream `plugin-rsc` supports `ExportAllDeclaration`

---

### Additional context

I realize this may ultimately require an upstream fix in `@vitejs/plugin-rsc`, but from a user perspective:

- vinext claims `antd` is default-optimized
- vinext claims barrel imports are rewritten in RSC/SSR
- App Router build still fails on an `antd` barrel export

That makes this worth tracking in vinext as a compatibility bug, even if the final long-term fix is shared with upstream.

Thanks — happy to provide a minimal reproduction if that helps.

llc1123:
## Summary
- fix `optimize-imports` so optimized package imports keep following local re-export chains to the concrete leaf module instead of stopping at an intermediate barrel like `antd/es/button/index.js`
- flatten local `export *` boundaries inside optimized packages before downstream graph analysis so `@vitejs/plugin-rsc` no longer sees the issue-845 client-boundary shape
- add both a targeted transform regression and an App Router production-build regression covering the `antd` repro path

## Validation
- `./node_modules/.bin/vp test run tests/optimize-imports.test.ts -t "multi-hop antd barrel"`
- `./node_modules/.bin/vp test run tests/optimize-imports-build.test.ts`

## Issue
Fixes #845.

## Current solution
The bug happened because `vinext:optimize-imports` used to stop at an intermediate local barrel (`antd/es/button/index.js`). That file is a `"use client"` boundary with `export *`, so `@vitejs/plugin-rsc` later failed with `unsupported ExportAllDeclaration`.

The fix keeps the rewrite generic rather than `antd`-specific:
- recursively resolve local re-exports until the concrete implementation file is reached
- safely flatten local wildcard re-exports inside optimized packages before plugin-rsc inspects them
- fall back to a conservative direct-export map when parsing a JSX-bearing `.js` leaf fails, so the local-barrel walk can still reach the leaf module

This means the app import now rewrites to the concrete leaf module instead of leaking the intermediate `export *` barrel into the RSC build.

### DIFF
--- a/packages/vinext/src/plugins/optimize-imports.ts
+++ b/packages/vinext/src/plugins/optimize-imports.ts
@@ -492,7 +492,7 @@ async function buildExportMapFromFile(
   if (cached) return cached;
 
   if (inProgress.has(filePath)) {
-    return cache.get(filePath) ?? new Map();
+    return new Map();
   }
   inProgress.add(filePath);
 
@@ -974,10 +974,6 @@ export function createOptimizeImportsPlugin(
               const source = entry.source.startsWith("/")
                 ? toRelativeModuleSpecifier(normalizedId, entry.source)
                 : entry.source;
-              if (!source) {
-                canRewrite = false;
-                break;
-              }
               const key = `${source}::${entry.isNamespace}`;
               let group = bySource.get(key);
               if (!group) {
@@ -1240,7 +1236,7 @@ export function createOptimizeImportsPlugin(
 
         return {
           code: s.toString(),
-          map: null,
+          map: s.generateMap({ hires: "boundary" }),
         };
       },
     },
--- a/packages/vinext/src/plugins/optimize-imports.ts
+++ b/packages/vinext/src/plugins/optimize-imports.ts
@@ -64,6 +64,8 @@ type DeclarationNode = {
   declarations?: Array<{ id: { name: string } }>;
 };
 
+type ReadFileFn = (filepath: string) => Promise<string | null>;
+
 /** Caches used by the optimize-imports plugin, scoped to a plugin instance. */
 type BarrelCaches = {
   /** Barrel export maps keyed by resolved entry file path. */
@@ -106,6 +108,131 @@ type AstBodyNode = {
 // This cast type is used consistently across resolveId and transform handlers
 // so that when Vite adds proper typing it can be removed in one place.
 type PluginCtx = { environment?: { name?: string } };
+
+function localModuleCandidates(modulePath: string): string[] {
+  return [
+    modulePath,
+    `${modulePath}.js`,
+    `${modulePath}.mjs`,
+    `${modulePath}.cjs`,
+    `${modulePath}.ts`,
+    `${modulePath}.tsx`,
+    `${modulePath}.jsx`,
+    `${modulePath}.mts`,
+    `${modulePath}.cts`,
+    `${modulePath}/index.js`,
+    `${modulePath}/index.mjs`,
+    `${modulePath}/index.cjs`,
+    `${modulePath}/index.ts`,
+    `${modulePath}/index.tsx`,
+    `${modulePath}/index.jsx`,
+    `${modulePath}/index.mts`,
+    `${modulePath}/index.cts`,
+  ];
+}
+
+function toRelativeModuleSpecifier(fromFile: string, toFile: string): string {
+  const relativePath = path.relative(path.dirname(fromFile), toFile).split(path.sep).join("/");
+  return relativePath.startsWith(".") ? relativePath : `./${relativePath}`;
+}
+
+function findOptimizedPackageForFile(id: string, packages: Set<string>): string | null {
+  const normalizedId = id.split("?")[0].split(path.sep).join("/");
+  let match: string | null = null;
+  for (const pkg of packages) {
+    if (
+      normalizedId.includes(`/node_modules/${pkg}/`) &&
+      (match === null || pkg.length > match.length)
+    ) {
+      match = pkg;
+    }
+  }
+  return match;
+}
+
+async function resolveLocalModuleFile(
+  modulePath: string,
+  readFile: ReadFileFn,
+): Promise<{ filePath: string; content: string } | null> {
+  for (const candidate of localModuleCandidates(modulePath)) {
+    const content = await readFile(candidate);
+    if (content !== null) {
+      return { filePath: candidate, content };
+    }
+  }
+  return null;
+}
+
+function buildFallbackExportMap(filePath: string, content: string): BarrelExportMap {
+  const exportMap: BarrelExportMap = new Map();
+
+  const recordDirectExport = (exportName: string, originalName = exportName) => {
+    exportMap.set(exportName, {
+      source: filePath,
+      isNamespace: false,
+      originalName,
+    });
+  };
+
+  const declarationPatterns = [
+    /^\s*export\s+(?:async\s+)?function\s+([A-Za-z_$][\w$]*)/gm,
+    /^\s*export\s+class\s+([A-Za-z_$][\w$]*)/gm,
+    /^\s*export\s+(?:const|let|var)\s+([A-Za-z_$][\w$]*)/gm,
+  ];
+  for (const pattern of declarationPatterns) {
+    for (const match of content.matchAll(pattern)) {
+      recordDirectExport(match[1]);
+    }
+  }
+
+  for (const match of content.matchAll(/^\s*export\s+default\s+([A-Za-z_$][\w$]*)\s*;?/gm)) {
+    recordDirectExport("default", "default");
+    if (!exportMap.has(match[1])) {
+      recordDirectExport(match[1]);
+    }
+  }
+
+  return exportMap;
+}
+
+async function buildSafeWildcardExportMap(
+  filePath: string,
+  readFile: ReadFileFn,
+  cache: Map<string, BarrelExportMap>,
+  visited = new Set<string>(),
+): Promise<BarrelExportMap | null> {
+  if (visited.has(filePath)) return null;
+  visited.add(filePath);
+
+  const content = await readFile(filePath);
+  if (!content) return null;
+
+  let ast: ReturnType<typeof parseAst>;
+  try {
+    ast = parseAst(content);
+  } catch {
+    const fallbackMap = buildFallbackExportMap(filePath, content);
+    return fallbackMap.size > 0 ? fallbackMap : null;
+  }
+
+  const fileDir = path.dirname(filePath);
+  for (const node of ast.body as AstBodyNode[]) {
+    if (node.type !== "ExportAllDeclaration" || node.exported) continue;
+    const rawSource = typeof node.source?.value === "string" ? node.source.value : null;
+    if (!rawSource || !rawSource.startsWith(".")) return null;
+
+    const resolved = await resolveLocalModuleFile(
+      path.resolve(fileDir, rawSource).split(path.sep).join("/"),
+      readFile,
+    );
+    if (!resolved) return null;
+
+    const nestedMap = await buildSafeWildcardExportMap(resolved.filePath, readFile, cache, visited);
+    if (!nestedMap) return null;
+  }
+
+  return buildExportMapFromFile(filePath, readFile, cache, new Set<string>(), content);
+}
 
 /**
  * Packages whose barrel imports are automatically optimized.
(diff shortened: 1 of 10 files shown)


llc1123:
Addressed the review items in three focused commits:

- `fix(plugins): parse JSX before optimize-import rewriting`
  - uses JSX-aware `parseAst` handling for `.tsx`/`.jsx` inputs while keeping the existing fallback behavior for unparseable `.js` leaf modules
  - deduplicates the inline local module candidate list via `localModuleCandidates()`
- `test: cover JSX-bearing optimize-import transforms`
  - adds a real TSX user-file regression so optimize-import rewriting is exercised with JSX in the transformed source
- `test: strengthen optimize-imports build assertions`
  - upgrades the build regression to read bundled output and assert the barrel shape no longer leaks into the build

Validation rerun:
- `./node_modules/.bin/vp test run tests/optimize-imports.test.ts -t "issue-845 multi-hop antd barrel|keeps optimizing issue-845 imports from JSX-bearing TSX user files"`
- `./node_modules/.bin/vp test run tests/optimize-imports-build.test.ts`

llc1123:
Follow-up fixes for the latest review:

- `fix(plugins): avoid phantom fallback exports`
  - prevents `buildFallbackExportMap` from recording phantom `"function"` / `"class"` exports when matching `export default function ...` fallback cases
  - clarifies that `buildSafeWildcardExportMap` bails on the entire file when any wildcard target is external
  - adds a focused regression covering the fallback path with a JSX-bearing `.js` module so we exercise the exact parse-failure shape under discussion

Validation rerun:
- `./node_modules/.bin/vp test run tests/optimize-imports.test.ts -t "does not record phantom function/class exports in fallback default-export parsing|issue-845 multi-hop antd barrel|keeps optimizing issue-845 imports from JSX-bearing TSX user files"`
- `./node_modules/.bin/vp test run tests/optimize-imports-build.test.ts`

I left the extra traversal note and the wildcard-merge `resolveConcreteLocalEntry` gap as follow-up scope for now, since they were called out as non-blocking / future-shape concerns rather than regressions in the current issue-845 path.

llc1123:
Follow-up update for the CI/unit failure:

I tracked the unit failure down to the optimize-imports export-map path rather than a single broken assertion.

What changed:
- `fix(plugins): guard export-map reentry during barrel analysis`
  - adds an `inProgress` guard in `buildExportMapFromFile(...)` so recursive barrel analysis does not re-enter the same file while its export map is still being built
- `test: split optimize-imports transform coverage`
- `test: split optimize-imports export-map basics`
- `test: split optimize-imports export-map recursion cases`
  - splits the old `tests/optimize-imports.test.ts` coverage into smaller focused files so the unit worker no longer has to load the entire optimize-imports surface into one file

Also synced a couple of expectations to current behavior while splitting:
- concrete local-file rewrites now assert explicit `.js` paths where that is the current normalized output
- the client-environment transform test now reflects the current Next.js-aligned behavior: direct concrete-file rewrites are still allowed on the client side for optimized packages

Validation rerun:
- `./node_modules/.bin/vp test run --project unit`

Result:
- full unit project passes locally again (`94` test files / `2895` tests)

llc1123:
Follow-up fix for the remaining CI blocker:

- `test: add local fixture for client optimize-imports path`
  - makes the client-environment lucide-react test self-contained instead of relying on an ambient workspace install
  - the test now creates a tiny local `lucide-react` barrel fixture under the temp app root and verifies the client transform rewrites to the concrete `dist/esm/icons/sun.js` leaf

This keeps the test aligned with the intended Next.js-style behavior we discussed:
- client-side optimize-package rewrites are still allowed when they resolve directly to concrete local files
- bare sub-package rewrites are still left alone for the client graph

Validation rerun:
- `./node_modules/.bin/vp test run tests/optimize-imports.test.ts -t "rewrites direct barrel imports on client environment when they resolve to concrete files"`
- `./node_modules/.bin/vp test run --project unit`

Result:
- unit project passes locally again (`94` test files / `2895` tests)

llc1123:
Follow-up for the last four inline comments:

- `fix(plugins): restore optimize-imports sourcemaps`
  - restores `s.generateMap({ hires: "boundary" })` in the optimize-imports transform so it matches the repo's other MagicString-based transforms
  - removes the dead `!source` guard after `toRelativeModuleSpecifier(...)`
  - simplifies the reachable-but-redundant `cache.get(filePath) ??` branch in the `inProgress` guard to `new Map()`
- `tests/optimize-imports-build.test.ts`
  - moves the temp app back to `os.tmpdir()` like the rest of the test suite
  - keeps the test self-contained by symlinking the repo `node_modules` into the temp app, matching the other build-test pattern so JSX runtime resolution still works

Validation rerun:
- `./node_modules/.bin/vp fmt --check packages/vinext/src/plugins/optimize-imports.ts tests/optimize-imports-build.test.ts`
- `./node_modules/.bin/vp test run tests/optimize-imports-build.test.ts`
- `./node_modules/.bin/vp test run --project unit`

Result:
- full unit project still passes locally (`94` test files / `2895` tests)

llc1123:
Follow-up for the latest review comments:

- `test: isolate optimize-imports build fixture writes`
  - removes the dead `canRewrite` scaffolding in the optimize-imports transform loop
  - keeps the build test on `os.tmpdir()` but stops writing custom `antd` fixture files through a repo-wide `node_modules` symlink
  - the temp app now creates its own `node_modules/antd` fixture and only symlinks the workspace packages it actually needs for the build (`react`, `react-dom`, `react-server-dom-webpack`)

That keeps the reviewer-requested temp-dir cleanup while avoiding writes into the real workspace `node_modules`.

Validation rerun:
- `./node_modules/.bin/vp fmt --check packages/vinext/src/plugins/optimize-imports.ts tests/optimize-imports-build.test.ts`
- `./node_modules/.bin/vp test run tests/optimize-imports-build.test.ts`
- `./node_modules/.bin/vp test run --project unit`

Result:
- full unit project still passes locally (`94` test files / `2895` tests)

llc1123:
One note on the two remaining non-blocking observations that I did **not** fold into this PR:

1. **`buildFallbackExportMap` and anonymous `export default function() {}`**
   - This only affects the conservative **fallback** path, i.e. `.js` files that fail AST parsing and fall back to regex-based export recovery.
   - The normal AST path already handles anonymous default exports correctly, so this is not part of the main issue-845 path and it is not a known blocker.
   - Expanding the fallback semantics here would increase the regex / inference surface more than I want to for this PR, so I would prefer to keep that as a separate follow-up if needed.

2. **`resolveConcreteLocalEntry` breaking on the first readable candidate**
   - This is the current resolution strategy: once a readable candidate file is found, we check that file for the target export instead of continuing across later candidates.
   - As noted in review, this seems unlikely to matter in practice, and it was not part of the issue-845 failure mode or any of the CI regressions we actually hit here.
   - If we want to broaden that behavior, I think it should be done as a separate follow-up, because it changes the candidate-resolution semantics and would deserve its own tests.

For this PR I tried to keep the scope tightly focused on:
- the issue-845 correctness fix itself,
- the concrete review items that affected correctness / CI / test isolation,
- and not broadening into additional resolution-semantics changes unless they were directly required.

llc1123:
Hi! Just following up on this PR since it’s been sitting for a while.
The review feedback should all be addressed now, and the latest checks passed on my side.
If there’s anything else you’d like me to adjust before merge, I’m happy to update it. Otherwise, would you mind taking another look when you get a chance? Thanks!

llc1123:
Closing because the issue this PR targeted (#845) no longer reproduces on current vinext main. A minimal App Router + antd repro now builds and runs successfully, so this PR is no longer needed.

PR #857, opened 2026-04-17
chore: set oxfmt printWidth to 90

### LINKED ISSUE #855: 80 or 85 print width formatting
It's a lot more tasteful and pleasant to the eye

NathanDrake2406:
## Summary

- Lowers `fmt.printWidth` in `vite.config.ts` from `100` to `90` (closes #855).
- Runs `vp fmt` across the repo so the tree matches the new width. The reformat lives in its own commit so it can be added to `.git-blame-ignore-revs` if desired.
- Converts two `// oxlint-disable-next-line …` suppressions to paired `/* oxlint-disable … */` / `/* oxlint-enable … */` blocks in `plugins/strip-server-exports.ts` and `server/middleware.ts`. The single-line variants previously covered expressions that fit on one line at width 100; after wrapping to 90, only the top of the wrap remained suppressed. Block pairs are stable across any formatter width and do not introduce new suppressions.

No behavioral changes. This is a formatting-and-config PR.

Refs #855

## Test plan

- [ ] `vp check` (fmt + lint + type) passes on CI
- [ ] `vp test` passes on CI
- [ ] Playwright E2E passes on CI

Running the full suite locally isn't useful here (nothing runtime-facing changed), so CI is the source of truth.


james-elicx:
I'm afraid these things come down to personal preference, and I often lean towards a larger print width.

I generally agree with oxfmt's [default](https://oxc.rs/docs/guide/usage/formatter/config-file-reference.html#printwidth) of 100.

It would be nice if we had tooling that worked seamlessly with supporting individuals having their own formatting config locally, and a different one when pushed to a branch...

PR #878, opened 2026-04-24
[codex] Fix usePathname hydration snapshot

### LINKED ISSUE #871: `usePathname` temporarily resolves to `"/"` on the client before navigating to the correct pathname
### Description

When using `usePathname`, the server correctly logs the expected pathname (e.g. `"/home"`), but on the client side it briefly resolves to `"/"` before moving to the correct path. This mismatch causes a **hydration error**.

---

### Steps to Reproduce

1. Use `usePathname` in a component
2. Add a `console.log`
3. Navigate to a route such as `/home`
4. Observe the server logs `/home` correctly, but the client initially logs `/` before correcting to `/home`

---

### Expected Behavior

`usePathname` should return the correct pathname on both server and client without any intermediate `"/"` state.

---

### Actual Behavior

- **Server:** `usePathname` → `"/home"` ✅
- **Client (initial):** `usePathname` → `"/"` ❌
- **Client (after correction):** `usePathname` → `"/home"` ✅
- Results in a hydration mismatch error

---

### Additional Context

This may be a regression of #688 — needs verification.

---

southpolesteve:
Fixes #871.

## Summary
- Share the browser hydration navigation context through a `Symbol.for` global so split `next/navigation` module instances read the same initial snapshot.
- Keep server-side ALS access unchanged while preventing client `usePathname()`, `useSearchParams()`, and `useParams()` from falling back to `/` during hydration.
- Add a regression test that loads separate setter/hook module instances and verifies the hydrated pathname, query, and params stay in sync.

## Validation
- `pnpm test tests/shims.test.ts -t "shares the hydrated navigation snapshot"`
- `pnpm test tests/nextjs-compat/use-client-page-pathname.test.ts tests/nextjs-compat/rsc-context-lazy-stream.test.ts`

### DIFF
--- a/packages/vinext/src/shims/navigation.ts
+++ b/packages/vinext/src/shims/navigation.ts
@@ -163,8 +163,29 @@ export const GLOBAL_ACCESSORS_KEY = Symbol.for("vinext.navigation.globalAccessor
 const _GLOBAL_ACCESSORS_KEY = GLOBAL_ACCESSORS_KEY;
 type _GlobalWithAccessors = typeof globalThis & { [_GLOBAL_ACCESSORS_KEY]?: _StateAccessors };
 
+// Browser hydration has the same module-split shape as SSR in Vite dev:
+// the browser entry seeds the snapshot before hydrateRoot(), but client
+// components can import a different module instance of this shim.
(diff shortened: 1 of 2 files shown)


NathanDrake2406:
@james-elicx not approved here. I think it forgot how to approve PR

Issue #874, opened 2026-04-23
Suspense streaming doesn't seem to work with `wrangler dev`


LubomirGeorgiev:
When I run my vinext app with `vinext dev` the Suspense blocks get resolved and I can see the initial fallback which is correct.

However if I build the app and run it with `wrangler dev` the longest suspense promise hangs the whole page and the fallback is never visible.

I made a reproduction repo here https://github.com/LubomirGeorgiev/vinext-next-safe-action

To reproduce it with the correct behaviour run `pnpm dev` and open http://localhost:3000/
To reproduce the wrong behaviour  run `pnpm preview` and open http://localhost:8787/

southpolesteve:
Thanks for the reports! I suspect this may be the same root issue https://github.com/cloudflare/workers-sdk/issues/8004

LubomirGeorgiev:
@southpolesteve I just deployed it here https://vinext-nextjs.lubomirgeorgievgeorgiev.workers.dev/ and it works fine, so if this is just an issue with `wrangler dev` I guess it's not so critical.

Btw does that mean that `wrangler dev` and the cloudflare worker production environment don't run the exact same runtime underneath and it's possible that one thing could work locally but not when deployed and vice versa?

southpolesteve:
They both use the same runtime: https://github.com/cloudflare/workerd. You should not expect any difference between local and production. If you see the attached wranglerissue it just seems that under wrangler dev it buffers larger chunks which causes issues. The team is looking into a fix.

LubomirGeorgiev:
Thanks

PR #879, opened 2026-04-24
[codex] Fix wrangler dev suspense streaming

southpolesteve:
## Summary

- Add a local Wrangler dev response wrapper that sets `Content-Encoding: identity` for streamed loopback App Router responses without explicit body headers.
- Replace whole-line RSC Flight hint buffering with a streaming transform so partial non-hint rows can flow immediately.
- Keep the SSR HTML/RSC embed flush independent of host timers while preserving tick-level batching.

## Root Cause

`wrangler dev`/Miniflare buffers streamed non-SSE responses unless compression is explicitly avoided. The behavior reproduces with a plain `ReadableStream` route, outside React and vinext rendering. Setting `Content-Encoding: identity` restores chunk delivery locally.

This aligns with upstream Cloudflare workers-sdk issue #8004, which describes Miniflare aggressively buffering responses and lists `Content-Encoding: identity` as the workaround.

## Validation

- `npx vp test run tests/app-ssr-stream.test.ts tests/rsc-streaming.test.ts tests/deploy.test.ts -t "applyLocalDevStreamingHeaders|host timers|createFlightHintFixTransform"`
- `npx vp check packages/vinext/src/server/worker-utils.ts packages/vinext/src/server/app-router-entry.ts packages/vinext/src/server/app-ssr-stream.ts packages/vinext/src/server/flight-hints.ts packages/vinext/src/entries/app-rsc-entry.ts tests/app-ssr-stream.test.ts tests/rsc-streaming.test.ts tests/deploy.test.ts`
- `npx vp run vinext#build`
- Manual Wrangler repro: before the workaround `/streaming` buffered until ~830ms; after the workaround the fallback arrived at ~47ms and resolved content at ~834ms.

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -68,6 +68,7 @@ const appPageRenderPath = resolveEntryPath("../server/app-page-render.js", impor
 const appPageResponsePath = resolveEntryPath("../server/app-page-response.js", import.meta.url);
 const cspPath = resolveEntryPath("../server/csp.js", import.meta.url);
 const appPageRequestPath = resolveEntryPath("../server/app-page-request.js", import.meta.url);
+const flightHintsPath = resolveEntryPath("../server/flight-hints.js", import.meta.url);
 const appRouteHandlerResponsePath = resolveEntryPath(
   "../server/app-route-handler-response.js",
   import.meta.url,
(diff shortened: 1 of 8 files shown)


southpolesteve:
This might be a sufficient workaround for the reported `wrangler dev` streaming problem, but I do not think we should treat it as the final answer yet.

The buffering reproduces with a plain `ReadableStream` response outside React/vinext, and it lines up with the upstream Miniflare buffering issue tracked in cloudflare/workers-sdk#8004. If this is fixed upstream, we should prefer that path and avoid keeping a vinext-specific workaround unless we decide the local-dev compatibility value is worth the extra response-header behavior.

LubomirGeorgiev:
Ref https://github.com/cloudflare/vinext/issues/874

PR #888, opened 2026-04-24
fix: centralize protocol-relative URL guard to handle percent-encoded delimiters

southpolesteve:
## Summary

Six places in the request pipeline inlined a variation of:

```ts
pathname.replaceAll("\\", "/").startsWith("//")
```

to reject protocol-relative paths before they reach the trailing-slash redirect emitter (which would otherwise echo them back into a `Location` header). The check handles literal delimiters but not their percent-encoded forms — `/%5Cevil.com/` survives the guard, then `normalizePathnameForRouteMatchStrict` re-encodes the decoded backslash back to `%5C` (preserving it as part of a single path segment), so the 308 trailing-slash redirect ends up with `Location: /%5Cevil.com`. Browsers percent-decode `Location` headers and WHATWG URL treats `\` as `/`, so the browser resolves that to a protocol-relative host.

## Changes

Factor the leading-segment shape check into a new `isOpenRedirectShaped` helper in `server/request-pipeline.ts` that recognises literal (`//`, `/\`) and percent-encoded (`%5C`, `%2F`) variants, and swap each call site over to it. Also add a defense-in-depth check in `normalizeTrailingSlash` so a future caller that skips the upstream guard still can't emit a bad `Location`.

Updated call sites:
- `server/request-pipeline.ts` — `guardProtocolRelativeUrl` + `normalizeTrailingSlash`
- `server/prod-server.ts` — App Router and Pages Router Node handlers
- `server/app-router-entry.ts` — default Cloudflare Worker entry
- `server/app-ssr-entry.ts` — SSR environment entry
- `index.ts` — Pages Router Vite dev middleware
- `deploy.ts` — generated Pages Router Cloudflare Worker (inlined copy kept in sync)

## Tests

Regression tests in `tests/request-pipeline.test.ts` cover the encoded variants plus the defense-in-depth path. Updates the pinned string assertion in `tests/deploy.test.ts` that was checking for the old inlined pattern.

### DIFF
--- a/packages/vinext/src/deploy.ts
+++ b/packages/vinext/src/deploy.ts
@@ -564,6 +564,20 @@ function stripBasePath(pathname: string, basePath: string): string {
   return pathname.slice(basePath.length) || "/";
 }
 
+// Mirror of isOpenRedirectShaped in server/request-pipeline.ts. Inlined here
+// because this worker runs in the Cloudflare Workers environment and can't
+// import from our local source at build time. Keep in sync.
+function isOpenRedirectShaped(rawPathname: string): boolean {
+  if (!rawPathname.startsWith("/")) return false;
+  const afterSlash = rawPathname.slice(1);
+  if (afterSlash.startsWith("/") || afterSlash.startsWith("\\")) return true;
+  if (afterSlash.length >= 3 && afterSlash[0] === "%") {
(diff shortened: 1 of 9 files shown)


southpolesteve:
/bigbonk review please. Also can you be sure this fix is comprehensive. Did we get it everywhere?

PR #895, opened 2026-04-25
try gpt 5.5 for bonk

james-elicx:


NathanDrake2406:
Xhigh/high variant!! @james-elicx

NathanDrake2406:
Max is for Claude only

PR #900, opened 2026-04-25
ci(bonk): downgrade bots from gpt-5.5 to gpt-5.4

NathanDrake2406:
## What
Roll bonk + bigbonk back to `cloudflare-ai-gateway/openai/gpt-5.4`. Remove the `opencode.json` workaround introduced in #899.

## Why
Every `/bonk` and `/bigbonk` invocation since #898 fails. Two stages:

1. Initial throw was `ProviderModelNotFoundError` because models.dev's `cloudflare-ai-gateway` entry doesn't list gpt-5.5. #899 patched that by registering the model in `opencode.json`.
2. After the opencode bump (1.14.22 → 1.14.25, #896) the registry lookup now passes, but a second throw surfaces during SDK init: `UnknownError: ProviderInitError`. opencode swallows the `cause`, so the underlying reason isn't visible in the run log.

Failing run after both #898 and #899 landed: https://github.com/cloudflare/vinext/actions/runs/24938979889.

gpt-5.4 has a known-good path through opencode + `ai-gateway-provider` + Cloudflare AI Gateway, so this restores bots immediately while the gpt-5.5 SDK-init issue gets diagnosed upstream.

## Approach
- Workflows + agent files: `openai/gpt-5.5` → `openai/gpt-5.4`.
- Reasoning effort stays at `xhigh`. opencode falls back to its default effort if gpt-5.4 rejects it; if that turns out to be a problem we can drop to `high` in a follow-up.
- Delete `opencode.json` (no longer needed).
- CONTRIBUTING.md updated to match.

## Validation
- Will comment `/bigbonk review` on this PR before merging to confirm bots come back up.
- Tracking upstream so we can roll forward to gpt-5.5: anomalyco/models.dev#1596 (registry entry) plus the still-unidentified opencode/ai-gateway-provider init regression.


james-elicx (approved):

NathanDrake2406:
@james-elicx update — pushed an opencode pin on top of the gpt-5.4 downgrade.

**Root cause confirmed:** the regression is opencode 1.14.25, not the model. After tracing through every CF repo using cf-ai-gateway, workers-sdk has an inline comment pinning `opencode_version: "1.4.6"` because newer versions throw `ProviderInitError`. vinext is the only CF repo running OpenAI through cf-ai-gateway, which is why the OAI-specific path stayed unobserved.

**This PR now does:**
- Pin `opencode_version: 1.4.6` in both bonk + bigbonk (matches workers-sdk)
- Roll model back to `cloudflare-ai-gateway/openai/gpt-5.4`, keep `variant: xhigh`
- Drop the `opencode.json` workaround from #899

**Treat this as the last attempt at OAI models on cf-ai-gateway.** If `/bigbonk review` on this PR still throws, next move is to revert the workflow + agent files back to `anthropic/claude-opus-4-7` + `variant: max` and stop chasing GPT until upstream opencode + ai-gateway-provider settle.

Tracking upstream: anomalyco/models.dev#1596 (registry entry), and the still-undiagnosed cf-ai-gateway init regression in opencode 1.14.25.

PR #907, opened 2026-04-26
fix(app-router): fixing cache request leaks

NathanDrake2406:
## What this changes

Fixes two App Router cache-safety issues around request-specific data:

- Enforces `dynamic = "force-static"` and `dynamic = "error"` for route-handler request APIs by stubbing or throwing for direct `NextRequest` fields and `next/headers`.
- Tracks direct `request.ip` and `request.geo` access as dynamic in auto mode.
- Re-checks dynamic usage after HTML stream consumption before writing ISR HTML/RSC cache entries.

## Why

Route handlers could previously expose request-specific IP/geo/header/cookie data through untracked request fields, and HTML ISR could cache page output before lazy streamed Server Components had finished touching `cookies()` or `headers()`. Both cases could put per-request output into shared cache.

## Approach

This mirrors Next.js' static-generation request policy instead of treating dynamic API access as observation-only state:

- `force-static` route handlers receive empty readonly stubs for request-specific APIs.
- `dynamic = "error"` throws when request-specific APIs are touched during static rendering.
- auto-mode direct request fields are tracked as dynamic instead of silently flowing into cached output.
- cache writes happen only after the relevant stream has been consumed and dynamic usage has been checked again.

## Next.js references

Checked against Next.js canary commit `ae61573e062e900050b8e6b24626e450accc4570`:

- App route modules set `workStore.forceStatic` for `dynamic = "force-static"` and proxy the request through `forceStaticRequestHandlers`: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/server/route-modules/app-route/module.ts#L907-L913
- That proxy returns sealed empty headers and sealed empty cookies for `request.headers` and `request.cookies`: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/server/route-modules/app-route/module.ts#L1038-L1056
- `headers()` under `forceStatic` returns `HeadersAdapter.seal(new Headers({}))` without tracking, while `dynamic = "error"` throws `StaticGenBailoutError`: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/server/request/headers.ts#L58-L101
- `cookies()` under `forceStatic` returns sealed empty request cookies, while `dynamic = "error"` throws `StaticGenBailoutError`: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/server/request/cookies.ts#L52-L62
- Next's readonly cookie adapter returns throwing callables for mutators, so method lookup is safe but invocation fails: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/server/web/spec-extension/adapters/request-cookies.ts#L36-L49
- Next's readonly headers adapter uses the same throwing-callable shape for `append`, `delete`, and `set`: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/server/web/spec-extension/adapters/headers.ts#L117-L132
- Next consumes dynamic access after prerender/static stream work before returning the result metadata: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/server/app-render/app-render.tsx#L7967-L7974

## Validation

- `vp check`
- `vp test run tests/app-route-handler-runtime.test.ts tests/app-route-handler-execution.test.ts tests/app-route-handler-cache.test.ts tests/app-route-handler-policy.test.ts tests/app-page-cache.test.ts tests/app-page-render.test.ts tests/app-page-response.test.ts tests/app-page-probe.test.ts tests/shims.test.ts`
- `vp test run tests/app-router.test.ts`
- `vp run vinext#build`
- GitHub CI passed on the latest pushed branch

## Risks / follow-ups

The behavior is deliberately stricter around cacheable App Router work. The main compatibility risk is code that relied on reading real request fields under `dynamic = "force-static"`, which conflicts with Next.js' empty-stub behavior and can leak per-request data into shared cache.

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -73,6 +73,10 @@ const appPageResponsePath = resolveEntryPath("../server/app-page-response.js", i
 const cspPath = resolveEntryPath("../server/csp.js", import.meta.url);
 const appPageRequestPath = resolveEntryPath("../server/app-page-request.js", import.meta.url);
 const appPageMethodPath = resolveEntryPath("../server/app-page-method.js", import.meta.url);
+const appStaticGenerationPath = resolveEntryPath(
+  "../server/app-static-generation.js",
+  import.meta.url,
+);
 const appRouteHandlerResponsePath = resolveEntryPath(
   "../server/app-route-handler-response.js",
   import.meta.url,
@@ -434,6 +438,9 @@ import {
 import {
   resolveAppPageMethodResponse as __resolveAppPageMethodResponse,
 } from ${JSON.stringify(appPageMethodPath)};
+import {
+  createStaticGenerationHeadersContext as __createStaticGenerationHeadersContext,
+} from ${JSON.stringify(appStaticGenerationPath)};
 import {
   applyRouteHandlerMiddlewareContext as __applyRouteHandlerMiddlewareContext,
 } from ${JSON.stringify(appRouteHandlerResponsePath)};
@@ -2185,6 +2192,7 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
           __clearRequestContext();
         },
         consumeDynamicUsage,
+        dynamicConfig: handler.dynamic,
         getCollectedFetchTags,
         handlerFn,
         i18n: __i18nConfig,
@@ -2201,7 +2209,11 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
         revalidateSeconds,
         routePattern: route.pattern,
         runInRevalidationContext: async function(renderFn) {
-          const __revalHeadCtx = { headers: new Headers(), cookies: new Map() };
+          const __revalHeadCtx = __createStaticGenerationHeadersContext({
+            dynamicConfig: handler.dynamic,
+            routeKind: "route",
+            routePattern: route.pattern,
+          });
           const __revalUCtx = _createUnifiedCtx({
             headersContext: __revalHeadCtx,
             executionContext: _getRequestExecutionContext(),
@@ -2214,6 +2226,7 @@ async function _handleRequest(request, __reqCtx, _mwCtx) {
           });
         },
         scheduleBackgroundRegeneration: __triggerBackgroundRegeneration,
+        setHeadersAccessPhase,
         setNavigationContext,
       });
       if (__cachedRouteResponse) {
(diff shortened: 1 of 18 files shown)


NathanDrake2406:

NathanDrake2406:

NathanDrake2406:
Addressed the latest bonk review in `850f8f10`.

Assessment:

- `cleanStaticUrl` preserving `username` / `password`: valid. `URL` keeps credentials when protocol/host are rewritten, so the force-static sanitized URL could still contain credentials if a credentialed `NextURL` reached the proxy. Fixed by clearing both fields and added a regression test for `request.url` and `request.nextUrl.href`.
- Force-static body readers falling through: valid for vinext. We apply the force-static request proxy at runtime independent of method, so body readers should not expose real per-request body data. Fixed by returning `null` for `body` and empty-body reader callables for `text()`, `arrayBuffer()`, and `blob()`, with `json()` / `formData()` matching empty-body failure behavior. Added regression coverage for a POST body.
- Dead `buildErrorMessage` closure parameter: valid cleanup. The expression-specific messages come from the request proxy path, so the headers context now constructs the generic access error directly.

Verification:

- `vp test run tests/app-route-handler-runtime.test.ts tests/app-route-handler-execution.test.ts`
- `vp check`

NathanDrake2406:
Assessed the latest bonk review and addressed the actionable parts in `1ae809d3`.

Assessment:

- Empty-body reader comment: valid as maintainability feedback. `readEmptyBodyAsJson()` and `readEmptyBodyAsFormData()` intentionally reject because the force-static stub behaves like a bodyless request, not synthetic JSON/form data. Added an inline comment beside those helpers.
- Background route-handler regeneration missing `setHeadersAccessPhase`: valid as a narrow consistency gap. It is safe today because the call was optional and the headers context already carries `forceStatic` / `accessError`, but the generated entry has the setter in scope and stale regeneration should use the same route-handler phase as the normal route-handler execution path. Threaded `setHeadersAccessPhase` through `readAppRouteHandlerCacheResponse()` and added regression coverage that fails without the wiring.
- Body reader binding concern: not a bug. The returned functions do not depend on `this`; they intentionally create a fresh empty `Response` per call. The rejection behavior for JSON/form data is deliberate and tested.

Verification:

- `vp test run tests/app-route-handler-cache.test.ts -t "sets the route-handler header access phase"` failed before the fix and passes now.
- `vp test run tests/app-route-handler-cache.test.ts tests/app-route-handler-runtime.test.ts tests/app-route-handler-execution.test.ts tests/entry-templates.test.ts tests/app-router.test.ts`
- `vp check`

PR #910, opened 2026-04-26
fix(app-router): action redirects

NathanDrake2406:
## What changed

- Treat no-JS multipart App Router server action form submissions as possible MPA server actions, decode them with React's `decodeAction`, and return a real HTTP 303 redirect when the action calls `redirect()`.
- Classify app route-handler POST form submissions as action-like for redirect handling, so `redirect()` and `permanentRedirect()` return 303 instead of replay-prone 307/308 semantics.
- Preserve pending `cookies()` / draft-mode cookies when a route handler throws a redirect, matching Next.js app-route behavior.
- Add Next.js-compat fixtures and tests for no-JS server action redirects and POST form route-handler redirects.

## Why

Browser form submissions without JavaScript send `multipart/form-data` and no RSC action header. The previous implementation only handled POSTs with `x-rsc-action`, so progressive-enhancement server action forms fell through to a normal render and the action was silently skipped.

For route handlers, `redirect()` encodes 307/308 in the digest, but Next.js overrides action-like POST redirects to 303 so the browser follows with GET instead of resubmitting the original POST.

## Next.js references

- [`getServerActionRequestMetadata()`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/server-action-request-meta.ts): marks POST `multipart/form-data`, POST `application/x-www-form-urlencoded`, and POST action-header requests as possible server actions.
- [`action-handler.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/app-render/action-handler.ts): multipart non-fetch actions call `decodeAction(formData, serverModuleMap)` and return through the full-page render path for progressive enhancement.
- [`app-route/module.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/route-modules/app-route/module.ts): app route redirects return `RedirectStatusCode.SeeOther` when `actionStore.isAction` is true, and append mutable cookies to the redirect response.
- [`app-action-progressive-enhancement.test.ts`](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/actions/app-action-progressive-enhancement.test.ts): verifies formData + redirect without JavaScript returns 303 and navigates.
- [`app-action.test.ts`](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/actions/app-action.test.ts): verifies POST route-handler `redirect()` / `permanentRedirect()` flows use 303.

## Validation

- `vp check`
- `vp test run tests/app-route-handler-policy.test.ts`
- `vp test run tests/nextjs-compat/app-routes.test.ts`
- `vp test run tests/app-router.test.ts -t "server action"`
- `vp test run tests/entry-templates.test.ts`
- `vp run vinext#build`
- `PLAYWRIGHT_PROJECT=app-router vp exec playwright test tests/e2e/app-router/server-actions.spec.ts`

Note: the local pre-commit hook failed in this worktree during Vitest suite initialization before running assertions (`Cannot read properties of undefined (reading 'config')`). The same snapshot update/test command passes when run directly through `vp`, and the updated snapshot is committed.

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -325,6 +325,7 @@ ${slotEntries.join(",\n")}
   return `
 import {
   renderToReadableStream as _renderToReadableStream,
+  decodeAction,
   decodeReply,
   loadServerAction,
   createTemporaryReferenceSet,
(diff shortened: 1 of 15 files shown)


NathanDrake2406:
@james-elicx Sorry for the PR dump. I switched from Claude to Codex and lost all restraint.

The code reads pretty good to me, curious about your impression on the code quality of these pr
I don't see much blasphemous error handling anymore

NathanDrake2406:
Addressed review feedback in ff424787.

What changed:
- Progressive multipart decoding now reads from `request.clone()`, so non-action multipart POSTs and successful non-redirect MPA actions can fall through without consuming the original request body. Added unit coverage for both cases.
- `NEXT_NOT_FOUND` and `NEXT_HTTP_ERROR_FALLBACK;<status>` thrown from progressive actions now return status responses instead of going through the generic 500 path. Added 404 and 403 coverage.
- `isProgressiveServerActionRequest` now normalizes method casing to match the route-handler classifier.
- Added coverage for the `next-action` fallback header and documented the strict `application/x-www-form-urlencoded` parity behavior.
- Removed optional chaining from the `getSetCookie()` assertion.

Left unchanged deliberately:
- The open-redirect defense-in-depth note is broader than this PR because the existing fetch-action redirect path has the same Next.js-parity surface. I did not patch only the new progressive helper and leave the older path divergent.

PR #913, opened 2026-04-26
fix(dev): backstop uncaughtException for socket errors pipe() misses

### LINKED ISSUE #905: Dev server crashes with Unhandled 'error' event (ECONNRESET) when editing files

james-elicx:
## Summary

Follow-up to #911. The connection-level guard installed there fixes the direct `res.socket` case but doesn't cover Node's `pipe()` re-emission path. When a `Readable` source errors and the destination has no `'error'` listener, `pipe()` emits the source's error onto the destination — and the destination isn't always the inbound connection socket.

Real call sites in vinext that hit this:

- `fromWeb(fetch().body).pipe(res)` in `proxyExternalRewriteNode` — source is the upstream fetch body stream, not an inbound socket.
- Streaming surfaces inside `@vitejs/plugin-rsc` that own their own pipe topology.
- Outbound sockets created by `fetch()` from middleware — these never fire `'connection'` on `server.httpServer`, so they sit entirely outside the existing guard.

Stack trace from the field (`vp dlx vinext@54c91f1 dev`, after a successful `GET /stage/oliver`):

```
Error: read ECONNRESET
    at Socket.onerror (node:internal/streams/readable:1035:14)
    at Socket.emit (node:events:509:28)
    ...
```

`Socket.onerror at internal/streams/readable:1035` is `pipe()`'s internal error-propagation function — same crash shape that #911 set out to fix, different installation site.

## Approach

Add a process-level `uncaughtException` handler scoped to the dev server. It drops **only** peer-disconnect codes (`ECONNRESET`, `EPIPE`, `ECONNABORTED`) and re-throws everything else on `nextTick`, which preserves Node's default crash semantics for real bugs. The handler is removed on `httpServer` `'close'` so it doesn't leak across restarts.

Why a process-level backstop:

- It's the only place that catches outbound socket errors from `fetch()` and pipe re-emissions to non-socket destinations.
- Filtering by error code keeps it safe — non-disconnect errors still surface exactly as Node would have surfaced them.
- Dev-only: lives inside `configureServer`, never installed in build/prod.

## Refs

- Refs #905 (original issue — re-occurring after #911 partially closed it)
- Builds on #911

## Test plan

- [ ] `vp check` passes (note: 12 pre-existing typecheck errors about missing peer deps are unchanged)
- [ ] Manual repro from #905 with `proxyExternalRewriteNode` in play: dev server no longer crashes after browser disconnect mid-stream
- [ ] Genuine errors thrown from request handlers still crash dev server with full stack (handler re-throws on nextTick)
- [ ] `httpServer.close()` followed by another `listen()` doesn't double-register the handler

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -2013,23 +2013,35 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
         // @vitejs/plugin-rsc). Outbound sockets created by fetch() also never
         // fire 'connection' on server.httpServer. Filtering by error code
         // keeps real bugs surfacing — only peer-disconnect codes are dropped.
-        const isPeerDisconnect = (err: unknown): boolean => {
-          const code = (err as { code?: string } | null)?.code;
-          return code === "ECONNRESET" || code === "EPIPE" || code === "ECONNABORTED";
-        };
-        const onUncaught = (err: Error) => {
-          if (isPeerDisconnect(err)) return;
-          // Re-throw on next tick so we don't shadow Node's default crash
-          // behavior for genuine errors — same shape as not having installed
-          // a handler at all.
-          process.nextTick(() => {
+        //
+        // Skipped in middleware mode (httpServer is null): the embedding host
+        // owns process-level handlers, and we have no reliable teardown hook
+        // to remove ours, so installation would leak.
+        if (server.httpServer) {
+          const isPeerDisconnect = (err: unknown): boolean => {
+            const code = (err as { code?: string } | null)?.code;
+            return code === "ECONNRESET" || code === "EPIPE" || code === "ECONNABORTED";
+          };
+          // Synchronous throw inside an uncaughtException listener aborts
+          // the process the same way no listener would (stack to stderr,
+          // non-zero exit). Re-throwing on nextTick instead would re-enter
+          // this same listener and loop indefinitely, silently swallowing
+          // genuine errors.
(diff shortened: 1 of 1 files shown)


james-elicx:
Thanks for the catch — all three addressed in 95cd980.

**Blocker (1):** Confirmed the loop locally — your repro reproduces verbatim. Switched to a synchronous throw inside the handler. Verified Node treats it as fatal:

```
$ node -e 'process.on("uncaughtException", (err) => { if (err.code === "ECONNRESET") return; throw err; }); process.nextTick(() => { throw new Error("genuine bug"); });'
[eval]:1
... uncaughtException listener ... throw err;
                                   ^
Error: genuine bug
    at [eval]:1:...
Node.js v24.15.0
```

Stack to stderr, non-zero exit — same observable as no listener.

**Minor (2):** Skipped installation entirely when `httpServer` is null. Wrapped the whole block in `if (server.httpServer)` so neither the listeners nor the close hook are registered in middleware mode. Embedding host stays in charge of process handlers, no leak.

**Nit (3):** Folded in — same `isPeerDisconnect` filter mirrored onto `unhandledRejection` with the same sync-throw shape, removed alongside the uncaughtException handler on close.

PR #916, opened 2026-04-26
fix(dev): hoist socket-error backstop to module top-level so lifecycle events can't remove it

### LINKED ISSUE #905: Dev server crashes with Unhandled 'error' event (ECONNRESET) when editing files

james-elicx:
## Summary

Real fix for the residual #905 crashes that #913 didn't catch. Confirmed via the opt-in trace below.

#913 installed the `uncaughtException` handler inside `configureServer` and tore it down on `httpServer` `'close'`. Vite emits `close` on dep re-optimization, full reloads, and other lifecycle events — leaving a window where the listener is absent when a stale stream errors. The exact pattern users were hitting:

```
GET /page 200 in 109ms
node:events:487
      throw er; // Unhandled 'error' event
Error: read ECONNRESET
```

Reproducer confirmed with the diagnostic flag from this PR:

```
$ VINEXT_DEBUG_SOCKET_ERRORS=1 vp dlx https://pkg.pr.new/vinext@916 dev
...
[vinext] absorbed uncaughtException ECONNRESET
```

The marker firing where the crash used to happen confirms the listener is now in place when the error fires.

## Changes

1. **Hoist install from `configureServer` to module top-level**, guarded by `Symbol.for("vinext.devSocketErrorBackstop")` to prevent double-install (e.g. when vinext is loaded by both a project and a sibling workspace package). No teardown — the listener lives for the process.

2. **Add opt-in `VINEXT_DEBUG_SOCKET_ERRORS=1` trace** that logs when the listener absorbs a peer-disconnect. Default off; useful for confirming the listener fires when users still see crashes in the field.

Filter codes (`ECONNRESET` / `EPIPE` / `ECONNABORTED`) and synchronous re-throw of non-peer-disconnect errors are unchanged from #913.

## Refs

- Refs #905 (residual crashes after #911 and #913)
- Closes #915 (the `EventEmitter.prototype.emit` patch was the wrong direction — domain can replace prototype emit on its load, so that approach was fragile)

## Test plan

- [x] `vp check` passes (1 pre-existing benchmarks `Cannot find module 'vinext'` error is unrelated)
- [x] Reproducer from #905: `VINEXT_DEBUG_SOCKET_ERRORS=1 vp dev` no longer crashes; `[vinext] absorbed uncaughtException ECONNRESET` printed where the crash used to be
- [x] Genuine programming errors still crash dev server with full stack (sync re-throw preserves Node default crash semantics)

### DIFF
--- a/packages/vinext/src/build/prerender.ts
+++ b/packages/vinext/src/build/prerender.ts
@@ -370,6 +370,7 @@ export async function prerenderPages({
 
   const previousHandler = getCacheHandler();
   setCacheHandler(new NoOpCacheHandler());
+  const previousPrerenderFlag = process.env.VINEXT_PRERENDER;
   process.env.VINEXT_PRERENDER = "1";
   // ownedProdServerHandle: a prod server we started ourselves and must close in finally.
   // When the caller passes options._prodServer we use that and do NOT close it.
@@ -642,7 +643,8 @@ export async function prerenderPages({
     return { routes: results };
   } finally {
     setCacheHandler(previousHandler);
-    delete process.env.VINEXT_PRERENDER;
+    if (previousPrerenderFlag === undefined) delete process.env.VINEXT_PRERENDER;
+    else process.env.VINEXT_PRERENDER = previousPrerenderFlag;
     if (ownedProdServerHandle) {
       await new Promise<void>((resolve) => ownedProdServerHandle!.server.close(() => resolve()));
     }
@@ -685,8 +687,12 @@ export async function prerenderApp({
   const previousHandler = getCacheHandler();
   setCacheHandler(new NoOpCacheHandler());
   // VINEXT_PRERENDER=1 tells the prod server to skip instrumentation.register()
-  // and enable prerender-only endpoints (/__vinext/prerender/*).
-  // The set/delete is wrapped in try/finally so it is always restored.
+  // and enable prerender-only endpoints (/__vinext/prerender/*). It also makes
+  // the socket-error backstop (server/socket-error-backstop.ts) re-throw
+  // peer-disconnect errors during prerender. Save the prior value so callers
+  // that already set the flag (run-prerender.ts) aren't clobbered when this
+  // function's finally block restores.
+  const previousPrerenderFlag = process.env.VINEXT_PRERENDER;
   process.env.VINEXT_PRERENDER = "1";
 
   const serverDir = path.dirname(rscBundlePath);
@@ -1113,7 +1119,8 @@ export async function prerenderApp({
     return { routes: results };
   } finally {
     setCacheHandler(previousHandler);
-    delete process.env.VINEXT_PRERENDER;
+    if (previousPrerenderFlag === undefined) delete process.env.VINEXT_PRERENDER;
+    else process.env.VINEXT_PRERENDER = previousPrerenderFlag;
     if (ownedProdServerHandle) {
       await new Promise<void>((resolve) => ownedProdServerHandle!.server.close(() => resolve()));
     }
--- a/packages/vinext/src/build/run-prerender.ts
+++ b/packages/vinext/src/build/run-prerender.ts
@@ -130,6 +130,7 @@ export async function runPrerender(options: RunPrerenderOptions): Promise<Preren
   // sets and clears this var around its own render passes, but we widen
   // the scope here to cover startProdServer / shared-server setup that
   // happens before those phases run. See server/socket-error-backstop.ts.
+  const previousPrerenderFlag = process.env.VINEXT_PRERENDER;
   process.env.VINEXT_PRERENDER = "1";
 
   // The manifest lands in dist/server/ alongside the server bundle so it's
@@ -270,6 +271,8 @@ export async function runPrerender(options: RunPrerenderOptions): Promise<Preren
     if (sharedProdServer) {
       await new Promise<void>((resolve) => sharedProdServer!.server.close(() => resolve()));
     }
+    if (previousPrerenderFlag === undefined) delete process.env.VINEXT_PRERENDER;
+    else process.env.VINEXT_PRERENDER = previousPrerenderFlag;
   }
 
   if (allRoutes.length === 0) {
--- a/packages/vinext/src/build/run-prerender.ts
+++ b/packages/vinext/src/build/run-prerender.ts
@@ -124,6 +124,14 @@ export async function runPrerender(options: RunPrerenderOptions): Promise<Preren
 
   if (!appDir && !pagesDir) return null;
 
+  // Mark the entire prerender orchestration so the socket-error backstop
+  // re-throws peer-disconnect errors during user fetch() calls instead of
+  // silently absorbing them and producing corrupt output. prerender.ts
+  // sets and clears this var around its own render passes, but we widen
+  // the scope here to cover startProdServer / shared-server setup that
+  // happens before those phases run. See server/socket-error-backstop.ts.
+  process.env.VINEXT_PRERENDER = "1";
+
   // The manifest lands in dist/server/ alongside the server bundle so it's
   // cleaned by Vite's emptyOutDir on rebuild and co-located with server artifacts.
   const manifestDir = path.join(root, "dist", "server");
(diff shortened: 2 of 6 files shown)


james-elicx:
Thanks — all four addressed in 455a61b.

**1 + 2 (dev-only framing & middleware-mode skip):** Moved the install back into `configureServer` so it genuinely only fires in dev (build / Vitest worker imports of `index.ts` skip it entirely), and restored the `if (server.httpServer)` middleware-mode skip. The lifecycle bug from #913 was specifically the `httpServer.once('close', ...)` teardown — dropping that and relying on the `Symbol.for` guard for idempotency keeps the listener alive across the session without any teardown surface. Verified via the diagnostic flag that `[vinext] absorbed uncaughtException ECONNRESET` still fires after the previously-failing reproducer.

**3 (listener ordering):** Documented in the JSDoc — by going back to `configureServer`, vinext's listener registers late in the queue again (after most user / tooling setup), so Sentry / structured logging observers still see non-peer-disconnect errors before our sync re-throw aborts iteration. I considered the `nextTick` re-entry-guard variant but kept sync throw to match #913's earlier review reasoning and because the regression bonk was concerned about is largely a top-level-install artifact that no longer applies.

**4 (Symbol.for caveat):** Added the "first-loaded-wins" line to the JSDoc.

**Smaller items:**
- Repeated `code` cast — collapsed into a `peerDisconnectCode(err)` helper that returns the matched code or `undefined`, so the debug logs reuse it without recasting.
- `unhandledRejection` signature — left as `(reason: unknown)`; the `promise` arg isn't used, and Node accepts the narrower signature.

james-elicx:
You were right — that broke it. Pushed [ffc5684](https://github.com/cloudflare/vinext/pull/916/commits/ffc5684) re-hoisting back to module load.

**Root cause** of the configureServer regression: in vite-plus's plugin lifecycle, `server.httpServer` is null when `configureServer` fires (it gets populated later). The `if (server.httpServer)` middleware-mode guard caused install to be skipped entirely. Requests are still served (httpServer is created and listens after) but our listener was never attached. Confirmed by the missing `[vinext] dev socket-error backstop installed` startup line in your last run.

**Bonk's concerns re-addressed without going back into the broken hook:**
- **build / test workers** — gate at module load on `process.env.VITEST === "true"` and `process.argv` containing `"build"`. `process.argv` rather than an env var because cli.ts imports index.ts before it could set one, and `configResolved` runs too late for module-load gating.
- **middleware-mode embedders** — there's no module-load-time signal for middleware mode. JSDoc now acknowledges they get the listeners and points them at `VINEXT_DEBUG_SOCKET_ERRORS` to detect interception + a workaround (lazy import).
- **listener ordering** — JSDoc now honest that we register early at module load, and explains the trade-off. Sync re-throw on non-peer-disconnect kept (later observers don't see the event, but the crash still surfaces with the right stack).
- **Symbol.for caveat** — kept.

Worth re-running once pkg.pr.new rebuilds:
```
VINEXT_DEBUG_SOCKET_ERRORS=1 vp dlx https://pkg.pr.new/vinext@916 dev
```
The startup `[vinext] dev socket-error backstop installed` line should be back, and the absorb marker should fire on disconnect like the first working iteration.

james-elicx:
Pushed [501dd95](https://github.com/cloudflare/vinext/pull/916/commits/501dd95) — extends coverage to `vinext start` for parity with Next.js.

**What Next.js does:** [router-server.ts:809-810](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/router-server.ts#L809) installs `process.on('uncaughtException')` and `process.on('unhandledRejection')` unconditionally — same handler runs for both their dev and prod Node.js servers. Per-request `req.on('error')` / `socket.on('error')` swallows at lines [289-294](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/router-server.ts#L289-L294) and [823-830](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/router-server.ts#L823-L830) are also unconditional. They install everywhere their Node server runs.

**What vinext now does to match:**

| Context | Layer 1 (per-connection) | Layer 2 (process-level) |
|---|---|---|
| `vinext dev` / `vp dev` | ✓ in `configureServer` | ✓ via `config()` hook |
| `vinext start` (self-hosted Node prod) | already none | ✓ via `startProdServer()` |
| Cloudflare Workers prod | n/a | n/a (runtime owns sockets) |
| `vinext build` | n/a | skipped |
| Vitest workers | n/a | skipped |

**Refactor:**
- Extracted install function into [packages/vinext/src/server/socket-error-backstop.ts](packages/vinext/src/server/socket-error-backstop.ts), exporting `installSocketErrorBackstop()` (renamed — no longer "Dev"-prefixed).
- Plugin's `config()` hook calls it gated on `command === "serve" && !isPreview`.
- `startProdServer()` calls it at startup (matching Next.js's router-server pattern of guarding any Node HTTP-serving entry).

**Where vinext intentionally diverges from Next.js:** their `logError` is log-only — they swallow every uncaught exception silently (with a stderr log). vinext keeps the strict filter on peer-disconnect codes (`ECONNRESET` / `EPIPE` / `ECONNABORTED`) and synchronously re-throws everything else, so genuine bugs still surface with full stacks. The parity is in *where* we install, not *what* we swallow.

Comments + docstring at the call sites and in the new module updated to match.

james-elicx:
Pushed [b3e8759](https://github.com/cloudflare/vinext/pull/916/commits/b3e8759) addressing bonk's blocker. The fix:

- Replace the install-time `NODE_ENV` gate with a fire-time `VINEXT_PRERENDER` check inside the listener. The listener is always installed, but during prerender it re-throws all errors unconditionally — acting as if no listener were present.
- Fire-time (vs install-time) is necessary because `index.ts` loads at Vite plugin import, well before prerender begins. The `Symbol.for` guard then makes any later install call a no-op, so a static install-time gate can't catch the prerender phase that follows.
- Set `VINEXT_PRERENDER=1` at the top of `runPrerender()` in `build/run-prerender.ts` so the flag covers the whole prerender orchestration including `startProdServer` setup. `prerender.ts` already sets the same var around its render passes; this widens the scope.
- Drops the `NODE_ENV === "production"` gate. Restores Next.js-parity install for `vinext start`. Test-runner skip stays install-time on `VITEST` / `NODE_ENV === "test"` since those contexts genuinely shouldn't have the listener.

Net behavior:

| Context | Listener installed? | Behavior |
|---|---|---|
| `vinext dev` | yes | filters peer-disconnect ✓ |
| `vinext start` | yes | filters peer-disconnect ✓ (Next.js parity) |
| `vinext build` (no prerender phase) | yes | listener present but no peer-disconnect-prone code runs |
| Prerender (within `vinext build`) | yes | re-throws everything via `VINEXT_PRERENDER` check ✓ |
| Vitest workers | no | normal Node behavior ✓ |

james-elicx:
Pushed [7894608](https://github.com/cloudflare/vinext/pull/916/commits/7894608) addressing the follow-up.

**#1 Stale comment in prod-server.ts** — replaced. Now describes the current behavior: idempotent install via the Symbol.for guard, kept at this site for entry points that load prod-server without index.ts (Next.js parity), prerender bypass is fire-time via `VINEXT_PRERENDER` not install-time.

**#2 prerender.ts clobbering runPrerender's set** — both `prerenderApp` and `prerenderPages` finally blocks now save `previousPrerenderFlag = process.env.VINEXT_PRERENDER` before setting and restore it (`delete` only when the prior value was `undefined`). Closes the gap windows around shared-server close where ECONNRESET would have been absorbed instead of re-thrown.

**#3 Orchestrator-induced ECONNRESET docs** — added a paragraph to the backstop JSDoc calling out that an orchestrator client abort during prerender will surface the build crash as `Error: read ECONNRESET` rather than the underlying route failure, and that `VINEXT_DEBUG_SOCKET_ERRORS=1` can disambiguate.

**#4 JSDoc wording nit** — reworded "by the time `VINEXT_PRERENDER` is set the listener would already be installed via the Symbol.for guard" to "the listener has already been installed (we cannot uninstall and re-install per phase)". The Symbol.for is doing de-dup, not timing — your phrasing is right.

**#5 Pre-existing prod-server.ts:999-1000 comment** — left for now; happy to fix opportunistically in a follow-up if useful but not touching this PR's scope.

PR #920, opened 2026-04-27
fix(fonts): google fonts building perfectly (it was just a regex 😭)

MrIago:
> ## "I" finally fixed google fonts 😭🙏🙌🙌 (it was just a regex!)
>
> Me writing great code

## What this is, in one breath

If you copy the **exact** sample from the official Next.js font docs into a vinext project, the build dies.

Not because your code is wrong. Not because vinext doesn't support it. Because of how your code formatter ends each line.

## Wait, what?

Modern formatters — **Prettier**, **Biome**, **oxfmt** — all ship without trailing semicolons by default. That's how `import { Inter } from 'next/font/google'` shows up in 90%+ of new Next.js projects today.

In vinext, the *absence* of those semicolons made the font plugin's regex misfire. The rewrite was silently skipped. Rolldown then complained about a missing export, and the build was dead.

```
   With trailing ";"            Default style (no ";")
   ─────────────────            ──────────────────────
        ✅ builds                    ❌ build dies
                                       MISSING_EXPORT "Inter"
```

Same code. Same import. Different format. One worked, one didn't.

## Why this matters

This wasn't a niche edge case. It was the **canonical** path:

- The exact line from the [Next.js font docs](https://nextjs.org/docs/app/api-reference/components/font#google-fonts).
- The output of every modern formatter on its default settings.
- The first thing a new vinext user types after `vinext init`.

Every one of them hit a wall.

## How small the actual fix was

Two characters. `\n` added to a single regex (twice — once for imports, once for exports).

The regex was written to stop at `;`. Now it also stops at line breaks. That's it. The rest of the font pipeline was already correct.

## Proof it works

A real Cloudflare Workers project that wouldn't deploy now deploys cleanly:

```
BEFORE   →   vinext deploy   →   ❌  MISSING_EXPORT "Inter"
AFTER    →   vinext deploy   →   ✅  Deployed and healthy
                                   ✅  Inter <link> emitted
                                   ✅  --font-inter applied to CSS
```

Tested with Inter, Geist, and Architects_Daughter — swapping the font name is now a no-op, exactly as the docs promise.

---

## For reviewers (the technical bit)

<details>
<summary>Click to expand — root cause walk-through</summary>

The plugin scans source files for `import … from 'next/font/google'` with a regex. The clause portion was `[^;]+?` — "any character that isn't a `;`, lazy".

When the previous line in the file lacks a `;`, that lazy match still has to eventually hit `next/font/google`. So it walks **across the `\n`** and consumes the previous import line on its way down. The captured clause looks like:

```
type { Metadata } from 'next'
import { Inter }
```

That mangled clause goes into `parseGoogleFontImportClause()`, returns no `fontImports`, and the rewrite branch is skipped. `hasChanges` stays `false`, the transform returns `null`, and `Inter` survives untouched into rolldown — which then can't find a named export on the Proxy-default shim.

The fix:

```diff
- /^[ \t]*import\s+([^;]+?)\s+from\s*(["'])next\/font\/google\2\s*;?/gm
+ /^[ \t]*import\s+([^;\n]+?)\s+from\s*(["'])next\/font\/google\2\s*;?/gm
```

Same change on `exportRe` for parity (`{ … }` clause).

A full ESM parser would have been overkill — the transform's `filter` already opts in only when `code: "next/font/google"` is present, and the regex sweep is what keeps the hot path Rust-side. The minimal regex change preserves that contract.

</details>

<details>
<summary>Why your existing tests didn't catch it</summary>

Every fixture in `tests/font-google.test.ts` writes imports with trailing `;`. Reasonable style, just not the modern default.

Added a regression test that mirrors a real Prettier-default 3-line layout. Asserts (a) the font import gets rewritten, (b) the unrelated preceding `import type` line is left untouched.

</details>

<details>
<summary>Validation</summary>

- ✅ `pnpm test run tests/font-google.test.ts` — 100 passed (1 added)
- ✅ `pnpm run check` — format, lint, typecheck green
- ✅ End-to-end deploy on a Cloudflare Workers project (`vinext@0.0.44`, `@vitejs/plugin-rsc@0.5.24`, `@cloudflare/vite-plugin@1.32.3`)

</details>

<details>
<summary>Refs</summary>

- Next.js's own `next-font-loader` parser is line-bounded by construction because SWC emits one font import per call into the resourceQuery: https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack/loaders/next-font-loader/index.ts#L25-L31
- AGENTS.md asks contributors to bring "here's what the agent found" — done above.

</details>

---

🧠 Thought by a Human ([@MrIago](https://github.com/MrIago)) · 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/vinext/src/plugins/fonts.ts
+++ b/packages/vinext/src/plugins/fonts.ts
@@ -686,7 +686,13 @@ export function createGoogleFontsPlugin(fontGoogleShimPath: string, shimsDir: st
         const fontLocals = new Map<string, string>();
         const proxyObjectLocals = new Set<string>();
 
-        const importRe = /^[ \t]*import\s+([^;]+?)\s+from\s*(["'])next\/font\/google\2\s*;?/gm;
+        // Negate `\n` along with `;` so the lazy clause doesn't roll across
+        // line breaks. Without it, a preceding semicolon-less import line
+        // (e.g. Prettier's `import type { Metadata } from 'next'`) is
+        // greedily swallowed into the clause until the regex finally hits
+        // 'next/font/google' on the next line — which then fails to parse
+        // as a single import clause and the rewrite is silently skipped.
+        const importRe = /^[ \t]*import\s+([^;\n]+?)\s+from\s*(["'])next\/font\/google\2\s*;?/gm;
         let importMatch;
         while ((importMatch = importRe.exec(code)) !== null) {
           const [fullMatch, clause] = importMatch;
@@ -739,7 +745,7 @@ export function createGoogleFontsPlugin(fontGoogleShimPath: string, shimsDir: st
           }
         }
 
-        const exportRe = /^[ \t]*export\s*\{([^}]+)\}\s*from\s*(["'])next\/font\/google\2\s*;?/gm;
+        const exportRe = /^[ \t]*export\s*\{([^}\n]+)\}\s*from\s*(["'])next\/font\/google\2\s*;?/gm;
         let exportMatch;
         while ((exportMatch = exportRe.exec(code)) !== null) {
           const [fullMatch, specifiers] = exportMatch;
--- a/packages/vinext/src/plugins/fonts.ts
+++ b/packages/vinext/src/plugins/fonts.ts
@@ -686,13 +686,17 @@ export function createGoogleFontsPlugin(fontGoogleShimPath: string, shimsDir: st
         const fontLocals = new Map<string, string>();
         const proxyObjectLocals = new Set<string>();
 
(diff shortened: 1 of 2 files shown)


MrIago:
Addressed the multi-line regression — clause now allows `\n` inside `{...}` via `\{[^}]*?\}`, but `[^}]` keeps it from spanning past the close brace, so the original swallow guard still holds. Added two regression tests covering multi-line bracket import + re-export.

The two `rsc-fetch-errors` failures look unrelated (RSC nav timing on 500 + redirect chain) — happy to look if it'd help, but it's outside the font path.

PR #925, opened 2026-04-28
test(app-router): stabilize RSC fetch error redirects

### LINKED ISSUE #901: Flaky E2E: rsc-fetch-errors.spec.ts "redirect chain to a non-ok endpoint hard-navs to the post-redirect URL"
## Symptom
`tests/e2e/app-router/rsc-fetch-errors.spec.ts:156` (`redirect chain to a non-ok endpoint hard-navs to the post-redirect URL`) times out on `page.waitForURL("/about", { timeout: 10_000 })` and fails both initial run and `Retry #1`. PR #900 isn't touching app-router code, so the failure is unrelated to the diff under test.

Failing run: https://github.com/cloudflare/vinext/actions/runs/24939159481/job/73029827222?pr=900

## Reproduction signal
Playwright trace shows the main frame navigated to `/redirect-src` three times in a row, never advancing to `/about`:

```
navigated to "http://localhost:4174/redirect-src"
navigated to "http://localhost:4174/redirect-src"
navigated to "http://localhost:4174/redirect-src"
TimeoutError: page.waitForURL: Timeout 10000ms exceeded.
```

The test's intercept chain depends on Playwright re-dispatching the route table on the 307 follow-up so both the `/redirect-src.rsc` (307 → `/about.rsc`) and the `/about.rsc` (500) handlers fire in sequence. The author already flagged this fragility in `tests/e2e/app-router/rsc-fetch-errors.spec.ts:168-172`:

> Intercept chain depends on Playwright re-entering the route table on the 307 follow-up so both handlers fire in sequence (browser fetch defaults to redirect:follow). Migrating to a mocker that follows redirects without re-dispatching would silently skip the /about.rsc intercept and the test would fall through to the real backend.

The CI failure looks like exactly this scenario inverted: Playwright followed the 307 to `/about.rsc` itself without re-dispatching the route, so the 500 handler never fired and the navigation guard kept retrying the original `/redirect-src` request.

## Hypotheses
1. Playwright/Chromium version drift changed `route.fulfill` 307 follow behaviour to internal-follow (no re-dispatch). Worth pinning the Playwright version and re-running the suite.
2. Race between `page.goto("/")`, hydration, and the intercept registration; if the intercept is replaced by the second registration before the first fetch lands, the 307 chain breaks. Less likely given `page.route` registers synchronously.
3. The redirect handler returns a `Location` header pointing at `/about.rsc`, but the request URL has a query string in some runs (`(\?|$)` regex match), which can desync the second intercept on the bare path.

## Suggested fix direction
Replace the dual-intercept-and-pray pattern with an explicit redirect-follow strategy: a single `page.route` matcher that distinguishes the two URLs by inspecting `request.url()`, then returns 307 or 500 deterministically without depending on Playwright's route-redispatch semantics. Alternatively, route through a real test server that issues the 307 and 500 directly so the browser follows the chain without test-side mocking.

## Severity
Blocks CI on unrelated PRs (e.g. #900). Should be quarantined or re-skipped until the redirect-chain mocking is reworked.

NathanDrake2406:
## What this changes
Stabilizes the App Router RSC fetch error E2E coverage by driving the navigation cases through real fixture links and a server-backed redirect-to-500 RSC chain.

Fixes #901.

## Why
The flaky test depended on direct calls to the private `__VINEXT_RSC_NAVIGATE__` global and a Playwright-mocked 307 chain. Under CI timing and Playwright behavior changes, the test could miss the intended follow-up `.rsc` request or call navigation before router state had committed.

## Approach
The fixture now exposes prefetch-disabled links for the missing-route and redirect-chain cases. Middleware handles RSC requests for `/rsc-fetch-redirect-src` by returning a 307 to `/rsc-fetch-error-target.rsc`, then returns a 500 HTML response for that RSC target. The spec observes those real network responses and asserts the browser hard-navigates to the post-redirect HTML page without flashing the source URL.

## Validation
- `vp check tests/e2e/app-router/rsc-fetch-errors.spec.ts tests/fixtures/app-basic/middleware.ts tests/fixtures/app-basic/app/page.tsx tests/fixtures/app-basic/app/rsc-fetch-error-target/page.tsx`
- `git diff --check`
- `CI=1 PLAYWRIGHT_PROJECT=app-router vp run test:e2e tests/e2e/app-router/rsc-fetch-errors.spec.ts --repeat-each=5`

### DIFF
--- a/tests/e2e/app-router/rsc-fetch-errors.spec.ts
+++ b/tests/e2e/app-router/rsc-fetch-errors.spec.ts
@@ -56,7 +56,7 @@ test.describe("RSC fetch non-ok response handling", () => {
       }
     });
 
-    await page.goto(`${BASE}/about`);
+    await page.goto(`${BASE}/`);
     await waitForAppRouterHydration(page);
 
     // Trigger RSC navigation to a route that does not exist (returns 404 HTML).
@@ -64,10 +64,7 @@ test.describe("RSC fetch non-ok response handling", () => {
     const navigationPromise = page.waitForURL(`${BASE}/this-route-does-not-exist`, {
       timeout: 10_000,
     });
-    await page.evaluate(() => {
-      void (window as any).__VINEXT_RSC_NAVIGATE__("/this-route-does-not-exist");
-    });
-    await navigationPromise;
+    await Promise.all([navigationPromise, page.getByTestId("missing-route-link").click()]);
 
     // The browser must land on the non-.rsc URL — never on the .rsc variant.
     expect(page.url()).toBe(`${BASE}/this-route-does-not-exist`);
@@ -111,10 +108,7 @@ test.describe("RSC fetch non-ok response handling", () => {
     await waitForAppRouterHydration(page);
 
     const navigationPromise = page.waitForURL(`${BASE}/about`, { timeout: 10_000 });
-    await page.evaluate(() => {
-      void (window as any).__VINEXT_RSC_NAVIGATE__("/about");
-    });
-    await navigationPromise;
+    await Promise.all([navigationPromise, page.locator('a[href="/about"]').click()]);
 
     expect(page.url()).toBe(`${BASE}/about`);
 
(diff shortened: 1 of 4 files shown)


NathanDrake2406:

NathanDrake2406:

PR #927, opened 2026-04-28
fix(router): ignore pre-root parallel slots for routed layouts

NathanDrake2406:
## Summary

- Stop attaching parallel slots discovered above a route's first/root layout to that route's `parallelSlots`.
- Add scanner coverage for `app/@modal/default.tsx` with the selected root layout at `app/(group)/layout.tsx`.
- Preserve the valid case where the slot is inside the selected root layout subtree, such as `app/(group)/@modal/default.tsx`.

## Root cause / Why

`discoverInheritedParallelSlots` initialized `layoutIdx` to `-1` when `app/layout.*` was absent, but immediately clamped that value to `0` for `appDir` and every segment before the first actual layout. A slot above the selected root layout, such as `app/@modal/default.tsx`, was therefore recorded as owned by `layoutIndex: 0` and injected into `app/(group)/layout.tsx`.

That diverges from Next.js multiple-root-layout semantics. Next.js documents that omitting `app/layout.js` lets subdirectory layouts become root layouts: [multiple root layout docs](https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/docs/01-app/03-api-reference/03-file-conventions/layout.mdx#L140-L145). The server tree construction also notes that parallel route children cannot exist above the root layout: [create-component-tree note](https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/server/app-render/create-component-tree.tsx#L1273-L1290).

The fix keeps the pre-layout sentinel while scanning and skips those slots once the route has a real layout. Layout-less scanner metadata is intentionally preserved here; scanner validation fixes are intentionally separate in another PR.

## Verification

- `vp test run tests/routing.test.ts`
- `vp check packages/vinext/src/routing/app-router.ts tests/routing.test.ts`

### DIFF
--- a/packages/vinext/src/routing/app-router.ts
+++ b/packages/vinext/src/routing/app-router.ts
@@ -711,24 +711,32 @@ function discoverInheritedParallelSlots(
   let currentDir = appDir;
   const dirsToCheck: { dir: string; layoutIdx: number }[] = [];
   let layoutIdx = findFile(appDir, "layout", matcher) ? 0 : -1;
-  dirsToCheck.push({ dir: appDir, layoutIdx: Math.max(layoutIdx, 0) });
+  dirsToCheck.push({ dir: appDir, layoutIdx });
 
   for (const segment of segments) {
     currentDir = path.join(currentDir, segment);
     if (findFile(currentDir, "layout", matcher)) {
       layoutIdx++;
     }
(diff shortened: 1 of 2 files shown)


NathanDrake2406:

NathanDrake2406:

PR #948, opened 2026-04-29
feat: add unstable_io shim for next/cache

### LINKED ISSUE #805: Add `unstable_io` shim for `next/cache` export
## Summary

Next.js added a new `unstable_io()` export from `next/cache` behind the `experimental.unstableIO` config flag. This is a metasyntactic IO boundary that, during prerendering with cache components, returns a hanging promise to prevent execution of code that follows it. In all other contexts it resolves as a fulfilled thenable that React can unwrap synchronously.

## Upstream

- Commit: https://github.com/vercel/next.js/commit/e22dc27d8f61eb4ca9dad762d4e46b1f812fdf86
- PR: https://github.com/vercel/next.js/pull/92521

## What changed

- New export `unstable_io` from `next/cache` (both server and client/browser paths)
- New config option `experimental.unstableIO` (boolean, default `false`)
- New env variable `process.env.__NEXT_UNSTABLE_IO`
- Server implementation in `packages/next/src/server/request/io.ts` with behavior varying by work unit type:
  - `request`: resolves immediately (production) or uses staged rendering (dev)
  - `prerender` / `prerender-client` / `prerender-runtime`: returns a hanging promise to prevent prerendering past the IO boundary
  - `cache` / `unstable-cache` / `generate-static-params` / `prerender-legacy`: resolves immediately
- Client/browser implementation always resolves immediately (returns a pre-resolved thenable)
- Throws if called without `experimental.unstableIO` enabled

## Impact on vinext

vinext shims `next/cache` and needs to add `unstable_io` to the shim. Since vinext primarily targets Cloudflare Workers (no prerendering context), the simplest correct implementation is:

1. Add `unstable_io` to the `next/cache` shim exports
2. Have it return a pre-resolved thenable (same as the browser implementation)
3. Optionally: gate behind `experimental.unstableIO` config check and throw if not enabled (matching Next.js behavior)

This is gated behind an experimental flag and not ready for general use, so this is low priority.

Divkix:
Fixes #805

## Summary
- Adds `unstable_io()` export to the `next/cache` shim
- Returns a pre-resolved fulfilled thenable matching Next.js's browser/client implementation
- React's `use()` hook can unwrap it synchronously without suspending
- No config gate needed — the guard was removed upstream in next.js#92923

## Test plan
- Added 3 unit tests in `tests/shims.test.ts`:
  - Verifies `unstable_io` is exported as a function
  - Verifies it returns a resolved `Promise<void>`
  - Verifies singleton behavior (same instance on multiple calls)

### DIFF
--- a/packages/vinext/src/shims/cache.ts
+++ b/packages/vinext/src/shims/cache.ts
@@ -453,8 +453,9 @@ const _resolvedIOPromise: Promise<void> = Promise.resolve(undefined);
  * Guard removed: https://github.com/vercel/next.js/pull/92923
  *
  * In Next.js, `unstable_io()` during prerendering contexts returns a hanging
- * promise to prevent execution past the IO boundary. vinext targets
- * Cloudflare Workers (no prerendering), so this always resolves immediately.
+ * promise to prevent execution past the IO boundary. vinext does support
+ * prerendering (static export, --prerender-all, TPR), but the hanging IO
+ * boundary behavior is not yet implemented, so this always resolves immediately.
  */
 export function unstable_io(): Promise<void> {
   return _resolvedIOPromise;
--- a/packages/vinext/src/shims/cache.ts
+++ b/packages/vinext/src/shims/cache.ts
@@ -450,6 +450,7 @@ const _resolvedIOPromise: Promise<void> = Promise.resolve(undefined);
  * Marks an IO boundary in server components by returning a resolved promise.
  *
  * See: https://github.com/vercel/next.js/pull/92521
+ * Guard removed: https://github.com/vercel/next.js/pull/92923
  *
  * In Next.js, `unstable_io()` during prerendering contexts returns a hanging
  * promise to prevent execution past the IO boundary. vinext targets
--- a/packages/vinext/src/shims/cache.ts
+++ b/packages/vinext/src/shims/cache.ts
@@ -435,6 +435,30 @@ export function unstable_noStore(): void {
 // Also export as `noStore` (Next.js 15+ naming)
 export { unstable_noStore as noStore };
 
+/**
+ * A fulfilled thenable that React can unwrap synchronously via `use()`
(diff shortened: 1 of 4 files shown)


packages/vinext/src/shims/cache.ts
```diff
@@ -435,6 +435,31 @@ export function unstable_noStore(): void {
...
+
+/**
+ * Marks an IO boundary in server components by returning a resolved promise.
+ *
+ * See: https://github.com/vercel/next.js/pull/92521
+ * Guard removed: https://github.com/vercel/next.js/pull/92923
+ *
+ * In Next.js, `unstable_io()` during prerendering contexts returns a hanging
+ * promise to prevent execution past the IO boundary. vinext targets
+ * Cloudflare Workers (no prerendering), so this always resolves immediately.
```
james-elicx: this isn't strictly true - workers is a deployment option, but it's not the only option

Divkix: oh yeah, should've taken care of that. will change it

packages/vinext/src/shims/cache.ts
```diff
@@ -435,6 +435,32 @@ export function unstable_noStore(): void {
...
+/**
+ * Marks an IO boundary in server components by returning a resolved promise.
+ *
+ * See: https://github.com/vercel/next.js/pull/92521
+ * Guard removed: https://github.com/vercel/next.js/pull/92923
+ *
+ * In Next.js, `unstable_io()` during prerendering contexts returns a hanging
+ * promise to prevent execution past the IO boundary. vinext does support
+ * prerendering (static export, --prerender-all, TPR), but the hanging IO
+ * boundary behavior is not yet implemented, so this always resolves immediately.
```
james-elicx, line 458: can we open an issues for this gap if we don't already have one (on my phone so not checked)

PR #966, opened 2026-04-29
refactor: extract app rsc manifest construction

NathanDrake2406:
## Summary

Extracts the App Router RSC generated-manifest construction out of `generateRscEntry()` into `app-rsc-manifest.ts`.

The generator now treats the app shape as an explicit construction boundary:

- route module import allocation
- route entry serialization
- root fallback/global-error module references
- metadata route entries
- `generateStaticParams` map entries

`generateRscEntry()` keeps request/runtime codegen ownership and consumes the manifest-construction result.

## Why

`app-rsc-entry.ts` had accumulated app-shape responsibilities inside the top-level RSC entry generator. That made the generated runtime path harder to review because file-system route shape, module import allocation, metadata route handling, and request lifecycle code all lived in one function.

This mirrors the separation Next.js keeps between its generated App Router tree shape and the runtime route module that consumes it:

- Next.js app-page template receives an injected `tree` and passes it as `userland.loaderTree`: [packages/next/src/build/templates/app-page.ts](https://github.com/vercel/next.js/blob/e1bb911c14e4ea3c6bb8cc98871ffdee317d513f/packages/next/src/build/templates/app-page.ts#L100-L138)
- Next.js `LoaderTree` makes segment modules and parallel routes explicit: [packages/next/src/server/lib/app-dir-module.ts](https://github.com/vercel/next.js/blob/e1bb911c14e4ea3c6bb8cc98871ffdee317d513f/packages/next/src/server/lib/app-dir-module.ts#L6-L28)
- Next.js parses loader-tree shape through a dedicated helper: [packages/next/src/shared/lib/router/utils/parse-loader-tree.ts](https://github.com/vercel/next.js/blob/e1bb911c14e4ea3c6bb8cc98871ffdee317d513f/packages/next/src/shared/lib/router/utils/parse-loader-tree.ts#L4-L20)
- Build-time static path logic traverses that tree, including parallel routes: [packages/next/src/build/static-paths/utils.ts](https://github.com/vercel/next.js/blob/e1bb911c14e4ea3c6bb8cc98871ffdee317d513f/packages/next/src/build/static-paths/utils.ts#L113-L193)

Vinext still uses its own scanned `AppRoute` model, but this PR makes the equivalent app-shape-to-generated-manifest step explicit and testable.

## Notes

Static metadata files are now read through the extracted manifest helper. If a discovered static metadata file cannot be read, the helper throws a build-time error with the file path instead of generating an empty runtime response.

## Tests

- `vp test run tests/entry-templates.test.ts`
- `vp check`
- `vp run vinext#build`
- commit hook: `vp check --fix`, `knip --no-progress`

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -7,7 +7,7 @@
  *
  * Previously housed in server/app-dev-server.ts.
  */
-import fs from "node:fs";
+import { buildAppRscManifestCode } from "./app-rsc-manifest.js";
 import { resolveEntryPath } from "./runtime-entry-module.js";
 import type {
   NextHeader,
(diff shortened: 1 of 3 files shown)


packages/vinext/src/entries/app-rsc-entry.ts
```diff
@@ -151,207 +151,18 @@ export function generateRscEntry(
...
+    imports,
+    routeEntries,
+    metaRouteEntries,
+    generateStaticParamsEntries,
+    rootNotFoundVar,
+    rootForbiddenVar,
+    rootUnauthorizedVar,
+    rootLayoutVars,
+    globalErrorVar,
+  } = manifestCode;
```
james-elicx, line 165: nit: could destructure inline instead of after

PR #973, opened 2026-04-29
fix: support enablePrerenderSourceMaps config with default true

### LINKED ISSUE #960: Apply sourcemaps by default during prerender in production build
## Next.js Change

Next.js now applies sourcemaps by default during prerender in `next build`. Previously this was gated behind an experimental config option (`experimental.applySourceMaps` or similar) which has now been removed.

**Commit:** [`59c73ae`](https://github.com/vercel/next.js/commit/59c73aecc9b26fdde5d6cd525ccc433123e7f55a)
**PR:** [#93280](https://github.com/vercel/next.js/pull/93280)

## What Changed

- `packages/next/src/server/config-shared.ts`: option type updated.
- `packages/next/src/server/config.ts`: 8 lines removed — likely the gating logic that previously toggled sourcemap application based on an experimental flag.
- Removed from documentation (`docs/01-app/02-guides/memory-usage.mdx`, `docs/01-app/03-api-reference/06-cli/next.mdx`).
- Removed from test fixtures' `next.config.js` files (`root-suspense-dynamic`, `server-source-maps/edge`), implying the flag is no longer needed to opt in.

The behavior is now: during `next build` prerender, server stack traces are rewritten through sourcemaps by default, producing more useful error output if a build-time render throws.

## Impact on vinext

vinext runs prerender via the App Router RSC build pipeline, with its own production server (`cloudflare/worker-entry.ts`) and prod build flow (`createBuilder` + `buildApp`). Two concerns:

1. **Default behavior parity**: Build-time prerender errors in vinext should also surface with sourcemap-resolved stack traces by default, matching Next.js DX. If vinext currently has no sourcemap-application step during prerender, errors will point to bundled output rather than user source.
2. **Config compatibility**: Users migrating from Next.js may have removed (or be about to remove) `experimental.applySourceMaps` (or whichever exact key was removed) from their config. vinext should not warn or break on that key's absence, and should not require an opt-in to get the now-default behavior.

## Suggested Action

- Identify the exact config option name removed (the diff trims 8 lines from `config.ts`; check the PR for the key) and ensure vinext's config ingestion silently accepts/ignores both the present and absent forms.
- Audit vinext's prerender pipeline (`vp build` / `createBuilder` flow + the App Router RSC build) for whether it applies sourcemaps to server-side stack traces. If not, add it as the default.
- Add a test that triggers a build-time render error in a fixture and asserts the printed stack frames reference user source files, not bundled output.

Divkix:
Fixes #960

## Summary
Adds the `enablePrerenderSourceMaps` config option matching Next.js behavior:
- Defaults to `true` — prerender error stack traces resolve through sourcemaps to original source files
- Error results in prerender output now include full stack traces when enabled
- Node.js sourcemap support (`process.setSourceMapsEnabled(true)`) is enabled before the prerender phase in both `cli.ts` (build) and `deploy.ts` (deploy)
- Config is silently accepted and recognized as "supported" in `vinext check` output

## Test plan
- 3 new unit tests in `tests/next-config.test.ts` verifying default `true`, explicit `false`, and explicit `true`
- All existing `next-config.test.ts` and `prerender.test.ts` tests continue to pass


Divkix (commented):
Summary of changes: adds enablePrerenderSourceMaps config option defaulting to true, matching Next.js. Prerender error results include full stack traces when enabled. Node.js sourcemap support enabled before prerender phase. Config silently accepted, recognized as "supported" in vinext check.

PR #984, opened 2026-04-30
perf: fuse RSC stream tees reducing allocs from 3 to 2

### LINKED ISSUE #981: perf: reduce stream tee() allocation from 3 to 2 in ISR cache path
### Problem

When ISR caching is enabled, a single App Router request can trigger up to 3 `ReadableStream.tee()` calls, each allocating 2 new ReadableStream objects:

| # | File:Line | Tee | Purpose |
|---|-----------|-----|---------|
| 1 | `app-page-execution.ts:323` | `stream.tee()` | ISR RSC data capture |
| 2 | `app-ssr-entry.ts:187` | `rscStream.tee()` | RSC embed transform for hydration |
| 3 | `app-page-cache.ts:236` | `response.body.tee()` | HTML cache write buffering |

Total: up to 5 ReadableStream objects alive simultaneously per request, each with internal buffers. On Cloudflare Workers (128MB hard limit), this creates meaningful memory pressure and GC overhead.

**Note:** Tee 1 already has a gate (`!shouldCapture` → skip), so non-ISR requests only have tee 2. The triple-tee only occurs in the cache-enabled path.

### What each tee does

**Tee 1** (`teeAppPageRscStreamForCapture`, `app-page-execution.ts:312-328`):
- `responseStream` → passed to SSR handler
- `captureStream` → accumulated into `ArrayBuffer` via `readAppPageBinaryStream()` for ISR cache write

**Tee 2** (SSR embed, `app-ssr-entry.ts:187-210`):
- `ssrStream` → fed to `createFromReadableStream()` for React SSR
- `embedStream` → consumed by `createRscEmbedTransform()` which injects inline `<script>` tags progressively

**Tee 3** (HTML cache, `app-page-cache.ts:236-287`):
- `streamForClient` → returned to browser with `no-store` headers
- `streamForCache` → fully buffered, then conditionally written to ISR cache (skipped if dynamic usage detected)

Tee 3 must remain separate — it operates on HTML (post-SSR), not RSC, and has a conditional write decision requiring full buffering.

### Proposed fix: Fuse Tee 1 + Tee 2

Tee 1 and Tee 2 both consume from the SAME original RSC stream. They can share a single tee:

**Current:**
```
RSC stream → tee1 → [responseStream → tee2 → [ssrStream + embedStream]]
                 → [captureStream → accumulate for cache]
```

**Proposed:**
```
RSC stream → fusedTee → [mainBranch → SSR]
                       → [sideBranch → embed transform + capture accumulate]
```

The side branch runs both the embed transform and binary accumulation from a single stream reader. This means `createRscEmbedTransform` must also pass through raw chunks to an accumulator.

### Estimated files touched: 4

| File | Change |
|------|--------|
| `app-page-execution.ts` | `teeAppPageRscStreamForCapture` → produce combined embed+capture from single tee |
| `app-ssr-stream.ts` | `createRscEmbedTransform` → also accumulate raw chunks |
| `app-ssr-entry.ts` | `handleSsr` → accept combined embed+capture |
| `app-page-render.ts` | Wire combined pipeline |

### Difficulty: Medium-Hard

The embed transform is progressive (must inject scripts in lockstep with React SSR), while capture is passive accumulation. The `createRscEmbedTransform` must be restructured to pass through raw bytes to an accumulator while also producing the embed output.

### Expected improvement

~33% less memory during SSR in ISR path. Fewer GC pauses. On Workers, this reduces OOM risk under concurrency.

### Risk: Medium

Touches the core SSR rendering pipeline. The embed transform timing is tightly coupled to React's render lifecycle. Must verify no regressions in progressive script injection (which enables out-of-order RSC chunk hydration).

Divkix:
Fixes #981

## Summary
- Fuse Tee 1 (ISR RSC capture) and Tee 2 (SSR embed) into a single `stream.tee()` call, reducing from 3 tees to 2 in the ISR cache path
- `createRscEmbedTransform` now accumulates raw `Uint8Array` chunks alongside decoded text, exposed via `getRawBuffer()`
- `teeAppPageRscStreamForCapture` returns `{ssrStream, sideStream}` — sideStream feeds both embed transform and byte accumulation
- `handleSsr` accepts optional `sideStream` + `capturedRscDataRef` out-parameter, skipping internal tee when pre-split
- ~33% less ReadableStream objects per ISR request, reducing GC pressure on Cloudflare Workers (128MB limit)

## Test plan
- New tests for `createRscEmbedTransform` raw buffer accumulation (`app-ssr-stream.test.ts`)
- Updated tests for fused `teeAppPageRscStreamForCapture` returning `ssrStream`/`sideStream` (`app-page-execution.test.ts`)
- Updated test mock to handle `capturedRscDataRef` out-param (`app-page-render.test.ts`)
- 6 snapshot updates for generated entry template changes (`entry-templates.test.ts`)
- 489 related tests pass across 10 test files


packages/vinext/src/server/app-ssr-stream.ts
```diff
@@ -78,6 +84,22 @@ export function createRscEmbedTransform(
       scripts += createInlineScriptTag("self.__VINEXT_RSC_DONE__=true", scriptNonce);
       return scripts;
     },
+
+    async getRawBuffer(): Promise<ArrayBuffer> {
+      await pumpPromise;
```
james-elicx, line 89: i was just looking at `pumpReader` above - it looks like it's swallowing errors that happen, so `getRawBuffer` could potentially return an incomplete buffer due to an error in `pumpReader`.

i think we should probably be re-throwing errors caught there inside these functions.

PR #995, opened 2026-04-30
fix: warm server action modules in dev

### LINKED ISSUE #994: Server actions fail with [vite-rsc] invalid server reference on cold-start in dev
Repro: App Router project, 'use client' component statically imports a 'use server' action, action invoked before browser has fetched the action file (e.g. dev-server restart while browser kept stale modules, or any race where the action call lands first).

Cause: @vitejs/plugin-rsc populates serverReferenceMetaMap lazily as files are transformed. Validation runs before the lazy import that would trigger the transform, so cold validation throws [vite-rsc] invalid server reference '/path/to/action.ts' and the action never runs. Tracked upstream as [#641](https://github.com/vitejs/vite-plugin-react/issues/641).

Workaround (in user's vite.config.ts):

```
environments: { rsc: { dev: { warmup: ['./lib/actions/**/*.ts'] } } }
```
Suggested fix: vinext could auto-inject dev.warmup for any file under app/ (and configurable action dirs) containing 'use server'. 
Since vinext already scans for App Router files, adding 'use server' detection during that scan is cheap. Users wouldn't need any config.

NathanDrake2406:
## Summary

Fixes #994.

In dev, `@vitejs/plugin-rsc` validates incoming server action references against its in-memory `serverReferenceMetaMap`. That map is populated when action modules are transformed, so a cold action POST can fail before the action file has been loaded. This PR has vinext discover module-level `"use server"` files under `app/` and add them to the RSC environment `dev.warmup`, preserving any user-specified warmup entries.

The behavior is intentionally implemented in a normal config helper instead of expanding generated entries: codegen continues to describe the app shape, while runtime/config behavior lives in importable modules.

## Next.js reference behavior

Next.js does not rely on first-request lazy module transforms to recognize server actions. The webpack client module loader identifies server action IDs from RSC module metadata and proxies them in production:

- https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/build/webpack/loaders/next-flight-client-module-loader.ts#L25-L31
- https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/build/webpack/loaders/next-flight-client-module-loader.ts#L31-L40

Next also builds action assets and emits the server action manifest from discovered action entries, rather than waiting for the first action request to discover the module:

- https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/build/webpack/plugins/flight-client-entry-plugin.ts#L1101-L1128
- https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/build/webpack/plugins/flight-client-entry-plugin.ts#L1130-L1158

For directive semantics, the warmup scanner follows Next.js's module directive model: only directive-prologue string literals count, so inline `"use server"` actions are not treated as module-level server action files. Reference:

- https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/server/typescript/utils.ts#L120-L175

## Tests

- `vp test run tests/server-action-warmup.test.ts`
- `vp check packages/vinext/src/config/server-action-warmup.ts packages/vinext/src/index.ts tests/server-action-warmup.test.ts`
- `vp test run tests/app-server-action-execution.test.ts`
- `vp test run tests/app-router.test.ts -t "RSC plugin auto-registration"`
- `vp run vinext#build`

`vp run vinext#build` still reports the existing virtual-module external warnings for `private-next-instrumentation-client`, `virtual:vinext-rsc-entry`, and `virtual:vite-rsc/client-references`, then exits successfully.

### DIFF
--- a/packages/vinext/src/config/server-action-warmup.ts
+++ b/packages/vinext/src/config/server-action-warmup.ts
@@ -1,0 +1,168 @@
+import { glob, readFile } from "node:fs/promises";
+import path from "node:path";
+import { normalizePageExtensions } from "../routing/file-matcher.js";
+
+const SERVER_ACTION_SOURCE_EXTENSIONS = ["js", "jsx", "ts", "tsx", "mjs", "mts", "cjs", "cts"];
+const SERVER_ACTION_SCAN_EXCLUDED_ROOTS = new Set([
+  ".git",
+  ".next",
+  ".output",
+  ".refs",
+  ".turbo",
+  ".vinext",
+  ".worktrees",
+  "build",
+  "coverage",
+  "dist",
+  "node_modules",
+  "out",
+]);
+
+type CollectServerActionWarmupEntriesOptions = {
+  root: string;
+  pageExtensions?: readonly string[] | null;
+};
+
+function buildExtensionGlob(extensions: readonly string[]): string {
+  return extensions.length === 1 ? extensions[0] : `{${extensions.join(",")}}`;
+}
+
+function toViteEntry(root: string, filePath: string): string {
+  return path.relative(root, filePath).split(path.sep).join("/");
+}
+
+function normalizeServerActionExtensions(pageExtensions?: readonly string[] | null): string[] {
+  return [
+    ...new Set([...SERVER_ACTION_SOURCE_EXTENSIONS, ...normalizePageExtensions(pageExtensions)]),
+  ];
+}
+
+function shouldExcludeServerActionScanPath(name: string): boolean {
+  const segments = name.split(/[\\/]+/).filter(Boolean);
+  return (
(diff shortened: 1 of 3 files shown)


james-elicx (changes requested):
scanning every file on the disk would be somewhat inefficient - we'd probably want to do this by walking through a module graph and only scanning modules depended on by app router page for directives. however, that could still be somewhat taxing in large projects... this probably needs some more thought... maybe some collaboration in the vite plugin as well - there may be something that could be improved there for this

i think this might not work during hmr, as in, new server actions wouldn't be caught as the scan is a one-time thing.

PR #1020, opened 2026-05-02
fix(rsc): exclude client shims from dep optimization

### LINKED ISSUE #1008: vinext shims with 'use client' aren't registered in clientReferenceMetaMap, causing @vitejs/plugin-rsc to crash.
## Versions

- `vinext`: 0.0.45
- `@vitejs/plugin-rsc`: 0.5.25
- `vite`: 8.0.10
- Node 24 / Bun 1.3.11
- macOS 15 (Apple Silicon)

## Symptom

Navigating to an App Router route that transitively imports any `vinext/shims/*` file with `'use client'` (e.g. via `next/link`, `next/navigation`, `next/image`, `next/form`, `next/script`) crashes the dev server with:

```ts
[vite] Internal server error: Cannot read properties of undefined (reading 'exportNames')
    at LoadPluginContext.handler (.../@vitejs/plugin-rsc/dist/plugin-BhzHKRFo.js:1395:110)
    at EnvironmentPluginContainer.load (.../vite/dist/node/chunks/node.js:30122:56)
    at async loadAndTransform (.../vite/dist/node/chunks/node.js:24465:21)
```

The route renders blank, with the error only visible in the wrangler/vite terminal. Other routes that don't pull a `'use client'` shim through the server-component graph load fine — which made it look route-specific and very confusing to diagnose.

## Root cause

`@vitejs/plugin-rsc` generates a `\0virtual:vite-rsc/client-package-proxy/<source>` virtual module for each `'use client'` node_modules package transitively imported by a server component. The `load` handler then looks up the matching entry in `clientReferenceMetaMap` to know which exports to re-export.

The following vinext shim files all start with `'use client'`:

- `vinext/dist/shims/layout-segment-context.js`
- `vinext/dist/shims/slot.js`
- `vinext/dist/shims/link.js`
- `vinext/dist/shims/navigation.js`
- `vinext/dist/shims/form.js`
- `vinext/dist/shims/script.js`
- `vinext/dist/shims/error-boundary.js`
- `vinext/dist/shims/client-hook-error.js`

When server code imports e.g. `next/link` (resolved by vinext to `vinext/shims/link`), plugin-rsc's `resolveId` hook records the package source. Vite's `depsOptimizer` then bundles `vinext/shims/*` into a consolidated chunk that includes unrelated vinext server code. The path of the optimized chunk drifts from what plugin-rsc scanned, so its `clientReferenceMetaMap` lookup misses, the load handler hits `undefined.exportNames`, and the entire dev server bundle for that route crashes.

After patching plugin-rsc with a defensive guard, the warning identified the actual offenders:

```
[vite-rsc PATCH] no clientReferenceMetaMap entry for "vinext/shims/layout-segment-context" — falling back to wildcard re-export
[vite-rsc PATCH] no clientReferenceMetaMap entry for "vinext/shims/slot" — falling back to wildcard re-export
```

## Reproduction shape

(I don't have a hosted repro since this is a private app, but the recipe is straightforward.)

A vinext App Router project where:
1. The route's server component renders a `'use client'` component (e.g. a chat UI)
2. That client component imports `next/link` and at least one other `'use client'` node_modules package (e.g. `sonner`, `swr`)
3. Vite's `depsOptimizer` runs in dev (default)

The route bundle crashes on first request with the `exportNames` error.

## Workaround we shipped

Patched `@vitejs/plugin-rsc/dist/plugin-BhzHKRFo.js` line 1395 with a defensive guard that warns + falls back to `export *` when the meta lookup misses, persisted via `bun patch`:

```js
const meta = Object.values(manager.clientReferenceMetaMap).find((v) => v.packageSource === source);
if (!meta) {
  console.warn(`[vite-rsc PATCH] no clientReferenceMetaMap entry for "${source}" — falling back to wildcard re-export`);
  return `export * from ${JSON.stringify(source)};\n`;
}
return `export {${meta.exportNames.join(",")}} from ${JSON.stringify(source)};\n`;
```

This unblocks the dev server, but `export *` doesn't preserve named-export semantics perfectly for every package. The right fix lives in vinext.

## Proposed fix (vinext side)

Two viable directions, either would resolve this:

1. **Pre-register shims with plugin-rsc.** When vinext sets up the `rsc()` plugin, also configure plugin-rsc to know about all `vinext/shims/*` files that ship `'use client'`. They're a known fixed list at vinext build time.
2. **Force-include shims in `optimizeDeps.exclude`.** Have vinext inject `optimizeDeps.exclude: ['vinext/shims/...']` (or per-shim entries) automatically so Vite doesn't bundle them into a consolidated chunk in the first place. Users hit this without knowing they need to do it themselves, and the per-shim list will keep growing as vinext adds shims.

Happy to PR option 2 if it'd help — needs guidance on the right place in `dist/index.js` to inject the config.

## Adjacent observation (worth a separate issue?)

`@vitejs/plugin-rsc`'s assumption at line 1395 that `find()` always succeeds is itself a defensive-coding gap. Frameworks built on top of plugin-rsc (vinext, Waku, custom) all stand to crash the same way if any client package isn't registered. The handler should fall back gracefully + log instead of throwing. Plugin-rsc's source already has a `warnInoncistentClientOptimization()` helper that recommends `optimizeDeps.exclude` — it should be paired with a non-fatal load-handler fallback.

(Filing this with vinext first since the actionable fix is here. Plugin-rsc maintainers want a CodeSandbox repro for upstream issues, which is harder to construct in isolation than as a vinext-on-real-app problem.)

NathanDrake2406:
## What this changes

Adds vinext's known RSC client shim subpaths to the Vite dep optimizer exclude set for App Router projects. The exclude set is merged into the top-level, RSC, SSR, and client optimizer configs while preserving user-provided excludes and `serverExternalPackages`.

Closes #1008.

## Why

Next.js treats public App Router APIs such as `next/link`, `next/form`, and `next/script` as client modules:

- `next/link`: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/client/app-dir/link.tsx#L1
- `next/form`: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/client/app-dir/form.tsx#L1
- `next/script`: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/packages/next/src/client/script.tsx#L1
- Next's App Router E2E suite explicitly verifies `next/link` in Server Components: https://github.com/vercel/next.js/blob/ae61573e062e900050b8e6b24626e450accc4570/test/e2e/app-dir/rsc-basic/rsc-basic.test.ts#L272-L286

vinext matches that shape with client shims. In dev, `@vitejs/plugin-rsc` records bare package sources resolved from the RSC graph and generates `client-package-proxy/<source>` modules from that metadata. The relevant source path is here: https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-rsc/src/plugin.ts

Before this change, vinext excluded the bare `vinext` package but not `vinext/shims/*`. That leaves known client shim subpaths eligible for dep optimization, which can make the package source recorded by plugin-rsc diverge from the optimized module that produced the client reference metadata. The result is the `clientReferenceMetaMap` miss reported in #1008.

## Approach

- Keep the fixed client shim list and exclude merge behavior in a normal plugin helper module.
- Reuse that helper from vinext's Vite config setup instead of spreading another inline `Set` expression through the config hook.
- Keep codegen out of the behavior. The generated entries continue to describe app shape, while the normal plugin module owns the optimizer policy.
- Scope the list to current vinext shims that are actual top-level client modules and can appear as package subpath imports: `error-boundary`, `form`, `layout-segment-context`, `link`, `script`, and `slot`.

## Validation

- `vp check packages/vinext/src/index.ts packages/vinext/src/plugins/rsc-client-shim-excludes.ts tests/build-optimization.test.ts`
- `vp test run tests/build-optimization.test.ts -t "optimizeDeps.exclude"`
- `vp test run tests/app-router.test.ts -t "renders next/link"`

## Risks / follow-ups

The main maintenance risk is list drift: if vinext adds another top-level `"use client"` shim that can be imported as `vinext/shims/<name>` from the RSC graph, it should be added to the helper list.

### DIFF
--- a/packages/vinext/src/plugins/rsc-client-shim-excludes.ts
+++ b/packages/vinext/src/plugins/rsc-client-shim-excludes.ts
@@ -8,6 +8,7 @@ const RSC_CLIENT_SHIM_OPTIMIZE_DEPS_EXCLUDE = Object.freeze([
   "vinext/shims/link",
   "vinext/shims/script",
   "vinext/shims/slot",
+  "vinext/shims/offline",
 ]);
 
 export const VINEXT_OPTIMIZE_DEPS_EXCLUDE = Object.freeze([
--- a/packages/vinext/src/plugins/rsc-client-shim-excludes.ts
+++ b/packages/vinext/src/plugins/rsc-client-shim-excludes.ts
@@ -1,0 +1,34 @@
+const RSC_CLIENT_SHIM_OPTIMIZE_DEPS_EXCLUDE = Object.freeze([
+  // @vitejs/plugin-rsc tracks package client references by the original
+  // bare source. If Vite pre-bundles these known client shims, the generated
+  // client-package proxy can lose the matching export metadata in dev.
+  "vinext/shims/error-boundary",
+  "vinext/shims/form",
+  "vinext/shims/layout-segment-context",
+  "vinext/shims/link",
+  "vinext/shims/script",
+  "vinext/shims/slot",
+]);
+
+export const VINEXT_OPTIMIZE_DEPS_EXCLUDE = Object.freeze([
+  "vinext",
+  "@vercel/og",
+  ...RSC_CLIENT_SHIM_OPTIMIZE_DEPS_EXCLUDE,
+]);
+
+export function mergeOptimizeDepsExclude(
+  ...excludeGroups: readonly (readonly string[])[]
+): string[] {
+  const seen = new Set<string>();
+  const merged: string[] = [];
+
+  for (const group of excludeGroups) {
+    for (const entry of group) {
+      if (seen.has(entry)) continue;
+      seen.add(entry);
+      merged.push(entry);
+    }
+  }
+
+  return merged;
+}
--- a/packages/vinext/src/plugins/rsc-client-shim-excludes.ts
+++ b/packages/vinext/src/plugins/rsc-client-shim-excludes.ts
@@ -20,15 +20,13 @@ export function mergeOptimizeDepsExclude(
   ...excludeGroups: readonly (readonly string[])[]
 ): string[] {
   const seen = new Set<string>();
-  const merged: string[] = [];
 
   for (const group of excludeGroups) {
     for (const entry of group) {
       if (seen.has(entry)) continue;
       seen.add(entry);
-      merged.push(entry);
     }
   }
 
-  return merged;
+  return [...seen];
 }
--- a/tests/build-optimization.test.ts
+++ b/tests/build-optimization.test.ts
@@ -112,6 +112,7 @@ describe("optimizeDeps.exclude for vinext", () => {
     "vinext/shims/link",
     "vinext/shims/script",
     "vinext/shims/slot",
+    "vinext/shims/offline",
   ];
 
   it("excludes vinext at top level for Pages Router builds", async () => {
--- a/tests/build-optimization.test.ts
+++ b/tests/build-optimization.test.ts
@@ -105,6 +105,15 @@ describe("clientManualChunks", () => {
 // ─── optimizeDeps.exclude — prevents esbuild scanning virtual module imports ─
 
 describe("optimizeDeps.exclude for vinext", () => {
+  const rscClientShimExcludes = [
+    "vinext/shims/error-boundary",
+    "vinext/shims/form",
+    "vinext/shims/layout-segment-context",
+    "vinext/shims/link",
+    "vinext/shims/script",
+    "vinext/shims/slot",
+  ];
+
   it("excludes vinext at top level for Pages Router builds", async () => {
     const vinext = (await import("../packages/vinext/src/index.js")).default;
     const plugins = vinext();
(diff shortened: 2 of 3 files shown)


packages/vinext/src/plugins/rsc-client-shim-excludes.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  for (const group of excludeGroups) {
+    for (const entry of group) {
+      if (seen.has(entry)) continue;
+      seen.add(entry);
+      merged.push(entry);
+    }
+  }
+
+  return merged;
```
james-elicx:
```suggestion
  const seen = new Set<string>();

  for (const group of excludeGroups) {
    for (const entry of group) {
      if (seen.has(entry)) continue;
      seen.add(entry);
    }
  }

  return [...seen];
```

packages/vinext/src/plugins/rsc-client-shim-excludes.ts
```diff
@@ -0,0 +1,34 @@
...
+  // @vitejs/plugin-rsc tracks package client references by the original
+  // bare source. If Vite pre-bundles these known client shims, the generated
+  // client-package proxy can lose the matching export metadata in dev.
+  "vinext/shims/error-boundary",
+  "vinext/shims/form",
+  "vinext/shims/layout-segment-context",
+  "vinext/shims/link",
+  "vinext/shims/script",
+  "vinext/shims/slot",
+]);
```
james-elicx, line 12:
```suggestion
  "vinext/shims/slot",
  "vinext/shims/offline",
]);
```

tests/build-optimization.test.ts
```diff
@@ -105,6 +105,15 @@ describe("clientManualChunks", () => {
...
 
 describe("optimizeDeps.exclude for vinext", () => {
+  const rscClientShimExcludes = [
+    "vinext/shims/error-boundary",
+    "vinext/shims/form",
+    "vinext/shims/layout-segment-context",
+    "vinext/shims/link",
+    "vinext/shims/script",
+    "vinext/shims/slot",
+  ];
```
james-elicx, line 116:
```suggestion
    "vinext/shims/slot",
    "vinext/shims/offline",
  ];
```

PR #1021, opened 2026-05-02
fix: add hydration-level onError replay in next/image via useLayoutEffect

### LINKED ISSUE #1007: next/image: hydration-level onError replay via useLayoutEffect
Identified during review of #1004.

## Context
The upstream Next.js PR (https://github.com/vercel/next.js/pull/93209) uses a `useLayoutEffect` + `img.src = img.src` pattern to replay error events that are lost before React hydration completes. The vinext implementation in #1004 uses a simpler `useRef`-based dedup that doesn't handle this case.

## Impact
Images that fail to load *before* React hydration completes (e.g., during SSR streaming or initial HTML parse) won't fire `onError` callbacks. This means error state management in parent components won't trigger for pre-hydration failures.

## Suggested approach
Add a `useLayoutEffect` that, when `onError` is provided, sets `img.src = img.src` on the ref to re-trigger the load/error cycle after hydration. This mirrors the upstream Next.js fix in vercel/next.js#93209.

Divkix:
Fixes #1007

## Summary
- Ports the upstream Next.js hydration replay pattern from vercel/next.js#93209
- Adds `useNonWarningLayoutEffect` + `img.src = img.src` to re-trigger the error trajectory when an image fails before React hydration completes (e.g. during SSR streaming or initial HTML parse)
- Adds `mergedRef` to capture the underlying img element and forward the parent ref simultaneously
- Forwards ref to `@unpic/react` Image elements (supports ref forwarding)
- Adds SSR rendering regression tests for onError, onLoad+onError, ref forwarding, and loader paths

## Test plan
- Added 5 new SSR rendering tests in `tests/shims.test.ts` covering onError, onLoad+onError, ref forwarding, and loader path
- Full test suite: 4668 passed (1 pre-existing unrelated `afterAll` timeout in `pages-router.test.ts`)
- Lint, format, and typecheck all pass

### DIFF
--- a/packages/vinext/src/shims/image.tsx
+++ b/packages/vinext/src/shims/image.tsx
@@ -12,7 +12,7 @@
  * `images.domains` from next.config.js. Unmatched URLs are blocked
  * in production and warn in development, matching Next.js behavior.
  */
-import React, { forwardRef, useRef } from "react";
+import React, { forwardRef, useCallback, useEffect, useLayoutEffect, useRef } from "react";
 import { Image as UnpicImage } from "@unpic/react";
 import { hasRemoteMatch, type RemotePattern } from "./image-config.js";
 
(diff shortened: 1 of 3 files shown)


packages/vinext/src/shims/image.tsx
```diff
@@ -236,13 +282,97 @@ const Image = forwardRef<HTMLImageElement, ImageProps>(function Image(
...
+
+  const mergedRef = useCallback((node: HTMLImageElement | null) => {
+    imgElementRef.current = node;
+    const currentRef = forwardedRef.current;
+    if (typeof currentRef === "function") {
+      currentRef(node);
+    } else if (currentRef) {
+      (currentRef as React.RefObject<HTMLImageElement | null>).current = node;
+    }
+  }, []);
```
james-elicx: to be honest i'm finding this a tad confusing when comparing to the next.js implementation if this is an inlined version of that pattern.

reading it, i feel like it might have different behaviour to next.js when the parent ref changes. i think we should just port their hook properly and use that here instead for better parity.

Divkix: Yeah you're right. The [] deps mean the callback identity never changes, so when a parent re-renders with a new inline ref callback React never re-invokes it. Next.js's version returns useCallback(..., [refA, refB]) — React sees the identity change and runs the ref lifecycle (old gets null, new gets the element). Also missing React 19 cleanup ref support.
I'll port useMergedRef directly from next/src/client/use-merged-ref.ts and use it here instead.

PR #1034, opened 2026-05-03
refactor(app-rsc): extract request normalization into server/app-rsc-request-normalization

NathanDrake2406:
## What this changes

Introduces `server/app-rsc-request-normalization.ts` and wires it into the App
Router RSC entry. The generated entry no longer owns the normalization logic
itself; it calls `normalizeRscRequest` and destructures the typed result.

Also extracts `normalizeMountedSlotsHeader` into a neutral module
(`server/app-mounted-slots-header.ts`) used by all three consumers: request
normalization (reads the incoming header), app-elements (outgoing header
construction), and isr-cache (RSC cache key generation). The generated RSC
entry no longer calls the cache-layer export directly; request normalization
now owns reading and canonicalising the incoming header.

## Why

The normalization pipeline is the most security and compatibility-sensitive
part of the request lifecycle: protocol-relative URL open redirects,
`%2F`-encoded path segment boundaries, basePath bypass via `/__vinext/`,
RSC content-type detection, and null-byte header injection all live here.
Having this logic inlined in a template string means:

- Zero unit test coverage for the security-sensitive paths
- Impossible to audit the ordering invariants (guard must fire before
  `normalizePath` collapses `//`) by reading a single file
- `normalizeMountedSlotsHeader` duplicated across `isr-cache.ts` and
  `app-elements.ts` (with the copy in `app-elements.ts` already diverging
  from the canonical `isr-cache.ts` version)

## Approach

`normalizeRscRequest(request, basePath)` encodes early exits as `Response`
returns (400/404) and success as `NormalizedRscRequest`. The discriminant is
`instanceof Response`, which callers already use for every other early-exit
check in the entry. No new control-flow patterns introduced.

Step ordering inside the function is documented with numbered comments that
explain the security constraint each ordering decision satisfies:

- Step 2 (protocol-relative guard) must precede step 4 (`normalizePath`):
  `normalizePath` collapses `//evil.com` to `/evil.com`, causing the guard
  to miss it. Source: [vercel/next.js — `server/lib/router-utils/guard-protocol-relative-url.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/router-utils/guard-protocol-relative-url.ts)

- Step 3 (strict percent-decode) must precede step 5 (basePath check):
  a `%2F`-encoded slash in the basePath position would decode to `/` and
  create a fake match. Source: [vercel/next.js — `server/lib/router-utils/decode-path-params.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/router-utils/decode-path-params.ts)

- Step 5 (`/__vinext/` bypass): internal prerender endpoints must be
  reachable regardless of the configured basePath. These are
  `/__vinext/prerender/*` endpoints consumed by `wrangler unstable_startWorker`
  during Cloudflare Workers builds.

`normalizeMountedSlotsHeader` moves to `server/app-mounted-slots-header.ts` and
is re-exported from `isr-cache.ts` for backward compatibility. Only the
`app-elements.ts` private copy is removed — the generated entry was already
importing it from `isr-cache.ts`.

## Validation

44 behavior tests in `tests/app-rsc-request-normalization.test.ts` cover:

- Protocol-relative variants: `//`, `/\`, `/%5C`, `/%2F`
- Ordering guarantee: guard fires before `normalizePath` (tested explicitly)
- Malformed percent sequences: `%GG`, truncated `%`, single-digit `%A`
- basePath enforcement: missing prefix, exact match, non-segment-boundary prefix
- `/__vinext/` bypass under a configured basePath
- `%2F` preservation (not decoded to path separator)
- RSC detection: `.rsc` suffix and `Accept: text/x-component`
- `cleanPathname` stripping
- Interception context header: absent, empty, legitimate value, null-byte injection
- Mounted slots: sort, dedup, absent, blank

395 tests pass across the 6 affected test files (new + isr-cache, app-elements,
request-pipeline, entry-templates, app-router).

## Risks / follow-ups

The `app-rsc-entry.ts` generator now has one fewer resolved path variable
(`normalizePathnameForRouteMatchStrict` and `guardProtocolRelativeUrl` no
longer need separate `resolveEntryPath` entries). The `normalizePathModulePath`
and `routingUtilsPath` constants are retained because `__normalizePath` and
`__normalizePathnameForRouteMatch` are still used in the `handler()` outer
function for `applyConfigHeadersToResponse`.

No behavior change. The generated RSC entry produces the same request handling
semantics; only the source of the logic has moved.

### DIFF
--- a/tests/app-rsc-request-normalization.test.ts
+++ b/tests/app-rsc-request-normalization.test.ts
@@ -1,0 +1,340 @@
+/**
+ * Behavior tests for normalizeRscRequest and normalizeMountedSlotsHeader.
+ *
+ * These functions sit at the security and compatibility boundary of the App
+ * Router request pipeline. Wrong behavior here produces:
+ *   - Open redirect vulnerabilities (protocol-relative bypass)
+ *   - Route confusion (malformed %2F interpreted as path separator)
+ *   - Cache fragmentation (non-canonical slot headers)
+ *   - 500s instead of 400s on bad client input
+ *
+ * Each test names the observable failure on regression, not the implementation
+ * detail being exercised.
+ */
+import { describe, it, expect } from "vite-plus/test";
+import {
+  normalizeRscRequest,
+  normalizeMountedSlotsHeader,
+  type NormalizedRscRequest,
+} from "../packages/vinext/src/server/app-rsc-request-normalization.js";
+
+function req(path: string, headers: Record<string, string> = {}): Request {
+  return new Request(`http://localhost${path}`, { headers });
+}
+
+function normalized(result: Response | NormalizedRscRequest): NormalizedRscRequest {
+  if (result instanceof Response) {
+    throw new Error(`Expected NormalizedRscRequest but got Response(${result.status})`);
+  }
+  return result;
+}
+
+// ── Protocol-relative URL guard ─────────────────────────────────────────────
+
+describe("normalizeRscRequest — protocol-relative URL guard", () => {
+  it("returns 404 for // path so trailing-slash redirect cannot emit open-redirect Location", () => {
+    // Regression for: trailing-slash 308 echoes //evil.com → open redirect.
+    const result = normalizeRscRequest(req("//evil.com/path"), "");
+    expect(result).toBeInstanceOf(Response);
+    expect((result as Response).status).toBe(404);
+  });
+
+  it("returns 404 for /\\ path (browsers normalize \\ to / in Location headers)", () => {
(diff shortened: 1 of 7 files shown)


tests/app-rsc-request-normalization.test.ts
james-elicx, line 1: these sorts of tests would benefit from using it.each

james-elicx:
Screenshot 2026-05-03 at 17 45 21

NathanDrake2406:
@james-elicx I was letting it polling until the thing is merged lol

PR #1050, opened 2026-05-04
fix(thenable-params): protect well-known properties and unify page/route handler implementation

NathanDrake2406:
## What this changes

Unifies thenable params construction for App Router pages and route handlers under a single shared implementation, and expands the protected property set from `then/catch/finally` to the full well-known property list that Next.js reserves.

## Why

Route handlers used a local `Object.assign(Promise.resolve(...), params)` helper that did not protect **any** well-known properties from shadowing. If a route segment were named `then`, `catch`, `finally`, or `status`, the resulting params object would shadow the Promise/React methods and break `await params`, React introspection, or spread behaviour.

Page params used a stronger Proxy, but it only guarded `then/catch/finally`, missing React's Promise `status` field and other well-known properties like `toString`, `hasOwnProperty`, etc.

This inconsistency meant pages and route handlers behaved differently for the same URL segments, and certain param names could silently break runtime behaviour.

## Approach

- Expanded `packages/vinext/src/shims/thenable-params.ts` to protect the full well-known property set defined in Next.js `reflect-utils.ts`:
  - Promise prototype: `then`, `catch`, `finally`
  - React Promise extension: `status`
  - Object prototype: `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable`, `toString`, `valueOf`, `toLocaleString`
  - React introspection: `displayName`, `_debugInfo`
  - Common tested properties: `toJSON`, `$$typeof`, `__esModule`, `@@iterator`
- Replaced the weak local `makeThenableParams` in `app-route-handler-dispatch.ts` with the shared shim, eliminating the duplicate implementation.
- Added explicit `ThenableParams<T>` return type to document the contract at the boundary.

## Validation

- Unit tests (`tests/thenable-params.test.ts`) now verify:
  - `await params` works when params contain `then`
  - Sync access to non-well-known params still works
  - `params.then` / `params.catch` / `params.finally` remain functions (Promise methods are not shadowed)
  - `params.status` is protected from React Promise shadowing
  - `Object.keys(params)` excludes well-known properties
  - After `await params`, the resolved object contains the actual param values

- App Router integration tests (`tests/app-router.test.ts`) now verify:
  - Pages render correctly with params named `then`, `catch`, `finally`, and `status`
  - Route handlers return correct JSON with the same param names
  - The params object remains thenable before awaiting

## Risks / follow-ups

None. This is a pure bugfix that makes route handler params behave identically to page params. No public API changes.

## References

- Next.js `wellKnownProperties` source: [packages/next/src/shared/lib/utils/reflect-utils.ts](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/utils/reflect-utils.ts)
- Next.js param shadowing tests: [test/e2e/app-dir/cache-components/cache-components.params.test.ts](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/cache-components/cache-components.params.test.ts)

### DIFF
--- a/packages/vinext/src/server/app-route-handler-cache.ts
+++ b/packages/vinext/src/server/app-route-handler-cache.ts
@@ -9,6 +9,7 @@ import {
   buildRouteHandlerCachedResponse,
 } from "./app-route-handler-response.js";
 import { markKnownDynamicAppRoute } from "./app-route-handler-runtime.js";
+import { makeThenableParams } from "vinext/shims/thenable-params";
 import {
   runAppRouteHandler,
   type AppRouteDebugLogger,
@@ -105,7 +106,7 @@ export async function readAppRouteHandlerCacheResponse(
             handlerFn: options.handlerFn,
             i18n: options.i18n,
             markDynamicUsage: options.markDynamicUsage,
-            params: options.params,
+            params: makeThenableParams(options.params),
             request: new Request(options.requestUrl, { method: "GET" }),
             routePattern: options.routePattern,
             setHeadersAccessPhase: options.setHeadersAccessPhase,
--- a/packages/vinext/src/server/app-route-handler-dispatch.ts
+++ b/packages/vinext/src/server/app-route-handler-dispatch.ts
@@ -37,6 +37,7 @@ import {
 } from "./app-route-handler-response.js";
 import { createStaticGenerationHeadersContext } from "./app-static-generation.js";
 import { buildPageCacheTags } from "./implicit-tags.js";
+import { makeThenableParams } from "vinext/shims/thenable-params";
 import { reportRequestError } from "./instrumentation.js";
 
 type AppRouteHandlerDispatchRoute = {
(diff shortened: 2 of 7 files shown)


NathanDrake2406:
@copilot-pull-request-reviewer Both review comments addressed in the amended commit:

**1. value/error omitted from well-known set:**
You are correct. Next.js comments out \"value\" and \"error\" in reflect-utils.ts because they use Promise.resolve(underlyingParams) directly in production — React mutations on the promise object are never shadowed. Our Proxy intercepts sync reads through a separate plain object, which means a param named value or error would shadow React's .status/.value/.error attachments that React adds to resolved promises for use() caching.

I added both value and error to the runtime well-known set, with a comment explaining this divergence from Next.js. Unit tests now verify that params.value and params.error return Promise internals, not param values, and that the resolved object contains the actual param values after awaiting.

**2. ThenableParams<T> overstates the synchronous shape:**
You are correct. The old Promise<T> & T type allowed callers to write params.status and have TypeScript believe it was T[\"status\"], when at runtime the Proxy returns undefined (or React internals).

I replaced the type with Promise<T> & Omit<T, WellKnownProperty>, where WellKnownProperty is a type-level union that stays in sync with the runtime set. This means:
- params.then is typed as the Promise method (not T[\"then\"])
- params.status is not accessible on the thenable (must await first)
- params.slug is still accessible as T[\"slug\"]
- After await params, the resolved object has all properties including well-known ones

The tests use Reflect.get(params, \"status\") for the well-known property checks, which TypeScript correctly allows because the type no longer claims those properties exist on the thenable.

PR #1051, opened 2026-05-04
fix(routing): discover nested parallel slot sub-routes from layout-only parents

NathanDrake2406:
## What this changes

Nested parallel routes now work when the parent segment has a layout but no `page.tsx`. Previously, `discoverSlotSubRoutes` in `app-route-graph.ts` skipped any parent route without a `pagePath`, so parallel slot sub-pages like `@parallelB/nested/page.tsx` under `home/layout.tsx` (with no `home/page.tsx`) were never discovered as valid URL routes.

This is a high-value App Router parity fix. Next.js explicitly tests this shape in `parallel-routes-and-interception`:
- `app/parallel-nested/home/layout.tsx`
- `app/parallel-nested/home/@parallelB/default.tsx`
- `app/parallel-nested/home/@parallelB/nested/page.tsx`

...creates a valid `/parallel-nested/home/nested` route even though there is no `home/page.tsx`.

## Approach

The fix makes four targeted changes to `discoverSlotSubRoutes`:

1. **Remove the hard `pagePath` requirement.** Layout-only routes with parallel slots can still own nested slot sub-routes.

2. **Compute `parentPageDir` from the innermost layout when `pagePath` is null.** This keeps slot ownership checks working correctly for layout-only routes.

3. **Relax the `childrenDefault` (`default.tsx`) requirement, but only for layout-only parent routes.** When a parent HAS a children page, `default.tsx` is still required as a fallback for the synthetic sub-route. Layout-only parents do not need this fallback because the children slot was never occupied at the parent level.

4. **Add structural-conflict detection before creating synthetic routes.** This prevents layout-only root routes from generating synthetic routes that collide with existing page routes differing only by param name (e.g. `/shop/:id` vs `/shop/:name`), which `validateRoutePatterns` rejects. The slot content for the existing route is handled by `findMirroredSlotPage` instead.

## Validation

- Added unit test in `tests/routing.test.ts`: "discovers nested parallel slot sub-routes from layout-only parent" (ported from Next.js test suite)
- Added integration test in `tests/app-router.test.ts`: "renders nested parallel route from layout-only parent"
- Added fixture files under `tests/fixtures/app-basic/app/parallel-nested/` matching the Next.js reference fixture
- All 419 routing-related tests pass:
  - `tests/routing.test.ts` (106 tests)
  - `tests/app-route-graph.test.ts` (11 tests)
  - `tests/app-router.test.ts` (301 tests)
  - `tests/intercepting-routes-build.test.ts` (1 test)
- Lint passes with 0 warnings/errors

## Risks / follow-ups

- The structural-conflict check uses pattern-part comparison. In theory, two routes could have the same shape but different param names and both be legitimate (e.g. one from a slot, one from a page). Next.js rejects this at build time, so skipping the synthetic route is the safe choice.
- This fix only affects route discovery. Render-time slot resolution via `findMirroredSlotPage` is unchanged.

## References

- Next.js test: [parallel-routes-and-interception.test.ts#L510](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/parallel-routes-and-interception/parallel-routes-and-interception.test.ts)
- Next.js fixture: [app/parallel-nested](https://github.com/vercel/next.js/tree/canary/test/e2e/app-dir/parallel-routes-and-interception/app/parallel-nested)

### DIFF
--- a/packages/vinext/src/routing/app-route-graph.ts
+++ b/packages/vinext/src/routing/app-route-graph.ts
@@ -371,6 +371,13 @@ function discoverSlotSubRoutes(routes: AppRoute[], matcher: ValidFileMatcher): A
     // the synthetic sub-route has a fallback for the children slot. Layout-only
     // parent routes (no page.tsx) do not need a default — the children slot was
     // never occupied at the parent level, so the sub-route simply renders null.
+    //
+    // TODO: layout-only synthetic routes have pagePath: null, which means they
+    // are skipped by the prerender phase (build/prerender.ts line 849). These
+    // routes work correctly in dev/SSR but won't be statically exported. This
+    // is pre-existing behaviour for all layout-only routes; fixing it requires
+    // teaching prerender to classify routes by their parallel slot pages, not
+    // just the children pagePath.
     const childrenDefault = findFile(parentPageDir, "default", matcher);
     if (parentRoute.pagePath && !childrenDefault) continue;
 
@@ -398,8 +405,10 @@ function discoverSlotSubRoutes(routes: AppRoute[], matcher: ValidFileMatcher): A
       // Skip synthetic routes that would structurally conflict with an existing
       // route (same shape, different param names). The slot content is handled
       // by findMirroredSlotPage for the existing route instead.
+      // Scan routesByPattern (not just the original routes array) so synthetic
+      // routes created earlier in this loop are also visible.
       const syntheticParts = [...parentRoute.patternParts, ...urlParts];
-      const hasStructuralConflict = routes.some((r) =>
+      const hasStructuralConflict = Array.from(routesByPattern.values()).some((r) =>
         patternsStructurallyEquivalent(r.patternParts, syntheticParts),
       );
       if (hasStructuralConflict) continue;
--- a/packages/vinext/src/routing/app-route-graph.ts
+++ b/packages/vinext/src/routing/app-route-graph.ts
@@ -302,6 +302,9 @@ function discoverSlotSubRoutes(routes: AppRoute[], matcher: ValidFileMatcher): A
   for (const parentRoute of routes) {
     if (parentRoute.parallelSlots.length === 0) continue;
 
+    // For page-bearing routes, the route directory is the page's directory.
+    // For layout-only routes (no page.tsx), proxy the route directory through
+    // the innermost layout — it lives at the same filesystem level as the route.
     const parentPageDir = parentRoute.pagePath
       ? path.dirname(parentRoute.pagePath)
       : path.dirname(parentRoute.layouts[parentRoute.layouts.length - 1]);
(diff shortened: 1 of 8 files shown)


NathanDrake2406 (commented):
Thanks for the review. Addressing each comment:

**Comment 1 (patternsStructurallyEquivalent literal vs dynamic):** Already fixed in commit d1fc3650 before this review was posted. Replaced the ad-hoc `startsWith(":")` check with `segmentTreeNodeType` which classifies each segment by its Next.js validator tree node (literal, dynamic, catchAll, optionalCatchAll). Literal and dynamic segments now correctly map to different types.

**Comments 2 & 4 (structural conflict misses synthetic routes):** Valid. Fixed in commit bfaf8113. Changed `routes.some(...)` to `Array.from(routesByPattern.values()).some(...)`. Since `routesByPattern` is updated with each synthetic route as it's created, this now catches conflicts between synthetic routes generated earlier in the same pass.

**Comment 3 (downstream build code skips !pagePath):** Partially valid observation, but this is pre-existing behaviour for ALL layout-only routes — not new with this PR. The `build/prerender.ts:849` skip has always affected routes like `/slot-collision` and `/parallel-sibling-modal/[username]` which are layout-only with parallel slots. Before this PR, nested parallel routes from layout-only parents were simply 404s (not discovered at all). Now they work correctly in dev/SSR. Static export support for layout-only routes is a broader gap that requires teaching `classifyAppRoute` and prerender to handle routes without a children `pagePath`, which is out of scope for this fix. Added a TODO comment in `app-route-graph.ts:369-375` documenting this known limitation.

All 418 routing-related tests continue to pass.

NathanDrake2406:
Thanks for the review. Addressing each comment:

**Comment 1 (patternsStructurallyEquivalent literal vs dynamic):** Already fixed in commit d1fc3650 before this review was posted. Replaced the ad-hoc `startsWith(":")` check with `segmentTreeNodeType` which classifies each segment by its Next.js validator tree node (literal, dynamic, catchAll, optionalCatchAll). Literal and dynamic segments now correctly map to different types.

**Comments 2 & 4 (structural conflict misses synthetic routes):** Valid. Fixed in commit bfaf8113. Changed `routes.some(...)` to `Array.from(routesByPattern.values()).some(...)`. Since `routesByPattern` is updated with each synthetic route as it's created, this now catches conflicts between synthetic routes generated earlier in the same pass.

**Comment 3 (downstream build code skips !pagePath):** Partially valid observation, but this is pre-existing behaviour for ALL layout-only routes — not new with this PR. The `build/prerender.ts:849` skip has always affected routes like `/slot-collision` and `/parallel-sibling-modal/[username]` which are layout-only with parallel slots. Before this PR, nested parallel routes from layout-only parents were simply 404s (not discovered at all). Now they work correctly in dev/SSR. Static export support for layout-only routes is a broader gap that requires teaching `classifyAppRoute` and prerender to handle routes without a children `pagePath`, which is out of scope for this fix. Added a TODO comment in `app-route-graph.ts:369-375` documenting this known limitation.

All 418 routing-related tests continue to pass.

PR #1054, opened 2026-05-04
fix(routing): accept any non-] chars in dynamic segment names (Next.js parity)

NathanDrake2406:
## Summary

Aligns vinext's dynamic segment name parsing with Next.js `PARAMETER_PATTERN`, which accepts **any non-`\]` characters** inside brackets. vinext was restricting param names to `[\w-]+`, silently treating valid Next.js segments like `[repo:name]`, `[variant.id]`, and `[user@domain]` as literal URL segments.

### Regex fixes (4 files)

| File | Before | After |
|------|--------|-------|
| `routing/app-route-graph.ts` | `/^\[\.\.\.([\w-]+)\]$/` | `/^\[\.\.\.([^\]]+)\]$/` |
| `routing/pages-router.ts` | same | same |
| `routing/route-validation.ts` | `/:([\w-]+)\+/g` | `/:([^/]+)\+/g` |
| `shims/router.ts` | `\[{1,2}(?:\.\.\.)?([\w-]+)\]{1,2}` | `\[{1,2}(?:\.\.\.)?([^\]]+)\]{1,2}` |

## Next.js reference

- **`PARAMETER_PATTERN`**: [packages/next/src/shared/lib/router/utils/get-dynamic-param.ts#L175](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/router/utils/get-dynamic-param.ts#L175)
  ```ts
  export const PARAMETER_PATTERN = /^([^[]*)\[((?:\[[^\]]*\])|[^\]]+)\](.*)$/
  ```
  The `[^\]]+` group matches any character except `\]`, which is why `[repo:name]` is valid in Next.js.

## Test coverage

Added a test in `tests/app-route-graph.test.ts` that verifies routes with dotted, colon, and at-sign param names are materialized correctly.

### DIFF
--- a/packages/vinext/src/routing/pages-router.ts
+++ b/packages/vinext/src/routing/pages-router.ts
@@ -113,29 +113,35 @@ function fileToRoute(file: string, pagesDir: string, matcher: ValidFileMatcher):
   for (let i = 0; i < segments.length; i++) {
     const segment = segments[i];
 
-    // Catch-all: [...slug] -> :slug+ (param names may contain hyphens)
-    const catchAllMatch = segment.match(/^\[\.\.\.([\w-]+)\]$/);
+    // Catch-all: [...slug] -> :slug+ (param names may contain any non-] chars)
+    // Matches Next.js PARAMETER_PATTERN.
+    const catchAllMatch = segment.match(/^\[\.\.\.([^\]]+)\]$/);
     if (catchAllMatch) {
       if (i !== segments.length - 1) return null;
+      // Guard: names ending in + or * would collide with internal pattern modifiers.
+      if (catchAllMatch[1].endsWith("+") || catchAllMatch[1].endsWith("*")) return null;
       isDynamic = true;
       params.push(catchAllMatch[1]);
       urlSegments.push(`:${catchAllMatch[1]}+`);
       continue;
     }
 
-    // Optional catch-all: [[...slug]] -> :slug* (param names may contain hyphens)
-    const optionalCatchAllMatch = segment.match(/^\[\[\.\.\.([\w-]+)\]\]$/);
+    // Optional catch-all: [[...slug]] -> :slug* (param names may contain any non-] chars)
+    const optionalCatchAllMatch = segment.match(/^\[\[\.\.\.([^\]]+)\]\]$/);
     if (optionalCatchAllMatch) {
       if (i !== segments.length - 1) return null;
+      if (optionalCatchAllMatch[1].endsWith("+") || optionalCatchAllMatch[1].endsWith("*"))
+        return null;
       isDynamic = true;
       params.push(optionalCatchAllMatch[1]);
       urlSegments.push(`:${optionalCatchAllMatch[1]}*`);
       continue;
     }
 
-    // Dynamic segment: [id] -> :id (param names may contain hyphens)
-    const dynamicMatch = segment.match(/^\[([\w-]+)\]$/);
+    // Dynamic segment: [id] -> :id (param names may contain any non-] chars)
+    const dynamicMatch = segment.match(/^\[([^\]]+)\]$/);
     if (dynamicMatch) {
+      if (dynamicMatch[1].endsWith("+") || dynamicMatch[1].endsWith("*")) return null;
       isDynamic = true;
       params.push(dynamicMatch[1]);
       urlSegments.push(`:${dynamicMatch[1]}`);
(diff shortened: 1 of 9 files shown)


NathanDrake2406:
## Review fixes applied

### 1. Guard against param names ending in `+` or `*` (both App Router and Pages Router)

As noted in the Copilot review, the broadened regex `[^\]]+` would match segments like `[id+]`, which would map to internal pattern `:id+` — indistinguishable from a catch-all. Added an `endsWith('+') || endsWith('*')` guard in both `app-route-graph.ts` and `pages-router.ts` that returns `null` (skips the route) when a param name would collide with internal pattern modifiers.

### 2. Pages Router test coverage

Added a `pagesRouter - dotted/colon param names` describe block in `tests/routing.test.ts` with three tests:
- Discovers dynamic segments with dots and colons
- Extracts params correctly for dotted dynamic segments
- Skips routes whose param names end in `+` or `*`

### 3. App Router test for skipped routes

Added a test in `tests/app-route-graph.test.ts` verifying that `[id+]/page.tsx` and `[id*]/page.tsx` produce zero routes (skipped entirely).

NathanDrake2406:
## Second review fixes applied

All three issues from ask-bonk's review are addressed:

### 1. `patternToNextFormat` backtracking bug (marked HIGH)

**Root cause:** The greedy `[^/]+` in `/:([^/]+)\+/g` would backtrack when param names contain internal `+` or `*`, splitting the name incorrectly:
- `:c++lang` → `[...c+]lang` (wrong)
- `:a+b` → `[...a]b` (wrong)

**Fix:** Non-greedy `[^/]+?` with `(?=\/|$)` lookahead so the regex requires `+`/`*` to be immediately before `/` or end-of-string. Added regression test: `/:c++lang` → `/[c++lang]`, `/utils/:a+b/c` → `/utils/[a+b]/c`.

### 2. `pages-server-entry.ts` duplicate `patternToNextFormat` (marked HIGH)

The generated entry had an inline copy using the old `[\w]+` regex — dev/prod would diverge for dotted/colon param names. Updated to match the non-greedy-lookahead regex from `route-validation.ts`. Since `route-validation.ts` is re-exported through `pages-router.ts` and imported by `pages-client-entry.ts`, the same function definition propagates to both dev and prod paths.

### 3. Nit fixes
- Comment wording: `+/* suffixes` instead of `+ / * suffixes`
- Removed unnecessary `mkdir` calls creating empty directories in Pages Router test
- Added Windows NTFS `:` comment on the `[repo:name]` test fixture

NathanDrake2406:
## Third review fixes applied

Both cleanup points are addressed:

### 1. Windows portability — removed `:` from filesystem fixtures

The `[repo:name]/page.tsx` filesystem fixture (which creates a `:` directory) is removed. Colon conversion is now tested through `patternToNextFormat` (pure string, no filesystem):
- `route-sorting.test.ts`: `expect(patternToNextFormat("/:repo:name")).toBe("/[repo:name]")`

### 2. Catch-all test coverage with broadened names

Added a dedicated test in `app-route-graph.test.ts` covering the catch-all surface:
- `[...variant.id]` → `/:variant.id+` (catch-all with dotted name)
- `[[...user@domain]]` → `/blog/:user@domain*` (optional catch-all with at-sign name)

packages/vinext/src/routing/pages-router.ts
```diff
@@ -113,29 +113,35 @@ function fileToRoute(file: string, pagesDir: string, matcher: ValidFileMatcher):
   for (let i = 0; i < segments.length; i++) {
     const segment = segments[i];
 
-    // Catch-all: [...slug] -> :slug+ (param names may contain hyphens)
-    const catchAllMatch = segment.match(/^\[\.\.\.([\w-]+)\]$/);
+    // Catch-all: [...slug] -> :slug+ (param names may contain any non-] chars)
+    // Matches Next.js PARAMETER_PATTERN.
+    const catchAllMatch = segment.match(/^\[\.\.\.([^\]]+)\]$/);
```
james-elicx, line 118: would be good to see if we can extract some of the route regexes and re-use across places instead of re-writing the pattern in each place

packages/vinext/src/routing/route-validation.ts
```diff
@@ -148,10 +148,13 @@ class UrlNode {
 export function patternToNextFormat(pattern: string): string {
```
james-elicx, line 148: use this util in pages server entry?

PR #1068, opened 2026-05-05
perf(pages): cache ISR misses from streamed render

NathanDrake2406:
## What this changes

Pages Router ISR cache misses now cache the same body stream used for the live HTML response instead of rendering the page a second time just to build cache HTML.

The response path keeps the existing render-before-head-collection order, tees the completed body stream when ISR is eligible, sends one branch to the client, and records the other branch into the Pages ISR cache. Nonce-bearing responses still skip ISR writes.

## Why

`renderPagesPageResponse` previously rendered the page once for the live response, then created a second page element and called `renderIsrPassToStringAsync()` to generate cached HTML. That second pass can repeat userland render work, data reads, Suspense work, and render-scoped state mutations on every Pages ISR MISS.

Next.js Pages ISR stores the generated `RenderResult` from the response cache path rather than constructing a second React tree for the cache fill. Vinext can do better for Workers by preserving streaming to the client while recording the same streamed body for persistence.

## Approach

- Keep the page render before shell/head collection so `next/head`, styled-jsx, and server-inserted HTML state are populated before shell assembly.
- For cacheable nonce-free Pages ISR responses, tee the body stream after shell assembly.
- Send branch A through the existing shell/body/shell composite response stream.
- Drain branch B with streaming UTF-8 decoding and write `shellPrefix + body + shellSuffix` to ISR cache.
- Report cache recording or cache write failures through the Pages Router render error path and avoid writing partial HTML.
- Remove the now-unused `renderIsrPassToStringAsync` dependency from the Pages response helper call.

## Validation

- `vp check packages/vinext/src/server/pages-page-response.ts packages/vinext/src/entries/pages-server-entry.ts tests/pages-page-response.test.ts tests/pages-router.test.ts`
- `vp test run tests/pages-page-response.test.ts tests/pages-page-data.test.ts`
- `vp test run tests/pages-router.test.ts -t "caches the streamed ISR render|Pages Router production fixture"`
- `vp test run tests/features.test.ts -t "ISR \\(Pages Router\\)"`
- `vp run vinext#build`

The package build exits 0 and still emits the existing unresolved virtual-module warnings for virtual RSC/instrumentation imports.

## Risks / follow-ups

This follows the existing App Router pattern of scheduling cache writes through `ExecutionContext.waitUntil()` when available. An immediate second request after a MISS can only hit once the cache backend has completed the write, which matches the asynchronous cache-write direction already used for App Router HTML caching.

### DIFF
--- a/packages/vinext/src/server/pages-page-response.ts
+++ b/packages/vinext/src/server/pages-page-response.ts
@@ -1,7 +1,10 @@
 import React, { type ComponentType, type ReactNode } from "react";
+import type { CachedPagesValue } from "vinext/shims/cache";
 import { withScriptNonce } from "vinext/shims/script-nonce-context";
+import { getRequestExecutionContext } from "vinext/shims/request-context";
 import { buildRevalidateCacheControl } from "./cache-control.js";
 import { createInlineScriptTag, createNonceAttribute, escapeHtmlAttr } from "./html.js";
+import { reportRequestError } from "./instrumentation.js";
 
 type PagesFontPreload = {
   href: string;
@@ -42,13 +45,7 @@ type RenderPagesPageResponseOptions = {
   isrRevalidateSeconds: number | null;
   isrSet: (
     key: string,
-    data: {
-      kind: "PAGES";
-      html: string;
-      pageData: Record<string, unknown>;
-      headers: undefined;
-      status: undefined;
-    },
+    data: CachedPagesValue,
     revalidateSeconds: number,
     tags?: string[],
     expireSeconds?: number,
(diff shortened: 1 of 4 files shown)


packages/vinext/src/server/pages-page-response.ts
```diff
@@ -197,6 +193,81 @@ async function buildPagesCompositeStream(
   });
 }
 
+async function recordStreamToString(stream: ReadableStream<Uint8Array>): Promise<string> {
+  const reader = stream.getReader();
+  const decoder = new TextDecoder();
+  const chunks: string[] = [];
+
+  try {
+    for (;;) {
```
james-elicx, line 202: alternatively, while(true)

PR #1092, opened 2026-05-06
fix(cache): scope use cache keys by deployment id

### LINKED ISSUE #1064: Include deploymentId in "use cache" cache key
## Upstream change

Next.js commit [07f76411b07de9417d4a6b816f3137cafe1045fc](https://github.com/vercel/next.js/commit/07f76411b07de9417d4a6b816f3137cafe1045fc) changes the cache key seed used by `"use cache"` (via `cacheHandlers`) to include `deploymentId`, falling back to `buildId` when `deploymentId` is unset.

Relevant change in `packages/next/src/server/use-cache/use-cache-wrapper.ts`:

```diff
-  const buildId = workStore.buildId
+  const buildId = workStore.deploymentId || workStore.buildId
```

The PR description notes this was previously broken in dev, where `buildId = "development"` meant the same cache key was used across deployments. `deploymentId` is now threaded through the build pipeline, work store, and `AppRouteSharedContext`.

Test fixture: `test/production/app-dir/use-cache-cross-deployment/` (asserts that changing `NEXT_DEPLOYMENT_ID` or `BUILD_ID` produces different cache keys, including across the default cache handler).

vinext implements `"use cache"` and ships a KV-based `CacheHandler` for Cloudflare Workers. To match Next.js parity:

- `"use cache"` cache keys must incorporate the deployment identifier so that different deployments do not share cached results.
- On Cloudflare Workers, the natural mapping is the Workers version ID / `CF_VERSION_METADATA` (or a build-time environment variable equivalent to `NEXT_DEPLOYMENT_ID`).
- The fallback to `buildId` should be preserved, and the existing `buildId = "development"` problem should be considered for vinext dev mode as well.

## Action items

- [ ] Plumb a `deploymentId` value through vinext's `"use cache"` work store / cache key derivation, mirroring `workStore.deploymentId || workStore.buildId`.
- [ ] Read the deployment id from a documented env var (e.g., `NEXT_DEPLOYMENT_ID`) and/or the Workers version metadata when running on Cloudflare.
- [ ] Verify that the dev server does not reuse cache entries across restarts when no deployment id is configured (or document the dev-mode behavior).
- [ ] Port the cross-deployment cache key test from `test/production/app-dir/use-cache-cross-deployment/`.

## References

- Commit: https://github.com/vercel/next.js/commit/07f76411b07de9417d4a6b816f3137cafe1045fc
- PR: https://github.com/vercel/next.js/pull/93453

NathanDrake2406:
## What this changes

Shared `"use cache"` entries now include a deployment-aware seed before falling back to the existing vinext build ID. This matches the upstream cache key precedence model for vinext's supported deployment ID sources while keeping Worker deployment identity request-scoped.

## Why

Next.js changed `use-cache-wrapper` so cache keys use `workStore.deploymentId || workStore.buildId`. Without this, two deployments can share the same `"use cache"` entry when the build ID is stable or when a runtime deployment ID changes independently of the build fallback.

References:

- Next.js source change: [packages/next/src/server/use-cache/use-cache-wrapper.ts](https://github.com/vercel/next.js/blob/07f76411b07de9417d4a6b816f3137cafe1045fc/packages/next/src/server/use-cache/use-cache-wrapper.ts#L1510-L1513)
- Next.js work store plumbing: [packages/next/src/server/async-storage/work-store.ts](https://github.com/vercel/next.js/blob/07f76411b07de9417d4a6b816f3137cafe1045fc/packages/next/src/server/async-storage/work-store.ts#L63-L86)
- Next.js regression test: [test/production/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts](https://github.com/vercel/next.js/blob/07f76411b07de9417d4a6b816f3137cafe1045fc/test/production/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts)
- Next.js `deploymentId` config docs: [next.config.js deploymentId](https://nextjs.org/docs/app/api-reference/config/next-config-js/deploymentId)
- Cloudflare version metadata binding: [Cloudflare Workers docs](https://developers.cloudflare.com/workers/runtime-apis/bindings/version-metadata/)
- Tracks #1064

## Approach

- Add a deployment-aware cache key seed in `cache-runtime`, with precedence: request-scoped deployment ID, explicit startup/test fallback, build-time resolved deployment ID, then vinext build ID.
- Resolve `next.config.js deploymentId` before `NEXT_DEPLOYMENT_ID`, validate its string shape, and expose the resolved value through `process.env.__VINEXT_DEPLOYMENT_ID`.
- Read Worker deployment identity from `env.NEXT_DEPLOYMENT_ID` or `env.CF_VERSION_METADATA.id` in the App Router Worker entry.
- Use AsyncLocalStorage for the Worker deployment ID so concurrent requests cannot overwrite one another.
- Generate a `version_metadata` binding in `wrangler.jsonc` so deployed Cloudflare Workers have `CF_VERSION_METADATA.id` available by default.

## Validation

- `vp test run tests/shims.test.ts -t 'use cache'`
- `vp test run tests/deploy.test.ts -t 'Wrangler Config Generation|generateAppRouterWorkerEntry'`
- `vp test run tests/next-config.test.ts -t 'deploymentId|generateBuildId'`
- `vp check packages/vinext/src/config/next-config.ts packages/vinext/src/shims/cache-runtime.ts packages/vinext/src/server/app-router-entry.ts packages/vinext/src/deploy.ts packages/vinext/src/index.ts packages/vinext/src/global.d.ts tests/shims.test.ts tests/deploy.test.ts tests/next-config.test.ts`
- Pre-commit hook ran `vp check --fix` and `knip --no-progress`

## Risks / follow-ups

This PR scopes the App Router Worker entry and generated deploy config. Custom Worker entries that bypass `vinext/server/app-router-entry` still need to provide their own deployment context if they invoke lower-level internals directly.

★ Insight
- Next.js seeds `"use cache"` with deployment identity before build identity because cacheHandlers can outlive a single build assumption.
- Cloudflare version metadata is request-time Worker `env` state, so reading it at module registration would miss or race the real deployment value.
- The fallback chain keeps existing build ID isolation intact for non-Cloudflare and local paths while adding stronger isolation for configured and Workers deployments.

### DIFF
--- a/packages/vinext/src/deploy.ts
+++ b/packages/vinext/src/deploy.ts
@@ -412,6 +412,11 @@ export function generateWranglerConfig(info: ProjectInfo): string {
     images: {
       binding: "IMAGES",
     },
+    // Exposes the Cloudflare Worker version ID at runtime. vinext uses it as
+    // the deployment seed for shared "use cache" entries when available.
+    version_metadata: {
+      binding: "CF_VERSION_METADATA",
+    },
   };
 
   if (info.hasISR) {
@@ -457,6 +462,10 @@ import handler from "vinext/server/app-router-entry";
 ${isrImports}
 interface Env {
   ASSETS: Fetcher;${isrEnvField}
+  NEXT_DEPLOYMENT_ID?: string;
+  CF_VERSION_METADATA?: {
+    id?: string;
+  };
   IMAGES: {
     input(stream: ReadableStream): {
       transform(options: Record<string, unknown>): {
--- a/packages/vinext/src/deploy.ts
+++ b/packages/vinext/src/deploy.ts
@@ -412,11 +412,6 @@ export function generateWranglerConfig(info: ProjectInfo): string {
     images: {
       binding: "IMAGES",
     },
-    // Exposes the Cloudflare Worker version ID at runtime. vinext uses it as
-    // the deployment seed for shared "use cache" entries when available.
-    version_metadata: {
-      binding: "CF_VERSION_METADATA",
-    },
   };
 
   if (info.hasISR) {
@@ -462,10 +457,6 @@ import handler from "vinext/server/app-router-entry";
 ${isrImports}
 interface Env {
   ASSETS: Fetcher;${isrEnvField}
-  NEXT_DEPLOYMENT_ID?: string;
-  CF_VERSION_METADATA?: {
-    id?: string;
-  };
   IMAGES: {
     input(stream: ReadableStream): {
       transform(options: Record<string, unknown>): {
--- a/packages/vinext/src/server/app-router-entry.ts
+++ b/packages/vinext/src/server/app-router-entry.ts
@@ -15,7 +15,6 @@
 // @ts-expect-error — virtual module resolved by vinext
 import rscHandler from "virtual:vinext-rsc-entry";
 import { runWithExecutionContext, type ExecutionContextLike } from "vinext/shims/request-context";
-import { runWithRequestRuntimeContext } from "vinext/shims/unified-request-context";
 import { resolveStaticAssetSignal } from "./worker-utils.js";
 import {
   cloneRequestWithHeaders,
@@ -34,8 +33,34 @@ type WorkerAssetEnv = {
   };
 };
 
+type AppRouterRuntimeContext = {
+  deploymentId?: string;
+  passThroughOnException?(): void;
+  waitUntil?(promise: Promise<unknown>): void;
+};
+
 function deploymentIdFromEnv(env: WorkerAssetEnv | undefined): string | undefined {
   return env?.NEXT_DEPLOYMENT_ID || env?.CF_VERSION_METADATA?.id || undefined;
+}
+
+function runtimeContextForRequest(
+  ctx: ExecutionContextLike | undefined,
+  deploymentId: string | undefined,
+): AppRouterRuntimeContext | undefined {
+  if (!deploymentId) return ctx;
+  return ctx
+    ? {
+        deploymentId,
+        passThroughOnException: ctx.passThroughOnException?.bind(ctx),
+        waitUntil: (promise) => ctx.waitUntil(promise),
+      }
+    : { deploymentId };
+}
+
+function isExecutionContextLike(
+  ctx: AppRouterRuntimeContext | undefined,
+): ctx is ExecutionContextLike {
+  return typeof ctx?.waitUntil === "function";
 }
 
 export default {
@@ -44,16 +69,14 @@ export default {
     env?: WorkerAssetEnv,
     ctx?: ExecutionContextLike,
   ): Promise<Response> {
-    return runWithRequestRuntimeContext({ deploymentId: deploymentIdFromEnv(env) }, () =>
-      handleRequest(request, env, ctx),
-    );
+    return handleRequest(request, env, runtimeContextForRequest(ctx, deploymentIdFromEnv(env)));
   },
 };
 
 async function handleRequest(
   request: Request,
   env: WorkerAssetEnv | undefined,
-  ctx: ExecutionContextLike | undefined,
+  ctx: AppRouterRuntimeContext | undefined,
 ): Promise<Response> {
   const url = new URL(request.url);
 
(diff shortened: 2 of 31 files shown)


packages/vinext/src/shims/cache-runtime.ts
```diff
@@ -90,20 +90,64 @@ type RscModule = {
...
+ * Set the process-wide fallback deployment ID used to seed shared "use cache" keys.
+ *
+ * This is for tests and startup-time configuration. Request handlers should use
+ * runWithUseCacheDeploymentId() so concurrent requests cannot overwrite each other.
+ */
+export function setUseCacheDeploymentId(deploymentId: string | undefined): void {
+  Reflect.set(globalThis, _DEPLOYMENT_ID_KEY, deploymentId || undefined);
+}
+
+export function runWithUseCacheDeploymentId<T>(
```
james-elicx: instead of using ALS, we should be able to inline this value with the same pattern we are using for build ids

james-elicx: i see we are inlining it - in that case, can we rip out the ALS?

NathanDrake2406:
@codex review

packages/vinext/src/server/app-router-entry.ts
```diff
@@ -27,69 +27,114 @@ type WorkerAssetEnv = {
...
+};
+
+type AppRouterRuntimeContext = {
+  deploymentId?: string;
+  passThroughOnException?(): void;
+  waitUntil?(promise: Promise<unknown>): void;
 };
 
+function deploymentIdFromEnv(env: WorkerAssetEnv | undefined): string | undefined {
+  return env?.NEXT_DEPLOYMENT_ID || env?.CF_VERSION_METADATA?.id || undefined;
```
james-elicx: this is a binding and not guaranteed to follow that var name

packages/vinext/src/server/app-router-entry.ts
```diff
@@ -27,69 +27,114 @@ type WorkerAssetEnv = {
...
   };
+  NEXT_DEPLOYMENT_ID?: string;
+  CF_VERSION_METADATA?: {
+    id?: string;
+  };
+};
+
+type AppRouterRuntimeContext = {
+  deploymentId?: string;
+  passThroughOnException?(): void;
```
james-elicx: why does deployment id need to live on the request context?

packages/vinext/src/shims/unified-request-context.ts
james-elicx, line 1: for simplicities sake, all we need for deployment ids at the moment if a function that return the value of `process.env.__VINEXT_DEPLOYMENT_ID` which is inlined at build time, and falls back to `process.env. NEXT_DEPLOYMENT_ID`, and then probably falls back to build id potentially.

deploymentid doesnt need to touch ALS, unless there's a reason I'm not seeing?

in the future when we support experimental.runtimeServerDeploymentId as well, i believe we would always default to the NEXT_DEPLOYMENT_ID var and error if there's a deploymentId in the next config.

packages/vinext/src/deploy.ts
```diff
@@ -412,6 +412,11 @@ export function generateWranglerConfig(info: ProjectInfo): string {
     images: {
       binding: "IMAGES",
     },
+    // Exposes the Cloudflare Worker version ID at runtime. vinext uses it as
+    // the deployment seed for shared "use cache" entries when available.
+    version_metadata: {
+      binding: "CF_VERSION_METADATA",
```
james-elicx: please can we exclude these changes for now, there's probably an open question around what's the right way to consume this at runtime, so would prefer to leave it out for now.

james-elicx (changes requested):

NathanDrake2406:
Updated to match this direction:

- removed the `CF_VERSION_METADATA` / `version_metadata` changes for now
- removed deployment ID from the app router runtime/request context
- removed the request-context cache seed path
- `cache-runtime` now uses the simple seed chain: `process.env.__VINEXT_DEPLOYMENT_ID || process.env.NEXT_DEPLOYMENT_ID || process.env.__VINEXT_BUILD_ID`

Focused checks passed locally: `vp check ...`, `vp test run tests/shims.test.ts -t "use cache"`, `vp test run tests/deploy.test.ts`, and `vp test run tests/next-config.test.ts -t "deploymentId"`.

james-elicx (approved):
thanks

Issue #563, opened 2026-03-16
Pre-render pipeline performance / configurability improvements


james-elicx:
There are a few opportunities that have already been identified to improve our pre-render pipeline:
- [ ] Configurable concurrency via CLI flag.
- [ ] Deduping RSC requests by extract RSC payloads from HTML responses.
- [ ] Add informational logs when a server is started for pre-rendering.
- [x] Replace Wrangler with Vite for builds.
- [ ] Improving the performance for `resolveParentParams` lookups.

PR #1062, opened 2026-05-05
feat(app-router): add artifact compatibility metadata

NathanDrake2406:
## What this changes
Implements `#726-COMPAT-01` plus `#726-COMPAT-02/03` from #726.

This PR adds the App Router artifact compatibility envelope, attaches concrete render-boundary metadata to outgoing record payloads, and defines a conservative compatibility evaluator for future cache/skip callers.

## Why
Issue #726 requires every reusable or skippable artifact to carry compatibility proof before cache reuse or skip transport can depend on it. The rule for this slice is deliberately strict: no proof, no reuse. Unknown compatibility is represented as a render-fresh fallback, never as positive compatibility proof.

## Approach
- Add `packages/vinext/src/server/artifact-compatibility.ts` with schema constants, `ArtifactCompatibilityEnvelope`, constructor/parser helpers, a route/root graph fingerprint helper, and `evaluateArtifactCompatibility()`.
- Keep legacy/malformed wire parsing fail-closed: missing legacy metadata defaults to an unknown-proof envelope; malformed or future-schema metadata is rejected.
- Attach compatibility metadata in `renderAppPageLifecycle` when the outgoing App Router payload is a record:
  - graph metadata from route pattern plus root boundary fingerprint
  - deployment metadata from `process.env.__VINEXT_BUILD_ID`
  - schema metadata from the envelope constants
  - `rootBoundaryId` from `__rootLayout` when available
  - `renderEpoch: null` until a real epoch owner exists
- Preserve current semantics: no cache reuse, no skip transport, and no rolling-deploy protocol completeness in this PR.

## Correctness oracle
Vinext internal invariant from #726: no reusable or skippable artifact should gain implicit compatibility proof from cache presence, artifact presence, or wire shape. Unknown graph/deployment/root/epoch compatibility must return a render-fresh fallback, and only fully known matching metadata can produce `compatible`.

A focused Next.js source/test search for `artifact compatibility`, `compatibility envelope`, `renderEpoch`, `deploymentVersion`, and `graphVersion` found no public Next.js behavior to port for this internal Vinext protocol. This PR does not claim public Next.js behavior parity for rolling deploys.

## Validation
- `vp test run tests/app-elements.test.ts tests/app-page-render.test.ts`
- `vp check packages/vinext/src/server/artifact-compatibility.ts packages/vinext/src/server/app-elements.ts packages/vinext/src/server/app-page-render.ts tests/app-elements.test.ts tests/app-page-render.test.ts`
- `vp run vinext#build`

[...]

PR #1097, opened 2026-05-06
perf(prerender): reuse embedded RSC payload

NathanDrake2406:
## What this changes
App Router prerender now reconstructs the canonical `.rsc` payload from the streamed RSC chunk scripts already embedded in the prerendered HTML response. When extraction succeeds, prerender avoids the previous second `RSC: 1` handler invocation.

## Why
The prerender pipeline already renders HTML first, and that HTML includes streamed Flight chunks for hydration. Reusing those chunks avoids duplicate App Router work during static output generation.

## Approach
- Parse Vinext's modern streamed RSC chunk protocol: `self.__VINEXT_RSC_CHUNKS__...push(<safeJsonStringify(chunk)>)` plus `self.__VINEXT_RSC_DONE__=true`.
- Preserve the HTML-safe JSON encoding semantics used by `createRscEmbedTransform`.
- Fall back to a second `RSC: 1` request only when HTML has no RSC chunk scripts at all, which covers middleware/custom HTML responses that bypass the App Router HTML stream.
- Treat partial or malformed embeds as errors instead of silently falling back.
- Fail the route when the fallback RSC request returns a non-OK response so prerender never writes an empty or corrupt `.rsc` artifact.

## Validation
- `vp test run tests/prerender.test.ts -t "RSC extraction|middleware short-circuit fallback RSC request fails|falls back to a second RSC"`
- `vp check tests/prerender.test.ts packages/vinext/src/build/prerender.ts`
- Commit hook `vp check --fix` plus `knip --no-progress`

## Notes
- This extractor intentionally targets the current modern chunk/done protocol; legacy browser bootstrap support is not part of the prerender extractor.
- `tests/app-elements.test.ts` was updated to include `artifactCompatibility` after merging current `main` so CI matches the current AppElements metadata contract.
- Full test suite was not run locally.

Refs #563

### DIFF
--- a/packages/vinext/src/build/prerender.ts
+++ b/packages/vinext/src/build/prerender.ts
@@ -293,6 +293,8 @@ export function extractRscPayloadFromPrerenderedHtml(html: string): ExtractRscPa
       continue;
     }
 
+    // Legacy embeds are parsed once for compatibility; modern chunk scripts take
+    // priority below if both protocols are present in the same HTML.
     if (legacyPayload === null) {
       legacyPayload = extractLegacyRscPayload(script);
       if (legacyPayload !== null) {
@@ -335,7 +337,9 @@ export function extractRscPayloadFromPrerenderedHtml(html: string): ExtractRscPa
   }
 
   if (sawRscMarker) {
-    throw new Error("[vinext] Malformed prerender RSC embed: no RSC chunks were found");
+    throw new Error(
+      "[vinext] Malformed prerender RSC embed: RSC protocol markers present but no data chunks found",
+    );
   }
 
   return {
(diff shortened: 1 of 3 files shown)


NathanDrake2406:
Bonk: the test change in `tests/app-elements.test.ts` is intentional for CI. After rebasing/merging current `main`, AppElements metadata now includes the `artifactCompatibility` envelope from #1062, so the legacy-compatible payload round-trip expectation needs to include that field. This is not changing runtime behavior; it just updates the stale test fixture to match the current metadata contract.

NathanDrake2406:
@codex review

PR #1100, opened 2026-05-06
fix(router): reject stale same-url server action commits

NathanDrake2406:
## What this changes

Implements #726-CORE-04 from #726 by rejecting stale same-URL server-action visible commits. A server action that started from an older visible router state can still return its action value, but it can no longer overwrite newer visible state once another commit has advanced the lifecycle version.

## Why

Issue #726 calls out that `activeNavigationId` is not enough for same-URL refresh and server-action races. Same URL does not mean same visible world. Before this change, two same-URL server actions resolving out of order could let the older RSC payload commit after the newer one had already updated the route.

That violated the lifecycle spine invariant from #726: no stale operation commits visible state, and no older server-action result overwrites newer visible state.

## Approach

- Extend the pending commit disposition gate to compare `startedVisibleCommitVersion` with the current router state `visibleCommitVersion`.
- Treat visible-version mismatch as stale before root-boundary hard-navigation decisions, so old action payloads do not trigger stale reload recovery.
- Read the current browser router state at approval time after async RSC payload resolution, instead of approving against only the state captured before the await.
- Preserve the existing same-URL server action return-value contract when the visible UI commit is rejected.

Non-goals: this does not model discarded revalidation, redirect lifecycle, cache reuse authority, or the broader NavigationPlanner work from issue #726.

## Validation

- `vp check packages/vinext/src/server/app-browser-navigation-controller.ts packages/vinext/src/server/app-browser-state.ts packages/vinext/src/server/app-browser-visible-commit.ts tests/app-browser-entry.test.ts`
- `vp run vinext#build`

`vp run vinext#build` completed with the existing virtual-module external warnings for `virtual:vinext-rsc-entry`, `virtual:vite-rsc/client-references`, and `private-next-instrumentation-client`.

## Review note

Bonk: please read issue #726 to see the big picture before reviewing this PR. This is specifically the #726-CORE-04 slice, not the full lifecycle, planner, cache, or compatibility architecture.

### DIFF
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -291,6 +291,7 @@ async function renderNavigationPayload(
 async function commitSameUrlNavigatePayload(
   nextElements: Promise<AppElements>,
   returnValue?: ServerActionResult["returnValue"],
+  actionInitiationState?: AppRouterState,
 ): Promise<unknown> {
   const navigationSnapshot = createClientNavigationRenderSnapshot(
     window.location.href,
@@ -300,6 +301,7 @@ async function commitSameUrlNavigatePayload(
     nextElements,
     navigationSnapshot,
     returnValue,
+    actionInitiationState,
   );
 }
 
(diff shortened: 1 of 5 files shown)


NathanDrake2406:
@codex review

PR #1102, opened 2026-05-06
fix(router): gate refresh and traverse outcomes

NathanDrake2406:
## What changed

Implements #726-CORE-05/06A and #726-CORE-06B from issue #726.

- Treat refresh, prefetch, and traverse completion as lifecycle outcomes instead of assuming every resolved payload is still allowed to affect the visible router tree.
- Gate pending refresh, traverse, navigation, and same-URL navigation commits by both active navigation id and the visible commit version captured when the operation started.
- Return explicit browser navigation payload outcomes: committed, no-commit, and hard-navigate. Stale and hard-navigation payloads now settle pending router state without dispatching visible UI updates.
- Only seed the visited-response cache after a payload actually commits, so late refresh or traverse results cannot masquerade as rendered responses.
- Make router.prefetch expose cache-seeded only after the RSC snapshot is ready and pending work has been cleared, and consume entries only in that settled state.
- Update the next/navigation ambient shim declarations to match the runtime prefetch cache shape.

## Why

Issue #726 calls out a class of stale async navigation results where an older operation can resolve after a newer visible commit. The previous gate only checked the active navigation id. That was not enough for operations such as refresh and traverse, because a newer visible commit can occur without changing the same operation token. The existing pending operation shape already captured startedVisibleCommitVersion, so this patch uses that value as the lifecycle boundary.

For prefetch, the required outcome is cache seed only. The cache entry now stays pending until the response snapshot is actually consumable, and prefetch still never calls the visible navigation entry point.

## Validation

- vp test run tests/app-browser-entry.test.ts tests/prefetch-cache.test.ts
- vp test run tests/shims.test.ts -t "useRouter|prefetch|refresh"
- vp check packages/vinext/src/server/app-browser-state.ts packages/vinext/src/server/app-browser-visible-commit.ts packages/vinext/src/server/app-browser-navigation-controller.ts packages/vinext/src/server/app-browser-entry.ts packages/vinext/src/shims/navigation.ts packages/vinext/src/shims/next-shims.d.ts tests/app-browser-entry.test.ts tests/prefetch-cache.test.ts
- git diff --check

## Non-goals

- Full refresh lane model or cache reuse authority.
- Full back/forward intent model or Activity preservation.

Bonk: please read issue #726 for the big picture before reviewing this patch. The lifecycle-gate work here is one slice of that larger router determinism plan.

### DIFF
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -43,6 +43,7 @@ import {
 import {
   createAppBrowserNavigationController,
   type HistoryUpdateMode,
+  type NavigationPayloadOutcome,
   type PendingBrowserRouterState,
 } from "./app-browser-navigation-controller.js";
 import {
@@ -263,7 +264,7 @@ async function renderNavigationPayload(
   useTransition = true,
   actionType: "navigate" | "replace" | "traverse" = "navigate",
   operationLane: OperationLane = "navigation",
-): Promise<void> {
+): Promise<NavigationPayloadOutcome> {
   try {
     return await browserNavigationController.renderNavigationPayload({
       actionType,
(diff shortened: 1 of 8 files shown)


NathanDrake2406:
@codex review

PR #1105, opened 2026-05-06
fix(config): handle ssr.external: true in App Router (closes #1103)

### LINKED ISSUE #1103: Dev: SSR React dispatcher null in 'use client' components when ssr.external: true
## Summary

In dev mode, App Router pages that contain a `'use client'` component which calls `React.useContext` from a third-party package (e.g. `nuqs`, `next-themes`, `launchdarkly-react-client-sdk`) crash with:

```
TypeError: Cannot read properties of null (reading 'useContext')
```

This happens when the user sets `ssr.external: true` in `vite.config.ts`. `pnpm/npm run build` is unaffected (Rollup bundles everything in one pass).

This is the same family as #848 and partially mitigated by #1066, but #1066 only handles vinext's *default* `ssr.external` array. When the user explicitly sets `ssr.external: true` (which is currently the only practical way to keep CJS-quirky packages like `pg`, `protobufjs`, `keyv`, `@launchdarkly/js-server-sdk-common`, `@opentelemetry/otlp-transformer` working in dev), the bug persists.

## Environment

- vinext `0.0.47` (with #1066 patched locally)
- `vite@8.0.9`, `@vitejs/plugin-rsc@0.5.25`, `@vitejs/plugin-react@6.0.1`
- `react@19.2.5`, `react-dom@19.2.5`, `react-server-dom-webpack@19.2.5`
- Single React install (verified via the diagnostic from #848)

## Minimal repro

```tsx
// app/repro-nuqs/page.tsx
import { SearchForm } from './SearchForm';

export default function ReproPage() {
  return (
    <div>
      <h1>nuqs repro</h1>
      <SearchForm />
    </div>
  );
}
```

```tsx
// app/repro-nuqs/SearchForm.tsx
'use client';

import { useQueryState, parseAsInteger } from 'nuqs';

export function SearchForm() {
  const [query, setQuery] = useQueryState('q', { defaultValue: '' });
  const [page, setPage] = useQueryState('page', parseAsInteger.withDefault(1));
  return <div>{query} {page}</div>;
}
```

```tsx
// app/layout.tsx — wrap children in <NuqsAdapter>
```

```ts
// vite.config.ts
import vinext from 'vinext';
import { defineConfig } from 'vite';

export default defineConfig({
  plugins: [vinext()],
  ssr: { external: true },
});
```

### Steps

1. `pnpm dev`
2. Visit `/repro-nuqs`. The page returns 200 with the error boundary rendered:

```text
TypeError: Cannot read properties of null (reading 'useContext')
    at SearchForm (about://React/Server/.../SearchForm.tsx?22:11:67)
    at resolveErrorDev (.../react-server-dom-webpack_client__browser.js)
```

This mirrors `tests/fixtures/ecosystem/nuqs/` which works in your test suite. The difference: that fixture does not set `ssr.external: true`.

## Status matrix

| Route shape | Result |
| --- | --- |
| `page.tsx` is a server component (no `'use client'`) | ✅ Works |
| `page.tsx` is a server component that renders a `'use client'` child using `nuqs.useQueryState` | ❌ `useContext null` |
| `page.tsx` itself has `'use client'` and uses any nuqs-style hook | ❌ `useContext null` |

## Root cause analysis

Reading `packages/vinext/src/index.ts` around line 1345 (App Router branch, SSR env):

1. With `userSsrExternal === true`, vinext sets `environments.ssr.resolve.external: true` and **does not** set `noExternal`.
2. `@vitejs/plugin-rsc` (auto-registered by vinext) populates `environments.ssr.optimizeDeps.include` with React + react-dom + the `server.edge` entries via `crawlFrameworkPkgs`.
3. Vite pre-bundles those into `node_modules/.vite/deps_ssr/react.js`. With `external: true`, this still happens because `optimizeDeps.include` overrides `external: true`.
4. SSR rendering paths now have **two** physical React copies:
   - vinext's runtime (`app-ssr-entry.ts`) is loaded by Node and gets `/node_modules/react/index.js`.
   - `'use client'` modules that go through Vite's transform resolve React via the optimizer to `/node_modules/.vite/deps_ssr/react.js`.
5. `react-dom-server.edge` (also in `deps_ssr/`) sets the dispatcher on its bundled React. Externalized callers (vinext's runtime) see a *different* React → `H` is null → any hook call from a `'use client'` module crashes.

Build path is unaffected because Rollup hoists everything into a single bundle and there's only one React copy by construction.

## Approaches that don't work

| Approach | Outcome |
| --- | --- |
| Default vinext (no user `ssr.external`), let `noExternal: true` mode bundle everything | Dev hits a long tail of CJS issues: `@protobufjs/inquire` (CJS dynamic require), `pg-native` virtual-module-resolution, `@opentelemetry/otlp-transformer` "default of undefined", `@launchdarkly/js-server-sdk-common/node_modules/semver` "Cannot read properties of undefined (reading '__cjs_module_runner_transform')". Each one breaks the SSR module graph. |
| User sets `ssr.external: true` (current workaround) | Build OK. Dev: server-component routes load. `'use client'` routes still crash with `useContext null`. |
| Apply #1066 locally to `node_modules/vinext/dist/index.js` | Same outcome — server-component routes work, `'use client'` paths still fail because plugin-rsc still pre-bundles a duplicate React for the SSR env. |
| Refactor `'use client'` page.tsx to server-page-shell + client-child pattern | Same `useContext null` error in the client child even with the canonical pattern. |
| Force-bundle React via user `ssr.noExternal: ['react', 'react-dom', 'nuqs', 'next-themes', 'use-sync-external-store']` | Optimizer hangs for 3+ minutes during initial bundling. |

## Proposed fix

When `userSsrExternal === true`, also strip React entries from `environments.ssr.optimizeDeps.include` (or set `optimizeDeps.exclude` for those entries) so plugin-rsc's `crawlFrameworkPkgs` cannot pre-bundle a duplicate React copy for the SSR env.

Sketch (in the App Router branch of the SSR env config):

```ts
ssr: {
  ...,
  optimizeDeps: {
    exclude: mergeOptimizeDepsExclude(
      incomingExclude,
      VINEXT_OPTIMIZE_DEPS_EXCLUDE,
      // When SSR is fully externalized, prevent plugin-rsc's crawlFrameworkPkgs
      // from pre-bundling React into deps_ssr. Externalized callers (vinext's
      // runtime) load React via Node's CJS resolution; if the optimizer also
      // bundles a copy, they're two distinct module records and the dispatcher
      // mismatch leaves React.H null in 'use client' components.
      ...(userSsrExternal === true
        ? [
            'react',
            'react-dom',
            'react-dom/server.edge',
            'react-dom/static.edge',
            'react-server-dom-webpack/client.edge',
            'react/jsx-runtime',
            'react/jsx-dev-runtime',
          ]
        : []),
    ),
    entries: optimizeEntries,
  },
},
```

The alternative (also from #848) — force-resolving vinext's runtime modules to use the project-root React — is more invasive and would change behavior for cases that are already working.

I have a working version of this fix locally and am happy to open a PR (branch `issue-NNN`).

## Why this matters

For non-trivial apps, `ssr.external: true` is currently the only practical way to make dev work, because the alternative (`noExternal: true`, the default) hits the long tail of CJS-quirky packages listed above. Without this fix, every authenticated route in any real-world app that uses nuqs / next-themes / launchdarkly-react / etc. is broken in dev mode.

jgeurts:
Closes #1103.

## What

When the user sets `ssr.external: true` in their `vite.config.ts`, ensure no React copy gets bundled into the SSR env. Two changes:

1. Strip React entries from `environments.ssr.optimizeDeps.exclude` so `@vitejs/plugin-rsc`'s `crawlFrameworkPkgs` cannot pre-bundle a duplicate React into `deps_ssr/`.
2. When `config.ssr.external === true`, vinext only sets `ssr: { external: true }` at the top level — no `noExternal: true` to leak into the per-env config (Vite uses top-level `ssr.*` as defaults for `environments.ssr.*`). Plus a defensive `configResolved` hook that strips React from `environments.ssr.resolve.noExternal` if anything re-added it.

## Why

With `ssr.external: true`, the SSR env loads everything via Node's resolver — including React from `/node_modules/react`. But `optimizeDeps.include` overrides `external: true`, and `noExternal: ['react', ...]` defeats `external` for those entries. The result is two distinct React module records:

- vinext's runtime (loaded via Node externals) → `/node_modules/react/index.js`
- 'use client' modules SSR'd through the bundled-React env → Vite-resolved React

`react-dom-server.edge` sets the dispatcher on its bundled React; externalized callers see a different React → `React.H` is null → every hook call from a `'use client'` module crashes with `Cannot read properties of null (reading 'useContext')` / `Invalid hook call`.

The default mode (`noExternal: true`) is unaffected because everything bundled in one pass = one React. Build is unaffected for the same reason.

## How

```ts
// SSR env's optimizeDeps.exclude — strip React when external is fully external:
optimizeDeps: {
  exclude: mergeOptimizeDepsExclude(
    incomingExclude,
    VINEXT_OPTIMIZE_DEPS_EXCLUDE,
    userSsrExternal === true ? SSR_EXTERNAL_REACT_ENTRIES : [],
  ),
  entries: optimizeEntries,
},

// Top-level ssr block — gate on user's external: true so noExternal: true
// doesn't leak into environments.ssr.resolve.noExternal as a default:
...(hasCloudflarePlugin || hasNitroPlugin
  ? {}
  : config.ssr?.external === true
    ? { ssr: { external: true as const } }
    : { ssr: { external: ['react', 'react-dom', 'react-dom/server'], noExternal: true } }),

// configResolved — defensive strip if anything re-added React:
if (hasAppDir) {
  const ssrEnv = config.environments?.ssr;
  if (ssrEnv?.resolve?.external === true && Array.isArray(ssrEnv.resolve.noExternal)) {
    ssrEnv.resolve.noExternal = ssrEnv.resolve.noExternal.filter(
      (entry) => typeof entry !== 'string' || !SSR_EXTERNAL_REACT_ENTRIES.includes(entry),
    );
  }
}
```

The RSC env is left alone (it still pre-bundles React with the `react-server` condition).

## Tests

Three new cases in `tests/build-optimization.test.ts`:

1. `excludes React from ssr optimizeDeps when ssr.external: true (App Router)` — verifies the strip applies, that RSC env still pre-bundles React, and that top-level `ssr.noExternal` is undefined.
2. `does NOT exclude React from ssr optimizeDeps when ssr.external is unset` — default mode unchanged.
3. `does NOT exclude React from ssr optimizeDeps when ssr.external is a string array` — only the blanket `true` form triggers the strip.

`pnpm test --project unit` ⇒ 131/131 files, 3,736 tests pass.
`pnpm run check` ⇒ clean (1 unrelated pre-existing warning).

## Manual verification

Applied locally to a real-world App Router app (~80 deps including nuqs, next-themes, launchdarkly-react, blocknote, opentelemetry, pg, protobufjs, keyv, etc.) configured with `ssr.external: true`. Before the fix: every `'use client'` route crashed with `useContext null` in dev. After the fix: routes render correctly, hooks fire as expected.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -1193,16 +1193,25 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
           //   Any user-provided `ssr.noExternal` is intentionally superseded
           //   by this setting; only `ssr.external` entries escape Vite's transform.
           // Skip when targeting bundled runtimes (Cloudflare/Nitro bundle everything).
+          // Also skip `noExternal: true` when the user opted into
+          // `ssr.external: true` — they've explicitly asked for everything
+          // external, and forcing `noExternal: true` here leaks down into
+          // `environments.ssr.resolve.noExternal` (Vite uses top-level
+          // `ssr.*` as the default for the per-env resolve config), which
+          // makes Vite bundle React despite the user's intent and produces
+          // the duplicate-React crashes documented in #1103.
           // This also resolves extensionless-import issues in packages like
           // `validator` (see #189) by routing them through Vite's resolver.
           ...(hasCloudflarePlugin || hasNitroPlugin
             ? {}
-            : {
-                ssr: {
-                  external: ["react", "react-dom", "react-dom/server"],
-                  noExternal: true,
-                },
-              }),
+            : config.ssr?.external === true
+              ? { ssr: { external: true as const } }
+              : {
+                  ssr: {
+                    external: ["react", "react-dom", "react-dom/server"],
+                    noExternal: true,
+                  },
+                }),
           resolve: {
             // Materialize simple tsconfig/jsconfig path aliases into resolve.alias
             // so Vite can transform import.meta.glob("@/...") and import(`@/...`).
@@ -1519,6 +1528,37 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
       },
 
       configResolved(config) {
+        // When the user sets `ssr.external: true`, strip React entries from
+        // `environments.ssr.resolve.noExternal`. @vitejs/plugin-rsc populates
+        // this list via crawlFrameworkPkgs, but `noExternal` overrides
+        // `external: true` for the listed packages. The result is that React
+        // gets bundled by Vite's transform pipeline despite the user opting
+        // for full externalization, producing a second React module record
+        // alongside the Node-loaded one used by externalized callers (vinext's
+        // runtime). 'use client' modules SSR'd through the bundled-React env
+        // then crash with `Invalid hook call` / `useContext null`. Stripping
+        // these entries forces the SSR env to load React via Node externals,
+        // matching the renderer's React. See #1103.
+        if (hasAppDir) {
+          const ssrEnv = config.environments?.ssr;
+          if (ssrEnv?.resolve?.external === true && Array.isArray(ssrEnv.resolve.noExternal)) {
+            // Strip React entries that @vitejs/plugin-rsc auto-adds to
+            // `environments.ssr.resolve.noExternal` via crawlFrameworkPkgs.
+            // With `ssr.external: true`, the SSR env loads React via Node's
+            // resolver, but `noExternal: ["react", ...]` overrides that for
+            // the listed packages — Vite bundles React anyway, producing a
+            // second module record alongside the Node-loaded one used by
+            // externalized callers (vinext's runtime). 'use client' modules
+            // SSR'd through that env then crash with `useContext null` /
+            // `Invalid hook call`. Stripping these entries forces the SSR
+            // env to load React via Node externals so the renderer and the
+            // runtime share a single React. See #1103.
+            ssrEnv.resolve.noExternal = ssrEnv.resolve.noExternal.filter(
+              (entry) => typeof entry !== "string" || !SSR_EXTERNAL_REACT_ENTRIES.includes(entry),
+            );
+          }
+        }
+
         // Detect double React plugin registration. When vinext auto-injects
         // @vitejs/plugin-react AND the user also registers it manually, the
         // React transform / refresh pipeline runs twice.
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -79,6 +79,7 @@ import { createOptimizeImportsPlugin } from "./plugins/optimize-imports.js";
 import { createOgInlineFetchAssetsPlugin, ogAssetsPlugin } from "./plugins/og-assets.js";
 import {
   mergeOptimizeDepsExclude,
+  SSR_EXTERNAL_REACT_ENTRIES,
   VINEXT_OPTIMIZE_DEPS_EXCLUDE,
 } from "./plugins/rsc-client-shim-excludes.js";
 import { createServerExternalsManifestPlugin } from "./plugins/server-externals-manifest.js";
@@ -651,7 +652,9 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
         });
       })
       .catch((cause) => {
-        throw new Error("vinext: Failed to load @vitejs/plugin-rsc.", { cause });
+        throw new Error("vinext: Failed to load @vitejs/plugin-rsc.", {
+          cause,
+        });
       });
   }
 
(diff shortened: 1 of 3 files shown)


jgeurts:
Minimal standalone reproduction available at https://github.com/jgeurts/vinext-issue-1103.

```
git clone https://github.com/jgeurts/vinext-issue-1103
cd vinext-issue-1103
npm install
node inspect.mjs
```

`inspect.mjs` reads the resolved Vite config (no dev server needed) and reports whether the bug surface is present. With stock `vinext@0.0.47` you'll see React in `environments.ssr.optimizeDeps.include` while `resolve.external: true` — the conflicting state that produces the duplicate React copy. After dropping a fork build of `packages/vinext/dist` into `node_modules/vinext/dist`, the inspect output flips and the React entries land in `optimizeDeps.exclude`.

Repro mirrors `tests/fixtures/ecosystem/nuqs/` from this repo; the only difference is the `ssr: { external: true }` line in `vite.config.ts`.

jgeurts:
Added an Open-in-StackBlitz button to the repro: https://stackblitz.com/github/jgeurts/vinext-issue-1103 (no checkout needed; click the link, wait for install, run `node inspect.mjs`).

PR #1141, opened 2026-05-08
fix: use loader for client reference

hyoban:
# Safari RSC Client Reference Crash Report

## Scope

The Safari/WebKit crash:

```text
undefined is not an object (evaluating 'r.call(t,e[2])')
```

## Root Cause Hypothesis

The minified crash maps to React RSC client-reference export lookup, roughly:

```js
hasOwnProperty.call(moduleExports, exportName)
```

The crash means `moduleExports` is `undefined`.

The likely failure chain is:

1. vinext groups many RSC client references under the virtual RSC server entry facade.
2. This can produce a large browser chunk like:

   ```text
   facade__virtual_vinext-rsc-entry-*.js
   ```

3. Safari/WebKit evaluates the resulting dynamic import graph in an order where a client reference can be observed before its module namespace is fully initialized.
4. React RSC later tries to read an export from `undefined` and crashes.

Chrome appears to tolerate this module graph timing better, which is why the same output can work there.

## Next.js Behavior

Next.js does not organize App Router client references as one large virtual RSC entry facade.

Next.js uses this pipeline:

1. `next-flight-loader` transforms `"use client"` modules into `registerClientReference(...)` references.
2. `FlightClientEntryPlugin` walks each App Router entry, collects reachable client components, and injects synthetic client entries.
3. `next-flight-client-entry-loader` eagerly imports those client modules.
4. `ClientReferenceManifestPlugin` records each client module against real webpack entrypoint chunks.
5. Next.js emits per-page client reference manifests, merged from the relevant layout hierarchy.

This makes client-reference chunks route/app-entry driven, not virtual-RSC-facade driven.

## vinext Difference

vinext currently relies on `@vitejs/plugin-rsc` default client-reference grouping by `meta.serverChunk`.

For App Router builds, many references can share:

```text
facade:virtual:vinext-rsc-entry
```

This is the artifact shape that likely exposes the Safari crash.

## Bug-Relevant Invariant

vinext should avoid grouping many unrelated RSC client references into one browser chunk keyed by the virtual RSC server entry facade.

A safer shape should make browser client-reference chunks depend on client module identity or route/app-entry identity.

The closer Next.js parity target is route/app-entry based synthetic client entries plus per-route client reference manifests.


hyoban:
resolved

PR #1146, opened 2026-05-09
test: harden ecosystem fixture startup

### LINKED ISSUE #1145: test: ecosystem fixture startup timeout can leak dev server
## Summary

The ecosystem integration test harness can report a generic Vitest `beforeAll` timeout while leaving the fixture dev server running. This showed up on PR #1130 in `Vitest (integration 1/3)` for the `better-auth` fixture.

## Evidence

The failed shard timed out in `tests/ecosystem.test.ts` at the `better-auth` `beforeAll` hook:

```txt
Error: Hook timed out in 90000ms.
  ❯ tests/ecosystem.test.ts:333:3
```

The GitHub cleanup step then terminated an orphaned process for the same fixture:

```txt
npm exec vp dev --port 4404 --strictPort
```

Running the same target locally passed quickly:

```sh
vp test run tests/ecosystem.test.ts -t "better-auth"
# passed in 8.2s
```

## Likely root cause

`STARTUP_TIMEOUT_MS` is used for both the internal fixture readiness deadline and the Vitest hook timeout. Those deadlines race, so Vitest can kill the hook before `startFixture()` rejects with the buffered server output. Because the `proc` variable is only assigned after `startFixture()` resolves, `afterAll(() => killProcess(proc))` cannot clean up the child process when startup fails.

The readiness probe also requests `/`, which includes first SSR compilation work, so CI load can make startup readiness much more variable than simple port binding.

## Expected fix

- Ensure `startFixture()` owns cleanup on startup failure and timeout.
- Give Vitest hooks a timeout larger than the internal readiness deadline so startup failures surface the diagnostic error.
- Preserve the existing fixture behavior while making failures deterministic and diagnosable.
- Consider fixing the `better-auth` fixture default `baseURL` mismatch (`4403` default while the test starts on `4404`).

NathanDrake2406:
## Summary

Fixes #1145.

The ecosystem test harness used the same 90s deadline for both its internal fixture readiness loop and the Vitest `beforeAll` hook. When startup stalled under CI load, Vitest could kill the hook before the helper rejected with its buffered server output. Because the child process was only assigned after successful startup, the `afterAll` cleanup path could not terminate the still-running fixture server.

This PR extracts the subprocess fixture launcher into `tests/fixture-dev-server.ts` and makes startup ownership explicit:

- kill the spawned dev server when readiness fails or times out
- keep the Vitest hook timeout above the internal startup deadline so failures surface the diagnostic error
- preserve buffered stdout/stderr in startup failures
- add a focused regression test for the timeout cleanup path
- fix the `better-auth` fixture fallback `baseURL` to match its test port (`4404`)

## Verification

```sh
vp test run tests/fixture-dev-server.test.ts tests/ecosystem.test.ts
vp check
```

`vp check` passes with the existing unrelated warning in `packages/vinext/src/server/request-pipeline.ts:604`.

### DIFF
--- a/tests/ecosystem.test.ts
+++ b/tests/ecosystem.test.ts
@@ -10,130 +10,23 @@
  * Run with: npx vitest run tests/ecosystem.test.ts
  */
 import { describe, it, expect, beforeAll, afterAll } from "vite-plus/test";
-import { spawn, type ChildProcess } from "node:child_process";
+import type { ChildProcess } from "node:child_process";
 import { readFileSync } from "node:fs";
 import path from "node:path";
+import {
+  FIXTURE_HOOK_TIMEOUT_MS,
+  startFixtureDevServer,
+  stopFixtureDevServer,
+} from "./fixture-dev-server.js";
 
 const FIXTURES_DIR = path.resolve(__dirname, "fixtures", "ecosystem");
-const STARTUP_TIMEOUT_MS = process.env.CI ? 90_000 : 30_000;
-const READY_POLL_INTERVAL_MS = 250;
 
-/**
- * Start a Vite dev server as a child process and wait for it to be ready.
- */
-async function startFixture(
-  name: string,
-  port: number,
-): Promise<{
-  process: ChildProcess;
-  baseUrl: string;
-  fetchPage: (pathname: string) => Promise<{ html: string; status: number }>;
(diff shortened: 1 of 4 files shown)


NathanDrake2406:
@codex review

PR #1152, opened 2026-05-10
fix(app-router): catch redirect diges, render loading.tsx during cross-route navigation and invalid with calling refresh()

### LINKED ISSUE #1150: Redirect from Server Component under route-level loading.tsx is surfaced as NEXT_REDIRECT error during RSC navigation
I’m using vinext in an App Router project and ran into an issue where `redirect()` from a Server Component can show up as an unhandled `NEXT_REDIRECT` error in the browser instead of performing a normal redirect.

After narrowing this down with a minimal repro, the issue appears to happen when the route has a route-level `loading.tsx`.

Demo repo: https://github.com/Yoinky3000/ViNext-Redirect-Demo

## Reproduction

1. Clone the demo repo:

   ```bash
   git clone https://github.com/Yoinky3000/ViNext-Redirect
   cd ViNext-Redirect
   bun install
   bun run dev:vinext
   ```

2. Open:

   ```txt
   http://localhost:3001
   ```

3. Click `Protected without loading boundary`.

4. Go back to the home page.

5. Click `Protected with loading boundary`.

## Route Setup

The demo has two protected routes:

- `/protected`
- `/protected-2`

Both routes do the same async auth check and then call:

```ts
redirect("/?auth=required");
```

The only meaningful difference is that `/protected-2` has a route-level `loading.tsx`.

## Actual Behavior

For `/protected`, vinext handles the redirect correctly:

```txt
GET /protected?_rsc=... -> 307
```

For `/protected-2`, vinext returns a `200` RSC/HTML response and serializes the redirect as an error payload:

```txt
GET /protected-2?_rsc=... -> 200
NEXT_REDIRECT:/?auth=required
```

In the browser, this appears as a dev error overlay / console error:

```txt
Error: NEXT_REDIRECT:/?auth=required
    at Protected2Page
    at resolveErrorDev
    at processFullStringRow
    at processFullBinaryRow
    at processBinaryChunk
    at progress
```

## Expected Behavior

Calling `redirect()` from a Server Component should perform the redirect consistently, regardless of whether the route has a `loading.tsx`.

The `/protected` and `/protected-2` routes should both redirect to:

```txt
/?auth=required
```

without surfacing `NEXT_REDIRECT` as a client/dev overlay error.

## Relevant Files

- `app/protected/page.tsx`
- `app/protected-2/page.tsx`
- `app/protected-2/loading.tsx`
- `app/page.tsx`

## Notes

This seems specifically tied to the route-level loading boundary.

Without `loading.tsx`, the RSC navigation returns a proper `307`.

With `loading.tsx`, the redirect appears to be treated as an RSC render error and gets sent to the client as `NEXT_REDIRECT`.

# Comparison
I have also created a pure nextjs demo that do the same thing: https://github.com/Yoinky3000/Next-Redirect-Demo
in this demo, both path redirects properly without showing error

### LINKED ISSUE #1151: Loading.tsx not shown during client-side navigation between two routes that both have a loading boundary
## Summary

When the user clicks a `<Link>` to navigate from one route to another, vinext does not display the
destination route's `loading.tsx` while the new page resolves on the server. The previous page stays
visible the whole time. This only happens when **both** the source and destination route have a
`loading.tsx` at the same tree position. If only one of them has a loading boundary, the destination's
fallback shows correctly. Or if you directly access the page that has loading.tsx, the loading will still show correctly.

## Reproduction

Demo repos (same shape as the ones used in #1150):

- vinext: https://github.com/Yoinky3000/ViNext-Redirect-Demo
- next.js (reference): https://github.com/Yoinky3000/Next-Redirect-Demo

The relevant routes:

- `app/page.tsx` — `await new Promise((r) => setTimeout(r, 5000))` and renders the link list. Has an
`app/loading.tsx` at the root level.
- `app/test/page.tsx` — `await new Promise((r) => setTimeout(r, 5000))` and renders test content. Has its
 own `app/test/loading.tsx`.

So both `/` and `/test` are wrapped in a route-level Suspense at the same tree position.

### Steps

1. Run `bun run dev:vinext` (or `vinext dev`) in the vinext demo. Wait for `/` to fully load.
2. From `/`, click the `test loading` link to navigate to `/test`.
3. Observe the page during the 5-second await on `/test/page.tsx`.

### Expected behavior (what the Next.js demo does)

`Loading test...` (`app/test/loading.tsx`) renders immediately on click and remains visible for ~5
seconds until the page's promise resolves.

DevTools shows the `/test.rsc` request streaming chunks in over the 5 s window with a fast TTFB.

### Actual behavior in the vinext demo
`/` stays fully rendered and interactive for the entire 5 seconds. No fallback appears. After the await
resolves, the new page suddenly replaces the old one with no intermediate loading state.

Yoinky3000:
## Summary

Fixes two related App Router client-navigation issues:

- **#1150** — `redirect()` from a server component inside a route with a `loading.tsx` threw an unhandled `NEXT_REDIRECT` digest in the browser.
- **#1151** — the destination route's `loading.tsx` never rendered during a cross-route navigation when both the source and destination routes had a `loading.tsx` at the same tree position; the previous page stayed visible until the new page resolved.

Both surfaced in the same demo apps:
- vinext: https://github.com/Yoinky3000/ViNext-Redirect-Demo
- next.js (reference): https://github.com/Yoinky3000/Next-Redirect-Demo

### `packages/vinext/src/shims/error-boundary.tsx`

Adds `RedirectErrorBoundary` (and `RedirectBoundary` wrapper). `getDerivedStateFromError` decodes both vinext-style (`NEXT_REDIRECT;<type>;<encodedUrl>[;status]`) and Next.js-style (`NEXT_REDIRECT;<type>;<url>;<status>;<isClient>`) digests, then renders a `HandleRedirect` effect that dispatches `router.push()` / `router.replace()` inside a transition.

### `packages/vinext/src/server/app-page-route-wiring.tsx`

Wraps `routeChildren` (the page slot + its loading-Suspense) in `RedirectBoundary` **per route segment**, sitting below the layouts. When the boundary catches a digest, only the page subtree unmounts; surrounding layouts (and any client React state they hold) stay mounted across the redirect-driven `router.replace()`. This mirrors Next.js's per-segment `LayoutRouter` placement.

Keys the route-level Suspense by `pageId`:

```tsx
<Suspense key={pageId} fallback={<RouteLoadingComponent />}>
{routeChildren}
</Suspense>
```

Without the key, React's `startTransition` reconciliation matches the source and destination Suspense boundaries as the same one and applies "keep old UI" transition semantics, so the destination's fallback never paints. Keying by `pageId` gives cross-route navigations a fresh Suspense identity (fallback shows) while same-route renders (search-param updates, refresh) share the key (no unnecessary fallback flash).

### `packages/vinext/src/server/app-browser-entry.ts`

- Imports a top-level `RedirectBoundary` as a safety net for redirects thrown above the route segment (e.g. from a layout or the dev recovery boundary). The per-segment boundary is the primary catch site for the common server-component case.
- Replaces the buffer-then-feed pattern in `navigateRsc` with a tee'd live-stream + background-buffer pattern. One branch of `navResponse.body.tee()` is wrapped in a fresh `Response` and handed straight to `createFromFetch` so the shell parses immediately and React commits as soon as the first chunks arrive — Suspense fallbacks render while the rest of the RSC stream resolves. The second branch is read into an `ArrayBuffer` in the background and `await`-ed only after commit to populate the visited-response cache.

### `packages/vinext/src/server/dev-error-overlay.tsx`

Adds `isNavigationSignalError()` and short-circuits all four reporting paths (`window.error`, `unhandledrejection`, `onCaughtError`, `onUncaughtError`) for `NEXT_REDIRECT;`, `NEXT_NOT_FOUND`, and `NEXT_HTTP_ERROR_FALLBACK;` digests. These are caught and handled by their dedicated boundaries, so they shouldn't appear in the dev overlay.

### Tests

- `tests/error-boundary.test.ts` — digest classification for `RedirectErrorBoundary` (vinext-style 3-part digest, Next-style 5-part digest, re-throws non-redirect errors).
- `tests/app-browser-entry.test.ts` — `devOnCaughtError` / `devOnUncaughtError` ignore navigation signal digests.
- `tests/e2e/app-router/routes.spec.ts` + `tests/fixtures/app-basic/app/delayed-protected-loading/` — Playwright test
exercising server-component `redirect()` inside a loading boundary; asserts no `NEXT_REDIRECT` page error and final URL settled at the redirect target.

## Why per-segment for `RedirectBoundary`

A top-level `RedirectBoundary` inside `BrowserRoot` would also resolve #1150, but its `getDerivedStateFromError` → `render() { return null }` path unmounts every descendant, including the surrounding layouts. Any client React state held in layout components (counters, theme toggles, form drafts, modal state, etc.) would be destroyed on every redirect. Per-segment placement keeps the catch local to the page slot — same shape Next.js uses in `packages/next/src/client/components/redirect-boundary.tsx` mounted inside the per-segment `LayoutRouter`.

## Validation

- `pnpm test` — full Vitest suite passes.
- `pnpm run check` — clean.
- `PLAYWRIGHT_PROJECT=app-router npx playwright test` — passes (incl. the new redirect+loading-boundary E2E case).
- Using this changes in my demo repo (https://github.com/Yoinky3000/ViNext-Redirect-Demo) solve the problems with no other issues

End-to-end manual verification against both demo repos: the vinext demo now matches the next.js reference for the redirect-inside-loading-boundary flow, layout state (a `Counter` mounted in the root layout) survives the redirect, and `loading.tsx` paints during cross-route navigation between two routes that both have a loading boundary.

## Closes

Closes #1150
Closes #1151


Yoinky3000:
Two new commits address ask-bonk's review and roll in an additional fix that surfaced during follow-up testing with claude code.

### 1. Review fixes — `026cd98`

Implemented all six suggestions:

- `RedirectErrorBoundary.reset()` clears both `redirect` and `redirectType`.
- `getDerivedStateFromError` re-throws on malformed `NEXT_REDIRECT;` digests instead of silently swallowing — mirrors the server-side guard in
`parseNextRedirectDigest` (`next-error-digest.ts:51`).
- Documented the `error.handled` branch as a Next.js parity placeholder.
- Updated the stale `snapshotRscResponse` reference in the cached-route comment to describe the current `body.tee() + createFromFetch` flow.
- Extracted `getErrorDigest` / `isNavigationSignalError` into a shared `utils/navigation-signal.ts`; both `error-boundary.tsx` and `dev-error-overlay.tsx` import it now (no more drift between the two copies).
- Strengthened the delayed-redirect E2E test to also assert the `loading.tsx` fallback paints during the server delay.

Also fixes the formatting issue on `tests/fixtures/app-basic/app/page.tsx` that was failing the `Check` CI job.

### 2. Additional fix: `router.refresh()` didn't invalidate the client RSC cache — `29b35db`

After a state change that requires re-evaluating the server (session change, role change, data mutation), a `router.refresh()` followed by a `<Link>` click to a previously visited route still rendered the **cached** RSC payload — `visitedResponseCache` happily satisfied the navigation locally and the server's route logic never ran.

**Root cause**: `router.refresh()` only re-fetched the current URL. Sibling routes' cached payloads (visited-response + prefetch caches) were never invalidated and stayed live until their TTL expired.

**Fix**: `router.refresh()` now invalidates the entire client RSC cache before re-fetching, matching Next.js's `refresh-reducer.ts`:

> *"During a refresh, we invalidate the segment cache but not the route cache. The segment cache contains the actual RSC data which needs to be re-fetched."*

Next does this via an O(1) version counter; vinext's bounded LRU is small enough that `Map.clear()` is equivalent in observable behavior. Implementation:
`app-browser-entry.ts` exposes `clearClientNavigationCaches` on `window`; `shims/navigation.ts` `refresh()` calls it synchronously before scheduling
`rscNavigate` inside `React.startTransition` (the synchronous ordering matters — clearing inside the transition would let prefetches kicked off during render repopulate the just-cleared cache).

PR #1157, opened 2026-05-11
feat(dev): detect 'use cache' module-scope deadlocks early in dev (#1126)

### LINKED ISSUE #1126: Detect 'use cache' module-scope deadlocks early in dev

Divkix:
## Summary

Implements a dev-only probe that detects when a `'use cache'` fill is stuck on module-scope shared state (for example, a top-level `Map` used to dedupe fetches), following the durable behavior from vercel/next.js#93500 and #93538.

## How it works

- **Early positive-signal probe:** after a fill has made no progress for 10 seconds, vinext reruns the same cached function in an isolated Vite `ModuleRunner`. A successful isolated completion surfaces `UseCacheDeadlockError`; probe import, decode, execution, or timeout failures are inconclusive and fall back to the normal cache timeout.
- **Isolated module scope:** each probe gets a fresh `EvaluatedModules` graph so module-level state is recreated. Active runners are closed on HMR invalidation and dev-server shutdown.
- **Request fidelity:** forwards headers/cookies, URL pathname and query, draft-mode secret, and root params. Probe arguments use RSC `encodeReply`/`decodeReply`, preserve exact cache variants, and retain private cached page `searchParams` while public cached pages keep Next-compatible omission behavior.
- **Request-scoped recursion isolation:** `UnifiedRequestContext._probeDepth` prevents probes from recursively scheduling probes without allowing concurrent requests to interfere.
- **Configurable timeout:** supports `experimental.useCacheTimeout`, defaulting to Next.js's 54 seconds.
- **HMR lifecycle:** existing App Router source edits, route additions/removals, environment replacement, and server shutdown reset the probe hook/runners.
- **Development only:** the probe path is gated to development and does not change production cache semantics.

## Scope note

Next.js currently schedules probes around cache-fill stream idleness and can discard a result if the main stream recovers mid-probe. Vinext does not currently expose an equivalent cache-fill stream seam, so this PR intentionally uses a flat 10-second function-promise probe and does **not** claim mid-probe stream recovery. Adding a general cache stream scheduler is outside this PR's scope.

## Validation

- `vp test run tests/shims.test.ts tests/next-config.test.ts` — 1,295 tests passed
- Scoped `vp check` on the changed probe/cache/config files
- Independent correctness and Next.js parity review loops on the cumulative diff

Closes #1126


Divkix:
@james-elicx does this pr look or are any changes needed?

james-elicx:
> _AI-generated review (OpenCode + Claude). Posted by a human but not human-authored — treat as a starting point, not a verdict. Verify against the diff and the upstream Next.js implementation before acting._

Comparing this PR against the Next.js upstream (`packages/next/src/server/use-cache/use-cache-wrapper.ts` and `packages/next/src/server/dev/use-cache-probe-worker.ts`), two issues stand out as legitimate behavioral defects worth blocking on. The rest of what I initially flagged either matches upstream or is design commentary.

### 1. Probe argument encoding diverges from Next.js (false-positive deadlocks)

In `cache-runtime.ts` the probe is invoked with:

\`\`\`ts
encodedArguments: (() => { try { return JSON.stringify(args); } catch { return \"[]\"; } })()
\`\`\`

…and the probe pool decodes with \`JSON.parse\`. Next.js encodes args via RSC \`encodeReply\` (producing \`string | FormData\`) and the worker decodes with \`decodeReply\` against the server module map (see \`use-cache-probe-worker.ts:117-144\`, \`use-cache-wrapper.ts:1842\`).

This matters because App Router \`'use cache'\` functions are typically called with \`params\`/\`searchParams\`, which are \`Object.assign(Promise.resolve(obj), obj)\` thenables. \`JSON.stringify\` on a Promise serializes to \`{}\`, and React elements throw. The probe will run the function with effectively empty args, which can complete quickly even when the real call wouldn't — producing **false-positive \`UseCacheDeadlockError\`s** on legitimate slow fills.

The same file already calls \`rsc.encodeReply(unwrapThenableObjects(args), …)\` a few lines above for cache-key generation. The probe should reuse that encoded form (and the pool should \`decodeReply\` it).

### 2. \`setInsideUseCacheProbe\` is process-global, not request-scoped

\`\`\`ts
(globalThis as ...)[INSIDE_PROBE_SYMBOL] = value ? current + 1 : Math.max(0, current - 1);
\`\`\`

Next.js doesn't need an equivalent because each probe runs in a separate worker thread (\`getUseCacheProbe()\` simply returns \`undefined\` inside the worker). Vinext runs the probe in-process via a fresh \`ModuleRunner\` sharing \`globalThis\`, so a recursion guard is required — but a process-global counter means two concurrent dev requests will see \`isInsideUseCacheProbe() === true\` for each other's probes. Open two tabs and hit two cache routes; the second request silently skips probe scheduling while the first is mid-probe.

Should be ALS-scoped (or stored on the request context) so the guard only applies within the call stack that triggered the probe.

### What I'd previously flagged but takes back after checking upstream

- **Unconditional 54s dev timeout** — this matches Next.js. \`use-cache-wrapper.ts:1211-1248\` installs the same timer in dev (\`case 'request':\` gated on \`__NEXT_DEV_SERVER\`), default \`useCacheTimeout = staticPageGenerationTimeout * 0.9 = 54s\`. Not a regression vs upstream. (Minor: vinext hardcodes 54s instead of reading \`experimental.useCacheTimeout\`, but that's config plumbing, not a behavioral defect.)
- **\"Double-wrapping\" the function in the probe** — Next.js does the same thing. The worker calls \`actionMod[msg.id]\` which is the registered wrapper (\`use-cache-probe-worker.ts:106-110\`). The reason it doesn't infinite-loop is \`workStore.useCacheProbeMode\`, which gates the dev-timeout/probe block off. Vinext's \`isInsideUseCacheProbe()\` plays the same role for the probe-scheduler block. The inner 54s timer still fires in vinext but gets pre-empted by the probe-pool's own \`Promise.race\` timeout. Functionally aligned.

### Smaller things worth noting (not blocking)

- **PR description doesn't match the diff.** Lists \`server/use-cache-probe-scheduler.ts\` and two fixture pages (\`use-cache-deadlock\`, \`use-cache-recovery\`) as new files — none are in the diff. The probe currently fires on a flat \`setTimeout(10_000)\` from call start, not on cache-fill stream idleness like Next.js's \`setupProbeScheduler\`. End-to-end coverage in the App Router dev server is missing; the two added unit tests use a hand-installed fake probe and don't exercise the real \`ModuleRunner\` path.
- **Branch name is \`fix/issue-1009\` but PR closes #1126.** Issue #1009 is the draft-mode Suspense streaming port, unrelated to this change — worth confirming with the author that #1009 isn't being accidentally claimed.

CI is green, but the existing tests don't cover the cases that would catch #1 or #2 (concurrent requests, real RSC-shaped params/searchParams as cache args). Porting one of the Next.js \`use-cache-deadlock-probe\` e2e fixtures would help.

james-elicx:
Sorry. Ran an agent over it and there's a couple of things that sound worth looking at

Divkix:
> Sorry. Ran an agent over it and there's a couple of things that sound worth looking at

I'll take a look at these and try to fix them

james-elicx:
/bigbonk review for issues and nextjs parity

james-elicx:
/bigbonk review for issues and nextjs parity

PR #1165, opened 2026-05-11
fix(app-router): append RSC vary headers in finalizer

NathanDrake2406:
## Overview

| Field | Details |
| --- | --- |
| Goal | Ensure App Router responses expose the RSC response variant key to HTTP caches. |
| Core change | `finalizeAppRscResponse` now merges `VINEXT_RSC_VARY_HEADER` for non-redirect, non-static-file App Router responses. |
| Primary files | `packages/vinext/src/server/app-rsc-response-finalizer.ts`, `packages/vinext/src/server/middleware-response-headers.ts` |
| Expected impact | App route handler and other custom App Router `Response`s preserve user `Vary` values while also varying on RSC navigation headers. |

## Why

App Router responses are selected by request headers such as `RSC`, `Next-Router-State-Tree`, `Next-Router-Prefetch`, and `Next-Router-Segment-Prefetch`. If the final response omits those fields from `Vary`, shared caches can reuse an HTML/navigation/route-handler variant for a request that needed a different variant.

| --- | --- | --- |
| App Router response finalization | The cache variant key belongs at the outer App Router boundary, not only page helpers. | The finalizer appends the shared vinext RSC vary header for non-redirect App Router responses. |
| User response headers | User and middleware `Vary` values must be preserved. | Reuses the existing deduping `Vary` merge helper rather than replacing the header. |
| Static assets | Public file signals should not gain RSC cache fragmentation. | Leaves `x-vinext-static-file` signal responses alone. |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| App route handler returns `Vary: User-Agent` | Final response only varied on `User-Agent`. | Final response varies on `User-Agent` plus vinext RSC headers. |
| App page helper already sets RSC `Vary` | It could be set by the page helper only. | Finalizer dedupes and keeps the same value. |
| Public file signal | No RSC `Vary`. | Still no RSC `Vary`. |

<details>
<summary>Validation</summary>

- `vp test run tests/app-rsc-response-finalizer.test.ts tests/app-rsc-handler.test.ts tests/app-page-response.test.ts tests/app-route-handler-response.test.ts`
- `vp check`

</details>

<details>

- No public API change.
- Redirect responses still bypass finalization to avoid immutable header mutation.
- Public static file signals still avoid the RSC `Vary` value to preserve asset cache efficiency.
- Existing app page responses that already set the shared `Vary` value are deduped.

</details>

## References

| --- | --- |
| [Next.js `BaseServer.setVaryHeader`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/base-server.ts#L2000-L2025) | Shows the App Router base `Vary` set. |
| [Next.js app page route module `getVaryHeader`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/route-modules/app-page/module.ts#L184-L200) | Shows the same base vary construction for the app page module. |
| [Next.js vary-header e2e test](https://github.com/vercel/next.js/blob/canary/test/e2e/vary-header/test/index.test.ts#L16-L31) | Confirms app route handlers preserve custom `Vary` values and append internal RSC headers. |

### DIFF
--- a/packages/vinext/src/server/app-rsc-response-finalizer.ts
+++ b/packages/vinext/src/server/app-rsc-response-finalizer.ts
@@ -1,6 +1,8 @@
 import type { NextHeader } from "../config/next-config.js";
 import type { RequestContext } from "../config/config-matchers.js";
 import { applyConfigHeadersToResponse } from "./request-pipeline.js";
+import { VINEXT_RSC_VARY_HEADER } from "./app-rsc-cache-busting.js";
+import { mergeVaryHeader } from "./middleware-response-headers.js";
 import { stripBasePath } from "../utils/base-path.js";
 import { normalizePath } from "./normalize-path.js";
 import { normalizePathnameForRouteMatch } from "../routing/utils.js";
@@ -18,7 +20,8 @@ type FinalizeAppRscResponseOptions = {
 };
 
 /**
- * Apply next.config.js response headers to an App Router response.
+ * Apply App Router response finalization that must happen outside individual
+ * route dispatchers.
  *
  * Called once per request in the outer handler() wrapper, after all route
  * handling, so that every response path (page, route handler, server action,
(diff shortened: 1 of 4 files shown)


packages/vinext/src/server/app-rsc-response-finalizer.ts
```diff
@@ -39,6 +42,10 @@ export function finalizeAppRscResponse(
     return response;
   }
 
+  if (!response.headers.has("x-vinext-static-file")) {
```
james-elicx, line 45: feels like we should pull all these header keys out into consts

PR #1170, opened 2026-05-11
fix: remove duplicate next-shims.d.ts from fixtures

### LINKED ISSUE #1169: Remove need for shims type declarations in fixtures
It appears we duplicate some of the `next-shims.d.ts` content in multiple fixtures - it would be good to re-use the declaraction file from the main package if possible.

Deepam02:
fixes #1169

three test fixtures had their own copy of `next-shims.d.ts` duplicating what already exists in the main package. Removed those and pointed each fixture's tsconfig at the single shared file instead.

Had to add a few missing declarations to the shared file first before removing the copies.


Deepam02:
@james-elicx Should I clean up the examples copies too or is that out of scope?

james-elicx:
That would be great!

PR #1178, opened 2026-05-12
fix(app-router): hard navigate stale RSC build payloads

NathanDrake2406:
## Overview

| Field | Detail |
| --- | --- |
| Goal | Prevent stale App Router tabs from applying RSC payloads produced by a different vinext build. |
| Core change | Browser-consumed App RSC responses now carry a public RSC compatibility ID, and the browser hard-navigates before decoding missing or mismatched payloads. |
| Main boundary | The RSC response compatibility check lives with RSC URL/header helpers, while the browser entry owns when to hard navigate. |
| Primary files | `packages/vinext/src/server/app-rsc-cache-busting.ts`, `packages/vinext/src/server/app-browser-entry.ts`, `packages/vinext/src/shims/navigation.ts` |
| Expected impact | Stale tabs recover with a document navigation after deploy instead of attempting a broken soft navigation. |

## Why

A browser-side App Router navigation must only apply RSC data that was produced by a server artifact compatible with the current client bundle. Next.js enforces this by initializing a navigation build/deployment ID during hydration and falling back to an MPA navigation when a later Flight response belongs to another build. Vinext had content-type and status checks, but no artifact compatibility check, so stale tabs could keep using soft navigation across deploy boundaries.

| --- | --- | --- |
| RSC responses | Every browser-consumed App RSC response should identify its public compatibility identity. | Adds `X-Vinext-RSC-Compatibility-Id` after middleware headers are merged so app middleware cannot override the framework value. |
| Public identity | The browser boundary should not expose the raw `generateBuildId()` value as a header. | Uses `deploymentId` when configured; otherwise injects a random per-build public compatibility token shared by the generated bundles. |
| Browser navigation | A payload from another build must not be decoded as the current app tree. | Checks fetched, prefetched, visited-cache, and server-action RSC responses before `createFromFetch`. |
| Cached payload replay | Cached RSC snapshots must preserve compatibility metadata. | Stores and restores the compatibility header alongside params and mounted-slot headers. |

## One PR vs split

This should be one PR. Splitting would either emit unused server metadata without protection, or add a client guard with no reliable compatibility ID to compare. The server header, snapshot preservation, and browser guard are one observable behavior.

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| Fresh matching RSC response | Soft navigation decoded and committed. | Same behavior. |
| Mismatched RSC response after deploy | Browser attempted to decode and render stale Flight data. | Browser hard-navigates to the normalized document URL before decoding. |
| Missing response compatibility ID while client has one | Browser treated the payload as usable. | Browser treats it as incompatible and hard-navigates. |
| Middleware sets `X-Vinext-RSC-Compatibility-Id` | Middleware could spoof the framework value. | Framework value is applied after middleware response headers, or the framework-owned header is removed when no framework token exists. |

<details>

1. `packages/vinext/src/index.ts`: injects a public RSC compatibility ID. It uses `deploymentId` when configured and otherwise generates a random per-plugin-instance token so raw build IDs are not exposed as response headers.
2. `packages/vinext/src/server/app-rsc-cache-busting.ts`: compatibility header constant, compatibility predicate, framework-owned header application, and RSC hard-navigation target normalization.
3. `packages/vinext/src/server/app-browser-entry.ts`: checks in server actions, visited response cache, fresh fetch, and prefetch paths before decoding RSC.
4. `packages/vinext/src/server/app-page-*.ts` and `app-server-action-execution.ts`: all App Router RSC response constructors apply the compatibility header after middleware header merging.
5. `packages/vinext/src/shims/navigation.ts`: cached/prefetched RSC snapshots preserve the compatibility header.
6. Tests listed below show config identity generation, metadata preservation, middleware spoof protection, and stale navigation behavior.

</details>

<details>
<summary>Validation</summary>

Ran locally:

- `vp check`
  Result: all files formatted; no warnings, lint errors, or type errors.
- `vp test run tests/app-rsc-cache-busting.test.ts tests/app-page-response.test.ts tests/app-page-cache.test.ts tests/app-server-action-execution.test.ts tests/app-browser-entry.test.ts tests/prefetch-cache.test.ts tests/pages-router.test.ts`
  Result: 7 files, 412 tests passed.
- Commit hook also ran formatting, lint/type checks, and `knip --no-progress`.

Review feedback addressed:

- Server action return-value-only RSC responses carry the compatibility header.
- Fresh navigation visited-cache snapshots preserve the compatibility header for replay.
- The browser-facing header no longer exposes the raw `generateBuildId()` value.
- Middleware cannot spoof the framework-owned compatibility header.

</details>

<details>

- Runtime: affects App Router browser RSC navigation and server-action RSC responses.
- Cache: existing cached RSC payloads without the new header hard-navigate when the client has a compatibility ID. That is intentional because missing metadata is not safe to decode after this change.
- Public API: no intended public API change. The ambient shim declaration is updated to match the internal cached response shape.
- Middleware: app middleware can still set user headers, but the vinext RSC compatibility header remains framework-owned.

</details>

<details>
<summary>Non-goals</summary>

- Does not change Pages Router data route build ID behavior.
- Does not redesign the App Router browser entry or RSC payload envelope.
- Does not add multi-zone routing semantics beyond using `deploymentId` as the public compatibility ID when configured.

</details>

## References

| --- | --- |
| [Next.js navigation build ID state](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/navigation-build-id.ts) | Stores the client-side build/deployment ID used for navigation compatibility checks. |
| [Next.js hydration initializes navigation build ID](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/app-index.tsx) | Sets the navigation build ID from the initial RSC payload or deployment ID before hydration. |
| [Next.js fetchServerResponse build/deployment check](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/components/router-reducer/fetch-server-response.ts) | Falls back to MPA navigation when a Flight response identity does not match the client identity. |
| [Next.js app render attaches build ID to RSC payloads](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/app-render/app-render.tsx) | Shows the server side source of default build ID metadata used by the browser. |
| [Next.js deployment ID header constant](https://github.com/vercel/next.js/blob/canary/packages/next/src/lib/constants.ts) | Defines `x-nextjs-deployment-id`, the public header Next uses for deployment skew protection. |


NathanDrake2406:
FULL TIME JAMES HAS ARRIVED

PR #1181, opened 2026-05-12
Fix spelling errors in README.md

azizbecha:
Corrected spelling of 'workerd' to 'workers' in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ export default async function Page() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


james-elicx:
Hi, thanks for the PR. This is intentional - the Vite [plugin](https://developers.cloudflare.com/workers/vite-plugin/) uses workerd as its environment to run code, which is the runtime that Workers uses.

PR #1190, opened 2026-05-14
chore(deps): bump vite-plus to 0.1.21

fengmk2:
- Bump vite, vite-plus, vitest catalog entries to 0.1.21
- Refresh pnpm-lock.yaml
- Regenerate AGENTS.md via vp config postinstall
- Drop unused imports (PHASE_PRODUCTION_BUILD, patternToNextFormat) flagged by upgraded lint rules
- Move react-hooks/rules-of-hooks suppression in tests/e2e/fixtures.ts to the use(errors) call site
- Add react/no-children-prop and react/no-danger-with-children suppressions on the intentional precedence test in tests/head.test.ts

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -515,87 +515,14 @@ When adding support for third-party Next.js libraries:
 
 # Using Vite+, the Unified Toolchain for the Web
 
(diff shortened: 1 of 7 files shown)


james-elicx (approved):
Thanks!

PR #1197, opened 2026-05-14
Implement segment-scoped useRouter bfcacheId semantics

manNomi:
## Summary
- implement segment-scoped `useRouter().bfcacheId` values for the App Router browser state
- persist bfcache id maps in history state so back/forward restores prior ids
- preserve ids across shared layouts, hash/search-param navigation, and refresh while minting fresh ids for fresh segment entries
- add a Next.js compat fixture and E2E coverage adapted from upstream commit `56d9513`

This is a follow-up to the placeholder API-surface PR. It keeps the public field added by #1192, but wires it into vinext's current App Router state model.

## Tests
- `pnpm exec vp check packages/vinext/src/server/app-browser-state.ts packages/vinext/src/server/app-browser-visible-commit.ts packages/vinext/src/server/app-browser-navigation-controller.ts packages/vinext/src/server/app-browser-entry.ts packages/vinext/src/server/app-ssr-entry.ts packages/vinext/src/shims/navigation.ts packages/vinext/src/shims/slot.tsx tests/app-browser-entry.test.ts tests/shims.test.ts tests/e2e/app-router/nextjs-compat/use-router-bfcache-id.spec.ts`
- `pnpm exec vp test run tests/shims.test.ts -t "useRouter"`
- `pnpm exec vp test run tests/app-browser-entry.test.ts -t "bfcacheId"`
- `pnpm exec vp run vinext#build`
- `PLAYWRIGHT_PROJECT=app-router pnpm exec playwright test tests/e2e/app-router/nextjs-compat/use-router-bfcache-id.spec.ts`

### DIFF
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -16,6 +16,7 @@ import {
   commitClientNavigationState,
   consumePrefetchResponse,
   createClientNavigationRenderSnapshot,
+  getBfcacheIdMapContext,
   getCurrentNextUrl,
   getCurrentInterceptionContext,
   getClientNavigationRenderContext,
@@ -62,6 +63,8 @@ import {
 } from "./app-elements.js";
 import {
   createHistoryStateWithPreviousNextUrl,
+  createInitialBfcacheIdMap,
+  readHistoryStateBfcacheIds,
   readHistoryStatePreviousNextUrl,
   resolveInterceptionContextFromPreviousNextUrl,
   resolveServerActionRequestState,
@@ -217,13 +220,14 @@ function clearClientNavigationCaches(): void {
 }
 
 function createNavigationCommitEffect(options: {
+  bfcacheIds: Readonly<Record<string, string>>;
   href: string;
   historyUpdateMode: HistoryUpdateMode | undefined;
   navId: number;
   params: Record<string, string | string[]>;
   previousNextUrl: string | null;
 }): () => void {
-  const { href, historyUpdateMode, navId, params, previousNextUrl } = options;
+  const { bfcacheIds, href, historyUpdateMode, navId, params, previousNextUrl } = options;
 
   return () => {
     // Only update URL if this is still the active navigation.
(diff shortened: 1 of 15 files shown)


manNomi:
@james-elicx Thanks for the review. I pushed a follow-up commit addressing the concrete items and added comments where the behavior is intentional.

Could you take another look when you have a chance?

NathanDrake2406:
@james-elicx Let's merge this, I'll try to build the foundation to the BFcache system before my sub runs out.
Here's a review for this from my agent.
**Verdict: Request changes.**

CI is green for `CI` and `Preview Release`; `Deploy Examples` is skipped. But I would not merge this as-is.

### Blocking finding

**Preserved parallel-route subtrees can lose their `bfcacheId`.**

`createNextBfcacheIdMap()` builds the next id map only from the incoming `AppElements` payload. But visible navigation commits later call `mergeElements(...)`, which can preserve old elements that were absent from the incoming payload, especially mounted parallel-route slots. The committed router state then installs `action.bfcacheIds` wholesale.

That creates a bad state:

1. old mounted slot remains alive via `mergeElements`
2. its `<Slot>` still provides its segment id
3. `BfcacheIdMapContext` no longer has an entry for that id
4. `useRouter().bfcacheId` falls back to `"0"`

So a preserved subtree can suddenly observe a different bfcache id even though React preserved the subtree. Any descendant using `key={router.bfcacheId}` will remount spuriously. This directly violates the semantic this PR is trying to implement.

Concrete reproduction shape:

```tsx
app/feed/@modal/default.tsx
// client component inside modal uses:
<form key={useRouter().bfcacheId}>...</form>
```

Then navigate within the owning route so the server omits the already-mounted modal slot and the client preserves it. The slot content remains mounted, but its bfcache id falls out of the map and becomes `"0"`.

The tests already cover “preserve mounted parallel slots” and “bfcache ids”, but not their intersection. That’s the hole.

### Fix shape

Do not treat `action.bfcacheIds` as final for merge-based commits. For `navigate` / `traverse`, compute or repair the bfcache map against the **post-merge element set**, not just the incoming payload.

Minimal safe repair:

```ts
const mergedElements = mergeElements(...);
const bfcacheIds = {
  ...action.bfcacheIds,
};

for (const id of collectBfcacheSegmentIds(mergedElements)) {
  if (!(id in bfcacheIds) && id in state.bfcacheIds) {
    bfcacheIds[id] = state.bfcacheIds[id];
  }
}
```

That helper probably belongs in `app-browser-state.ts`, since the segment-id filtering logic is already private there.

Add a regression test beside the existing parallel-slot preservation test:

```ts
expect(nextState.elements["slot:modal:/feed"]).toBe(mountedSlot);
expect(nextState.bfcacheIds["slot:modal:/feed"]).toBe(state.bfcacheIds["slot:modal:/feed"]);
```

### Otherwise

The broad architecture is good: context layering is clean, history-state persistence is in the right place, traversal restoration is correctly prioritised, and the E2E coverage for fresh pushes, back/forward, search/hash, refresh, and programmatic navigation is solid. This is close — but the preserved-slot map hole is a real edge, not a theoretical neatness complaint.

PR #1201, opened 2026-05-14
fix: expose window.next global for Next.js parity

james-elicx:
## Summary

Adds the `window.next` global that Next.js publishes on the client from both the [Pages Router bootstrap](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/next.ts#L13) and the [App Router bootstrap](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/app-bootstrap.ts#L13). vinext did not install this global, which broke any caller that reads `window.next.router` directly — including the Next.js deploy test suite, Pages Router tests in userland, and several third-party libraries.

## Deploy suite impact

Before this PR the Next.js deploy suite reported:
- **~422 console errors** mentioning `window.next.*`
- **30+ runtime test failures**
- Most cited error: `TypeError: Cannot read properties of undefined (reading 'router')`. `__internal_src_page` and `__pendingUrl` ranked next.

After this PR those callers find the expected shape on `window.next` and the errors clear.

## What was added

A typed `installWindowNext()` helper lives in `packages/vinext/src/client/window-next.ts`. Both router bootstraps invoke it, so the field shape stays in one place per AGENTS.md's guidance to keep generated entries thin and put real logic in `client/` / `server/` typed modules.

| `window.next.*` field    | Owner             | Ported from                                                                                                                                                                                              |
| ------------------------ | ----------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `version`                | Pages + App       | `packages/next/src/client/next.ts` line 13 (Pages); `packages/next/src/client/app-bootstrap.ts` line 13 (App). Reports the vinext package version, injected via Vite `define` for `process.env.__NEXT_VERSION` (mirroring Next.js's own DefinePlugin). |
| `appDir`                 | App only          | `packages/next/src/client/app-bootstrap.ts` line 15 (`appDir: true`)                                                                                                                                     |
| `router` (App Router)    | App only          | `packages/next/src/client/components/app-router-instance.ts` line 510 (`window.next.router = publicAppRouterInstance`). Wired to vinext's `appRouterInstance` (the same singleton `useRouter()` returns). |
| `router` (Pages Router)  | Pages only        | `packages/next/src/client/next.ts` line 16-18 (live-binded getter). Wired to vinext's Pages `Router` singleton from `next/router`.                                                                       |
| `router.events`          | Pages only        | `packages/next/src/shared/lib/router/router.ts` line 372 (`NextRouter` type). Tests like `test/development/pages-dir/client-navigation/index.test.ts:457` rely on `window.next.router.events.on(...)`.   |
| `__pendingUrl`           | App Router (set during navigation) | `packages/next/src/client/components/app-router-instance.ts` line 296; read by `packages/next/src/client/components/nav-failure-handler.ts`                                                              |
| `__internal_src_page`    | App Router (set during navigation) | `packages/next/src/client/components/app-router.tsx` line 204                                                                                                                                            |

## Version wiring

The Vite plugin reads `packages/vinext/package.json#version` at plugin load (`getVinextVersion()` in `index.ts`) and exposes it via `define` as `process.env.__NEXT_VERSION`. The `window-next.ts` module reads that value and uses it as the default for `installWindowNext({ version })`. Mirrors how Next.js substitutes the same constant via its webpack `DefinePlugin`. Library code that reads `process.env.__NEXT_VERSION` directly now also sees a real value instead of `undefined`.

The standalone `"vinext"` literal remains as a safety fallback for any code path that imports `client/window-next.ts` without going through the Vite plugin (unit tests that don't exercise the build).

## Fields intentionally omitted

- **`emitter`** (Pages, `packages/next/src/client/next.ts` line 19) — this is the page-load `mitt()` emitter, distinct from `router.events`. vinext does not currently fire those events anywhere, so exposing it would be a fake surface. Left as a follow-up if it shows up in the deploy suite.
- **`router.hmrRefresh`** and **`router.experimental_gesturePush`** (App Router) — vinext does not implement either, so they remain undefined. Library callers that probe `typeof router.hmrRefresh === "function"` skip the branch, matching production Next.js behavior.
- **`turbopack`** flag from `packages/next/src/client/app-next-turbopack.ts` — vinext doesn't use Turbopack, so this flag is not meaningful.

## Open questions / follow-ups

1. `__pendingUrl` and `__internal_src_page` are populated only when the App Router's per-navigation code paths get wired up to write them. Initial install leaves both `undefined` (matching Next.js's initial-load state), but full parity requires hooking the writes inside `server/app-browser-entry.ts`'s navigation controller. If the deploy suite still complains about these specific fields after this PR, that follow-up should land next.

## Tests

Added 6 focused tests under `tests/shims.test.ts` in a new `window.next debug global` describe block:

- `installWindowNext` populates `version`, `router`, and `appDir`
- repeat calls merge fields (App Router bootstrap wins for `router` when both routers are present)
- Pages Router shim attaches the expected `NextRouter` surface at module load
- `window.next.router.events` forwards Pages Router events
- `appRouterInstance` named export has the App Router public surface
- `installWindowNext` is a no-op when `typeof window === "undefined"`

Full `tests/shims.test.ts` (866 tests) and `tests/link.test.ts`, plus `tests/app-browser-entry.test.ts`, `tests/app-router.test.ts`, and `tests/pages-router.test.ts` (605 tests) all pass. `tests/build-optimization.test.ts` (73 tests) also passes.

`vp run knip` is clean. `vp check` is clean for code in this PR. There is one pre-existing lint error in `tests/e2e/fixtures.ts` (also present on `origin/main`) that is unrelated.

### DIFF
--- a/packages/vinext/src/client/window-next.ts
+++ b/packages/vinext/src/client/window-next.ts
@@ -1,0 +1,189 @@
+/**
+ * Install the `window.next` debug/diagnostic global that Next.js exposes
+ * on the client.
+ *
+ * Next.js publishes a small per-app object on `window.next` from its
+ * client bootstraps and uses it for two distinct purposes:
+ *
+ *   1. An external debugging / test-automation surface. Pages Router tests
+ *      and userland code routinely call `window.next.router.push(...)` and
+ *      `window.next.router.events.on(...)` directly, and the App Router
+ *      bootstrap sets `appDir: true` so consumers can branch on which
+ *      router is active.
+ *      - Pages Router: `packages/next/src/client/next.ts`
+ *      - App Router: `packages/next/src/client/app-bootstrap.ts`
+ *      - App Router public surface:
+ *        `packages/next/src/client/components/app-router-instance.ts`
+ *        (`window.next.router = publicAppRouterInstance` at line 510)
+ *
+ *   2. Internal navigation bookkeeping read by Next.js itself. The App
+ *      Router's <Router> component writes `window.next.__internal_src_page`
+ *      whenever the active source-page changes, and the router instance
+ *      writes `window.next.__pendingUrl` at the start of a programmatic
+ *      navigation so nav-failure-handler.ts can fall back to a hard
+ *      navigation if a render fails.
+ *      - `packages/next/src/client/components/app-router.tsx` (line ~204)
+ *      - `packages/next/src/client/components/app-router-instance.ts`
+ *        (line ~296)
+ *      - `packages/next/src/client/components/nav-failure-handler.ts`
+ *
+ * Without this global, third-party libraries and a large fraction of the
+ * Next.js deploy test suite crash with
+ * `TypeError: Cannot read properties of undefined (reading 'router')`.
+ *
+ * Both routers in vinext share this installer so the field shape stays in
+ * sync and only one source of truth describes the supported keys.
+ */
+
+/**
+ * The minimum App Router public router surface that Next.js exposes on
+ * `window.next.router`. Mirrors the `publicAppRouterInstance` shape from
+ * `packages/next/src/client/components/app-router-instance.ts`.
+ *
+ * `hmrRefresh` and `experimental_gesturePush` are intentionally omitted —
+ * vinext does not implement them. Library callers that branch on their
+ * presence (`typeof router.hmrRefresh === "function"`) will skip the
+ * branch, matching what they would do on a production Next.js build.
+ */
+export type AppRouterPublicInstance = {
+  push: (href: string, options?: { scroll?: boolean }) => void;
+  replace: (href: string, options?: { scroll?: boolean }) => void;
+  back: () => void;
+  forward: () => void;
+  refresh: () => void;
+  prefetch: (href: string) => void;
+  /** Default placeholder, matches Next.js. */
+  bfcacheId?: string;
+};
+
+/**
+ * Pages Router singleton surface — matches `NextRouter` from
+ * `packages/next/src/shared/lib/router/router.ts` (line 372).
+ *
+ * Typed loosely here because the helper is consumed by both the pages
+ * router shim (which exports the strict `NextRouter` type) and tests that
+ * only care about the runtime shape. The structural fields below are the
+ * subset that downstream callers (Next.js tests, third-party libs) read
+ * synchronously off `window.next.router`.
+ */
+export type PagesRouterPublicInstance = {
+  push: (...args: unknown[]) => unknown;
+  replace: (...args: unknown[]) => unknown;
+  back: () => void;
+  reload: () => void;
+  prefetch: (...args: unknown[]) => unknown;
+  beforePopState: (cb: (...args: unknown[]) => boolean) => void;
+  events: {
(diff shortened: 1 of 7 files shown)


packages/vinext/src/client/window-next.ts
```diff
@@ -0,0 +1,189 @@
...
+// module type-checks standalone without depending on the global.d.ts
+// augmentation (which itself imports WindowNext from this file). Matches the
+// pattern Next.js uses in `packages/next/src/client/next.ts` lines 7-11:
+//   declare global { interface Window { next: any } }
+declare global {
+  // oxlint-disable-next-line typescript/consistent-type-definitions
+  interface Window {
+    next?: WindowNext;
+  }
+}
```
james-elicx, line 93: can this be moved to the global type decls file potentially?

packages/vinext/src/client/window-next.ts
```diff
@@ -0,0 +1,189 @@
...
+  replace: (...args: unknown[]) => unknown;
+  back: () => void;
+  reload: () => void;
+  prefetch: (...args: unknown[]) => unknown;
+  beforePopState: (cb: (...args: unknown[]) => boolean) => void;
+  events: {
+    on: (event: string, handler: (...args: unknown[]) => void) => void;
+    off: (event: string, handler: (...args: unknown[]) => void) => void;
+    emit: (event: string, ...args: unknown[]) => void;
+  };
```
james-elicx: do these types not already exist elsewhere?

packages/vinext/src/client/window-next.ts
```diff
@@ -0,0 +1,189 @@
...
+ * `next.ts` and overwrites `window.next.router`.
+ */
+export function installWindowNext(fields: Partial<WindowNext>): WindowNext {
+  if (typeof window === "undefined") {
+    // SSR: nothing to do. Returning an empty object keeps the call signature
+    // total without requiring callers to wrap each call in `if (window)`.
+    return { version: fields.version ?? FALLBACK_VERSION, ...fields };
+  }
+
+  const existing = window.next ?? installedNext;
```
james-elicx: would installedNext ever actually have a value when window.next doesn't?

james-elicx:
/bigbonk

Addressed all review points:

1. Dropped `export` from `AppRouterPublicInstance` and `WindowNext` (knip flagged them). `PagesRouterPublicInstance` stays exported because `shims/router.ts` imports it.
2. Removed `getWindowNext()` and its export — tests now read `window.next` directly, same surface runtime callers use.
3. Wired up the real version: `packages/vinext/package.json#version` is read once at plugin load and injected as `process.env.__NEXT_VERSION` via Vite `define`. This mirrors Next.js's own DefinePlugin substitution, so library code that branches on `process.env.__NEXT_VERSION` also gets a real value.
4. Removed the `installedNext` module-level fallback. `installWindowNext` now reads and writes through `window.next` directly so deleting `window.next` cleanly resets state.

`vp run knip` is clean. `vp check` is clean for code in this PR (the pre-existing lint error in `tests/e2e/fixtures.ts` is on `origin/main` and unrelated). All targeted vitest suites pass (shims/link/app-router/pages-router/app-browser-entry/build-optimization).

PR #1189, opened 2026-05-13
ci: add Next.js deploy suite harness

james-elicx:
## Summary

Adds the scripts and GitHub Actions workflow for running the Next.js e2e deploy test suite against vinext, structured per the [Next.js adapter testing docs](https://nextjs.org/docs/app/api-reference/adapters/testing-adapters).

Extracted from #922 (which also contains the Pages Router compatibility fixes that make the suites pass).

## How it works

Next.js ships a test harness (`run-tests.js`) that creates isolated temp apps, calls an adapter's deploy/logs/cleanup scripts, then runs Playwright tests against the running server. Our scripts wire that harness up to vinext: each test app gets vinext injected as a local dependency, built with `vinext build`, and served with `vinext start`.

## What's included

**Workflow** (`.github/workflows/nextjs-deploy-suite.yml`):
- `build` job: builds vinext, prepares Next.js checkout, generates the deploy manifest
- `test` job: restores the build cache, shards tests 16 ways via `run-tests.js`
- `workflow_dispatch` with `vinext-ref` (branch to test), `next-ref` (Next.js version, default v16.2.6), `suite-filter` (app/pages/all, default app), `test-concurrency`
- Nightly cron at 02:00 UTC against `main`

**Scripts** (named per Next.js docs convention):
- `scripts/e2e-deploy.sh` — per-test deploy: injects vinext into the test app, builds with `vinext build`, starts with `vinext start`, prints URL to stdout
- `scripts/e2e-cleanup.sh` — kills the server process, persists debug logs
- `scripts/e2e-logs.sh` — replays build/server logs (includes required `BUILD_ID:` / `DEPLOYMENT_ID:` / `IMMUTABLE_ASSET_TOKEN:` markers)
- `scripts/run-nextjs-deploy-suite.sh` — convenience wrapper for local runs

**Manifest generator** (`scripts/nextjs-deploy-manifest.mjs`):
- `--filter app|pages|all` to select which suites to include from the Next.js `deploy-tests-manifest.json`

**Other:**
- `.gitignore` updated to ignore `reports/`

## pnpm in the scripts

The scripts use pnpm (not `vp`) in two places, both intentionally:
- **Next.js checkout** — Next.js is a pnpm workspace; its build/test tooling requires pnpm
- **Isolated test app temp dirs** — throwaway apps created by the Next.js harness, outside the vinext monorepo; `vp` has no project context here

`vp` is already used where it should be (building vinext itself).

PR #1209, opened 2026-05-14
fix(deploy-suite): extract CJS->ESM converter to standalone script

james-elicx:
## Summary

- Move the `next.config.{js,ts}` CJS→ESM converter out of an inlined `node -e '…'` blob in `scripts/e2e-deploy.sh` and into a standalone `scripts/cjs-to-esm-config.mjs`. Same regex pipeline, no behavioural change.
- Add a `bash -n scripts/*.sh` gate to the **Check** job so this class of regression can't ship green.

## Why

The deploy-suite run [25869359604](https://github.com/cloudflare/vinext/actions/runs/25869359604) reported `329 passed, 2480 failed, 653 skipped` — a 9.5% pass rate. Investigating, **every single failure** was byte-identical:

```
Error: Custom deploy script failed:  undefined (2)
  at NextDeployInstance.deployUsingCustomScript (next.js/test/lib/next-modes/next-deploy.ts:70:13)
```

…and the runner logs showed the real cause:

```
scripts/e2e-deploy.sh: line 490: syntax error near unexpected token `('
scripts/e2e-deploy.sh: line 490: `      // TODO: This doesn't perfectly handle all CJS patterns (e.g. dynamic'
```

`bash -n scripts/e2e-deploy.sh` reproduces the error locally on `843df83`.

### Root cause

The CJS→ESM converter (added in #1189) was passed to `node -e` as a single-quoted bash argument. The JS body used the `'\"'\"'` quote-escape trick repeatedly inside regex literals. The net effect rebalanced quoting such that some later lines — notably the comment `// TODO: … (e.g. dynamic …` on line 490 — sat **outside** a single-quoted bash region. Bash then saw `(e.g. dynamic …)` as shell syntax and refused to parse the whole script.

The deploy step exited with status 2 before doing any work, so **every deploy-suite test failed at `createNext()` setup** before any vinext code was exercised. The 329 "passes" were almost entirely `should skip next deploy` placeholders, not real assertions — effective real-test pass rate this run was 0%.

### What the fix does

- `scripts/cjs-to-esm-config.mjs`: new standalone Node script with the identical regex pipeline (handles `module.exports = X`, `const X = require('mod')`, `const X = require('mod')(args)`, `const { a, b } = require('mod')`, and bare `require('mod')` in expressions). Lives in a `.mjs` file so its source is never re-parsed by bash.
- `scripts/e2e-deploy.sh`: the old 60-line `node -e '…'` blob is replaced by `node "${VINEXT_DIR}/scripts/cjs-to-esm-config.mjs" "${config_file}"`. Net `-65 / +13` lines in the bash script.
- `.github/workflows/ci.yml`: new step in the **Check** job runs `bash -n` on every `scripts/*.sh`. If anyone reintroduces a syntax error in a deploy-critical shell script, the Check gate fails on the PR.

## Verification

Locally:

```
$ bash -n scripts/e2e-deploy.sh && echo OK
OK
$ for f in scripts/*.sh; do bash -n "$f" || echo FAIL $f; done
(all clean)
```

Smoke-tested the converter against a representative `next.config.js`:

```js
// input
const path = require('path')
const withPWA = require('next-pwa')({ dest: 'public' })
const { withSentryConfig } = require('@sentry/nextjs')

module.exports = withPWA({
  reactStrictMode: true,
  basePath: '/myapp',
})
```

```js
// output (validated with `node --check --input-type=module`)
import _cjsImport0 from 'next-pwa';
import path from 'path';
import { withSentryConfig } from '@sentry/nextjs';

const withPWA = _cjsImport0({ dest: 'public' })

export default withPWA({
  reactStrictMode: true,
  basePath: '/myapp',
})
```

The regex pipeline is byte-for-byte the one that was inlined; only the wrapping changed.

## Follow-up (not in this PR)

The harness reported `undefined` for stderr in every failure (`Custom deploy script failed: undefined (2)`), which made the root cause invisible in the test report — you have to dig into raw runner logs to find the bash syntax error. The `deploy-debug-N` artifacts also only contain `context.txt` and `package.json`, no stdout/stderr capture from the deploy script itself. Worth a separate PR to fix the harness so future failures of this class are visible in the test report.

After this lands, re-run the deploy suite to get the first real parity signal. With the harness fixed, the run will surface actual vinext bugs and parity gaps instead of one repeated harness error.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,9 +25,6 @@ jobs:
         run: vp run build
       - run: vp run check
       - run: vp run knip
-      # Catch bash syntax errors in CI-critical shell scripts. The deploy-suite
-      # harness (scripts/e2e-deploy.sh) once shipped with a syntax error that
-      # caused 2480 deploy-suite failures before anyone noticed — see #1189.
       - name: Bash syntax check
         run: |
           for f in scripts/*.sh; do
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,14 @@ jobs:
         run: vp run build
       - run: vp run check
       - run: vp run knip
+      # Catch bash syntax errors in CI-critical shell scripts. The deploy-suite
+      # harness (scripts/e2e-deploy.sh) once shipped with a syntax error that
+      # caused 2480 deploy-suite failures before anyone noticed — see #1189.
+      - name: Bash syntax check
+        run: |
+          for f in scripts/*.sh; do
+            bash -n "$f" || { echo "Syntax error in $f"; exit 1; }
+          done
 
   test-unit:
     name: Vitest (unit)
--- a/scripts/cjs-to-esm-config.mjs
+++ b/scripts/cjs-to-esm-config.mjs
@@ -1,0 +1,90 @@
+#!/usr/bin/env node
+// Convert a CommonJS Next.js config file (next.config.{js,ts}) to ESM in-place.
+//
+// Used by scripts/e2e-deploy.sh after `vinext init` adds "type": "module" to
+// the test app's package.json — at that point Node treats .js as ESM, but
+// Next.js doesn't accept .cjs for its config file, so we have to rewrite the
+// CJS syntax to ESM equivalents.
+//
+// This was previously inlined into e2e-deploy.sh via `node -e '…'`, but the
+// JS body contained the `'"'"'` quote-escape pattern enough times that
+// rebalancing got broken (one unquoted `(` in a comment was enough to make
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -25,6 +25,14 @@ jobs:
         run: vp run build
       - run: vp run check
       - run: vp run knip
+      # Catch bash syntax errors in CI-critical shell scripts. The deploy-suite
+      # harness (scripts/e2e-deploy.sh) once shipped with a syntax error that
+      # caused 2480 deploy-suite failures before anyone noticed — see #1189.
```
james-elicx:
```suggestion
```

PR #1239, opened 2026-05-15
feat(apps/web): /compatibility page backed by D1 + deploy-suite ingest

james-elicx:
## Summary

Adds a `/compatibility` page to `apps/web` that visualises Next.js compatibility over time, populated by the nightly deploy-suite workflow.

- **D1 + Drizzle**: new `vinext-metrics` D1 database with namespaced tables (`compat_runs`, `compat_file_results`) so future metric kinds can live alongside.
- **POST `/api/compatibility`**: ingest endpoint authenticated by a `COMPAT_INGEST_SECRET` worker secret. Upserts by `(kind, runKey)`, so retried CI runs replace stale rows cleanly.
- **Visualisation** (both pure SVG, no chart libs):
  - GitHub-style contribution grid — one dot per test file, green/orange/red/gray for pass/partial/fail/skipped-by-Next.js.
  - Pass-rate over time line chart. Skipped tests don't count against the pass rate (denominator is `passed + failed`).
- **Shared chrome**: header/footer extracted into `app/_components` and rendered by the root layout. Nav uses `next/link` for soft client navigation.
- **Workflow change**: the existing per-test aggregation in `nextjs-deploy-suite.yml` now also POSTs the results to the ingest endpoint. Guarded to only run when (a) `filter=all` and (b) the run targets `main`, so partial runs and branch spot-checks don't pollute the historical record.

## Status

Draft because:
- The remote D1 is migrated but empty. The first scheduled nightly run after merge will seed it.
- `COMPAT_INGEST_SECRET` repo secret needs to be set before the workflow can submit.
- The `vinext-web` worker hasn't been redeployed yet; the new route handler and D1 binding ship on the next deploy.

## Setup required (one-time)

1. Set the repo secret: `gh secret set COMPAT_INGEST_SECRET`
2. Set the same value as a worker secret: `wrangler secret put COMPAT_INGEST_SECRET` (run from `apps/web/`)
3. (Optional) Override the ingest URL: `gh variable set COMPAT_INGEST_URL --body "https://your-worker/api/compatibility"`

Already done:
- D1 database `vinext-metrics` created in the `vinext` Cloudflare account, id `63e7cd24-9d47-4f17-a41d-b5806babc406`.
- Migration `0000_awesome_the_order.sql` applied to both local and remote.

## Testing

- Local dev (`vinext dev --port 3001`) reads from a local SQLite D1 emulator and renders the page with the historical run from `25915834604` already ingested.
- `vp check`, `vp run knip` both pass.

### DIFF
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -1,3 +1,3 @@
-	# Required by POST /api/compatibility to authenticate ingest submissions.
+# Required by POST /api/compatibility to authenticate ingest submissions.
 # Generate with: node -e "console.log(require('node:crypto').randomBytes(32).toString('hex'))"
 COMPAT_INGEST_SECRET=""
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -1,0 +1,3 @@
+	# Required by POST /api/compatibility to authenticate ingest submissions.
+# Generate with: node -e "console.log(require('node:crypto').randomBytes(32).toString('hex'))"
+COMPAT_INGEST_SECRET=""
--- a/.github/workflows/nextjs-deploy-suite.yml
+++ b/.github/workflows/nextjs-deploy-suite.yml
@@ -202,6 +202,15 @@ jobs:
       - name: Generate report
         id: report
         uses: actions/github-script@v7
+        env:
+          # Resolve the GitHub Actions expressions to env vars once, instead
+          # of splicing `${{ ... }}` into JS string literals further down.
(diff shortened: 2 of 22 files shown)


apps/web/.env.example
```diff
@@ -0,0 +1,3 @@
+	# Required by POST /api/compatibility to authenticate ingest submissions.
```
james-elicx:
```suggestion
# Required by POST /api/compatibility to authenticate ingest submissions.
```

PR #1275, opened 2026-05-16
fix(config): load CJS next.config.js under "type": "module"

james-elicx:
## Summary

- `loadNextConfig`'s `createRequire` fallback fails on `.js` files in `"type": "module"` packages because Node resolves the file as ESM (extension + parent `type` wins over the `require()` caller). The Next.js deploy suite worked around this by rewriting every fixture's config in-place via `scripts/cjs-to-esm-config.mjs` after `vinext init`.
- Add a final fallback: copy the source to a sibling temp `.cjs` (the explicit extension overrides the parent `type` field), `require()` that, clean up. Relative imports inside the config still resolve against the original directory.
- Drop the rewrite step from `scripts/e2e-deploy.sh` and delete the 145-line `cjs-to-esm-config.mjs` shim + its tests.

## Why this matters

238 fixtures under Next.js's `test/e2e/` use `module.exports = {...}` in `next.config.js`. Every one of them previously needed the syntactic rewrite before vinext could build. With this change vinext just loads them natively.

## Scope

Handles `.js` only. CJS in `next.config.ts` is rare (3 hits across all of vercel/next.js, all unit-test fixtures; 0 in `test/e2e/`). If a real fixture appears, the same trick won't work — Node can't `require()` `.ts` — and we'd need transpile-then-require.

## Test plan

- [x] `vp check` (1286 files clean)
- [x] `vp test run tests/next-config.test.ts` — 91/91 incl. 4 new tests covering bare `module.exports + require`, `__dirname`/`__filename`, `require('mod')(args)` plugin-wrapper shape, and tmp-file cleanup
- [x] `vp test run tests/deploy.test.ts tests/init.test.ts` — 281/281
- [ ] Next.js Deploy Suite workflow run on this branch (the actual integration test bench — needs CI)


james-elicx:
## End-to-end validation against Next.js v16.2.6 fixtures

Cloned `vercel/next.js@v16.2.6` (the ref the deploy-suite workflow targets) and ran two validation passes.

### 1. `loadNextConfig` pre-fix vs post-fix

Smoke-tested 11 real `test/e2e/*/next.config.js` fixtures by copying each into a temp dir with a `{"type":"module"}` package.json (i.e. the exact shape `vinext init` produces).

| State | Pass | Fail |
| --- | --- | --- |
| **Pre-fix** (createRequire fallback only) | 0 | 11 |
| **Post-fix** (temp `.cjs` sibling) | 11 | 0 |

Pre-fix failures were all `module is not defined in ES module scope` or `require is not defined in ES module scope` — the exact errors `scripts/cjs-to-esm-config.mjs` was working around.

Fixtures covered:
- `test/e2e/define` — plain `module.exports = {...}`
- `test/e2e/legacy-link-behavior`, `test/e2e/url`, `test/e2e/next-analyze` — same shape, different content
- `test/e2e/trailingslash-with-rewrite/app`, `test/e2e/app-dir-export`, `test/e2e/i18n-disallow-multiple-locales`, `test/e2e/edge-runtime-uses-edge-light-import-specifier-for-packages`, `test/e2e/optimized-loading` — varied content
- `test/e2e/app-dir/metadata-font` — gated `require('@next/bundle-analyzer')({...})` plugin-wrapper pattern
- `test/e2e/deferred-entries` — heavy `require('fs')` + `require('path')` + `__dirname` usage

(Three other fixtures — `app-dir/app-css`, `app-dir/modularizeimports`, `twoslash` — failed both pre and post with `Cannot find module '@next/mdx'` / `typescript/package.json` because the smoke harness skips `pnpm install`. Real `e2e-deploy.sh` runs install deps and these load fine.)

### 2. Full `e2e-deploy.sh` against a scaffolded app

Built a minimal Next.js scaffold matching what Next.js's `createNext` helper produces (an `app/page.tsx`, an `app/layout.tsx`, and a CJS `next.config.js` exercising `module.exports`, top-level `require('node:path')` + `require('node:fs')`, `__dirname`, and a gated `require('@next/bundle-analyzer')({...})`), then ran `scripts/e2e-deploy.sh` against it from cold:

- Build completed, BUILD_ID emitted
- Server came up at `http://127.0.0.1:65481`
- `curl /` returned the rendered page with `tag=cjs-via-vinext` — confirming the env var defined in the CJS `next.config.js` reached the server
- `next.config.js` is byte-identical to the original after the run (the deleted converter was not invoked — there is nothing to invoke it)
- No stray `.vinext-next-config.*.cjs` temp files left in the project root

The real CI deploy-suite run is the proof of full coverage across all 238 fixtures, but the pre/post comparison plus the e2e scaffold give me high confidence the loader change is sufficient.

PR #1278, opened 2026-05-16
fix(config): provide CJS globals when loading next.config.ts

james-elicx:
## Summary

Several upstream `next-config-ts*` fixtures fail to even build in vinext with errors like:

- `[vinext] Failed to load next.config.ts: module is not defined`
- `ReferenceError: __filename is not defined`
- `ReferenceError: __dirname is not defined`

Next.js evaluates `next.config.ts` in a context where the CJS-style globals (`__filename`, `__dirname`, `module`, `exports`, `require`) are available even though the file itself uses ESM syntax. Concretely, `packages/next/src/build/next-config-ts/transpile-config.ts` transforms the source to CommonJS via SWC ([`module.type: 'commonjs'`](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/next-config-ts/transpile-config.ts#L18-L43)) and then feeds it through `requireFromString`, which calls Node's `Module._compile` ([require-hook.ts](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/next-config-ts/require-hook.ts#L50-L56)) - and `Module._compile` naturally wraps the code in `(function(exports, require, module, __filename, __dirname) { ... })`.

vinext loads the config through Vite's `runnerImport`, which uses an ESM-only evaluator. Without the CJS globals, any TypeScript config that references them throws.

This PR adds a small Vite plugin scoped to the config file that prepends a preamble declaring the missing globals as local `const`s. The preamble also installs a wrapper `module = { exports: {} }` and re-exports it as a named export so that CJS-style `module.exports = { ... }` reassignments inside the config remain observable. `unwrapConfig` prefers the reassigned `module.exports` value over the ESM `default` export, mirroring Node's behaviour. An initial sentinel symbol distinguishes "did not touch CJS exports" from "reassigned to `{}`".

The plugin is intentionally narrow:
- Runs only on the resolved `next.config.{ts,mts,cts}` path (gated by extension and by symlink-resolved path comparison).
- Does not touch `next.config.{js,mjs,cjs}` - those continue to be loaded by Vite/`createRequire` as before.
- No behavioural change for ESM-only configs.

### Affected upstream test fixtures

`.nextjs-ref/test/e2e/app-dir/`:
- `next-config-ts-native-ts/import-js-extensions-cjs`
- `next-config-ts-native-mts/{node-api-cjs,import-js-extensions-cjs,nested-imports,...}`
- `next-config-ts/node-api-cjs`

CI run that surfaced these: https://github.com/cloudflare/vinext/actions/runs/25951111875

## Test plan

- [x] Added 4 unit tests in `tests/next-config.test.ts` covering `__dirname`, `__filename`, `require()`, and `module.exports = ...` inside a `next.config.ts`. All four fail on `main` and pass with this fix.
- [x] `pnpm test tests/next-config.test.ts` - 91/91 pass.
- [x] `pnpm test` - full suite, 5493 passed / 4 skipped.
- [x] `pnpm run check` - clean (formatter, lint, types).

### DIFF
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -380,13 +380,23 @@ async function resolveConfigValue(
 }
 
 /**
- * Sentinel marker placed on the injected `module.exports` initial object so we
- * can detect whether a config file reassigned `module.exports = {...}` (CJS
- * style) versus only using `export default` (ESM style). See
- * `cjsGlobalsInjectorPlugin` for how the marker gets attached.
+ * Named export attached by `cjsGlobalsInjectorPlugin` when the source
+ * statically looks like it assigns to `module.exports`. Holds the wrapper
+ * `module` object so {@link unwrapConfig} can read back the user's CJS-style
+ * export. Pure-ESM configs skip the wrapper entirely and rely on the ESM
+ * `default` export instead.
  */
 const VINEXT_CJS_EXPORTS_KEY = "__vinext_cjs_exports";
-const VINEXT_CJS_INITIAL_MARKER = Symbol.for("vinext:cjs-initial");
+
+/**
+ * Companion named export pointing at the initial empty `{}` that the wrapper
+ * is constructed with. Lets {@link unwrapConfig} distinguish "user reassigned
+ * or mutated module.exports" from "module.exports is still the untouched
+ * empty wrapper" — the latter happens when {@link reassignsModuleExports}
+ * matches inside a string or comment (a harmless false positive that should
+ * still fall through to the ESM `default` export).
+ */
+const VINEXT_CJS_INITIAL_KEY = "__vinext_cjs_initial_exports";
 
 /**
  * Unwrap the config value from a loaded module namespace.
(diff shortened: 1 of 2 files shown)


james-elicx:
Pushed two follow-up commits that tighten the injection cost based on a static read of the config source. The original `Module._compile`-mirroring shape is preserved for files that actually need it; everything else gets a cheaper path.

**Why bother**

Every upstream `test/e2e/app-dir/next-config-ts/*` fixture except `node-api-cjs` is pure ESM (`export default { ... }`) — so is nearly every real-world `next.config.ts` I've seen. Running a transform plus tagging the initial exports object with a `Symbol.for("vinext:cjs-initial")` on every load was paying full price for a case almost no user hits.

**Commit 1 — `referencesCjsGlobals`**

Skip the transform entirely when the source contains no bareword match for `__filename` / `__dirname` / `require` / `module` / `exports`. Substring matching is the trade-off:

- False positives (a comment, string literal, or `node:module` import specifier containing the word) cost at most one wasted, idempotent transform.
- False negatives would be a correctness regression, so the regex deliberately errs on the loose side.
- `\b` word boundaries already exclude `requireSomething`, `myModule`, `exporter`, and — importantly — `export default` does not match `\bexports\b`.

**Commit 2 — drop the unconditional sentinel**

Static check `reassignsModuleExports(source)` looks for `module.exports = …`, `module.exports.foo = …`, or `module.exports[…] = …`. Only when that matches do we emit the wrapper `module` object and the named export `unwrapConfig` reads back. Pure-ESM configs return the ESM `default` directly with no wrapper, no symbol-tagged property left on the final exports object.

The sentinel-symbol-on-`module.exports` hack is gone, but I kept a thin referential-identity check as belt-and-braces: if `reassignsModuleExports` gets a false positive (matches `"module.exports = legacy"` inside a string literal), `unwrapConfig` sees that `module.exports` is still the original `{}` reference and falls back to ESM `default` instead of returning an empty object. Test `falls back to ESM default when module.exports reference is only a false positive` pins this behaviour.

**Verification**
- `pnpm test tests/next-config.test.ts` — 108 passed (was 95; +5 helper-unit tests, +1 pure-ESM integration test, +7 module.exports-shape tests).
- `pnpm test` (full suite) — 5534 passed, 4 skipped.
- `pnpm run check` — clean.

PR #1279, opened 2026-05-16
fix(middleware): expose URLPattern global to middleware runtime

james-elicx:
## Summary

Next.js middleware fixtures that import `URLPattern` from `next/server` and call `new URLPattern(...)` at module scope fail to start under the Next.js deploy suite on Node 22 with:

```
TypeError: URLPattern is not a constructor
   at <tmp>/.next/server/middleware.js
```

Node 22 (vinext's minimum supported version) does not expose `URLPattern` globally — the global only appeared as `--experimental-url-pattern` in Node 24+. The previous shim in `packages/vinext/src/shims/server.ts` fell through to a plain arrow function, which is not constructible.

## Fix

- Add `urlpattern-polyfill` (^10.1.0) as a runtime dependency of `vinext`.
- The `next/server` shim now resolves `URLPattern` to the polyfilled class when the host runtime lacks the global, and installs it on `globalThis.URLPattern` so user code referencing the bare global keeps working.
- Cloudflare Workers and Node 24+ continue to use the platform implementation (the polyfill is only used as a fallback).

Next.js's own re-export does **not** polyfill — its middleware runs inside the edge runtime that already provides `URLPattern`:

https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/spec-extension/url-pattern.ts

vinext runs middleware on the host JS runtime (Node or Workers), so it has to polyfill itself when the host lacks it.

## Affected upstream tests

- [`test/e2e/middleware-general/app/middleware.js`](https://github.com/vercel/next.js/blob/canary/test/e2e/middleware-general/app/middleware.js) — imports `URLPattern` from `next/server` and constructs it at module top-level.
- [`test/e2e/middleware-trailing-slash/app/middleware.js`](https://github.com/vercel/next.js/blob/canary/test/e2e/middleware-trailing-slash/app/middleware.js) — same pattern.

## Reference

- Failing CI run: https://github.com/cloudflare/vinext/actions/runs/25951111875

## Test plan

- [x] Added `tests/pages-router.test.ts > middleware that constructs URLPattern at module scope builds and runs` which writes a middleware identical in shape to the Next.js fixtures, runs `vinext build`, imports the bundled entry, and asserts the middleware executes.
- [x] Updated `tests/shims.test.ts` URLPattern test to assert the shim is always constructible (regression on Node 22 would have been silent because the global is available on Node 24 where the CI test machine runs).
- [x] `pnpm test tests/shims.test.ts tests/pages-router.test.ts` — 1111 passed.
- [x] `pnpm test tests/deploy.test.ts tests/prod-server-logs.test.ts tests/app-post-middleware-context.test.ts` — 228 passed.
- [x] `pnpm run check` — format/lint/types clean.
- [ ] Re-run the Next.js Deploy Suite against the middleware-general / middleware-trailing-slash suites once this branch is available.

### DIFF
--- a/packages/vinext/package.json
+++ b/packages/vinext/package.json
@@ -69,6 +69,7 @@
     "image-size": "catalog:",
     "ipaddr.js": "catalog:",
     "magic-string": "catalog:",
+    "urlpattern-polyfill": "catalog:",
     "vite-plugin-commonjs": "catalog:",
     "vite-tsconfig-paths": "catalog:",
     "web-vitals": "catalog:"
--- a/packages/vinext/src/shims/server.ts
+++ b/packages/vinext/src/shims/server.ts
@@ -9,6 +9,7 @@
  * rather than bug-for-bug parity with Next.js internals.
  */
 
+import { URLPattern as URLPatternPolyfill } from "urlpattern-polyfill/urlpattern";
 import {
   MIDDLEWARE_NEXT_HEADER,
   MIDDLEWARE_REWRITE_HEADER,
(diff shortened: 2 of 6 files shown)


james-elicx:
Superseded by #1283 — going with a Node 24 bump instead of carrying a polyfill dep. Closing this in favour of that.

PR #1285, opened 2026-05-17
fix(server): clear browser globals before SSR user modules

### LINKED ISSUE #1222: vinext SSR provides a partial `window` global, breaking `typeof window === 'undefined'` checks
Under `vinext dev`, the SSR module runner exposes a `window` global, but it's a partial polyfill — `typeof window` returns `'object'` and `typeof document` returns `'object'`, while most DOM APIs (e.g. `window.getComputedStyle`, `window.history`) are undefined.

This breaks **every place that uses the canonical SSR-detection idiom**:

```js
if (typeof window === 'undefined') return /* SSR fallback */;
```

The check passes, the code proceeds as if it were on the client, then crashes on the next missing DOM call.

Two concrete crashes already observed in a single app, one in a third-party library, one in vinext's own shims.

## Versions

- vinext: 0.0.50
- @cloudflare/vite-plugin: 1.37.0
- @vitejs/plugin-rsc: 0.5.26
- vite: 8.0.13
- react / react-dom: 19.2.6

## Crash #1 — third-party library (`sonner`)

```
[vite] Internal server error: window.getComputedStyle is not a function
    at getDocumentDirection (.../sonner/dist/index.mjs?v=8096fa8a:831:23)
    at Toaster (.../sonner/dist/index.mjs?v=8096fa8a:920:167)
    at Object.react_stack_bottom_frame (.../react-dom-server.edge.development...)
    at renderWithHooks
    at renderElement
    ...
```

Sonner's source ([sonner/dist/index.mjs:826](node_modules/sonner/dist/index.mjs:826)):

```js
function getDocumentDirection() {
    if (typeof window === 'undefined') return 'ltr';
    if (typeof document === 'undefined') return 'ltr';
    const dirAttribute = document.documentElement.getAttribute('dir');
    if (dirAttribute === 'auto' || !dirAttribute) {
        return window.getComputedStyle(document.documentElement).direction;
    }
    return dirAttribute;
}
```

Both `typeof` guards pass under vinext+workerd SSR. The `getComputedStyle` call then throws.

### Minimal repro

```tsx
// app/providers.tsx
'use client';
import { Toaster } from 'sonner';
export default function Providers({ children }: { children: React.ReactNode }) {
  return (<><Toaster />{children}</>);
}
```

```tsx
// app/layout.tsx
import Providers from './providers';
export default function RootLayout({ children }: { children: React.ReactNode }) {
  return (
    <html lang="en"><body><Providers>{children}</Providers></body></html>
  );
}
```

Run `vinext dev`, hit `/` → 500. (Intermittent — triggers when a toast is queued during render.)

## Crash #2 — vinext's own shims

Same root cause, hits user code without any third-party dep involved:

```
[vite] Internal server error: Cannot read properties of undefined (reading 'pushState')
    at runInRunnerObject (workers/runner-worker/index.js:107:3)
    at Object.createHtmlResponse (.../vinext/src/server/app-page-boundary-render.ts:254:26)
    at buildAppPageSpecialErrorResponse (.../vinext/src/server/app-page-execution.ts:221:30)
    at probeAppPageLayouts (.../vinext/src/server/app-page-execution.ts:305:12)
    at probeAppPageBeforeRender
    at renderAppPageLifecycle
    at dispatchAppPage
    at route (.../vinext/src/server/app-rsc-handler.ts:582:24)
    at handleRequest (.../vinext/src/server/app-router-entry.ts:85:18)
```

The unguarded `window.history.pushState` calls are in vinext's own shims:

- [vinext/dist/shims/router.js:404, 415, 607, 617](node_modules/vinext/dist/shims/router.js:404)
- [vinext/dist/shims/link.js:284](node_modules/vinext/dist/shims/link.js:284)

Other paths in those same files **do** have `typeof window === 'undefined'` guards (link.js:52, 63, 83, 129, 203) — these specific lines were missed.

Symptom: intermittent 500 on app-router pages. Clean dev restart clears it because HMR cache state can push the failing code path into the SSR pre-render lifecycle.

## Root cause (both crashes)

The SSR module runner provides a `window` global that's *not* genuinely absent. I grep'd vinext, `@cloudflare/vite-plugin`, and `@vitejs/plugin-rsc` for explicit `globalThis.window = ...` assignments — found none. The partial `window` is most likely coming from workerd's web-compat globals (workerd exposes `WebSocket`, `EventTarget`, `fetch`, etc., and very likely aliases `window = globalThis` somewhere in its compat layer).

Net effect: the `typeof window === 'undefined'` short-circuit that's load-bearing across the React ecosystem doesn't fire.

## Why this matters

`typeof window === 'undefined'` is the **dominant** SSR-detection idiom in the React ecosystem. Libraries that use it include:

- sonner (Crash #1)
- vinext's own `next/router` / `next/link` shims (Crash #2 — partially, hence the missed-guard bug)
- react-aria-components
- @floating-ui/react
- All major date pickers
- Animation libs (Framer Motion, Motion, etc.)
- Most components that touch DOM measurements

Under vinext, any of these can hit the same class of crash. We hit two in one app within an hour — the next user will hit a third.

The React team has spent years convincing libraries to use this pattern over feature-detection alternatives. The runtime should respect it.

## Asks (ordered by leverage)

1. **Make `typeof window === 'undefined'` truthy in the SSR module runner.** At runner-worker boot, `delete globalThis.window` (or whatever specific path produces the partial polyfill). One change, fixes every existing and future occurrence of this class of bug — third-party libs *and* the missed shims in vinext itself.

2. **As a secondary safety net**, add the missing guards to [router.js:404, 415, 607, 617](node_modules/vinext/dist/shims/router.js:404) and [link.js:284](node_modules/vinext/dist/shims/link.js:284). 5-line change. Worth doing regardless of (1) since other vinext code might add similar lines later.

3. **If a full `window` is intentional**, ensure it provides the full DOM API surface (effectively a jsdom-level polyfill). Heavy and almost certainly not the right tradeoff. Listed only for completeness.

## Workaround for affected users today

Each affected library needs to be told *not* to do the unsafe DOM call:

- **sonner**: pass `dir="ltr"` to `Toaster` → skips `getDocumentDirection()`
- **vinext router/link**: clean restart clears it, but the underlying code path is still unsafe
- **Other libs**: generally `dynamic(() => import(...), { ssr: false })` or a mounted-flag wrapper

Each workaround is per-library and doesn't scale. (1) above is the only fix that does.

NathanDrake2406:
## Overview

| Field | Detail |
| --- | --- |
| Goal | Preserve Next.js SSR semantics when the underlying server runtime exposes partial browser globals. |
| Core change | Add a shared server-global bootstrap that clears `window` and `document` before SSR user modules evaluate, while preserving the existing `AsyncLocalStorage` edge global. |
| Primary review path | `packages/vinext/src/server/server-globals.ts`, generated App/Pages entry wiring, and the focused regression tests. |
| Expected impact | `typeof window` and `typeof document` guards behave like Next.js during server module evaluation. |

Fixes https://github.com/cloudflare/vinext/issues/1222

## Why

Server-rendered Next.js code relies on browser globals being absent. If a runtime provides a partial `window` or `document`, user module evaluation can incorrectly enter client-only branches and crash before vinext gets to render or route.

| --- | --- | --- |
| SSR globals | Server code must observe browser globals as absent. | Deletes, undefines, or shadows `window` and `document`, and fails fast if a non-removable runtime global remains visible. |
| Entry ordering | Global correction must happen before user modules evaluate. | Imports the bootstrap before generated App Router and Pages Router user imports. |
| Edge compatibility | Existing edge global behavior should remain available. | Keeps `installEdgeGlobals()` as a compatibility alias and still installs global `AsyncLocalStorage`. |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| Runtime exposes partial `window` | `typeof window !== "undefined"` branches can run during SSR and touch missing browser APIs. | `window` is removed or shadowed before user modules evaluate. |
| Runtime exposes partial `document` | Server code can misclassify itself as browser code. | `document` is removed or shadowed before user modules evaluate. |
| App and Pages generated entries | User imports could appear before server-global setup. | Template tests assert the bootstrap import comes first. |
| App Router dev route | No high-level guard proved module-evaluation behavior through the dev server. | A fixture route throws on leaked browser globals and is fetched through the normal fixture server. |

<details>
<summary>Validation</summary>

- `vp check packages/vinext/src/server/server-globals.ts tests/edge-globals.test.ts tests/server-globals.test.ts tests/fixtures/app-basic/app/nextjs-compat/typeof-window-ssr/page.tsx`
- `vp test run tests/edge-globals.test.ts tests/server-globals.test.ts tests/entry-templates.test.ts`
- `vp run vinext#build`
- Commit hook: checked-file formatting/type/lint, `knip --no-progress`, and rebuilt `packages/vinext/dist`

</details>

<details>

- Public API: no new public API. `installEdgeGlobals()` remains exported.
- Runtime: intentionally changes server runtime global visibility for `window` and `document` to match Next.js SSR expectations.
- Compatibility: if a runtime exposes a non-removable browser global that remains visible, vinext now throws a targeted error instead of allowing silent SSR misclassification.

</details>

## References

| --- | --- |
| [Next.js `typeof-window` App Router fixture](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/typeof-window/app/page.js) | Shows a client import guarded by `typeof window !== "undefined"` must not run during server render. |
| [Next.js `typeof-window` e2e test](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/typeof-window/typeof-window.test.ts) | Confirms the guarded App Router page renders successfully. |
| [Next.js production `typeof window` replacement test](https://github.com/vercel/next.js/blob/canary/test/production/typeof-window-replace/typeof-window-replace.test.ts) | Confirms server bundles treat `typeof window` as `"undefined"` and client bundles as `"object"`. |
| [Next.js Babel loader config](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/babel/loader/get-config.ts) | Shows Next.js rewrites `typeof window` according to server/client compilation target. |


packages/vinext/src/server/edge-globals.ts
james-elicx, line 1: if we're replacing this file can we remove it?

packages/vinext/src/server/server-globals.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (typeof Reflect.get(globalThis, "AsyncLocalStorage") === "undefined") {
+    Object.defineProperty(globalThis, "AsyncLocalStorage", {
+      configurable: true,
+      value: AsyncLocalStorage,
+      writable: true,
+    });
+  }
+}
+
+installServerGlobals();
```
james-elicx, line 55: some usages call the function themselves, while others dont - can we align on either all consumers calling, or none? all consumers calling would probably be better as it would provide more explicitness about what's happening, rather than it happening here

PR #1296, opened 2026-05-18
fix(config): recognize next.config.mts and interop CJS imports in config loader

james-elicx:
## Summary

vinext's `next.config.ts` loader silently failed for two categories of inputs from the Next.js `test/e2e/app-dir/next-config-ts-native-mts/` test family:

1. **`next.config.mts` was not recognized at all.** The loader's local `CONFIG_FILES` list omitted `next.config.mts`, so `findNextConfigPath` returned `null` and `loadNextConfig` silently returned `null` — manifesting downstream as missing config (no `env`, no `basePath`, etc.).
2. **Imports of `.cjs` / `.cts` files from the config failed with `module is not defined`.** Vite's `runnerImport` evaluates every module as ESM, so a nested `import x from './foo.cjs'` where `foo.cjs` uses `module.exports = …` crashes the first time it hits that statement.

The canonical `CONFIG_FILES` constant in `shims/constants.ts` already mirrors Next.js's accepted set (`.js`/`.mjs`/`.ts`/`.mts` when `process.features.typescript` is set), but `config/next-config.ts` had its own divergent copy. Same issue in `deploy.ts` and `check.ts`.

## Changes

1. **Extend the loader's local `CONFIG_FILES` to include `next.config.mts`.** Keeps `.cjs` for backwards-compat with existing vinext users (Next.js itself does not accept it; `findNextConfigPath` returns the first match so the extra entry is harmless). Same extension in `deploy.ts` and `check.ts`.

2. **Add `cjsImportsInteropPlugin`.** A new Vite plugin installed on the inline `runnerImport` instance for config loading only. It wraps `.cjs`/`.cts` files (unconditionally CJS by spec) and `.js`/`.ts` files that statically reassign `module.exports` so that `module.exports` becomes the ESM default export. This mirrors how Next.js's SWC pipeline auto-converts CJS imports brought in from `next.config.ts`. The plugin is scoped to the config-load Vite instance, so it never affects the user's application module graph.

## Tests

New file: `tests/next-config-extensions.test.ts`. Four cases ported from:

- `.nextjs-ref/test/e2e/app-dir/next-config-ts-native-mts/nested-imports/` (×2 — `.ts` and `.mts` config)
- `.nextjs-ref/test/e2e/app-dir/next-config-ts-native-mts/import-js-extensions-{cjs,esm}/`
- `.nextjs-ref/test/e2e/app-dir/next-config-ts-native-mts/import-from-node-modules/`

All four fail on `main` (the first two with silent-null, the third/fourth with `module is not defined` for `.cjs`/`.cts` imports). All four pass on this branch.

## Verification

```
vp test run tests/next-config-extensions.test.ts          # 4 passed
vp test run tests/next-config.test.ts                     # 108 passed (regression check)
vp test run tests/shims.test.ts                           # all green (regression check)
vp test run tests/deploy.test.ts tests/features.test.ts   # all green
vp test run tests/check.test.ts                           # 97 passed
vp check                                                  # clean
```

## Prior work

This continues the line of small loader-parity fixes:

- #1211 — jiti injection for TS configs
- #1213 — missing optional deps
- #1215 — tsconfig `compilerOptions.paths`
- #1275 — CJS `next.config.js` under `"type": "module"`
- #1278 — CJS globals in `next.config.ts`

None of the above addressed extension probing or the `.cjs`/`.cts` interop path.

### DIFF
--- a/packages/vinext/src/check.ts
+++ b/packages/vinext/src/check.ts
@@ -383,7 +383,16 @@ export function scanImports(root: string): CheckItem[] {
  * Analyze next.config.js/mjs/ts for supported and unsupported options.
  */
 export function analyzeConfig(root: string): CheckItem[] {
-  const configFiles = ["next.config.ts", "next.config.mjs", "next.config.js", "next.config.cjs"];
+  // Mirror the Next.js-compatible set in shims/constants.ts. Accepts both
+  // `.ts`/`.mts` (Next.js-recognized) and `.cjs`/`.cts` (defensive — Next.js
+  // does not, but if a user has them we should still scan and report).
+  const configFiles = [
+    "next.config.ts",
+    "next.config.mts",
+    "next.config.mjs",
+    "next.config.js",
+    "next.config.cjs",
+  ];
   let configPath: string | null = null;
   for (const f of configFiles) {
     const p = path.join(root, f);
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -312,7 +312,18 @@ export type ResolvedNextConfig = {
   sassOptions: Record<string, unknown> | null;
 };
 
-const CONFIG_FILES = ["next.config.ts", "next.config.mjs", "next.config.js", "next.config.cjs"];
+// Mirrors Next.js's accepted set in packages/next/src/shared/lib/constants.ts
+// (`.js`/`.mjs`/`.ts`/`.mts`) and adds `.cjs` for parity with vinext's own
+// loader, which has historically accepted CJS configs as well. The order is
+// significant: findNextConfigPath returns the first match, so prefer the more
+// modern flavours first.
+const CONFIG_FILES = [
+  "next.config.ts",
+  "next.config.mts",
+  "next.config.mjs",
+  "next.config.js",
+  "next.config.cjs",
+];
 const DEFAULT_EXPIRE_TIME = 31_536_000;
 
 /**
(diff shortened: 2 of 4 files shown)


james-elicx:
Thanks for the review. Addressed item 1 (precise CJS `exports`/`module.exports` comment) in 0dfd600. Rebased onto current main while I was here. Item 2 (.cts omission) is intentional as you noted — no change. Re-requesting review.

james-elicx:
Good call — replaced `cjsImportsInteropPlugin` with `vite-plugin-commonjs` in cf20845. Dropped the bespoke plugin entirely; net -59 lines. Two small config knobs were needed for parity: (1) `resolve.extensions` explicitly includes `.cjs`/`.cts` (the plugin keys off it and Vite's default list omits both), and (2) a `filter` that excludes the config file itself (to avoid double `module` injection with `cjsGlobalsInjectorPlugin`) and opts back in for `node_modules` (which `commonjs()` skips by default — needed for bare-import CJS plugin wrappers). All 112 tests in `tests/next-config*.test.ts` still pass; broader suites green (1637 passing).

/bigbonk review

james-elicx:
Re-verified the refactor against the originating Next.js fixtures end-to-end: **12/12 passed**. No regressions vs the original `cjsImportsInteropPlugin` approach.

Ran the full `next-config-ts-native-{mts,ts}` × `{nested-imports,import-js-extensions,import-from-node-modules}` × `{cjs,esm}` matrix against the `vite-plugin-commonjs`-based loader (cf20845) via `run-tests.js` against the prepared Next.js canary checkout. Each fixture builds, prerenders, and serves through real vinext; sample direct-jest run confirms the assertion executes:

```
PASS Turbopack test/e2e/app-dir/next-config-ts-native-mts/import-from-node-modules/.../next-config-ts-import-from-node-modules-cjs.test.ts (11.769 s)
  next-config-ts-import-from-node-modules-cjs
    ✓ should import from node_modules (CJS) (10 ms)
```

| Family | nested-imports cjs | nested-imports esm | import-js-extensions cjs | import-js-extensions esm | import-from-node-modules cjs | import-from-node-modules esm |
| --- | --- | --- | --- | --- | --- | --- |
| `next-config-ts-native-mts` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `next-config-ts-native-ts` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |

(Each fixture runs ~13–20s end-to-end including a fresh `vinext build` and prerender.)

PR #1299, opened 2026-05-18
fix(prerender): inject CJS-globals banner into ESM server bundles

james-elicx:
## Summary

Inject a small ESM-compatible banner into the RSC, SSR, and Pages Router server bundles that synthesises the missing CJS globals (`__filename`, `__dirname`, `require`) from `import.meta.url`. This fixes `ReferenceError: __filename is not defined in ES module scope` during the prerender phase of `vinext build` whenever the bundled app pulls in a CJS-style package that reads those globals at module load time (sqlite3, typescript, graceful-fs, node-pre-gyp, …).

## Root cause

`vinext init` adds `"type": "module"` to the host project's `package.json`. Node then evaluates every `.js` file under `dist/server/` as ESM, which doesn't expose the CJS globals. vinext bundles user code with `noExternal: true` by default, so any transitive CJS dep that touches `__filename` at module-load time crashes when the bundle is dynamically imported by the prerender phase.

Repro path from `shard-logs/10.log`:

```
✓ built in 98ms
  Pre-rendering all routes...
[ReferenceError: __filename is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension
and '/tmp/next-test-…/package.json' contains "type": "module".]
```

PR #1217 (CJS→ESM converter for `next.config.{js,ts}`) and PR #1278 (`cjsGlobalsInjectorPlugin` for `next.config.${ts,cts,mts}`) only target the config-load path. Neither helps when a transitive CJS dep ends up inside the bundled server output. PR #1275 superseded the converter script with a `.cjs` temp-copy strategy for plain CJS configs — that still doesn't cover bundled user/app code.

## What this PR does

1. Adds `packages/vinext/src/build/server-build-config.ts` exporting:
   - `cjsGlobalsBanner()` — returns the canonical banner string with a marker comment for idempotency.
   - `isCjsGlobalsBanner(s)` — detects a previously-injected banner.
   - `mergeBanner(existing)` — composes the shim with user-provided banners (string or Rolldown `AddonFunction`).
   - `withCjsGlobalsBanner(bundlerOptions)` — non-mutating helper that returns a new bundler-options object with `output.banner` set; handles both single `output` and array `output` shapes.

2. Wires `withCjsGlobalsBanner` into the four server-bundle output sites in `packages/vinext/src/index.ts` (App Router RSC, App Router SSR, Pages Router server) and `packages/vinext/src/cli.ts` (hybrid Pages bundle).

3. Skips the banner for Cloudflare Workers and Nitro deploy targets — those adapters handle runtime globals separately, and Workers don't have `__filename` semantics at all.

## Tests

`tests/server-bundle-cjs-globals.test.ts` covers:
- Shape of the banner string (CJS bindings, aliased imports, marker).
- `isCjsGlobalsBanner` round-trip.
- Banner appears in RSC, SSR, and Pages Router env build outputs.
- Banner is absent from client builds.
- User-provided `output.banner` is composed, not clobbered.
- Calling the config hook twice does not stack the banner.
- **Integration**: builds a synthetic App Router fixture whose `app/page.tsx` imports a server-only helper that references `String(__filename)` / `String(__dirname)` / `typeof require === "function"` at module scope, then dynamically imports the built `dist/server/index.js` — without the banner this throws `ReferenceError: __filename is not defined`, with the banner the import succeeds.

Verified RED→GREEN: stashing the source change makes the integration test fail with the expected `ReferenceError`; unstashing makes all 10 tests pass.

## Validation

```
vp check tests/server-bundle-cjs-globals.test.ts \
        packages/vinext/src/build/server-build-config.ts \
        packages/vinext/src/index.ts packages/vinext/src/cli.ts
# pass: formatted, no lint/type errors

vp test run tests/server-bundle-cjs-globals.test.ts
# 10 / 10 passed

vp test run tests/build-optimization.test.ts tests/prerender.test.ts \
            tests/deploy.test.ts tests/pages-router.test.ts \
            tests/app-router.test.ts tests/standalone-build.test.ts \
            tests/intercepting-routes-build.test.ts
# 826 / 826 passed (no regressions)

vp check
# pass: all 1303 files formatted, 589 files lint/type-clean
```

## Originating fixtures

The error pattern shows up across multiple Next.js deploy-suite fixtures whose user code (transitively) pulls a CJS native-addon loader into the server bundle, e.g.:

- `test/e2e/app-dir/turbopack-reports` — imports `sqlite3` from a server component.
- `test/e2e/app-dir/asset-prefix-absolute-no-path` — referenced in the failure analysis.

Both should now reach the prerender phase without the `__filename` crash; subsequent failures (if any) would be category-A1/A2 fixture-config issues handled by the other deploy-suite agents.

### DIFF
--- a/packages/vinext/src/build/server-build-config.ts
+++ b/packages/vinext/src/build/server-build-config.ts
@@ -1,0 +1,149 @@
+/**
+ * Server bundle build config helpers.
+ *
+ * vinext emits ESM server bundles (`dist/server/index.js`,
+ * `dist/server/ssr/index.js`, `dist/server/entry.js`) because the host project
+ * is initialised with `"type": "module"` (added by `vinext init`). Node's ESM
+ * loader does not expose the CommonJS globals `__filename`, `__dirname`, or
+ * `require`, so any third-party package whose code is bundled into the server
+ * entry and which reads these at module-load time (sqlite3, typescript,
+ * graceful-fs, node-pre-gyp, etc.) throws `ReferenceError: __filename is not
+ * defined in ES module scope` the first time the bundle is imported. That
+ * typically surfaces during the prerender phase of `vinext build`, which is
+ * where the freshly-built bundle is loaded.
+ *
+ * Next.js avoids this by emitting CommonJS for its server runtime; vinext
+ * can't take that shortcut because the project's `"type": "module"` makes
+ * every `.js` file under `dist/server/` ESM by default. Renaming outputs to
+ * `.cjs` would force a CJS evaluator but would also break:
+ *   - Vite plugins that hard-code `index.js` (plugin-rsc emits to `index.js`)
+ *   - `@cloudflare/vite-plugin`, which expects an ESM worker entry
+ *   - Top-level `await` inside the bundle (Vite uses it for dynamic imports)
+ *
+ * Instead, prepend a small banner that synthesises the missing CJS globals
+ * from `import.meta.url`. Rolldown / Rollup write `output.banner` verbatim at
+ * the top of the chunk, so the bindings live in the module's top-level scope
+ * and shadow Node's missing globals for every nested function in the bundle.
+ *
+ * The banner runs in O(1) per build, has zero impact on bundle size in
+ * production (~250 bytes pre-minification, ~0 bytes after gzip with any
+ * realistic server bundle), and is idempotent — repeated injection is
+ * de-duplicated by {@link mergeBanner}.
+ */
+
+/**
+ * Marker string embedded in the banner so {@link mergeBanner} and tests can
+ * detect a previously-injected vinext banner. Kept short and unique to avoid
+ * accidental collisions with user banners (license headers, etc.).
+ */
+const BANNER_MARKER = "// vinext:cjs-globals-banner";
+
+/**
+ * Return the ESM-compatible CJS-globals banner string to prepend to server
+ * bundles. The banner is a single multi-line block; callers paste it verbatim
+ * into Rolldown / Rollup's `output.banner` option.
+ *
+ * The synthesised bindings exactly match Node's CJS wrapper semantics:
+ *   - `__filename` — absolute path of the current bundle file
+ *   - `__dirname`  — directory of the current bundle file
+ *   - `require`    — a `createRequire`-scoped resolver bound to the bundle URL
+ *
+ * The named imports are aliased (`createRequire as __vinext_createRequire`,
+ * etc.) so user code that imports the same names from `node:module` /
+ * `node:url` / `node:path` is unaffected. Only `__filename`, `__dirname`, and
+ * `require` are introduced into the top-level scope — those are intentional
+ * since they replace the missing CJS globals.
+ *
+ * The marker comment on the first line lets us detect a previously-injected
+ * banner and skip re-injection (see {@link mergeBanner}).
+ */
+export function cjsGlobalsBanner(): string {
+  return (
+    `${BANNER_MARKER}\n` +
+    `import { createRequire as __vinext_createRequire } from "node:module";\n` +
+    `import { fileURLToPath as __vinext_fileURLToPath } from "node:url";\n` +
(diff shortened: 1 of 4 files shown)


james-elicx:
Audited this PR against the just-merged #1296 (which adopted `vite-plugin-commonjs` for the `next.config` loader's CJS sibling-imports problem) to check whether either of the simplifications suggested in review would apply here:

> Could we (a) make the offending packages external so they're loaded by Node CJS, or (b) use `vite-plugin-commonjs` during the server bundle build to translate `__filename` references into `import.meta.url` shims at bundle time?

**Verdict: keep banner as-is.** Both alternatives are unworkable; banner is the right shape. Rebased onto main (clean — no conflicts) and posting this comment so future maintainers don't re-litigate.

## Why `vite-plugin-commonjs` doesn't help here

It's the wrong tool for this problem. Reading its source and README:

> ✅ alias
> ✅ bare module(node_modules)
> ✅ dynamic-require similar to webpack
> ✅ require statement
> ✅ module.exports

It transforms `require()` / `module.exports` into ESM equivalents. It does **not** synthesise `__filename` / `__dirname` bindings — those references survive its transform untouched. Grepped its dist for `__filename` and `__dirname`: zero hits. So the banner would still be needed even with `vite-plugin-commonjs` applied to the server bundle.

(`@rollup/plugin-commonjs` *does* rewrite `__filename`, but it's not in the dep tree and adopting it for the server bundle would be a much bigger refactor than this PR — vinext uses Rolldown for output and `vite-plugin-commonjs` for transforms.)

## Why externalizing the offending packages doesn't work

1. The set is unbounded — `typescript`, `sqlite3`, `better-sqlite3`, `graceful-fs`, `node-pre-gyp`, `sharp`, `fsevents`, and many more reference `__filename` / `__dirname` at module load. A maintained allowlist would be brittle.
2. `noExternal: true` is **intentional** (`packages/vinext/src/index.ts:1316`): vinext wants non-JS imports (CSS, images) from `node_modules` to flow through Vite's transform pipeline rather than Node's native ESM loader. Selectively externalizing problem packages partially defeats that.
3. The primary deploy target is **Cloudflare Workers**, which has no Node CJS runtime. Externalizing would either fail at build time (when bundling for Workers) or push the problem into the Worker bundle in a worse shape.
4. The PR already correctly gates the banner to plain Node builds only (`hasCloudflarePlugin || hasNitroPlugin ? ... : withCjsGlobalsBanner(...)` at index.ts:1528, 1535, 1545+) — for Workers, the banner is intentionally skipped because V8 isolates don't expose `__filename` and the offending references never reach runtime.

## Banner is well-established prior art

- esbuild has `--banner:js` for exactly this case
- Rollup / Rolldown's `output.banner` is the documented escape hatch
- Vite recipes recommend this for ESM bundles bundling CJS code

## Rebase

Cleanly rebased onto main. No conflicts. All 10 tests still pass:

```
Test Files  1 passed (1)
Tests       10 passed (10)
```

The integration test ('imports a built App Router server bundle that references __filename') is the critical case — it builds a real RSC bundle with a server-only helper that references `__filename` at the top level, dynamic-imports the built bundle, and proves Node can load it. Without the banner this throws `ReferenceError: __filename is not defined in ES module scope`.

## Relationship to #1296 (for future maintainers)

These two PRs sound related ("both about CJS in vinext") but address fundamentally different layers:

- **#1296** — Handles CJS code at **transform/eval time** inside Vite's module runner, for the `next.config.*` loader. Uses `vite-plugin-commonjs` to translate `require()` / `module.exports` shapes into ESM before Vite evaluates them.
- **#1299** (this PR) — Handles CJS code at **runtime** in the **built bundle output**. The output is ESM but transitively-bundled CJS source references `__filename` / `__dirname` / `require` as bare identifiers; the banner re-creates them from `import.meta.url` at the top of the bundle file.

`vite-plugin-commonjs` operates in the first layer and does not synthesise the globals required by the second, which is why both PRs ship side-by-side.

PR #1249, opened 2026-05-16
feat(router): promote intercepted preservation through planner

NathanDrake2406:
## Overview

| Item | Details |
| --- | --- |
| Goal | Implement `#726-CORE-15` from #726 by promoting intercepted route preservation into the planner and lifecycle path. |
| Core change | Interception matching and wire metadata now carry proof, but the planner decides whether source UI and unrelated slots may remain visible. |
| Main boundary | The AppElements payload is transport, not authority. No proof means no reuse, no skip, and no visible commit. |
| Primary files | `navigation-planner.ts`, `app-browser-state.ts`, `app-browser-entry.ts`, `app-elements-wire.ts`, app page dispatch/build wiring, and the intercepted route tests. |
| Expected impact | Soft intercepted navigation keeps proven source UI and unrelated slots, while direct load and refresh resolve to the normal target page unless the visible world proves interception is current. |

Bonk reviewer note: please read #726 before reviewing this PR so the Core-15 slice is evaluated in the architectural big picture.

## Why

This implements `#726-CORE-15` by promoting intercepted route preservation through the planner/lifecycle path. The invariant is that route state and lifecycle authority decide what previous UI may remain visible. Context suffixes, wire-key suffixes, mounted-slot headers, cache partitioning, and missing payload entries are evidence only.

| Area | Principle / invariant | What this PR changes |
| --- | --- | --- |
| Route matcher | Finds a possible intercepted render target | Requires a matching source route and carries the target slot id as evidence. |
| Wire payload | Transports render data and metadata | Adds explicit interception proof metadata without making it commit authority. |
| Planner | Owns visible-world preservation decisions | Approves current-root intercepted preservation only when source, target, root, and slot proof line up. |
| Lifecycle | Rejects stale or wrong-context results | Clears stale same-URL interception state when normal target payloads commit and syncs current-context action/refresh state. |
| Visible commit | Applies the approved world change | Preserves proven source route and unrelated slots while replacing the intercepted slot. |

## What Changed

| Scenario | Before | After |
| --- | --- | --- |
| Soft intercepted navigation | Preservation could fall out of context-bearing route ids, missing entries, or slot overrides. | The planner approves preservation from explicit source and slot proof. |
| Missing or stale proof | Legacy interception context could still influence preservation. | The planner rejects preservation and falls back to hard navigation. |
| Direct load and refresh | Stale history context could be treated as current interception context. | Direct target payloads clear previous-next state and render the normal target page. |
[...]

PR #1302, opened 2026-05-18
fix(router): gate intercept matching on Next-URL source pattern

james-elicx:
## Overview

| Item | Details |
| --- | --- |
| Goal | Restrict App Router interception matching to navigation contexts whose `Next-URL` header actually satisfies the intercepting-route prefix, matching Next.js' rewrite semantics. |
| Core change | The route matcher now tracks each intercept's *intercepting route* URL pattern (\`sourceMatchPattern\`) and refuses to fire when the request's source pathname does not match (or is absent). |
| Primary files | \`routing/app-route-graph.ts\`, \`server/app-rsc-route-matching.ts\`, \`entries/app-rsc-manifest.ts\`, plus matching unit + integration tests. |
| Expected impact | Direct \`.rsc\` deep-loads, refreshes, and unrelated cross-route navigations now render the underlying page instead of the intercept modal. Soft client navigation from inside the intercepting route is unchanged. |

## Why

Vinext was modeling interception as \"if the URL matches the target pattern, render the modal\". Next.js models it as a rewrite gated on the Next-URL header (\`^<interceptingRoute>(?:/.*)?$\`), so the modal only ever renders when the navigator was already inside the slot's owning route. This PR brings the matcher into parity with that contract.

Categories addressed in the Next.js deploy suite:

- \`parallel-routes-and-interception-basepath\` — \`@slot/(.)nested\` no longer mismatches when the source is outside root.
- \`parallel-routes-and-interception-from-root\` — \`@modal/(...)[locale]/intercepted\` requires \`/[locale]/example\` as the source.
- \`parallel-routes-catchall-specificity\` — modal slot only attaches when the user came from a covered route.
- \`interception-dynamic-single-segment\` — direct deep links to deeply nested settings pages now render the full page, intercepts only fire from sibling navigations.

The remaining deploy-suite cases (\`interception-segments-two-levels-above\`, \`interception-routes-multiple-catchall\`) use interception markers placed *outside* a \`@slot\` directory. Vinext does not yet treat those as intercepting routes — fixing that requires scanner-level synthesis of an implicit \"children\" slot and is left for a follow-up PR.

## What Changed

| Layer | Before | After |
| --- | --- | --- |
| Scanner | Recorded \`convention\`, \`targetPattern\`, layouts, and params. | Also captures \`sourceMatchPattern\` from the slot owner's directory (route groups + \`@slot\` segments stripped, matching \`normalizeAppPath\`). |
| Manifest | Emitted intercept entries without source proof. | Emits \`sourceMatchPattern\` alongside the existing target metadata. |
| Matcher | \`findIntercept\` iterated all entries and returned the first target-pattern match. | \`findIntercept\` returns \`null\` when (a) no source pathname is provided or (b) the source pathname doesn't satisfy the entry's \`sourceMatchPattern\` (with descendants allowed). |
| Tests | Asserted current-buggy matcher behavior. | Asserts the rewrite-gated behavior; adds cross-test coverage for root-level, static-parent, dynamic-parent, two-level, and root-from-nested cases. |

### Behaviors verified

- Soft navigation that supplies \`X-Vinext-Interception-Context\` matching the intercepting route still renders the modal and the underlying source-route tree.
- Direct \`.rsc\` fetches without a Next-URL header render the direct page.
- RSC requests with an unrelated context (e.g. \`/feed\` against \`/team/[teamId]/members\` modal) render the direct page.
- Catch-all and optional catch-all segments inside intercepting routes match descendants correctly.

## References

- Next.js: [\`shared/lib/router/utils/interception-routes.ts\`](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/router/utils/interception-routes.ts) — \`extractInterceptionRouteInformation\` (intercepting route = pre-marker path normalized).
- Next.js: [\`lib/generate-interception-routes-rewrites.ts\`](https://github.com/vercel/next.js/blob/canary/packages/next/src/lib/generate-interception-routes-rewrites.ts) — rewrite source/destination + \`has[Next-URL]\` regex.
- Next.js: [\`lib/generate-interception-routes-rewrites.test.ts\`](https://github.com/vercel/next.js/blob/canary/packages/next/src/lib/generate-interception-routes-rewrites.test.ts) — authoritative table of source/header regex shapes we now mirror at the matcher boundary.
- vinext: PR #1249 (\`feat(router): promote intercepted preservation through planner\`) — orthogonal. It carries interception proof through the planner once the matcher has already fired; this PR ensures the matcher only fires in the right context. The two changes are compatible; #1249 still needs this gate to avoid passing stale-but-target-matching intercepts into the planner.

## Validation

\`\`\`bash
vp check
vp test run tests/app-rsc-route-matching.test.ts tests/app-route-graph.test.ts \\
  tests/intercepting-routes-build.test.ts tests/app-router.test.ts \\
  tests/app-page-request.test.ts tests/app-page-dispatch.test.ts \\
  tests/app-server-action-execution.test.ts tests/route-sorting.test.ts \\
  tests/entry-templates.test.ts tests/app-browser-entry.test.ts \\
  tests/app-elements.test.ts tests/app-page-element-builder.test.ts \\
  tests/navigation-planner.test.ts
vp run vinext#build
\`\`\`

All targeted tests pass (682 tests across the listed files).

## Risk / Compatibility

| Surface | Notes |
| --- | --- |
| Public API | None — \`InterceptingRoute\` is an internal manifest type. |
| RSC payload | No change. Intercept metadata in \`__AppElementsWire\` is unchanged. |
| Soft navigation | Unchanged when the client correctly emits \`X-Vinext-Interception-Context\`. Browser-state pathways always include the header on intercepted navigations. |
| Direct deep links | Now correctly render the underlying route instead of the modal — this is a *fix*, but applications that intentionally relied on the buggy behavior will see different RSC output for direct \`.rsc\` deep links. None observed in the in-repo fixtures or examples. |
| Cache keys | Unchanged. Cache keying still uses the explicit interception context. |

## Non-goals / Follow-ups

- Supporting interception markers placed outside \`@slot\` directories (children-slot intercept) for the remaining two deploy-suite categories.
- Parallel-route slot precedence regressions in \`parallel-routes-catchall-dynamic-segment\` (no interception involved; tracked separately).

### DIFF
--- a/packages/vinext/src/entries/app-rsc-manifest.ts
+++ b/packages/vinext/src/entries/app-rsc-manifest.ts
@@ -118,6 +118,7 @@ function buildRouteEntries(routes: AppRoute[], imports: ImportAllocator): string
         (ir) => `        {
           convention: ${JSON.stringify(ir.convention)},
           targetPattern: ${JSON.stringify(ir.targetPattern)},
+          sourceMatchPattern: ${JSON.stringify(ir.sourceMatchPattern)},
           interceptLayouts: [${ir.layoutPaths.map((layoutPath) => imports.getImportVar(layoutPath)).join(", ")}],
           page: ${imports.getImportVar(ir.pagePath)},
           params: ${JSON.stringify(ir.params)},
--- a/packages/vinext/src/routing/app-route-graph.ts
+++ b/packages/vinext/src/routing/app-route-graph.ts
@@ -16,6 +16,22 @@ export type InterceptingRoute = {
   convention: string;
   /** The URL pattern this intercepts (e.g. "/photos/:id") */
   targetPattern: string;
+  /**
+   * URL pattern of the *intercepting route* — the path that owns the slot
+   * containing this interception marker, with route groups and `@slot`
+   * segments stripped. Mirrors Next.js' `interceptingRoute` from
+   * `extractInterceptionRouteInformation`.
+   *
+   * Used at request time to gate `findIntercept` against the Next-URL /
+   * interception-context header: an intercept only fires when the source
+   * pathname matches `^<sourceMatchPattern>(?:/.*)?$`. Without this gate
+   * a direct RSC fetch to the intercept target would render the modal
+   * instead of the underlying page.
+   *
+   * @see https://github.com/vercel/next.js/blob/canary/packages/next/src/lib/generate-interception-routes-rewrites.ts
+   * @see https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/router/utils/interception-routes.ts
+   */
+  sourceMatchPattern: string;
   /** Absolute path to the intercepting page component */
   pagePath: string;
   /** Absolute layout paths inside the intercepting route tree, outermost to innermost */
@@ -1777,7 +1793,10 @@ function scanForInterceptingPages(
       const restOfName = entry.name.slice(interceptMatch.prefix.length);
       const interceptDir = path.join(currentDir, entry.name);
 
-      // Find page files within this intercepting directory tree
+      // Find page files within this intercepting directory tree.
+      // `currentDir` is the *parent* of the marker dir — used by
+      // computeInterceptSourceMatchPattern to derive the intercepting-route
+      // URL (the path that owns the slot containing the marker).
       collectInterceptingPages(
         interceptDir,
         interceptDir,
@@ -1785,6 +1804,7 @@ function scanForInterceptingPages(
         restOfName,
         routeDir,
         appDir,
+        currentDir,
         results,
         matcher,
       );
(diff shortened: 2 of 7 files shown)


james-elicx:
Bumping to retrigger CI (workflow did not fire on initial push).

james-elicx:
Note: GitHub Actions CI workflow has not triggered on any of the pushes to this branch (initial push, two follow-up commits, close/reopen). Only the CodeQL workflow runs. Other PRs from the same branch family (#1295–#1304) all triggered CI normally. Posting /bigbonk for review of the diff; CI status needs a maintainer to look at why workflows aren't dispatching here.

/bigbonk review

PR #1310, opened 2026-05-18
fix(build): pin cssTarget so esbuild preserves `max-width` media-query syntax

james-elicx:
## What

Pin `build.cssTarget` to `["chrome87", "edge88", "firefox78", "safari15"]` in vinext's top-level build config.

## Why

Vite's default `build.cssTarget` follows `build.target` (modern evergreens), which lets esbuild's CSS minifier rewrite `@media (max-width: 768px)` to the Media Queries Level 4 range syntax `@media (width <= 768px)`.

The two forms are semantically equivalent in modern browsers, but the rewrite is observable to:
- user code that inspects `cssText` on `CSSMediaRule` objects
- tools that pattern-match the raw query string
- design-system / CSS-in-JS code that round-trips queries through string operations

Next.js's CSS pipeline (webpack + cssnano, or lightningcss only when opted in via `experimental.useLightningcss`) preserves the original syntax by default. Pages migrated to vinext that relied on this default were silently rewritten.

Pinning `cssTarget` to a baseline that predates Safari 16.4 — the first browser to support MQ Level 4 range syntax — causes esbuild to skip the lowering. Only literal media-query syntax is affected; runtime matching behavior is unchanged.

## Failure this fixes

Found via the Next.js Deploy Suite. The Next.js fixture `test/e2e/app-dir/css-media-query/css-media-query.test.ts` ("should preserve max-width media query syntax instead of transpiling to range syntax") asserts:

```ts
expect(mediaQueryRule).toContain('max-width: 768px')
expect(mediaQueryRule).not.toContain('width <= 768px')
expect(mediaQueryRule).not.toContain('width<=768px')
```

Today vinext outputs `@media screen and (width<=768px)` and the assertion fails with `Received: undefined` (the test's `.find(rule => /@media/.test(rule.cssText) && /max-width/.test(rule.cssText))` returns undefined because the substring no longer appears).

## Tests

- New `tests/css-media-query-target.test.ts` builds a minimal fixture with vinext's production build pipeline, then reads the produced CSS asset and asserts `max-width: 768px` is preserved and `width<=768px` does NOT appear. Verified this test fails without the cssTarget change and passes with it.
- All existing CSS / SCSS / build-optimization / nextjs-compat app-css tests continue to pass.

## Considered alternatives

- **`build.cssMinify: 'lightningcss'`** — would also preserve the syntax (lightningcss is more conservative by default), but vinext already stubs lightningcss out for Workers builds in `deploy.ts:374`. Adopting it as the CSS transformer would conflict with that decision and add a new dependency surface.
- **Disable CSS minification entirely (`cssMinify: false`)** — too aggressive; loses the size win on every other CSS optimisation esbuild does.
- **Per-environment override** — `build.cssTarget` is a top-level config that fans out to all environments naturally, so a single top-level pin is the smallest change.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -1327,6 +1327,27 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
             // the same reason; this mirrors that for `ssr`. Affects only
             // the build pipeline.
             ssrEmitAssets: true,
+            // CSS minification target. Vite/esbuild defaults to the same target
+            // as JS (modern evergreens), which lets esbuild's CSS minifier rewrite
+            // `@media (max-width: 768px)` to the Media Queries Level 4 range
+            // syntax `@media (width <= 768px)`. Both forms are semantically
+            // equivalent in modern browsers, but the rewrite is observable to
+            // user code that inspects `cssText` of `CSSMediaRule`s and breaks
+            // tools that pattern-match the raw query string. Next.js does not
+            // perform this rewrite by default (its webpack/lightningcss CSS
+            // pipeline preserves the original syntax), so user code carried
+            // over from Next.js can break when migrating to vinext.
+            //
+            // Pin a baseline old enough to suppress the rewrite — Safari 15
+            // predates support for CSS MQ Level 4 range syntax (added in
+            // Safari 16.4). esbuild checks the configured target and skips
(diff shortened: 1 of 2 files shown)


james-elicx:
/bigbonk how does this compare against nextjs for parity?

PR #1311, opened 2026-05-18
feat(config): implement assetPrefix

james-elicx:
## What

Implements the Next.js `assetPrefix` config option in vinext. `assetPrefix` prepends a URL (CDN origin or path prefix) to every emitted JS/CSS/image/static asset URL while leaving routing under `basePath` untouched.

Closes the gap noted in the diagnostic: prior to this PR the config loader silently dropped `assetPrefix`, Vite's `base` was derived from `basePath` only, and the Pages/App prod servers had no way to recognise the prefixed URLs.

Reference: https://nextjs.org/docs/app/api-reference/config/next-config-js/assetPrefix
Next.js source: `packages/next/src/lib/load-custom-routes.ts` (`assetPrefix` handling + the automatic `/<prefix>/_next/:path+` rewrite).

## How

Layered approach split across three commits — config + build, runtime serving, tests:

1. **Config (`packages/vinext/src/config/next-config.ts`)** — adds `assetPrefix` to `NextConfig` and `ResolvedNextConfig`. A new exported `normalizeAssetPrefix(value)` helper trims trailing slashes, accepts path prefixes (`/cdn`) and absolute URLs (`https://cdn.example.com[/sub]`), validates parseability, and rejects non-string values.

2. **Asset-prefix helpers (`packages/vinext/src/utils/asset-prefix.ts`)** — new module collects the shared logic so the build plugin and runtime servers stay aligned:
   - `resolveAssetsDir(prefix)` → on-disk layout under `outDir`.
   - `resolveAssetUrlPrefix(prefix)` → emitted URL prefix (always ends with `_next/static/`, matching Next.js's URL convention).
   - `assetPrefixPathname(prefix)` → path portion used for runtime URL matching.
   - `isAbsoluteAssetPrefix(prefix)` → helper for branching.

3. **Vite plugin (`packages/vinext/src/index.ts`)** — when `assetPrefix` is configured:
   - `build.assetsDir` → `<prefix>/_next/static` for path prefixes, `_next/static` for absolute URLs. Disk layout matches the URL so the Cloudflare ASSETS binding and any static file server serve the assets without a runtime rewrite.
   - `experimental.renderBuiltUrl` → overrides Vite's default `assetsURL = base + url` emission so `assetPrefix` URLs are emitted directly. This keeps `basePath` and `assetPrefix` independent — routes anchor at `base = basePath + '/'` as today, asset URLs anchor at the configured asset prefix.

4. **Runtime serving (`packages/vinext/src/server/prod-server.ts`)** — both routers share a new `resolveAppRouterAssetPath(pathname, assetPathPrefix, assetPrefix)` helper that translates request URLs into on-disk lookup paths. Three URL shapes are recognised: historical `/assets/`, `<assetPathPrefix>/_next/static/...` for path prefixes / pathful absolute URLs, and `/_next/static/...` for plain absolute-URL prefixes. The App Router loads `assetPrefix` from a new `vinext-runtime-config.json` sibling (the RSC entry only embeds `__basePath`); the Pages Router reads it from the embedded `vinextConfig`.

5. **Cache headers (`packages/vinext/src/server/static-file-cache.ts`)** — files under `_next/static/` join `assets/` as "content-hashed" for the long-lived `immutable` Cache-Control. Same change in the slow-path branch of `tryServeStatic`.

## Tests

- **`tests/asset-prefix.test.ts`** (new — 25 cases) covers:
  - Unit tests for every helper (`normalizeAssetPrefix`, `isAbsoluteAssetPrefix`, `resolveAssetsDir`, `resolveAssetUrlPrefix`, `assetPrefixPathname`, `resolveAppRouterAssetPath`).
  - End-to-end `vinext build` + `startProdServer` tests for each `assetPrefix` shape:
    - Path prefix (`/custom-asset-prefix`) → SSR HTML references `/custom-asset-prefix/_next/static/...`, on-disk layout matches, fetching the bootstrap URL returns 200.
    - Absolute URL (`https://cdn.example.com`) → emitted asset URLs are fully qualified, no `/assets/` leaks.
    - Combined `basePath: '/app'` + `assetPrefix: '/cdn-prefix'` → routes stay under `/app/`, asset URLs under `/cdn-prefix/` (proves independence).
    - Unset (default) → historical `/assets/` layout preserved.
  - Result: ✅ 25/25 passing.

- **Regression sweep** — re-ran:
  - `tests/build-optimization.test.ts` (78 ✓)
  - `tests/ssr-css-assets.test.ts` (incl.)
  - `tests/next-config.test.ts` (108 ✓)
  - `tests/pages-router.test.ts` (209 ✓)
  - `tests/app-router.test.ts` (317 ✓)
  - `tests/deploy.test.ts` (220 ✓)
  - `tests/shims.test.ts` (915 ✓)
  - `tests/font-google-build.test.ts` (1 ✓)
  - `tests/nextjs-compat/` (288 ✓)
  - Misc build tests
  - No regressions.

- **Next.js fixture parity** — ran `VINEXT_BUILD=1 ./scripts/run-nextjs-deploy-suite.sh .nextjs-ref test/e2e/app-dir/asset-prefix/asset-prefix.test.ts`. Of the 7 cases in that suite, **4 now pass** that did not before:
  - ✅ `should redirect route when requesting it directly`
  - ✅ `should render link`
  - ✅ `rewrites that do not start with assetPrefix should still work`
  - ✅ `should respect rewrites that start with assetPrefix`
  - ❌ `should redirect route when requesting it directly by browser` — Playwright not installed locally (env issue, unrelated)
  - ❌ `should redirect route when clicking link` — same Playwright env issue
  - ❌ `bundles should return 200 on served assetPrefix` — see "Deliberate divergences" below

## Deliberate divergences from Next.js

* **`<script src="...">` vs `<script>import("...")</script>`** — the upstream `bundles should return 200 on served assetPrefix` test iterates over `$('script').toArray()` and reads `script.attribs.src`. Vinext uses `@vitejs/plugin-rsc`'s bootstrap which emits inline `<script>import("X")</script>` instead of `<script src="X" type="module">`, so the upstream selector returns 0 matches. The asset prefix logic itself is functionally correct — emitted URLs are prefixed, on-disk layout matches, and fetching the URLs returns 200 (verified in our own tests via the same regex extraction). This is a pre-existing architectural divergence in how vinext emits the bootstrap script and was not changed here.

## Open questions / follow-ups

* **Dev server parity** — Vite's `base` is still derived from `basePath` only; dev URLs aren't yet `assetPrefix`-aware. The production code path is what the documented use case targets, but a follow-up could thread `assetPrefix` through Vite's connect middleware too.
* **`<script src>` bootstrap emission** — emitting `<script src=...>` tags (perhaps via React's `bootstrapScripts` array instead of `bootstrapScriptContent`) would unlock the upstream `bundles should return 200` test directly. Out of scope here.
* **Automatic `/<prefix>/_next/:path+` rewrite** — Next.js auto-injects a rewrite from `/<assetPrefix>/_next/...` to `/_next/...` when `basePath === assetPrefix`. Vinext's prod server already serves the assets directly from disk so this rewrite isn't required for correctness, but parity for user-defined rewrites against the asset prefix path could be tightened up.

## Notes — updates since the original description

The original "Deliberate divergences" and "Open questions" sections were written against the very first round of this PR. Several items have since been resolved:

- **`<script src>` bootstrap is now emitted** (round 3, commit `3a2c886`). The App Router SSR pipeline passes `bootstrapModules: [url]` to React's `renderToReadableStream` instead of inline `bootstrapScriptContent`, so the streamed HTML contains a real `<script type="module" src="…">` tag. The Next.js `bundles should return 200 on served assetPrefix` fixture test now finds the bundle URLs via `$('script').toArray()`.

- **`vinext-runtime-config.json` sidecar removed** (round 4, commit `1218a81`). `assetPrefix` is now embedded as `export const __assetPrefix` in the generated RSC entry, matching the existing `__basePath` embed pattern and the Pages Router's `vinextConfig` export.

- **`basePath → assetPrefix` parity fallback added** (round 4, commit `bac8425`). When `basePath` is set and `assetPrefix` is left empty, vinext now mirrors Next.js's behaviour (`packages/next/src/server/config.ts:528-531`) and treats `assetPrefix = basePath`. This shifts the default asset URL layout for `basePath`-only apps from `/app/assets/…` to `/app/_next/static/…` — matching Next.js's expected layout.

### CSP behavioural change

The switch from inline `bootstrapScriptContent` to `bootstrapModules` (above) has CSP implications worth calling out:

- Apps no longer need `script-src 'unsafe-inline'` for the App Router bootstrap (improvement — inline imports previously required it).
- Apps that restrict script sources need `'self'` for the common case, or the CDN origin when `assetPrefix` is an absolute URL like `https://cdn.example.com`.
- React still threads the request nonce through to the emitted `<script type="module" src=…>` tag, so nonce-based CSP (`script-src 'nonce-…' 'strict-dynamic'`) keeps working unchanged.

See the comment at `packages/vinext/src/server/app-ssr-entry.ts` near the `bootstrapModules` call for the same notes inline.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -3955,11 +3955,19 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
             const workerEntry = path.join(workerOutDir, "index.js");
             if (!fs.existsSync(workerEntry)) return;
 
-            // Fallback: scan dist/client/assets/ for the client entry chunk.
-            // Pages Router uses "vinext-client-entry", App Router uses
-            // "vinext-app-browser-entry".
+            // Fallback: scan the on-disk assets directory for the client entry
+            // chunk when the SSR manifest lookup didn't surface one. Pages Router
+            // uses "vinext-client-entry", App Router uses "vinext-app-browser-entry".
+            //
+            // When `assetPrefix` is configured, chunks live under
+            // `<prefix>/_next/static/` (path-prefix) or `_next/static/`
+            // (absolute-URL prefix) — NOT `assets/`. Resolve the actual
+            // subdirectory from the same helper that drives `build.assetsDir`
+            // and the prod-server lookup path, so this fallback works for every
+            // layout supported by the rest of the pipeline.
             if (!clientEntryFile) {
-              const assetsDir = path.join(clientDir, "assets");
+              const assetsSubdir = resolveAssetsDir(nextConfig?.assetPrefix);
+              const assetsDir = path.join(clientDir, assetsSubdir);
               if (fs.existsSync(assetsDir)) {
                 const files = fs.readdirSync(assetsDir);
                 const entry = files.find(
@@ -3967,7 +3975,8 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
                     (f.includes("vinext-client-entry") || f.includes("vinext-app-browser-entry")) &&
                     f.endsWith(".js"),
                 );
-                if (entry) clientEntryFile = manifestFileWithBase("assets/" + entry, clientBase);
+                if (entry)
+                  clientEntryFile = manifestFileWithBase(`${assetsSubdir}/${entry}`, clientBase);
               }
             }
 
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -79,6 +79,11 @@ import { buildRequestHeadersFromMiddlewareResponse } from "./server/middleware-r
 import { detectPackageManager } from "./utils/project.js";
 import { manifestFileWithBase, manifestFilesWithBase } from "./utils/manifest-paths.js";
 import { hasBasePath, removeTrailingSlash } from "./utils/base-path.js";
+import {
+  ASSET_PREFIX_URL_DIR,
+  resolveAssetUrlPrefix,
+  resolveAssetsDir,
+} from "./utils/asset-prefix.js";
 import { asyncHooksStubPlugin } from "./plugins/async-hooks-stub.js";
 import { clientReferenceDedupPlugin } from "./plugins/client-reference-dedup.js";
 import { createRscClientReferenceLoadersPlugin } from "./plugins/rsc-client-reference-loaders.js";
@@ -1363,6 +1368,19 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
             // test/e2e/app-dir/css-media-query/css-media-query.test.ts which
             // asserts `cssText` preserves `max-width: 768px`.
             cssTarget: ["chrome111", "edge111", "firefox114", "safari15"],
+            // Direct Vite to write build output under `<assetPrefix>/_next/static/`
+            // (path-prefix form) or `_next/static/` (absolute-URL form) when
+            // `assetPrefix` is configured. Keeps the default of `assets/`
+            // when no prefix is set so existing on-disk layouts are stable
+            // for projects that haven't opted in.
+            //
+            // Pair with `experimental.renderBuiltUrl` above: the on-disk
+            // layout matches the URL path so the Cloudflare ASSETS binding
+            // and any static file server can resolve `<assetPrefix>/_next/static/...`
+            // requests without a runtime rewrite.
+            ...(nextConfig.assetPrefix
+              ? { assetsDir: resolveAssetsDir(nextConfig.assetPrefix) }
+              : {}),
             ...withBuildBundlerOptions(viteMajorVersion, {
               // Suppress "Module level directives cause errors when bundled"
               // warnings for "use client" / "use server" directives. Our shims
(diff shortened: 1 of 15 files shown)


james-elicx:
Round 2 review addressed in 165c0e9 and 00c00b6:

**Fixed**

1. `precompressAssets` hardcoded `"assets"` path — `precompressAssets` now takes an `assetsDir` option, the Vite plugin caller passes `resolveAssetsDir(nextConfig.assetPrefix)`, and two new tests in `tests/precompress.test.ts` cover the path-prefix and absolute-URL layouts (and assert the legacy `assets/` location is NOT walked when a custom dir is in effect).
2. Redundant `startsWith` in `prod-server.ts` — dropped (`pathname` always has a leading slash, so `includes("/_next/static/")` subsumes it).
3. Test cleanup race in `tests/asset-prefix.test.ts` — `buildFixtureWithConfig` now registers cleanup synchronously right after `mkdtempSync`, before any throwable work.

**Verified but not changed**

- `static-file-cache.ts` `startsWith("_next/static/")` (no leading slash) was kept — `relativePath` in that file has no leading slash, so `startsWith("_next/static/")` covers the absolute-URL prefix layout that `includes("/_next/static/")` (requires leading slash) cannot match. Comment updated to explain.
- `envConfig.build?.assetsDir` at the `_headers` write site — confirmed populated by the plugin's `config()` hook before the `_headers` generator runs, so the emitted rule is e.g. `/cdn/_next/static/*` for `assetPrefix: "/cdn"`. No code change needed.

**Deferred (follow-up)**

`basePath` → `assetPrefix` fallback parity gap. Next.js auto-sets `assetPrefix = basePath` when `basePath` is set and `assetPrefix` is empty ([next.js/packages/next/src/server/config.ts:555-564](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/config.ts#L555-L564)). vinext currently does not — when only `basePath` is set, asset URLs continue to live under `/assets/` rather than `/<basePath>/_next/static/`. The common case is still handled because `basePath` already sets `base = basePath + "/"` for Vite, so emitted asset URLs land under `<basePath>/assets/`. Will file a follow-up issue.

Tests: `vp check` clean across 1328 files. 15/15 precompress + 157/157 asset-prefix/next-config/static-file-cache passing.

packages/vinext/src/index.ts
```diff
@@ -3496,6 +3557,20 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
...
+          // file narrowly scoped — anything that affects request handling
+          // outside of the RSC handler should live here so dev/prod parity
+          // is easy to audit.
+          const runtimeConfig = {
+            assetPrefix: nextConfig?.assetPrefix ?? "",
+          };
+          fs.writeFileSync(
+            path.join(outDir, "vinext-runtime-config.json"),
+            JSON.stringify(runtimeConfig),
+          );
```
james-elicx: why cant we embed this as well? isn't the entire config already embedded?

PR #1321, opened 2026-05-19
chore: split e2e compatibility by App Router vs Pages Router

james-elicx:
## Summary

The `/compatibility` page previously showed one undifferentiated grid of ~1000 Next.js test files. This PR classifies each test by which router(s) its fixture exercises (App, Pages, both, parity) and surfaces the breakdown in the UI.

## Design

**Two-table model.** Classifications live in their own table, decoupled from per-run results:

| Table                 | Holds                                                | Lifecycle                         |
| --------------------- | ---------------------------------------------------- | --------------------------------- |
| `compat_runs`         | One row per CI run (totals)                          | Append per run                    |
| `compat_file_results` | One row per (run, suite) with status + counts        | Append per run                    |
| `compat_suite_meta`   | **One row per test file**: router, next_ref, when    | Upsert when classifier re-runs    |

The `/compatibility` page does a `LEFT JOIN compat_suite_meta ON suite = suite`. Suites without a classification render as `unknown` (test added between classifier runs, or fresh DB).

**Why decoupled:** classifications describe *test files*, not *runs*. A Next.js ref bump, an override fix, or a heuristic improvement should be able to re-classify everything without re-running tests. Conversely, a partial test run still gets fresh classifications because the build job submits them before any shards run.

**Two ingest endpoints**, sharing auth via a small `_auth.ts` helper:

| Endpoint                          | Purpose                                            |
| --------------------------------- | -------------------------------------------------- |
| `POST /api/compatibility`         | Per-run results (unchanged shape from main)        |
| `POST /api/compatibility/classify` | Per-suite router classification (new)             |

## Classification heuristic

For each Next.js e2e test we walk the test's fixture directory looking for `app/` and/or `pages/` subdirs that contain *real* route files (page / route / layout / default for app, any non-special file for pages — `_app` / `_document` / `_error` don't count):

| Kind      | Meaning                                                                       |
| --------- | ----------------------------------------------------------------------------- |
| `app`     | Fixture has real App Router routes only                                       |
| `pages`   | Fixture has real Pages Router routes only                                     |
| `both`    | Real routes in both — parity / interop test                                   |
| `unknown` | No on-disk fixture (config / build / edge-runtime tests or inline-only setup) |

A path-prefix-only rule ("everything under `test/e2e/app-dir/` is App Router") is wrong for ~50 suites that nest pure Pages Router tests under `app-dir/` for organizational reasons, and fails to catch ~76 parity tests outside `app-dir/`. The fixture-scan classifier handles both.

Edge cases handled:
- `pageExtensions` naming (`layout.page.tsx` → `layout`)
- Tests living in a `test/` subdir alongside the fixture
- Inline-fixture suites under `test/e2e/app-dir/` (no on-disk routes; path convention says App Router)
- `APP_ROUTER_NON_APP_DIR_SUITES` curated list (mirrors `scripts/nextjs-deploy-manifest.mjs`)
- `node_modules` / `.next` excluded from scan

**Parity tests count toward both router buckets.** A failure in a parity test is a real failure for both routers. The "How this works" card explains that App + Pages exceeds the total.

## What's in the PR

- **Schema** (`apps/web/app/lib/db/schema.ts`, migration `0001_lucky_xavin.sql`): new `compat_suite_meta` table
- **Classifier** (`scripts/classify-nextjs-suites.mjs` + `.d.mts`): bounded directory walk with all the edge cases above
- **Enumerator** (`scripts/list-nextjs-e2e-suites.mjs`): lists every test file under `test/e2e/` — input to the classifier
- **Auth helper** (`apps/web/app/api/compatibility/_auth.ts`): shared constant-time-compare across both endpoints
- **Classify endpoint** (`apps/web/app/api/compatibility/classify/route.ts`): upserts in chunks (25 rows / INSERT)
- **Workflow**: build job runs the classifier and POSTs to `/api/compatibility/classify` immediately; report job no longer needs the classification artifact
- **UI** (`apps/web/app/compatibility/page.tsx`, `contribution-grid.tsx`): new "By router" stat card row; filter chips on the grid (All / App / Pages / Parity / Other)
- **Tests** (`tests/classify-nextjs-suites.test.ts`): 14 unit tests covering each bucket and edge case

## Numbers (local sanity check vs `.nextjs-ref` HEAD)

```
{ "totalSuites": 1036, "counts": { "app": 576, "pages": 325, "both": 113, "unknown": 22 } }
```

The 22 "unknown" suites are config tests (`config-promise-export`, `config-schema-check`), edge-runtime tests (`edge-can-use-wasm-files`), and inline-fixture middleware tests — all genuinely unclassifiable from filesystem inspection alone.

## Trade-offs of the meta-table design

- ✅ **Retroactive corrections heal history.** Fix a misclassification or an override, and every past run's view updates on the next classifier POST.
- ✅ **Decoupled cadence.** Re-classify without re-running tests; partial test runs still publish fresh classifications.
- ✅ **Smaller results payload.** No `router` field per file row.
- ⚠️ **Trend chart split-by-router isn't strictly point-in-time** — re-classifying a suite shifts how it appears in historical runs. Provenance (`next_ref`, `classified_at`) on the meta row makes this auditable.
- ⚠️ **LEFT JOIN on every page load.** Cheap at ~1000 rows; negligible.

## Verification

- `vp check` clean
- `vp test run tests/classify-nextjs-suites.test.ts` — 14/14 pass
- `apps/web` `pnpm build:vinext` succeeds; both `/api/compatibility` and `/api/compatibility/classify` show up in the route list
- Classifier sanity-checked against a local Next.js checkout — spot-checked classifications (404-page, middleware-basic, app-dir/actions, app-dir/scss/*, app-dir/app-css-pageextensions, app-dir/pnpm-workspace-root) match expectations

## Follow-ups (not in this PR)

- Per-router trend line on the compatibility-over-time chart. Currently `compat_runs` only stores aggregate totals; splitting the trend requires either aggregating from `compat_file_results JOIN compat_suite_meta` per render or denormalising. Worth doing once we have enough nightly history to make the chart interesting.
- Hand-curated `scripts/nextjs-suite-overrides.json` for edge cases the heuristic misclassifies in practice. The classifier already consults this file when present, so adding entries is a no-code change.
- Tests for the `/api/compatibility/classify` endpoint itself (currently the GH workflow is the only integration coverage). The validation logic is small and the schema upsert is well-trodden, but a couple of unit tests wouldn't hurt.

Still draft pending review of the meta-table design and the parity-counts-in-both convention.


james-elicx:
Thanks for the reviews. Addressed in `893e993` and the preceding `36899ec`:

## Actionable items

**`scanFixture` recursion (PR review #2)** — fixed, with a twist. The naive fix (skip descent into any checked `app/` / `pages/`) caused ~10 regressions: many Next.js test fixtures use a wrapping directory literally named `app` as the test app's project root, with the real `app/` and `pages/` nested *inside* (e.g. `test/e2e/og-api/app/{app/,pages/}`, `test/e2e/middleware-static-files/app/{app/,pages/}`). The classifier needs to walk into those wrappers.

Final fix: a directory named `app`/`pages` is treated as a wrapper (and recursed into) if it contains a top-level `next.config.*` OR an inner `app/` alongside an inner `pages/`. Otherwise it's handed to the dedicated route checker. The wrapper detector intentionally does NOT treat `middleware.{js,ts}` as a wrapper signal — Next.js tests put noop `middleware.js` files inside real App Router `app/` dirs to assert the file is ignored at that level (`test/e2e/app-dir/app-middleware`).

Verified the classifier output is unchanged against the full `test/e2e/` tree of a current Next.js checkout: 1036 suites, 576/325/113/22, diff-vs-pre-fix is empty. Three new regression tests cover the wrapper, route-group, and noop-middleware cases.

**`list-nextjs-e2e-suites.mjs` import guard** — added the same `import.meta.url === \`file://${process.argv[1]}\`` guard the classifier has.

**`_auth.ts` empty-header early-exit** — short-circuit when the header is missing/empty before paying two SHA-256 digests. Constant-time-compare guarantee preserved (we only short-circuit on values we already know can't match a non-empty secret).

**Migration trailing newline** — added.

## Items resolved by the UI refactor (commit `36899ec`)

**Naming asymmetry between `both`/`parity`** — addressed by lifting filter state into a shared `CompatibilityViews` wrapper with consistent vocabulary, and removing the standalone "Parity" stat card. The chart and grid now share one filter (Kumo `<Tabs variant="segmented">`), so the chips and stat cards no longer disagree.

**"Other" stat card asymmetry** — removed entirely. Unclassified suites are still reachable via the segmented Tabs "Other" filter.

**Active-pill weird styling** — replaced the hand-rolled `bg-kumo-default + text-kumo-base` pills with Kumo's segmented Tabs primitive. The original was using the text token for the background, producing a saturated black/white inversion.

## Items I'm not addressing

- **`.d.mts` / `.mjs` extension coupling** — fragile but works under the project's moduleResolution config. Noted, leaving as-is.
- **Workflow inline `node -e` using CJS `require()`** — `node -e` defaults to CJS; converting to ESM with `--input-type=module` or extracting to a script wouldn't materially improve things. Noted, leaving as-is.
- **`readdirSync` swallowing non-ENOENT errors** — fair point. Would be nice to log on `EACCES`/`ENOTDIR` for local debugging, but in CI we don't see those and it'd add noise. Leaving as-is for now.

Per-router trend line is now in: the chart receives `app`/`pages`/`both`/`unknown`/`all` series from a single SQL aggregate (`SUM(CASE WHEN m.router IN (...))` over the `compat_file_results JOIN compat_suite_meta`), and the active filter picks one client-side.

james-elicx:
Addressed in `5e09f69`:

1. **`sqlExcluded` type-narrowing** — parameter is now `"router" | "classified_at"` instead of `string`, so the "no user input ever reaches `sql.raw`" invariant is compile-checked.

2. **SVG negative dimensions on empty filter** — the `<svg>` is now gated on `visibleCells.length > 0` rather than rendering alongside the placeholder, and dimensions are clamped to `>= 0` defensively.

3. **Trend retroactivity note in "How this works"** — added a one-sentence callout: *"Per-router trend lines use the latest classification snapshot, so reclassifying a suite updates how it appears across every historical run. The aggregate 'All' line is unaffected."*

4. **`scanFixture` early-return readability** — changed `if (hasApp && hasPages)` to `if (hasPages)` / `if (hasApp)` since the other flag was just set on the line above. No behavioural change; classifier output against the full Next.js `test/e2e/` tree is byte-identical (1036 suites, 576/325/113/22).

5. **`classifiedAt` bounds check** — reject NaN/Infinity and any timestamp before 2023-01-01 UTC. Catches the seconds-vs-millis mistake and accidental 0/-1 values cheaply.

Not addressed: the `trendRowsDesc` rename — the comment itself notes "very minor" and renaming would muddle the diff without any real win.

james-elicx:
Addressed in `44471e9`:

1. **`GROUP BY r.id` portability comment** — added an inline note explaining that selecting `r.created_at` without aggregating it is a SQLite functional-dependency affordance, and what to change if the query is ever ported to postgres.

2. **Unified router bucketing** (resolves items 3 and 6 together) — extracted a new `apps/web/app/compatibility/router-buckets.ts` module that owns the canonical filter / counting / pass-rate logic in one place. All three call sites (`contribution-grid.tsx`, `compatibility-views.tsx`, `page.tsx`) now import from it. The `parity` / `other` naming in the stat cards is gone — they use `byRouter.both` / `byRouter.unknown` to match `RouterKind`. The "Mixed counts in both App and Pages" rule is documented once at the top of the new module instead of three times.

4. **`passRate` function rename** — renamed the line chart's local `passRate(c)` helper to `computePassRateRatio(c)`. Distinguishes it from `bucketPassRate` in `router-buckets` (which returns a percentage, not a ratio) and avoids shadowing the `passRate` variable in `page.tsx`.

## Not addressed (intentionally)

- **"Mixed (both)" label vs "Parity" / "Interop"** (item 2) — the user explicitly asked to rename "Parity pass rate" to "Mixed pass rate" earlier in this PR's review. Keeping their choice.

- **`trendRowsDesc` rename** (item 5a) — reviewer self-tagged "very minor". Skipping to keep the diff focused.

- **`deriveSuiteGroup` inconsistent levels** (item 5b) — pre-existing tooltip-only behaviour, predates this PR.

The double-counting concern (item 6) is now resolved by the shared module: there's exactly one place that knows the rule, and `countByFilter` / `bucketByRouter` use the same accumulator pattern.

james-elicx:
Thanks. No new code changes needed for this round — the three items are all explicitly self-tagged as non-actionable / not-worth-deduplicating-yet:

1. **`next_ref` on `compat_suite_meta`** — deliberately dropped earlier this PR. Workflow log + `classified_at` are the audit trail today; revisit if the override loop gets frequent.
2. **`scanFixture` recursing into empty `app/`/`pages/`** — intentional, required for fixtures like `og-api` where the real App Router lives one level deeper.
3. **`FULL_DATETIME` duplication** — two instances; reviewer's own bar ("three would be worth deduplicating") not hit yet.

CI is green (36/36). Marking ready for review.

PR #1373, opened 2026-05-20
fix(rsc): resolve `export * from` in `use client` modules

### LINKED ISSUE #1352: Build failure: `export * from` in `"use client"` files not supported by RSC plugin

## Problem

`@vitejs/plugin-rsc`'s `use-client` transform generates client reference proxies for each export, but it does not handle `export * from '...'` (namespace re-exports). It only supports named exports and default exports.

```
[plugin rsc:use-client] components/export-all/index.js
Error: unsupported ExportAllDeclaration
    at transformProxyExport (.../plugin-rsc/dist/transforms/index.js:338:90)
```

Next.js's Webpack plugin resolves `export *` to individual named exports before generating proxies.

## Estimated Impact

~1 build failure.

## Affected Test Suites

- An `app-dir` test using `export * from '...'` in a `"use client"` module

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test with a `"use client"` file that uses `export * from './other'` and verify the build fails.

2. **This is likely an upstream `@vitejs/plugin-rsc` issue.** File an issue on the RSC plugin repo if one doesn't exist, or implement a workaround in vinext's build pipeline that pre-resolves `export *` declarations in `"use client"` files before the RSC transform runs.

3. **As a workaround**, users can replace `export * from` with explicit named re-exports in `"use client"` files.

james-elicx:
## Summary

`@vitejs/plugin-rsc`'s `rsc:use-client` transform throws `unsupported ExportAllDeclaration` whenever a `"use client"` module contains `export * from '...'` (or `export * as Ns from '...'`). This breaks any app that follows Next.js's `test/e2e/app-dir/rsc-basic` pattern of putting `export *` re-exports inside client boundaries.

Adds a new pre-pass Vite plugin, `vinext:use-client-export-all`, that runs in the RSC environment before `rsc:use-client` and rewrites:

- `export * from '...'` → `export { a, b, c, ... } from '...'` (named exports enumerated by resolving and parsing the target module; nested `export *` chains are recursed; `default` is intentionally not forwarded, matching ESM)
- `export * as Ns from '...'` → `import * as Ns from '...'; export { Ns };` (a single named export the upstream transform can wrap as a client-reference proxy)
- External / unresolvable targets are left alone — the upstream error then surfaces with its original wording, and users can switch to explicit named re-exports.

The plugin is registered before the auto-injected `@vitejs/plugin-rsc` in vinext's plugin array, at the default enforce position so it runs after Vite's built-in TS/JSX transform (parseAst doesn't accept TS/JSX).

## Design notes

- Reference: Next.js's flight loader handles `export *` differently — it sets the module's `assumedSourceType` back to `module` so webpack handles the re-export natively (see [`next-flight-loader/index.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack/loaders/next-flight-loader/index.ts#L25-L54)). We can't do the same in Rolldown/Vite, so we pre-expand the declaration instead. The visible behavior matches: each re-exported named export becomes its own client-reference proxy.
- Test fixture mirrors `.nextjs-ref/test/e2e/app-dir/rsc-basic/components/export-all/*` (chained `export *`, renamed re-export, and a namespace re-export case).
- The upstream transform location: `node_modules/@vitejs/plugin-rsc/dist/transforms/index.js:338` — `if (!options.ignoreExportAllDeclaration && node.type === "ExportAllDeclaration") throw new Error("unsupported ExportAllDeclaration");`. The plugin exposes no opt-out, so a pre-pass is the only workaround without patching `node_modules`.

## Upstream tracking

- vitejs/vite-plugin-react#1233 — filed for this PR. Even if upstream lands a fix, our pre-pass is a safe no-op on already-supported code paths.

## Test plan

- [x] `pnpm test tests/use-client-export-all-build.test.ts` — both new cases pass; both fail on `main` with the expected error
- [x] `pnpm test tests/app-router.test.ts tests/build-optimization.test.ts tests/standalone-build.test.ts` — 402 existing tests still pass
- [x] `pnpm run check` — clean (format, lint, types)
- [ ] CI green

Closes #1352

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -125,6 +125,7 @@ import {
   type BundleBackfillChunk,
 } from "./build/ssr-manifest.js";
 import { stripServerExports } from "./plugins/strip-server-exports.js";
+import { createUseClientExportAllPlugin } from "./plugins/use-client-export-all.js";
 import { hasMdxFiles } from "./utils/mdx-scan.js";
 import { scanPublicFileRoutes } from "./utils/public-routes.js";
 import tsconfigPaths from "vite-tsconfig-paths";
@@ -2269,6 +2270,11 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
     },
     // Stub node:async_hooks in client builds — see src/plugins/async-hooks-stub.ts
     asyncHooksStubPlugin,
+    // Expand `export * from '...'` in 'use client' modules to explicit named
+    // re-exports before @vitejs/plugin-rsc's `rsc:use-client` transform runs.
+    // The upstream transform throws "unsupported ExportAllDeclaration" otherwise.
+    // See src/plugins/use-client-export-all.ts.
+    createUseClientExportAllPlugin(),
     createInstrumentationClientTransformPlugin(() => instrumentationClientPath),
     // Dedup client references from RSC proxy modules — see src/plugins/client-reference-dedup.ts
     ...(options.experimental?.clientReferenceDedup ? [clientReferenceDedupPlugin()] : []),
--- a/packages/vinext/src/plugins/use-client-export-all.ts
+++ b/packages/vinext/src/plugins/use-client-export-all.ts
@@ -197,7 +197,7 @@ function quickHasExportAll(code: string): boolean {
  * declaration to an explicit `export { a, b, c } from '...'` form that the
  * RSC plugin already understands.
  *
- * Tracked upstream: https://github.com/vitejs/vite-plugin-react/issues/...
+ * Tracked upstream: https://github.com/vitejs/vite-plugin-react/issues/1233
  *
  * Reference: Next.js's flight loader treats `export *` similarly (it sets the
  * module source type back to `module` so webpack handles the re-export, rather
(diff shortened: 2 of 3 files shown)


james-elicx:
Thanks @ask-bonk — addressed your review:

- **Blocking (ReDoS in `HAS_USE_CLIENT_RE`)**: fixed in 36776a9. Replaced the backtracking regex with a single-pass manual scanner (`quickHasUseClient`) that visits each character at most once. CodeQL re-ran on that commit and is now green.
- **Non-blocking #1 (destructured exports)**: fixed in 53343b3. `collectNamedExports` now walks `ObjectPattern` / `ArrayPattern` / `RestElement` / `AssignmentPattern` declarator ids, and there's a new regression test that builds a `"use client"` barrel re-exporting `export const { First, Second } = ...` and `export const [Third] = ...`.

On the `hasReactDirective()` reuse suggestion: agreed the duplication isn't ideal, but the two functions have intentionally different semantics — `hasReactDirective` matches `"use client" | "use server"` and walks past prior directives like `"use strict"`, while `quickHasUseClient` is a `"use client"`-only fast filter. Extracting a shared helper would mean either widening the plugin's behaviour (and re-running on `"use server"` files that the RSC transform handles fine) or parameterising the helper. Happy to do the refactor in a follow-up if you'd rather; the immediate ReDoS concern is resolved.

/bigbonk

PR #1385, opened 2026-05-20
fix(build): avoid __dirname shim conflicts and cover node_modules

### LINKED ISSUE #1345: Build failure: `__dirname`/`__filename` shim injection conflicts and runtime errors

## Problem

```
   ╭─[ next.config.ts:3:7 ]
```

```
```

## Estimated Impact

## Affected Test Suites

## Recommendation

james-elicx:
## Summary

Two related fixes to vinext's CommonJS-global shim injection that surfaced on the Next.js v16 deploy suite:

- **next.config.ts.** When the user's config already declared `const __dirname = fileURLToPath(import.meta.url)` (a real pattern in Next.js v16-style configs), the injector still emitted its own `const __dirname = ...` and OXC rejected the duplicate `const`. New `findUserDeclaredCjsGlobals` helper masks strings/comments and detects existing `const/let/var __dirname` / `__filename` bindings (plain, comma-continued, or object-destructure); the injector skips those names.
- **node_modules ReferenceError.** Bundled third-party packages reference `__dirname` / `__filename`, but the server entry is emitted as ESM (`type: module`) where those CJS globals do not exist. New `vinext:cjs-globals-shim` plugin runs in the `ssr` / `rsc` environments only, targets `node_modules` `.js/.mjs/.cjs` files containing bare references, and prepends a per-module preamble that declares both bindings via `import.meta.url` + `fileURLToPath` + `dirname` (helpers imported under `__vinext_*` aliases to avoid colliding with user imports).

The shim only prevents the `ReferenceError`. Libraries that load sibling on-disk assets through `__dirname` (native addons, embedded wasm) still need to be externalized via `serverExternalPackages` — `import.meta.url` after bundling points at the bundle, not the original module location.

## Test plan

- [x] `pnpm test tests/next-config.test.ts` (125 tests pass, including new regressions for user-declared `__dirname` / `__filename` and helper-level coverage of `findUserDeclaredCjsGlobals`)
- [x] `pnpm test tests/cjs-globals-shim.test.ts` (18 new tests covering `detectCjsGlobalReferences`, `buildCjsGlobalsShimPreamble`, plugin filter, environment gating, and full transform output)
- [x] `pnpm test tests/build-optimization.test.ts tests/deploy.test.ts tests/standalone-build.test.ts` (304 pass, no regressions)
- [x] `pnpm run check` clean (format + lint + types)

Closes #1345

### DIFF
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -582,6 +582,65 @@ export function reassignsModuleExports(source: string): boolean {
 }
 
 /**
+ * Detect which of the CJS-style identifiers (`__filename`, `__dirname`) the
+ * user already declares themselves with `const`, `let`, or `var`. Used by the
+ * injector to skip duplicate declarations that would otherwise fail with
+ * "Identifier `__dirname` has already been declared" under OXC/Rolldown.
+ *
+ * Skips identifiers that appear inside strings, template literals, and
+ * comments so a docstring like `/* sets __dirname *\/` doesn't suppress the
+ * shim. Recognizes plain bindings (`const __dirname = ...`), comma-continued
+ * declarators (`const x = 1, __dirname = ...`), and identifier destructuring
+ * (`const { __dirname } = ...` / `const { foo: __dirname } = ...`). Bare
+ * declarations without initializer (`let __dirname;`) are also caught — they
+ * still create a binding that collides with the injected `const`.
+ *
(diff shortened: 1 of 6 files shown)


james-elicx:
Rebased on main. Conflict in `packages/vinext/src/index.ts` was a trivial adjacent-import collision (this PR added the `cjsGlobalsShimPlugin` import next to lines main had also added). `pnpm run check` passes. CI re-running.

PR #1404, opened 2026-05-21
Guard Pages Router navigation during SSR

### LINKED ISSUE #1353: Pages Router navigation shim references `window` during SSR/prerender

## Problem

The Pages Router `withRouter` / navigation shim references `window` inside `performNavigation()` during server-side rendering. This code path executes during prerendering (`VINEXT_PRERENDER=1`), causing a crash.

```
ReferenceError: window is not defined
    at performNavigation (file:///.../dist/server/entry.js:1288:76)
    at Object.push (file:///.../dist/server/entry.js:1407:30)
    at RouterMethodSSR (file:///.../dist/server/entry.js:5257:44)
```

The crash also causes a cleanup error: `TypeError: Cannot read properties of undefined (reading 'destroy')` in afterAll.

## Estimated Impact

~3 test failures (1 runtime crash, 3 assertions).

## Affected Test Suites

- `test/e2e/with-router/index.test.ts` (3 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test that uses `withRouter` in a Pages Router page and verify it renders during SSR without crashing. Confirm it fails.

2. **Add a server-side guard.** The `performNavigation()` function should check `typeof window === 'undefined'` and no-op (or throw a developer-friendly error) when called during SSR/prerendering instead of accessing `window` directly.

3. **Check all navigation shim code paths for `window` references.** Any code in `shims/router.ts` or related files that accesses browser globals needs SSR safety guards.

AjTheSpidey:
﻿Fixes #1353.

`performNavigation()` now returns early on the server before touching `window`, so Pages Router `router.push()` / `router.replace()` calls during SSR or prerender do not crash with `window is not defined`.

I added a regression test that renders a `withRouter` component on the server and calls `router.push()` during render. It should not throw, and the no-op navigation resolves to `false`.

Tested:
- pnpm exec vp test run tests/shims.test.ts -t "withRouter router.push is safe during server rendering"
- pnpm exec vp fmt --check packages/vinext/src/shims/router.ts tests/shims.test.ts
- pnpm run build

Note: `pnpm run check` reports formatting issues across many untouched files on my Windows checkout, so I verified formatting on the changed files directly.

### DIFF
--- a/packages/vinext/src/shims/router.ts
+++ b/packages/vinext/src/shims/router.ts
@@ -840,6 +840,8 @@ async function performNavigation(
   mode: "push" | "replace",
   onStateUpdate?: () => void,
 ): Promise<boolean> {
+  if (typeof window === "undefined") return false;
+
   const navigationLocale = resolveTransitionLocale(options?.locale);
   let resolved = resolveNavigationTarget(url, as, navigationLocale);
 
--- a/tests/shims.test.ts
+++ b/tests/shims.test.ts
@@ -1731,6 +1731,28 @@ describe("next/router withRouter HOC", () => {
     expect(receivedRouter.events).toBeDefined();
   });
 
+  it("withRouter router.push is safe during server rendering", async () => {
+    delete (globalThis as any).window;
+    vi.resetModules();
+
+    const { withRouter, wrapWithRouterContext } =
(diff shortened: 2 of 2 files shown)


james-elicx:
I believe this was implemented in #1378

Issue #1336, opened 2026-05-20
i18n routing: `locale: false` on rewrites/redirects not implemented


james-elicx:
> *This issue was created by an agent analysing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/26163751905) (vinext `main` vs Next.js `v16.2.6`, 2026-05-20).*

## Problem

Several i18n routing features are not properly implemented:

1. **`locale: false` on rewrite/redirect sources not implemented.** When a rewrite or redirect in `next.config.js` specifies `locale: false` on the source, the pattern should match the raw path without locale prefix handling. vinext does not honor this flag, so rewrites/redirects fail to match when they should.

2. **Locale detection re-runs on every request instead of being sticky.** After the initial locale is established, client-side navigations should preserve it. vinext appears to re-detect from `Accept-Language` on each server request, causing the locale to flip unexpectedly.

3. **Locale prefix not stripped for API routes.** Requests like `/fr/api/ok` need the locale prefix stripped before API route matching. vinext does not strip it, causing 404s.

4. **Default locale prefix stripping missing.** Requests to `/en/page` (where `en` is the default locale) should redirect to `/page`. This redirect is not happening.

```
Expected: "en"
Received: "id"  // locale flipped to browser preference on client nav
```

## Estimated Impact

~24 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/i18n-ignore-rewrite-source-locale/rewrites.test.ts` (8 failures, all 8 tests)
- `test/e2e/i18n-ignore-redirect-source-locale/redirects-with-basepath.test.ts` (4 failures)
- `test/e2e/i18n-ignore-redirect-source-locale/redirects.test.ts` (4 failures)
- `test/e2e/ignore-invalid-popstateevent/with-i18n.test.ts` (3 failures)
- `test/e2e/ignore-invalid-popstateevent/without-i18n.test.ts` (2 failures)
- `test/e2e/i18n-preferred-locale-detection/i18n-preferred-locale-detection.test.ts` (1 failure)
- `test/e2e/i18n-default-locale-redirect/i18n-default-locale-redirect.test.ts` (1 failure)
- `test/e2e/middleware-custom-matchers-i18n/test/index.test.ts` (1 failure)

## Recommendation

[...]

PR #1407, opened 2026-05-21
fix(i18n): make locale sticky across client navigations

james-elicx:
Refs #1336 (item 2).

## Summary

Locale stickiness for the Pages Router client runtime. Three behaviours that previously diverged from Next.js:

1. **History state was empty (`{}`).** Next.js writes `{ url, as, options, __N: true, key }` on every `pushState`/`replaceState` (see [`Router.changeState`](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/router/router.ts) around L1916). Without this shape, popstate handlers can't tell our entries apart from third-party history writes and can't recover the active locale on back/forward.

2. **Popstate ignored stale state filtering.** Next.js drops a popstate event when it's the first one observed and its `state.options.locale` + `state.as` match the current page (Safari tab-restore / BFCache replay). vinext's handler always re-fetched.

3. **Popstate re-detected locale from `window.__VINEXT_LOCALE__`** rather than the locale recorded in the history entry, so back/forward across locale changes used the wrong fetch URL.

This PR:

- Updates `updateHistory()` in `packages/vinext/src/shims/router.ts` to write the Next.js-shaped state, populated with the active locale resolved via the existing `resolveTransitionLocale` helper.
- Adds an `isFirstPopStateEvent` flag and an `isNextRouterState` guard in the popstate handler. Foreign state (no `__N: true`) is ignored. A first-event replay with matching locale+`as` is ignored. `state.options.locale` is preferred over the window global when computing the popstate fetch URL.
- Carries the `key` field through replaces (preserving the prior key) and mints a new key on each push.

## Next.js parity references

Local clone at `.nextjs-ref`:

- `packages/next/src/shared/lib/router/router.ts` — `onPopState` (L920–999) and `Router.changeState` (L1900–1931).
- `test/e2e/ignore-invalid-popstateevent/with-i18n.test.ts` and `without-i18n.test.ts` — exercise the stale-state filter, including the "Don't ignore event with different locale" branch.
- `test/e2e/i18n-preferred-locale-detection/i18n-preferred-locale-detection.test.ts` — covers the client-side stickiness behaviour we're trying to match (Link to `/` from a non-locale-prefixed page must not flip the locale to the browser preference).

## Tests added

`tests/pages-router-i18n-sticky-locale.test.ts` — 7 tests across three describe blocks:

- **Pages Router history state shape** (3 tests): asserts `Router.push` writes `{ __N: true, options: { locale } }` and that the locale slot is populated from the active locale, explicit `locale: 'fr'`, or `defaultLocale` when `locale: false`.
- **Pages Router popstate stale-state filter (i18n parity)** (3 tests): first popstate matching current locale+`as` is ignored; first popstate with a different locale is processed; foreign state (no `__N`) is ignored.
- **Pages Router locale stickiness on programmatic navigation** (1 test): `Router.push` with no explicit locale carries the current locale in history state.

I also relaxed 5 existing assertions in `tests/shims.test.ts` (in the "Pages Router router helpers" and "Pages Router concurrent navigation" blocks) that pinned `pushState` to `{}` — they now use `expect.objectContaining({ __N: true })` so the assertions stay focused on what they're actually testing (URL normalisation / locale-qualified fetch) without depending on the empty-state shape.

## Scope boundaries with parallel PRs

- **Item 3 (API-route locale stripping)**: nothing touched. No changes under `packages/vinext/src/server/api*` or route matching.
- **Item 4 (default-locale prefix normalisation)**: nothing touched. `packages/vinext/src/config/config-matchers.ts`, `pages-i18n.ts`, `dev-server.ts`, and `prod-server.ts` are unchanged.
- **NEXT_LOCALE cookie write on server-side detection**: explicitly out of scope. The existing `parseCookieLocaleFromHeader` already gives the cookie priority over Accept-Language. Whether vinext should automatically `Set-Cookie: NEXT_LOCALE` on initial detection is a separate question — Next.js itself doesn't do this (cookie is user-set, e.g. by a language switcher), and moving it would overlap with item 4's redirect work. Documented as a follow-up.

## Verification

```sh
vp test run tests/pages-router-i18n-sticky-locale.test.ts   # 7/7 pass
vp test run tests/shims.test.ts                              # 944/944 pass
vp test run tests/link.test.ts tests/link-navigation.test.ts \
            tests/pages-router.test.ts tests/pages-router-concurrency.test.ts \
            tests/pages-i18n.test.ts tests/pages-i18n-prod.test.ts \
            tests/router-javascript-urls.test.ts tests/features.test.ts
# 1036/1036 pass

vp check packages/vinext/src/shims/router.ts \
         tests/pages-router-i18n-sticky-locale.test.ts \
         tests/shims.test.ts
# clean
```

### DIFF
--- a/packages/vinext/src/shims/client-locale.ts
+++ b/packages/vinext/src/shims/client-locale.ts
@@ -22,9 +22,15 @@ export function getCurrentBrowserLocale({
   );
   if (pathnameLocale) return pathnameLocale;
 
+  // Prefer the actually-active locale (set by SSR for the rendered page) over
+  // the configured default. This matters for the i18n-sticky-locale flow
+  // (issue #1336): a default-locale path served under a non-default locale —
+  // e.g. `id` rendered at `/about` with no `/id` prefix — must still report
+  // its active locale so Router.push can stamp it into history state. Falling
+  // back to the default would erase that stickiness on every client nav.
   return (
     detectDomainLocale(domainLocales, hostname ?? undefined)?.defaultLocale ??
-    window.__VINEXT_DEFAULT_LOCALE__ ??
-    window.__VINEXT_LOCALE__
+    window.__VINEXT_LOCALE__ ??
+    window.__VINEXT_DEFAULT_LOCALE__
   );
 }
--- a/packages/vinext/src/shims/pages-router-runtime.ts
+++ b/packages/vinext/src/shims/pages-router-runtime.ts
@@ -1,10 +1,20 @@
 type PagesRouterPopStateHandler = (event: PopStateEvent) => void;
 
 let pagesRouterPopStateHandler: PagesRouterPopStateHandler | undefined;
+let stampInitialHistoryStateFn: (() => void) | undefined;
 let pagesRouterRuntimeInstalled = false;
 
 export function setPagesRouterPopStateHandler(handler: PagesRouterPopStateHandler): void {
   pagesRouterPopStateHandler = handler;
+}
+
+/**
+ * Register the function that stamps Next.js-shaped state onto the initial
+ * document entry (called once at install time). Router.ts registers this so
+ * the runtime module stays free of router internals.
+ */
+export function setStampInitialHistoryState(fn: () => void): void {
+  stampInitialHistoryStateFn = fn;
 }
 
 export function installPagesRouterRuntime(): void {
@@ -17,5 +27,9 @@ export function installPagesRouterRuntime(): void {
   }
 
   pagesRouterRuntimeInstalled = true;
+  // Stamp the initial document entry with router-shaped state *before* the
+  // listener attaches so a back-navigation popstate carries the active locale
+  // and passes the foreign-state filter.
+  stampInitialHistoryStateFn?.();
   window.addEventListener("popstate", pagesRouterPopStateHandler);
 }
--- a/packages/vinext/src/shims/router.ts
+++ b/packages/vinext/src/shims/router.ts
@@ -38,7 +38,7 @@ import {
   toBrowserNavigationHref,
   toSameOriginAppPath,
   getWindowOrigin,
-  withBasePath as withBasePathPrefix,
+  withBasePath,
 } from "./url-utils.js";
 import { stripBasePath } from "../utils/base-path.js";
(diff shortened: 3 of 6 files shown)


james-elicx:
Rebased on main and squashed the merge commit. Conflicts in `packages/vinext/src/shims/router.ts` were adjacent-import collisions; dropped an unused `withBasePath as withBasePathPrefix` alias that crept in during one of the intermediate commits. `pnpm run check` passes. CI re-running — the prior E2E failure (`router-events-not-routeChange`) may also be resolved by the rebase.

james-elicx:
Addressed all three failing checks with one commit:

- **Vitest** (`pages-router-i18n-sticky-locale.test.ts` 'Router.push without explicit locale'): `getCurrentBrowserLocale` was preferring `__VINEXT_DEFAULT_LOCALE__` over `__VINEXT_LOCALE__` when no path prefix was present, erasing the active locale on default-locale paths. Flipped the fallback order.
- **Vitest** (6 `tests/shims.test.ts` cases): updated `pushState({}, ...)` assertions to `expect.objectContaining({ __N: true })` to match the new state shape this PR introduces.
- **E2E** (`router-events.spec.ts` hash-only back/forward, 30s timeout): the Safari-replay filter on the first popstate was running even after a hash-only `Router.push`, where `state.as` (no hash) matched `_lastPathnameAndSearch` (also no hash) and incorrectly dropped a genuine `goBack()`. Gated the filter behind a new `_routerDidNavigate` flag so it only fires during the actual initial-load replay window.

`vp check` runs clean locally on the touched paths.

james-elicx:
Vitest unit failure after the rebase: two tests in `tests/link.test.ts` and `tests/shims.test.ts` (originally added in #1442) still asserted the pre-sticky priority — "non-locale-prefixed path means default locale". That contradicts the intentional flip in ab662c07 (`getCurrentBrowserLocale` now prefers `__VINEXT_LOCALE__` over `__VINEXT_DEFAULT_LOCALE__` to keep sticky locale on default-locale paths served under a non-default locale).

Updated both tests to assert the new behavior (`/id` is preserved). No production code touched.

Issue #1330, opened 2026-05-20
Pages Router: `_next/data` JSON endpoint not implemented


james-elicx:
> *This issue was created by an agent analysing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/26163751905) (vinext `main` vs Next.js `v16.2.6`, 2026-05-20).*

## Problem

The Pages Router production server does not handle `/_next/data/[buildId]/[page].json` requests. In Next.js, client-side navigations in the Pages Router fetch page data from this endpoint, which returns a JSON object like `{"pageProps":{"message":"Hello"}}`.

When vinext receives these requests, it either:
1. Returns **404** — does not recognize the `_next/data` URL pattern
2. Returns **HTML** — serves the full rendered page instead of the JSON data envelope

Additionally, middleware URL normalization is missing: middleware sees the raw `/_next/data/.../page.json` path instead of the normalized page path (e.g., `/page`).

```
FetchError: invalid json response body at
  http://127.0.0.1:33447/_next/data/e6777d6d-.../es/about.json
  reason: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```

```
Expected: "/send-url/"
Received: "/_next/data/49d16a23-.../send-url.json"
```

## Estimated Impact

~30 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/middleware-trailing-slash/test/index.test.ts`
- `test/e2e/middleware-general/test/index.test.ts`
- `test/e2e/middleware-general/test/node-runtime.test.ts`
- `test/e2e/middleware-rewrites/test/index.test.ts`
- `test/e2e/middleware-redirects/test/index.test.ts`
- `test/e2e/middleware-redirects/test/node-runtime.test.ts`
- `test/e2e/edge-pages-support/index.test.ts`
- `test/e2e/prerender.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test case that makes a request to `/_next/data/[buildId]/[page].json` on a Pages Router app and asserts it returns JSON with the correct `pageProps`. Confirm it fails before implementing the fix.

2. **Implement a `_next/data` route handler in `server/prod-server.ts`.** Intercept requests matching `/_next/data/[buildId]/[page].json`, extract the page path, call the page's `getServerSideProps`/`getStaticProps`, and return the result as `{ pageProps: ... }` JSON.

[...]

PR #1412, opened 2026-05-21
feat(pages-router): consume _next/data JSON endpoint from the client

james-elicx:
The server-side `/_next/data/<buildId>/<page>.json` endpoint landed in #1384, but the vinext client was still fetching the full HTML page on every navigation and regex-extracting `__NEXT_DATA__` from it. This wires the client to actually use the JSON endpoint, matching what Next.js does and avoiding the HTML round trip.

Closes the follow-up half of #1330.

## What changed

**Client navigation** (`shims/router.ts`). `navigateClient` is split into two paths:

- `navigateClientData` (new): when `window.__VINEXT_PAGE_LOADERS__` has a registered code-split loader for the target route AND we have a buildId, fetch `/_next/data/<buildId>/<page>.json` with `Accept: application/json` + `x-nextjs-data: 1`. On 200 parse the envelope, look up the page module via the in-memory loader map, and re-render.
- `navigateClientHtml`: the previous HTML-fetch + regex-extract logic, unchanged. Used in dev (where the inline hydration script does not populate the loader map) and as a fallback for any route not in the client loader map.

Failure modes — 404, 5xx, network error, malformed JSON, soft redirect via `x-nextjs-redirect`, missing loader — all queue a hard navigation (`window.location.href = url`). This is the deploy-skew safety net: when the server's buildId has rotated, the data endpoint returns 404 and the client lands on the new build via a full document load.

**Code-split loader manifest** (`entries/pages-client-entry.ts`). Vite already code-splits each page into its own chunk via the dynamic imports in `pageLoaders`. The generated client entry now exposes that map on `window.__VINEXT_PAGE_LOADERS__` (route pattern → loader thunk), plus `window.__VINEXT_PAGE_PATTERNS__` (ordered list) and `window.__VINEXT_APP_LOADER__` (the `_app` thunk, when present). This replaces Next.js' `_buildManifest.js` indirection — vinext already generates the equivalent at build time, we just had to expose it.

**Data URL construction** (`shims/internal/pages-data-url.ts`, new). Pure helper module mirroring `getAssetPathFromRoute` + `getDataHref` from Next.js:

| Route | Data URL |
|---|---|
| `/` | `/_next/data/<id>/index.json` |
| `/about` | `/_next/data/<id>/about.json` |
| `/index` | `/_next/data/<id>/index/index.json` |
| `/blog/foo` | `/_next/data/<id>/blog/foo.json` |

Round-trip tests verify the server's `parseNextDataPathname` agrees with the client's `buildPagesDataPath` for every shape — and surfaced a latent server bug: `/_next/data/<id>/index/foo.json` was parsing back to `/index/foo` instead of `/foo`. Fixed in `pages-data-route.ts` so the round-trip is now lossless.

**Prefetch** (`shims/link.tsx` + `shims/router.ts`). `<Link>` prefetch and `Router.prefetch()` now prefer the data path:

- Inject `<link rel=\"prefetch\" as=\"fetch\" crossorigin href={dataUrl}>` for the JSON, matching the actual navigation's request shape.
- Invoke the loader thunk to warm the chunk in parallel. Vite's `import()` caches the result, so the eventual click is a double cache hit.

When no loader matches (dev, unmapped route), fall back to the legacy `<link rel=\"prefetch\" as=\"document\">` so the browser at least preloads the HTML.

This also fixes a long-standing prefetch bug as a side effect: the previous Pages branch of `link.tsx`'s prefetch check (`__NEXT_DATA__.__vinext?.pageModuleUrl`) was only populated by the dev server's inline hydration script, so prefetch in production was a no-op.

## Ports from Next.js

- `getDataHref` in `.nextjs-ref/packages/next/src/client/page-loader.ts:154`
- `fetchNextData` in `.nextjs-ref/packages/next/src/shared/lib/router/router.ts:498`
- `getAssetPathFromRoute` in `.nextjs-ref/packages/next/src/shared/lib/router/utils/get-asset-path-from-route.ts`

## Tests

- `tests/pages-data-url.test.ts` (new): 27 tests covering URL construction and the client/server round trip.
- `tests/shims.test.ts` (+8 new tests): JSON-path navigation success, 404 hard-reload (build-skew), `x-nextjs-redirect` soft-redirect handling, HTML fallback when no loader matches, dynamic param extraction, malformed JSON hard-reload, prefetch JSON path, prefetch fallback.

Verified locally: 1393 vitest tests pass across shims, pages-router, link/navigation, prefetch, data-route, data-url, entry-templates, and routing test files.

## Out of scope (potential follow-ups)

- Dev server JSON-path navigation. Dev injects per-page inline hydration scripts and does not populate the loader map, so dev navigation still uses the HTML path. To bring dev up to parity we would emit a small loaders manifest alongside the dev hydration script — a separate small PR.
- Soft-redirect handled as a client-side `Router.replace` instead of a hard reload. Today the `x-nextjs-redirect` header hard-loads; a future PR can wire it through `Router.replace` to keep the redirect a client-side navigation.


james-elicx:
Thanks for the review. Addressed in 9029611f:

- **`__NEXT_DATA__.query` missing search params** — `navigateClientData` now merges parsed search params with the dynamic route params (route params win on key collision, mirroring `getPathnameAndQuery()` / Next.js). Added a regression test in `tests/shims.test.ts` that navigates to `/blog/my-post?slug=spoofed&q=hello&tag=a&tag=b` and asserts `query` is `{ slug: "my-post", q: "hello", tag: ["a", "b"] }`.
- **`__VINEXT_PAGE_PATTERNS__` ordering** — confirmed `pagesRouter()` already sorts via `compareRoutes` (static → dynamic → catch-all → optional catch-all), so `Object.keys(pageLoaders)` inherits that order and `matchPagesPattern()` can trust first-match. The comment in `pages-client-entry.ts` was misleading about *where* the sort happens — updated it to reference the build-time sort.
- **`x-nextjs-redirect` dead code** — agree, follow-up; the current hard-reload-on-bad-JSON path is functionally correct but indirect. Tracking as out-of-scope for this PR (along with the soft-redirect-as-client-side-replace work already called out in the description).
- **`if (!root)` duplication** — leaving as-is; the guard is in two places but each only runs once per nav, and hoisting requires both navigation paths to share a pre-check site that today doesn't exist.

/bigbonk review

james-elicx:
Added an E2E regression guard in 5529ccc5 ([`tests/e2e/cloudflare-pages-router/navigation.spec.ts`](../blob/james/_next/data-client/tests/e2e/cloudflare-pages-router/navigation.spec.ts)). Unit tests covered the `navigateClientData` logic, but none of the existing browser-level tests asserted that a Link click on a hydrated app *actually* hits `/_next/data/<buildId>/<page>.json` — they only checked the URL changed and the new page rendered, both of which still pass if the loader manifest silently regresses and we fall back to the HTML path. The new test fails loudly in that scenario by sniffing the network: ≥1 request to the JSON endpoint, 0 document-type requests to `/ssr`.

/bigbonk review

james-elicx:
Moved the `_next/data` request handling out of the example worker and into the generated server entry per feedback — the user-written worker template stays thin. `runMiddleware` and `renderPage` now auto-detect data URLs via a shared `__normalizePagesDataRequest` helper in the virtual entry; middleware sees the normalized page path, the worker sees the normalized URL via `result.rewriteUrl`, and `renderPage` falls back to detecting from `request.url` when there's no middleware. Buildid mismatches short-circuit to a JSON 404 in both paths.

Merged latest main and resolved the conflict in `pages-server-entry.ts` (kept both my `__dataNorm.request` change and main's new `trailingSlash` arg to `__runGeneratedMiddleware`).

All 37 CI checks green including the new E2E regression guard.

/bigbonk review

james-elicx:
Addressed two of the three review nits in 0a2760aa:

- **Prefetch duplication** — extracted `resolvePagesDataNavigationTarget` + `prefetchPagesData` into `shims/internal/pages-data-target.ts`. Both `Router.prefetch()` and `<Link>` prefetch now go through the shared helper; the navigateClient hot path uses the same resolver, so there is only one decision site for "can we use the JSON fast path?".
- **Locale transitions skip the data path** — the target resolver now strips the locale prefix before matching `__VINEXT_PAGE_PATTERNS__`, so navigations like `/fr/about` resolve the locale-unaware `/about` loader and take the JSON fast path. The data URL keeps the locale prefix so the server still picks locale-specific gSSP data. When the URL implies a different locale than the current `__NEXT_DATA__.locale`, the synthesised `__NEXT_DATA__` now refreshes its `locale` field (the JSON envelope only carries `pageProps`, so the locale has to come from the URL) — without this, `useRouter().locale` stayed stale after a client-side locale transition. Added two regression tests in `tests/shims.test.ts`: locale-prefixed nav (`/fr/about` → `/_next/data/<id>/fr/about.json`, page `/about`, locale `"fr"`) and default-locale-restore (`Router.push("/about", undefined, { locale: "en" })` from `/fr` drops the locale prefix and resets `locale` back to `"en"`).
- **gSSP/gSP redirect protocol** — leaving as-is. The PR description and the prior reviews already track this as out-of-scope; the real fix is server-side (emit `200` + `{ __N_REDIRECT, __N_REDIRECT_STATUS }` JSON for data-request redirects instead of raw 3xx), which belongs in a separate PR.

All 4738 unit tests pass locally including the new locale regression tests.

/bigbonk review

PR #1591, opened 2026-05-25
feat: payloadcms

z-zahi:
Payloadcms example working with vinext working with local sqlite, can be configured to use cloudflare d1 or other database using  adapters

https://payloadcms.com/docs/database/overview

Screenshot 2026-05-25 at 6 54 06 PM


james-elicx (changes requested):
i would be interested in us getting the payload cms cloudflare d1 working as an ecosystem fixture.

any payload fixture should be based on their create-payload-app output and include their auto-generated tests that run in CI.

I pushed some changes to https://github.com/cloudflare/vinext/compare/main...james/payload-cms to show what I mean.

It currently fails to run due to the `file-type` dep used in Payload.

z-zahi:
I made the changes against [main...james/payload-cms](https://github.com/cloudflare/vinext/compare/main...james/payload-cms), it now works with d1 and r2, fixed the file-type and `@payloadcms/ui/elements/HTMLDiff/diff` temporaraly, unfortunately vinext is not detecting exports correclty so I added aliases, the tests pass and the app works very smoothly, I'd like to know your opinion and if there's something missing @james-elicx

james-elicx:
Would you be able to do it in separate commits so we can track exactly what needs to change from the Payload-created template for their apps to work? It's not clear in a single commit how much things change from their template.

> unfortunately vinext is not detecting exports correclty so I added aliases

This sounds like a potential bug or issue that we should ideally look at fixing

PR #1597, opened 2026-05-26
fix(app-router): preserve encoded prerender params

### LINKED ISSUE #1544: Prerender: URL-encoded route params decoded before rendering

## Problem

When a route param contains URL-encoded characters (e.g. `sticks%20%26%20stones`), vinext's prerender output decodes the value to `sticks & stones` before passing it to the component. Next.js preserves the encoded form so user code receives the canonical encoded string.

```
Expected: "params.id is sticks%20%26%20stones"
Received: "params.id is sticks & stones"
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/app-dir/prerender-encoding/prerender-encoding.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Render `[id]/page.tsx` with `id = 'sticks%20%26%20stones'` and assert `params.id` is the encoded string.

2. **Preserve URL encoding in params.** Do not URL-decode dynamic segment values before passing into the component. Apply both at prerender time and at request time.

---

Part of #1328.

NathanDrake2406:
## Overview

| Item | Detail |
| --- | --- |
| Goal | Match Next.js for App Router prerendered dynamic params that require URL encoding. |
| Core change | Preserve encoded `generateStaticParams` values through prerender rendering, while keeping decoded params for `dynamicParams = false` validation. |
| Runtime boundary | Seed prerendered App Router artifacts under the same normalized cache key used by live production requests. |
| Route-param proof | The prerender-only encoded params are only used when they prove both the final route pattern and the final decoded matched params. |
| Primary files | `build/prerender.ts`, `server/app-rsc-handler.ts`, `server/seed-cache.ts`, `server/prerender-route-params.ts` |
| Impact | The upstream `prerender-encoding` deploy test now serves the prerendered HTML instead of falling through to a decoded dynamic render. |

## Why

App Router prerender has two separate path shapes that both need to remain true: the artifact path on disk is URL-encoded, while the live production request path is normalized before cache lookup. The page render itself also needs the encoded param value for the prerendered artifact, but static-param validation must still compare decoded matched params against the raw `generateStaticParams` result.

| --- | --- | --- |
| Prerender render | The prerendered page should receive the encoded dynamic params associated with the concrete output path. | Adds a trusted prerender-only route-param header that is created by the build prerenderer and stripped from external input. |
| Route matching authority | Encoded prerender params are route-relative and must describe the final matched route after middleware/config rewrites. | Binds the header payload to the source route pattern and only applies it when decoded payload params equal the final matched params, including catch-all arrays. |
| Static validation | `dynamicParams = false` should validate against the decoded route match, not against the encoded render value. | Passes a separate decoded validation param object through page dispatch. |
| Production cache seed | The first live request must hit the artifact generated at build time. | Reads artifacts from encoded paths but seeds ISR cache keys using the runtime-normalized pathname. |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| `/sticks%20%26%20stones` from `generateStaticParams()` | Build/runtime route matching decoded params, and production cache seeding missed the live request key. The page rendered `sticks & stones`. | The prerender render sees `sticks%20%26%20stones`, and production startup seeds the cache under `/sticks & stones`, so the first request serves the prerendered artifact. |
| Rewrites during prerender render | Encoded params could float across rewrite boundaries if they were not tied to the final route match. | Encoded params are ignored unless their route pattern matches and their decoded params exactly match the final route params. |
| Forged prerender route-param headers | No such header existed. | Header is only honored during `VINEXT_PRERENDER=1` with the prerender secret, is filtered from normal requests, and is stripped from external rewrites. |

## Validation

- `vp test run tests/app-rsc-handler.test.ts -t "same-pattern rewrite"` (red before the decoded-match guard, green after)
- `vp test run tests/prerender-route-params.test.ts tests/app-rsc-handler.test.ts -t "prerenderRouteParamsPayloadMatchesRoute|same-pattern rewrite|encoded prerender route params|different rewritten route pattern"`
- `vp test run tests/seed-cache.test.ts tests/prod-server-prerender-seeder.test.ts tests/app-rsc-handler.test.ts tests/prerender-route-params.test.ts tests/static-export.test.ts tests/prerender.test.ts tests/request-pipeline.test.ts`
- `vp test run --project unit`
- `vp check`
- `vp run knip`
- `VINEXT_DIR="$PWD" ADAPTER_DIR="$PWD" vp env exec --node 24 ./scripts/run-nextjs-deploy-suite.sh /Users/nathan/Projects/vinext/.refs/nextjs-v16.2.6 --retries 0 -c 1 --debug test/e2e/app-dir/prerender-encoding/prerender-encoding.test.ts`

## References

| --- | --- |
| [Next.js prerender encoding test](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/prerender-encoding/prerender-encoding.test.ts) | Upstream observable behavior this ports. |
| [Next.js fixture page](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/prerender-encoding/app/%5Bid%5D/page.tsx) | Shows `dynamicParams = false` and `generateStaticParams()` returning `sticks & stones`. |
| [Next.js static path replacements](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/static-paths/app.ts#L723-L738) | Shows encoded and non-encoded path construction for App Router prerender routes. |
| [Next.js app prerendered route generation](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/static-paths/app.ts#L968-L1055) | Shows how prerendered routes retain concrete path data. |
| [Next.js route matcher decoding](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/router/utils/route-matcher.ts#L17-L43) | Confirms normal live route matching decodes params, which must remain distinct from prerender artifact params. |
| [Next.js `generateStaticParams` docs](https://github.com/vercel/next.js/blob/canary/docs/01-app/03-api-reference/04-functions/generate-static-params.mdx) | Documents that returned params are used to statically generate dynamic routes. |

Closes #1544

### DIFF
--- a/packages/vinext/src/build/prerender.ts
+++ b/packages/vinext/src/build/prerender.ts
@@ -40,7 +40,15 @@ import { runWithHeadersContext, headersContextFromRequest } from "vinext/shims/h
 import { createValidFileMatcher, findFileWithExtensions } from "../routing/file-matcher.js";
 import { normalizeStaticPathsEntry, type StaticPathsEntry } from "../routing/route-pattern.js";
 import { navigationRuntimeRscBootstrapExpression } from "../server/app-ssr-stream.js";
-import { VINEXT_PRERENDER_SECRET_HEADER } from "../server/headers.js";
+import {
+  VINEXT_PRERENDER_ROUTE_PARAMS_HEADER,
+  VINEXT_PRERENDER_SECRET_HEADER,
+} from "../server/headers.js";
+import {
+  encodePrerenderRouteParams,
+  serializePrerenderRouteParamsHeader,
+  type PrerenderRouteParams,
+} from "../server/prerender-route-params.js";
 import { startProdServer } from "../server/prod-server.js";
 import { readPrerenderSecret } from "./server-manifest.js";
 import { getOutputPath, getRscOutputPath } from "../utils/prerender-output-paths.js";
@@ -1014,6 +1022,7 @@ export async function prerenderApp({
       urlPath: string;
       /** The file-system route pattern this URL was expanded from (e.g. `/blog/:slug`). */
       routePattern: string;
+      prerenderRouteParams: PrerenderRouteParams | null;
       revalidate: number | false;
       isSpeculative: boolean; // 'unknown' route — mark skipped if render fails
     };
@@ -1158,6 +1167,7 @@ export async function prerenderApp({
             urlsToRender.push({
               urlPath,
               routePattern: route.pattern,
+              prerenderRouteParams: encodePrerenderRouteParams(route.pattern, params),
               revalidate,
               isSpeculative: false,
             });
@@ -1178,6 +1188,7 @@ export async function prerenderApp({
         urlsToRender.push({
           urlPath: route.pattern,
           routePattern: route.pattern,
+          prerenderRouteParams: null,
           revalidate: false,
           isSpeculative: true,
         });
@@ -1186,6 +1197,7 @@ export async function prerenderApp({
         urlsToRender.push({
           urlPath: route.pattern,
           routePattern: route.pattern,
+          prerenderRouteParams: null,
           revalidate,
           isSpeculative: false,
         });
@@ -1203,6 +1215,7 @@ export async function prerenderApp({
     async function renderUrl({
       urlPath,
       routePattern,
+      prerenderRouteParams,
       revalidate,
       isSpeculative,
     }: UrlToRender): Promise<PrerenderRouteResult> {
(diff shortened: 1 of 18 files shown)


NathanDrake2406:
@codex review

NathanDrake2406:
Thanks for the reviews. Rebased onto latest `main` (resolved the `tests/static-export.test.ts` conflict by keeping both new `it()` blocks) and pushed `8e3be5cc`.

### Addressed

- **`readTrustedPrerenderRouteParams` skips secret validation** (ask-bonk): Confirmed this is intentional. `prod-server`'s `nodeToWebRequest` is the only entry point that touches raw external input, and it calls `readTrustedPrerenderRouteParamsFromHeaders` *with* `expectedSecret`, then re-serializes the validated payload onto a clean header. Every downstream read operates on an already-trusted request. Took option 2: added a comment on the wrapper documenting that verification happens once at the outer boundary and warning against calling it on unverified input.
- **Double read-then-reattach coupling** (ask-bonk): Added a comment in `createAppRscHandler` explaining why the secret header must stay out of `VINEXT_INTERNAL_HEADERS` for the second read in `handleAppRscRequest` to succeed.
- **Empty-params to null contract** (ask-bonk, line 36): Added a doc comment on `serializePrerenderRouteParamsHeader` noting the empty-params bail-out and why it is safe given the only producer.

### Not changing (with rationale)

- **`for...in` in `isPrerenderRouteParams`** (Copilot): False positive, as ask-bonk already noted. The code uses `Object.entries()`, which only visits own enumerable properties, and `isUnknownRecord` rejects arrays/null. No change needed.
- **`decodeHtmlText` decodes `&amp;` last** (Copilot): The concern is real for arbitrary nested entities, but this is a test-only helper and the actual fixture data (`sticks & stones` rendered as `sticks &amp; stones`) has no nested encoding, so the assertions are correct and stable. Decoding `&amp;` first would just introduce the inverse bug; a full entity decoder is overkill for this fixture. Leaving as-is.
- **Pattern-parsing duplication** and **`withEnv` test helper** (ask-bonk, non-blocking): Agreed these are reasonable cleanups, but both are cross-cutting refactors (the `extractParamNamesFromPattern` extraction touches three routing files) that are out of scope for this fix. Tracking as follow-ups rather than expanding this PR.
- **Shared `textContentByTestId`/`decodeHtmlText` test util** (Copilot, low confidence): Same reasoning, a test-helper consolidation that is better done as its own change.

`vp check`, `vp run knip`, and the targeted unit suites pass.

PR #1598, opened 2026-05-26
fix(pages): preserve import.meta.url source paths

### LINKED ISSUE #1505: `import.meta.url` resolves to bundled entry path instead of user source path

## Problem

`import.meta.url` should resolve to the user's source file path (e.g., `file:///ROOT/pages/index.tsx`) on the server, and a stable URL on the client. vinext returns the bundled entry path `file:///tmp/.../dist/server/entry.js?t=...` because Rolldown/Vite collapses all modules into one file.

```
import.meta.url returns bundled entry path, not source path
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/import-meta/import-meta.test.ts` (2 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add `console.log(import.meta.url)` in a page and assert it resolves to the source path.

2. **Preserve per-module `import.meta.url`.** Either keep modules split during the build, or replace `import.meta.url` references with the user source path during the transform pass.

NathanDrake2406:
## Overview

| Field | Detail |
| --- | --- |
| Goal | Match Next.js production behavior for direct `import.meta.url` in Pages Router modules. |
| Core change | Rewrite direct `import.meta.url` reads in user modules to source-file URLs during Vite transforms. |
| Runtime boundary | Server/RSC/SSR builds receive real `file:///.../source` URLs, while client builds receive Turbopack-style `file:///ROOT/...` URLs. |
| Primary files | `packages/vinext/src/plugins/import-meta-url.ts`, `packages/vinext/src/server/prod-server.ts`, `tests/pages-router.test.ts` |
| Expected impact | Pages Router production and prerendered output no longer expose generated bundle URLs for direct `import.meta.url`. |

## Why

`import.meta.url` is module identity. For Next.js compatibility, user code should observe the source module URL, not the generated server entry or bundled output URL. This matters in both SSR HTML and hydrated browser output, because the same component can render once on the server and again from the client page chunk.

| --- | --- | --- |
| User module transforms | Direct `import.meta.url` is a source-module value. | Adds `vinext:import-meta-url` to replace direct reads before bundling changes the module identity. |
| Asset handling | `new URL("./asset", import.meta.url)` remains owned by Vite and existing vinext asset plugins. | The AST rewrite skips `import.meta.url` when it is the `new URL()` base argument. |
| Pages prerender | Prerendered Pages HTML must hydrate with the page client chunk. | Node production startup now reads `.vite/manifest.json` and publishes `__VINEXT_CLIENT_ENTRY__`, matching the Cloudflare worker injection path. |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| Pages SSR render | `import.meta.url` could resolve to `dist/server/entry.js`. | Server render sees a real source `file:///.../pages/index.tsx` URL. |
| Pages browser hydrate | The page chunk did not match Next Turbopack's `/ROOT` source URL. | Client page chunks render `file:///ROOT/pages/index.tsx`. |
| `vinext build --prerender-all` | Prerendered Pages HTML could omit the client entry and never hydrate. | Node production/prerender servers inject the Pages client entry from the build manifest. |

<details>

1. `packages/vinext/src/plugins/import-meta-url.ts`: check the AST traversal, source URL selection, symlink canonicalization, and `new URL()` skip.
2. `packages/vinext/src/index.ts`: check plugin placement before `vinext:og-inline-fetch-assets` still leaves asset bases available for that pre-transform.
3. `packages/vinext/src/server/prod-server.ts`: check build-manifest parsing and `__VINEXT_CLIENT_ENTRY__` setup for Node Pages Router production and prerender.
4. `tests/pages-router.test.ts`: check the ported Next.js behavior test, including prerender and browser hydration.
5. `tests/import-meta-url.test.ts`: check low-level guard coverage for `/ROOT`, real source URLs, and asset base preservation.

</details>

<details>
<summary>Validation</summary>

- `vp test run tests/import-meta-url.test.ts`
- `vp test run tests/pages-router.test.ts -t "Pages Router import.meta.url in production"`
- `vp test run tests/og-inline.test.ts`
- `vp check`
- `vp env exec --node 24 ./scripts/run-nextjs-deploy-suite.sh /Users/nathan/Projects/vinext/.refs/nextjs-v16.2.6 --retries 0 -c 1 --debug test/e2e/import-meta/import-meta.test.ts`

</details>

<details>

- Public API: none.
- Runtime: the transform is limited to user modules under the project root and excludes `node_modules`, `dist`, `.next`, and `.vinext-local-package`.
- Asset compatibility: `new URL("./asset", import.meta.url)` is intentionally left intact so Vite and existing vinext asset transforms continue to own asset resolution.
- Pages production: Node server startup now mirrors the existing Cloudflare worker behavior by exposing the Pages client entry globally from the build manifest.

</details>

## References

| --- | --- |
| [Next.js import-meta deploy test](https://github.com/vercel/next.js/blob/v16.2.6/test/e2e/import-meta/import-meta.test.ts) | Defines the upstream server and browser expectations ported here. |
| [Next.js import-meta fixture](https://github.com/vercel/next.js/blob/v16.2.6/test/e2e/import-meta/pages/index.tsx) | Shows the exact direct `import.meta.url` user-code shape under test. |
| [Turbopack `import.meta` binding](https://github.com/vercel/next.js/blob/v16.2.6/turbopack/crates/turbopack-ecmascript/src/references/esm/meta.rs) | Shows Turbopack replacing `import.meta.url` from the source module path. |
| [Turbopack `/ROOT` path helper](https://github.com/vercel/next.js/blob/v16.2.6/turbopack/crates/turbopack-ecmascript/src/references/mod.rs) | Explains why browser output uses `file:///ROOT/...`. |
| [Next.js non-root import-meta assertions](https://github.com/vercel/next.js/blob/v16.2.6/test/e2e/app-dir/non-root-project-monorepo/non-root-project-monorepo.test.ts) | Confirms the split between real server file URLs and client `/ROOT` URLs. |

Closes #1505


NathanDrake2406:
@codex review

NathanDrake2406:
Addressed review feedback in c4d53f40 and b76cf19f:

1. **Double `realpathSync` per module** (ask-bonk #1, Copilot): `transformableModuleCanonicalId` now returns the canonical id and threads it through to `importMetaUrlValue`, so each rewritten module canonicalizes once instead of twice. The early `!code.includes("import.meta.url")` bail and the cached root paths Copilot suggested were already in place.

2. **`runPrerender` in the top-level-await test** (ask-bonk #2): removed. Verified the test asserts identical output whether the routes are prerendered or rendered on demand, so the prerender step was unrelated build work. The new `__VINEXT_CLIENT_ENTRY__` behavior reads `.vite/manifest.json` (build output), not prerender output, so it does not depend on this step.

3. **`isNewUrlExpression` scope** (ask-bonk #3, Copilot): added a comment noting it intentionally matches only the bare `URL` identifier, matching Vite's own asset-detection scope.

4. **Redundant `code.includes` check**: removed the third check inside the shared rewrite core; the `transform` hook and the public `rewriteImportMetaUrl` export each bail early on their own.

5. **`.mts` skipped** (Copilot): widened the extension match to also cover `.cts`, `.mts`, and `.cjs`.

The remaining notes (the `Parameters<typeof computeLazyChunks>[0]` coupling and the defensive global cleanup) were observations rather than requested changes, so no code change there.

packages/vinext/src/plugins/import-meta-url.ts
```diff
@@ -0,0 +1,242 @@
...
+  const canonicalId = canonicalizePath(id);
+  const normalizedId = normalizePath(canonicalId);
+  if (!isPathWithin(normalizedId, rootPaths.normalizedRoot)) return null;
+
+  const relativePath = normalizePath(path.relative(rootPaths.canonicalRoot, canonicalId));
+  const firstSegment = relativePath.split("/")[0];
+  if (
+    firstSegment === ".next" ||
+    firstSegment === ".vinext-local-package" ||
+    firstSegment === "dist"
```
james-elicx: what about if someone uses a custom output directory at some point? should we also include .vinext here? should we just be handling everything gitignored here?

packages/vinext/src/plugins/import-meta-url.ts
```diff
@@ -0,0 +1,242 @@
...
+  };
+}
+
+// Returns the canonical module id when the module is eligible for rewriting,
+// or null otherwise. Threading the canonical id back to the caller avoids a
+// second realpathSync when computing the replacement value.
+function transformableModuleCanonicalId(id: string, rootPaths: RootPaths): string | null {
+  if (!id || id.startsWith("\0")) return null;
+  if (!path.isAbsolute(id)) return null;
+  if (id.includes("/node_modules/") || id.includes("\\node_modules\\")) return null;
```
james-elicx: we could just do this against the normalise path that we're normalising below couldn't we?

james-elicx:
> _This review was generated by AI agents. Treat findings as starting points to verify, not conclusions._

## Real issues worth fixing

**1. `findClientEntryFile` is missing the on-disk fallback that the Cloudflare worker path has.**

In `packages/vinext/src/index.ts:4225-4238` the Cloudflare build explicitly falls back to scanning the on-disk assets directory for `vinext-client-entry`/`vinext-app-browser-entry` chunks "when the SSR manifest lookup didn't surface one." That fallback exists because the manifest-only lookup *does* sometimes miss the entry.

The new `packages/vinext/src/server/prod-server.ts:1382-1393` only does the manifest-only lookup. In any configuration where the Cloudflare fallback would have fired, the Node prod-server (and prerender, which now also runs through it) will produce HTML with `__VINEXT_CLIENT_ENTRY__ === undefined` — i.e. no hydration `<script>`. That's the exact regression this PR is trying to fix.

Suggest extracting the Cloudflare logic into a shared helper and using it from both call sites.

**2. The `assetBase` argument passed to `findClientEntryFile` differs from what the Cloudflare path passes.**

`prod-server.ts:1473` passes `assetBase = basePath ? \`\${basePath}/\` : "/"`, while `index.ts:4136` uses `envConfig.base ?? "/"`. They happen to agree today because vinext sets Vite `base` from `basePath` (`index.ts:1626`), so this is latent — but if `assetPrefix` ever drives Vite `base` separately from `basePath`, the two emitted entry URLs diverge. Not a blocker; worth a comment so the next person knows they're meant to track.

**3. `import.meta?.url` (optional chain) won't be rewritten.**

`import-meta-url.ts:223-232` requires `value.type === "MemberExpression"`. Acorn wraps `import.meta?.url` in a `ChainExpression` containing an `OptionalMemberExpression`/`MemberExpression` with `optional: true`. Rare in practice but real, and silently passing through leaks the bundle URL.

**4. Aliasing breaks the asset-detection skip.**

\`\`\`js
const u = import.meta.url;
const a = new URL("./asset", u);
\`\`\`

The `import.meta.url` on line 1 is bare (not arg 1 of a `new URL`), so it gets rewritten to a string literal. That kills any downstream asset detection. This was *already* broken pre-PR (Vite's asset transform requires the literal pattern too), so it's not a regression — but worth a docs callout that asset URLs must use the literal `new URL("./asset", import.meta.url)` form.

**5. `import.meta` destructuring isn't handled.**

`const { url } = import.meta;` leaves `url` pointing at the bundle. Same severity as #3.

## Smaller things

- `import-meta-url.ts:185` — `if (index === 1 && isImportMetaUrlNode(args[index])) continue;` correctly skips arg-1 `import.meta.url`, but the loop still visits `value.callee` afterwards. For `new URL` the callee is the `URL` Identifier — harmless, just dead work.
- `import-meta-url.ts:46` — pre-check uses `code.includes("import.meta.url")`, which misses `import.meta . url` with whitespace. Plausibly fine; minified user code won't insert whitespace.
- `prod-server.ts:1480-1481` — assigning `undefined` to `globalThis.__VINEXT_CLIENT_ENTRY__` is fine in normal operation, but two prod servers spun up in the same process (as tests do) clobber each other's globals. Pre-existing pattern; the PR just extends it.
- `enforce: "post"` is the right call (Vite's `vite:asset` transform must run first), but the plugin returns a fresh sourcemap without composing the upstream one. Vite's plugin runner usually chains these, but worth verifying source maps survive into a real build.
- The regex `/\.[cm]?[jt]sx?$/` skips `.mdx`, `.vue`, `.svelte`. Almost certainly intentional for a Next-compat layer; just noting it.
- `tests/e2e/pages-router-prod/production.spec.ts:171-173` — `decodeHtmlText` only decodes `&quot;` and `&amp;` and applies them in an order that mishandles `&amp;quot;` (would decode the inner `&quot;` first). Fine for the JSON shape under test, but brittle if the fixture ever contains a literal `&`.

## Overall

The core approach (post-transform AST rewrite of bare `import.meta.url`, with `new URL` arg-1 left intact) is sound and matches what Turbopack does. Issue #1 is the only one I'd block on — it directly undermines the PR's prerender-hydration goal in any project that lands in the manifest-miss case.

james-elicx:
/bigbonk review feedback addressed in c2ac3888 — please re-review for issues

PR #1612, opened 2026-05-27
fix(app-router): dev-only favicon.ico short-circuit to avoid expensive 404 render

### LINKED ISSUE #1550: Favicon short-circuit returns `text/plain` instead of falling through to user `favicon` route

## Problem

When a user has their own `app/favicon.ico` route returning HTML, Next.js does not short-circuit and serves the user content (Content-Type `text/html; charset=utf-8`). vinext short-circuits the favicon request with `text/plain;charset=UTF-8`, bypassing the user's route entirely.

```
Expected: Content-Type `text/html; charset=utf-8` from user favicon route; got `text/plain;charset=UTF-8`
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/favicon-short-circuit/favicon-short-circuit.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add an App Router route at `app/favicon.ico` returning HTML and assert the response is the user content.

2. **Only short-circuit when no user route exists.** When the App Router or Pages Router defines a favicon route, fall through to it instead of returning vinext's default text response.

---

Part of #1328.

Divkix:
Browsers auto-request /favicon.ico on every page load. Without this short-circuit, vinext compiles and renders the full not-found page for every favicon request in development.

Next.js has the same behavior in `packages/next/src/server/lib/router-server.ts`.

- In dev: return an empty 404 when no route matches `/favicon.ico`
- In prod: fall through to the normal 404 page as before
- User-defined `app/favicon.ico/route.ts` and metadata favicon files still take priority

Fixes #1550


james-elicx:
@ask-bonk thanks for the review. Addressed in [`4e7de9b4`](https://github.com/cloudflare/vinext/commit/4e7de9b4):

**Re: cleanPathname → canonicalPathname (minor precision)** — applied. The short-circuit now checks `canonicalPathname` (the original browser-requested URL), matching Next.js's `parsedUrl.pathname === '/favicon.ico'` check in `router-server.ts`. A middleware rewrite to `/favicon.ico` will now fall through to the normal not-found render rather than short-circuit.

**Re: Pages Router dev-server.ts:337 parity** — not applicable here, the `dev-server.ts` `!match → renderErrorPage` path is **unreachable** for `/favicon.ico` because `handlePagesMiddleware` in `packages/vinext/src/index.ts` (around line 2946, the comment is "Skip requests for files with extensions") forwards any path containing `.` (except `.html`) to `next()` *before* the SSR pipeline runs. So `/favicon.ico` is filtered upstream and never reaches `createSSRHandler`, regardless of whether `pages/404.tsx` exists.

I verified this empirically — a test that fetched `/favicon.ico` against a Pages Router dev server (with `pages/404.tsx` present, no `public/favicon.ico`) returned Vite's default fallback "Cannot GET /favicon.ico" rather than the rendered custom 404 HTML. Adding a defensive short-circuit at `dev-server.ts:337` would be dead code under the current architecture.

The architectural difference from Next.js: Next.js's `router-server.ts` is the single entry point for both routers, so its favicon check has to cover both. In vinext, the App Router's RSC handler (registered by `@vitejs/plugin-rsc`) is a separate middleware that does *not* short-circuit dot-extension paths — that's why the App Router needs the fix, and why the test passes there.

PR #1619, opened 2026-05-27
fix(security): bound x-vinext-mounted-slots cache key cardinality (F-PROD-1)

james-elicx:
## Severity

**HIGH** — production cache flooding / Cloudflare KV wallet attack.

## Vulnerability

Vinext invents two internal headers (`X-Vinext-Mounted-Slots`, `X-Vinext-Interception-Context`) that are legitimately set by the browser's RSC navigation client and flow into:

- The ISR RSC cache key (`appIsrRscKey` in `packages/vinext/src/server/isr-cache.ts:271-284`, via `fnv1a64`)
- Optimistic-routing template / source keys (`getOptimisticRouteTemplateKey`, `getOptimisticPrefetchSourceKey`)

Because these headers were accepted verbatim from external requests, an attacker who controls the browser can fan out unbounded distinct values:

```bash
for i in $(seq 1 100000); do
  curl -s "https://target.example/blog.rsc?_rsc=0" \\
    -H "X-Vinext-Mounted-Slots: attacker-$i" \\
    -H "RSC: 1" > /dev/null &
done
```

Each unique header value maps to a unique cache-key hash, producing one Cloudflare KV write per request. Cloudflare KV bills per write — this is a wallet attack — and the cache also fragments and pollutes content for legitimate users.

## Why we can't just strip the headers

The browser-to-server RSC fetch IS the user's external request. There is no separate internal hop where these headers are added server-side. Stripping them at the request boundary (the approach used for `INTERNAL_HEADERS` / `VINEXT_INTERNAL_HEADERS`) would break legitimate parallel-route and intercepting-route navigation.

## Fix

Input-shape validation + bounds at the boundary, before the value reaches any cache key.

**`x-vinext-mounted-slots`** (`normalizeMountedSlotsHeader`):
- Raw header value capped at **4096 bytes** (oversized → null).
- Each token must match the legitimate `slot:<name>:<treePath>` wire format produced by `encodeSlotId` (name non-empty and contains no `:`, treePath starts with `/`). Anything else (e.g. arbitrary `attacker-N` bytes) is dropped.
- Each token capped at **256 bytes**.
- Token count capped at **16** after dedup + sort.

**`x-vinext-interception-context`** (new `normalizeInterceptionContextHeader`):
- Null bytes stripped (header-injection defense, preserved from prior behavior).
- Value capped at **1024 bytes**.
- Must start with `/` (legitimate values are same-origin pathnames like `/feed`, `/photos/42` — see `resolveManifestNavigationInterceptionContext`).
- Raw whitespace is rejected.

Anything that fails validation is treated as if the header were absent, so the request is still served (just without interception or with the no-slots cache key). This matches the prior forgiving behavior for browsers that may send legitimate but stale formats during rolling deploys.

## Files changed

- `packages/vinext/src/server/app-mounted-slots-header.ts` — strict shape validation + length/count caps.
- `packages/vinext/src/server/app-interception-context-header.ts` (new) — pathname-shape validation + length cap.
- `packages/vinext/src/server/app-rsc-request-normalization.ts` — wire the new interception-context normalizer.
- `tests/app-rsc-request-normalization.test.ts` — updated existing tests to use the real wire format; added security-focused tests.
- `tests/isr-cache.test.ts` — updated existing tests to use the real wire format; added cardinality-bound tests.

## Test coverage added

| Test | What it asserts |
| --- | --- |
| `bounds RSC cache-key cardinality against attacker-supplied mounted-slot values` | 1000 distinct `attacker-${i}` values all collapse to the same `app:/feed:rsc` cache key. |
| `caps cache-key cardinality when an attacker pads the mounted-slots header` | 100 batches of 1000 legitimately-shaped tokens (each exceeding the 4 KiB cap) all collapse to the same cache key. |
| `drops tokens that do not match the legitimate slot:<name>:<treePath> wire format` | Only legitimately-shaped tokens reach the cache key. |
| `bounds the number of mounted-slot tokens that reach the cache key` | Token count cap is enforced. |
| `rejects an oversized mounted-slots header value` | Raw length cap is enforced. |
| `rejects non-pathname interception context values` | Non-pathname values are dropped. |
| `bounds cache-key cardinality across many attacker-supplied values` (interception) | 1000 distinct oversized interception values collapse to a single normalized result. |
| Updated existing tests | Confirm legitimate browser-emitted values (`slot:modal:/`, `/feed`) continue to round-trip. |

All targeted test files pass locally:

- `tests/app-rsc-request-normalization.test.ts` (56 tests)
- `tests/isr-cache.test.ts` (60 tests)
- `tests/app-rsc-cache-busting.test.ts`, `tests/app-rsc-handler.test.ts`, `tests/app-page-cache.test.ts`, `tests/app-page-element-builder.test.ts`, `tests/app-page-route-wiring.test.ts`, `tests/app-page-dispatch.test.ts`, `tests/prefetch-cache.test.ts`, `tests/request-pipeline.test.ts`, `tests/app-server-action-execution.test.ts` (284 tests combined)
- `tests/app-optimistic-routing.test.ts`, `tests/app-browser-interception-context.test.ts` (12 tests combined)

## Follow-up (out of scope)

The audit also suggests validating slot IDs against the route's declared parallel-route slots and rejecting unknown slot IDs (matching Next.js's `Next-Router-State-Tree` approach). That requires route-manifest access at the cache-key construction point and is a larger structural change tracked separately.

Audit also flagged **F-PROD-6** (`x-vinext-static-file`) as a similar response-side issue — not in this PR's scope.

---

Identified in internal security audit — see `SECURITY-AUDIT-2026-05.md` finding **F-PROD-1**.

### DIFF
--- a/packages/vinext/src/server/app-interception-context-header.ts
+++ b/packages/vinext/src/server/app-interception-context-header.ts
@@ -1,0 +1,45 @@
+/**
+ * Normalize the `x-vinext-interception-context` header from inbound requests.
+ *
+ * The browser sends the current pathname (e.g. `/feed`) as interception context
+ * so the server can decide whether to render an intercepted parallel route.
+ * The legitimate value is always a same-origin URL pathname produced by the
+ * vinext browser entry — never an arbitrary string.
+ *
+ * Security: this value flows into cache-key construction (via
+ * `getOptimisticRouteTemplateKey`, `getOptimisticPrefetchSourceKey`, and
+ * outbound RSC payload cache keys). Without bounds, an attacker who controls
+ * this header can fabricate unbounded distinct values to fragment the cache
+ * or drive per-write KV billing. See `SECURITY-AUDIT-2026-05.md` finding
+ * F-PROD-1.
+ *
+ * Bounds applied:
+ *   - Null bytes are stripped (header-injection defense).
+ *   - The value must start with `/` (a pathname).
+ *   - Whitespace is rejected (real pathnames do not contain raw whitespace;
+ *     legitimate spaces would be percent-encoded).
+ *   - Length capped at MAX_INTERCEPTION_CONTEXT_LENGTH bytes. Values that
+ *     exceed the cap are treated as absent so the request is still served,
+ *     just without interception.
+ *
+ * Anything that fails validation returns null, matching the prior behavior of
+ * an absent header. This is intentionally more permissive than rejecting the
+ * whole request — interception is a progressive enhancement.
+ */
+
+/** Hard cap on the byte length of the interception-context header value. */
+const MAX_INTERCEPTION_CONTEXT_LENGTH = 1024;
+
+export function normalizeInterceptionContextHeader(raw: string | null | undefined): string | null {
+  if (!raw) return null;
+  // Strip null bytes first so length bounds can't be evaded by padding with \0.
+  const stripped = raw.replaceAll("\0", "");
+  if (stripped.length === 0) return null;
+  if (stripped.length > MAX_INTERCEPTION_CONTEXT_LENGTH) return null;
+  // Must look like a same-origin pathname. Anything else (a full URL, a token,
+  // junk bytes) is not a legitimate value the browser would emit.
+  if (!stripped.startsWith("/")) return null;
+  // Raw whitespace is not legitimate inside a pathname.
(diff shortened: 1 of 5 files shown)


NathanDrake2406:
Is this mythos?

james-elicx:
> Is this mythos?

Opus 4.7

PR #1625, opened 2026-05-28
fix(shims/head): dedupe Pages Router <Head> tags on re-render

### LINKED ISSUE #1473: Pages Router: `next/head` manager re-executes inline scripts on re-render

## Problem

When the Pages Router head manager updates `<Script>` tags during component re-renders, Next.js dedupes by `src` and does not re-execute already-loaded scripts. vinext re-executes them, causing the test's side-effect array to grow beyond `["src-1.js"]`. The same issue applies whether or not a CSP nonce is set.

```
After re-render, scripts re-executed: side-effect array exceeds expected length
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/nonce-head-manager/index.test.ts` (2 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a Pages Router page that toggles state and re-renders a `<Script src="./once.js">`. Assert the script executes exactly once.

2. **Track loaded `<Script>` sources in the head manager.** Maintain a `Set<src>` of already-injected scripts and skip re-injection during head reconciliation.

3. **Cover the nonce variant.** Same dedup must apply when scripts carry a CSP nonce.

james-elicx:
## Summary

The Pages Router `<Head>` client manager previously purged every `[data-next-head]` element and re-appended a fresh DOM node on every re-render. For `<script src="...">` (and `<link>`) tags inside `<Head>`, this caused the browser to re-execute the script on each parent re-render — even when nothing in the head projection had actually changed.

Mirror Next.js's [`head-manager.ts:updateElements`](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/head-manager.ts) reconciliation:

- Bucket desired and existing managed tags by tag type.
- For each desired tag, look for an equal existing tag (`isEqualNode`) and keep the live DOM node when matched.
- Only remove the genuinely-stale existing tags and only append the genuinely-new ones.

Also handle the browser's nonce-stripping quirk (Chrome/Firefox strip `nonce` from the serialised attribute once an element is in the document) so CSP-tagged scripts still compare equal across renders — matches the upstream behaviour.

## Test plan

- [x] New unit tests in `tests/head.test.ts` cover:
  - Identical script tags preserved across re-renders (same DOM node).
  - Tags with a different `src` produce a fresh node.
  - Nonce-tagged scripts dedupe across re-renders.
  - Stale managed tags removed when no longer desired.
  - Unmanaged head children (no `data-next-head`) left untouched.
  - End-to-end shape of the upstream e2e: render → re-render → change src → change back.
- [x] `pnpm test tests/head.test.ts tests/script.test.ts tests/shims.test.ts` — 1068 tests pass.
- [x] `pnpm run check` — format, lint, type check clean.
- [ ] CI green.

## References

- Fixes #1473
- Mirrors the canonical Next.js e2e: `.nextjs-ref/test/e2e/nonce-head-manager/index.test.ts`
- Upstream impl: `.nextjs-ref/packages/next/src/client/head-manager.ts`

### DIFF
--- a/packages/vinext/src/shims/head.ts
+++ b/packages/vinext/src/shims/head.ts
@@ -341,6 +341,11 @@ export function _applyHeadPropsToElement(
  * See: https://bugs.chromium.org/p/chromium/issues/detail?id=1211471#c12
  */
 export function isEqualHeadNode(oldTag: Element, newTag: Element): boolean {
+  // In production this only runs on the client (the reconciler is invoked
+  // from `syncClientHead`, gated by `typeof window !== "undefined"`). The
+  // `typeof HTMLElement` guard exists for the Node test environment, where
+  // unit tests drive `_reconcileClientHead` directly with DOM doubles and
+  // `HTMLElement` may be undefined.
   if (
     typeof HTMLElement !== "undefined" &&
     oldTag instanceof HTMLElement &&
@@ -370,8 +375,10 @@ export type HeadDocumentLike = {
   createElement(tag: string): Element;
 };
 type HeadEl = {
+  querySelector(selector: string): Element | null;
   querySelectorAll(selector: string): ArrayLike<Element>;
   appendChild(el: Element): unknown;
+  prepend(el: Element): unknown;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


james-elicx:
CI was failing on formatting only — ran `vp check --fix tests/head.test.ts` and committed the result. No logic change.

PR #1633, opened 2026-05-28
fix(pages-router): invoke Document.getInitialProps so docProps reach SSR

### LINKED ISSUE #1361: Async modules (top-level `await`) render empty pages

## Problem

Pages that use top-level `await` (async modules / TLA) render empty content. The page module likely resolves to a Promise that is not being awaited before rendering, causing the page component and its data to be `undefined` or empty.

```
Expected: "hello"
Received: ""
// $('#app-value').text() is empty
```

Client-side rendering (CSR) also times out waiting for elements to appear.

## Estimated Impact

~6 test failures.

## Affected Test Suites

- `test/e2e/async-modules/index.test.ts` (6 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test with a page module that uses `const data = await fetchSomething()` at the top level and verify the page renders with the fetched data. Confirm it renders empty.

2. **Ensure async module imports are awaited.** The SSR/RSC entry needs to `await` the module import for pages that use top-level `await`. Check whether Vite/Rolldown handles async module resolution during SSR, or if vinext needs to explicitly handle this in its entry generation code.

3. **Check the Vite build output.** Verify whether Rolldown preserves top-level `await` in the SSR bundle and whether the entry properly awaits the module before accessing its exports.

james-elicx:
## Summary

- Wire `static async getInitialProps()` on user `pages/_document.tsx` into the Pages Router SSR pipeline so `this.props` includes the resolved values at render time. Both the dev-server and the production response builder now invoke the hook and spread the result on the rendered Document element, matching Next.js's `<Document {...htmlProps} {...docProps} />` in `packages/next/src/server/render.tsx`.
- Extracts a small shared helper `loadUserDocumentInitialProps` so the dev page render, dev custom-error render, and production `buildPagesShellHtml` all go through the same code path. The helper short-circuits when the user did not override the base shim, so the fast path keeps its existing number of awaits.

## Background

The Next.js [`test/e2e/async-modules`](https://github.com/vercel/next.js/blob/canary/test/e2e/async-modules/pages/_document.jsx) fixture's `_document.jsx` uses the exact pattern:

```jsx
class MyDocument extends Document {
  static async getInitialProps(ctx) {
    const initialProps = await Document.getInitialProps(ctx)
    return { ...initialProps, docValue }
  }
  render() { return <Html>…<div id="doc-value">{this.props.docValue}</div>…</Html> }
}
```

PR #1381 fixed the class-construction crash (the shim's default export now extends as a class). After that, the deploy-suite still reported "csr async page modules" failing because `<div id="doc-value">` came through empty: the render call was `React.createElement(DocumentComponent)` with no props, so `getInitialProps` was never invoked.

This PR closes that gap. Only `docProps` are forwarded; the full `DocumentContext` (`renderPage`, `defaultGetInitialProps`, `pathname`, `req`, `res`, etc.) is still not plumbed through and is called out in the helper's docstring as a follow-up. The common upstream pattern — `await Document.getInitialProps(ctx)` + spread — works because the base shim ignores `ctx`.

## Test plan
- [x] `pnpm test tests/document.test.ts` — covers `loadUserDocumentInitialProps` for override, no-override, and throw paths.
- [x] `pnpm test tests/pages-router.test.ts` — 253/253 pass, including:
  - the existing "Pages Router top-level await (async modules) in production" block, now with a `_document.tsx` fixture that uses `static async getInitialProps` + module-level `await`, plus a new `404.tsx` TLA case.
- [x] `pnpm test tests/pages-page-response.test.ts tests/pages-default-404.test.ts` — unaffected suites stay green.
- [x] `pnpm run check` — formatting, lint, and type checks pass.
- [ ] Full Vitest / Playwright via CI.

Fixes #1361

### DIFF
--- a/packages/vinext/src/server/dev-server.ts
+++ b/packages/vinext/src/server/dev-server.ts
@@ -48,6 +48,7 @@ import {
 import { buildDefaultPagesNotFoundResponse } from "./pages-default-404.js";
 import { resolvePagesPageMethodResponse } from "./pages-page-method.js";
 import { isSerializableProps } from "./pages-serializable-props.js";
+import { loadUserDocumentInitialProps } from "./pages-document-initial-props.js";
 
 /**
  * Render a React element to a string using renderToReadableStream.
@@ -131,7 +132,10 @@ async function streamPageToResponse(
   let shellTemplate: string;
 
   if (DocumentComponent) {
-    const docElement = React.createElement(DocumentComponent);
+    const docProps = await loadUserDocumentInitialProps(DocumentComponent);
+    const docElement = docProps
+      ? React.createElement(DocumentComponent, docProps)
+      : React.createElement(DocumentComponent);
     let docHtml = await renderToStringAsync(docElement);
     // Replace __NEXT_MAIN__ with our stream marker
     docHtml = docHtml.replace("__NEXT_MAIN__", STREAM_BODY_MARKER);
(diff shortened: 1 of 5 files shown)


james-elicx:
Rebased on main; conflict in `packages/vinext/src/server/dev-server.ts` was a trivial import-list overlap — kept both groups.

PR #1634, opened 2026-05-28
fix(build): preserve user-source semantics of import.meta.url

### LINKED ISSUE #1505: `import.meta.url` resolves to bundled entry path instead of user source path

james-elicx:
## Summary

- Rolldown bundles every server module into one entry, so at runtime `import.meta.url` resolves to the bundled entry path instead of the page's source file. Both Turbopack and Webpack preserve the user source path; that's the Next.js parity we want for code like `new URL('./data.json', import.meta.url)`.
- Add a build-only Vite transform (`vinext:import-meta-url`) that substitutes `import.meta.url` in user project source files with `pathToFileURL(id).href` before bundling. Skips `node_modules`, virtual modules, and vinext's own packaged source (where the token is used to locate sibling helpers).
- Dev is unaffected — Vite already serves each module from its own URL.

## Plugin ordering

Registered AFTER `og-assets` so its `fetch(new URL(..., import.meta.url))` matching for `@vercel/og` and friends still runs first on `node_modules`. My plugin skips `node_modules` anyway, so the two are complementary in practice.

## Test plan

- [x] `tests/import-meta-url.test.ts` — new unit + integration regression test:
  - Unit coverage for the textual rewrite (`rewriteImportMetaUrl`): no-op, replace, multi-occurrence, identifier guard, sibling-property preservation
  - Dev parity: a page using `import.meta.url` renders the expected `file:///.../pages/import-meta-url.tsx`
  - Production-build regression for the #1505 case: the SSR bundle embeds the page's source file URL as a literal
- [x] `vp check` (format + lint + types)
- [x] `vp test run tests/pages-router.test.ts` (252 passed)
- [x] `vp test run tests/app-router.test.ts` (331 passed)
- [x] `vp test run tests/og-inline.test.ts tests/og-font-patch.test.ts tests/asset-prefix.test.ts` (52 passed)
- [x] `vp test run tests/features.test.ts tests/static-export.test.ts` (325 passed)
- [x] `vp test run tests/cjs.test.ts` (3 passed)
- [ ] Full Vitest + Playwright via CI

Fixes #1505

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -102,6 +102,7 @@ import {
 import { createMiddlewareServerOnlyPlugin } from "./plugins/middleware-server-only.js";
 import { createOptimizeImportsPlugin } from "./plugins/optimize-imports.js";
 import { createOgInlineFetchAssetsPlugin, ogAssetsPlugin } from "./plugins/og-assets.js";
+import { createImportMetaUrlPlugin } from "./plugins/import-meta-url.js";
 import { generateRouteTypes } from "./typegen.js";
 import {
   mergeOptimizeDepsExclude,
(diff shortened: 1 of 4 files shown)


james-elicx:
Addressed CodeQL alerts in `tests/import-meta-url.test.ts`:
- Replaced sequential entity replacements with a single-pass decoder (no double-unescape).
- Loop the comment-strip regex to a fixpoint (no nested-comment re-injection).
Behaviour unchanged for the actual test inputs.

james-elicx:
Pushed biome formatting fix for `tests/import-meta-url.test.ts` — `Check` was failing on formatting. `pnpm vp check` now passes locally.

Issue #1332, opened 2026-05-20
Trailing slash configuration not enforced


james-elicx:
> *This issue was created by an agent analysing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/26163751905) (vinext `main` vs Next.js `v16.2.6`, 2026-05-20).*

## Problem

The `trailingSlash` configuration from `next.config.js` is not properly enforced in several areas:

1. **Middleware redirect `Location` headers ignore the config.** When `trailingSlash: true`, `NextResponse.redirect(new URL('/somewhere', req.url))` should produce `Location: /somewhere/` but vinext emits `Location: /somewhere`.

2. **No auto-redirect for missing/extra trailing slash.** Requests without trailing slash should receive a 307 redirect to the trailing-slash version (when `trailingSlash: true`), but vinext returns 200 with the page content directly.

3. **Shallow navigation triggers full server render.** URL changes via `router.push(url, undefined, { shallow: true })` cause a server roundtrip instead of a client-only URL update, detectable because the server-rendered random number changes.

```
Expected: "/somewhere/"
Received: "/somewhere"

Expected: 307
Received: 200
```

## Estimated Impact

~58 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/trailing-slashes/with-trailing-slash.test.ts` (20 failures)
- `test/e2e/trailing-slashes/without-trailing-slash.test.ts` (20 failures)
- `test/e2e/middleware-trailing-slash/test/index.test.ts` (14 failures)
- `test/e2e/app-dir/trailingslash/trailingslash.test.ts` (3 failures)
- `test/e2e/basepath/trailing-slash.test.ts` (1 failure)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add test cases that configure `trailingSlash: true` and verify: (a) requests without trailing slash get 307 redirected, (b) middleware redirect Location headers include trailing slash, (c) the inverse for `trailingSlash: false`. Confirm they fail.

2. **Study Next.js trailing slash handling.** Search `.nextjs-ref/packages/next/src/server/` for `trailingSlash` to understand where and how the normalization is applied.

3. **Apply trailing slash normalization at the server level.** Before route matching, check the `trailingSlash` config and issue a 307 redirect if the URL doesn't conform.

4. **Normalize middleware redirect targets.** When `NextResponse.redirect()` produces a Location header, apply the `trailingSlash` config to the target URL.

[...]

PR #1635, opened 2026-05-28
test(trailing-slash): pin core enforcement contract for App Router (#1332)

james-elicx:
## Summary

- Adds end-to-end regression tests covering the core trailing-slash enforcement contract for the App Router: 308 redirects in both directions and `<Link>` `href` normalisation.
- Refs #1332. The remaining items on that issue (canonical-URL metadata, browser-side routing for the deploy suite) are out of scope for this small change.

## What this protects

The dev server, prod-server, app-rsc-handler, and the Workers deploy entry all run `normalizeTrailingSlash` before route matching (#1395). `<Link>` normalises its `href` via `normalizePathTrailingSlash` (#1282) using the build-time-injected `__VINEXT_TRAILING_SLASH`. Until now, no test exercised the full path end-to-end through the dev server — so a regression in any of those layers would silently slip past the unit tests.

The new file boots the App Router fixture twice (once with `trailingSlash: true`, once with the default) and asserts:

| Config | Request | Expected |
|---|---|---|
| `trailingSlash: true` | `GET /about` | 308 → `/about/` |
| `trailingSlash: true` | `GET /about/` | 200 |
| `trailingSlash: true` | `GET /` | rendered `href="/about/"` |
| `trailingSlash: false` | `GET /about/` | 308 → `/about` |
| `trailingSlash: false` | `GET /about` | 200 |
| `trailingSlash: false` | `GET /` | rendered `href="/about"` |

## Test plan

- [x] `pnpm test tests/trailing-slash-1332.test.ts` — all 6 tests pass
- [x] `pnpm lint tests/trailing-slash-1332.test.ts` — clean
- [x] `pnpm fmt:check` — clean

Fixes nothing on its own; it's an explicit regression net around the existing #1395 / #1413 fixes for #1332.

### DIFF
--- a/tests/trailing-slash-1332.test.ts
+++ b/tests/trailing-slash-1332.test.ts
@@ -1,0 +1,137 @@
+/**
+ * Regression coverage for issue #1332 — trailing slash configuration not
+ * enforced. These tests cover the core enforcement contract documented in
+ * Next.js test/e2e/trailing-slashes/* and test/e2e/app-dir/trailingslash/*:
+ *
+ *   - With `trailingSlash: true`, a request to `/foo` returns 308 → `/foo/`
+ *   - With `trailingSlash: false` (default), a request to `/foo/` returns 308 → `/foo`
+ *   - App Router pages and route handlers obey the redirect
+ *   - <Link href="/foo"> renders as href="/foo/" when trailingSlash is true
+ *
+ * Refs cloudflare/vinext#1332
+ */
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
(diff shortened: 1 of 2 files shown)


james-elicx:
Both failing tests were timing out at the 5s default while the dev server compiled `/about` on first hit. Bumped to 30s (same headroom we use in `tests/app-router.test.ts`). No behaviour change.

james-elicx:
Pushed biome formatting fix for `tests/trailing-slash.test.ts` — `Check` was failing on formatting. `pnpm vp check` now passes locally.

PR #1595, opened 2026-05-25
feat(app-router): support inline CSS parity

NathanDrake2406:
## Overview

| Field | Details |
| --- | --- |
| Goal | Match Next.js `experimental.inlineCss` App Router production behavior. |
| Core change | Resolve `experimental.inlineCss`, collect emitted client CSS assets, and inline matching stylesheet resources in HTML without leaking CSS bodies into RSC payloads. |
| Main boundary | App Router production HTML/RSC delivery, including Node prod-server and Workers bundle globals. |
| Primary files | `packages/vinext/src/build/inline-css.ts`, `packages/vinext/src/server/app-ssr-stream.ts`, `packages/vinext/src/server/app-rsc-request-normalization.ts`, `tests/e2e/app-router/nextjs-compat/inline-css.browser.spec.ts` |
| Expected impact | The upstream `app-inline-css` deploy test file now passes against vinext. |

## Why

Next.js treats inline CSS as a production delivery contract: HTML responses inline CSS resources, dynamic RSC navigation payloads do not include CSS bodies, and client navigation must not add duplicate stylesheet resources for CSS that is already present inline. Vinext was missing that boundary, so the deploy-suite fixture saw linked CSS in HTML/navigation and treated `RSC: 1` plus `_rsc` as an HTML request.

| Area | Principle / invariant | What this PR changes |
| --- | --- | --- |
| Config/build | `experimental.inlineCss` must affect production CSS delivery. | Resolves the flag and embeds it in the generated RSC entry. |
| HTML stream | Non-RSC HTML may inline CSS bodies, but RSC payloads must not. | Rewrites matching React stylesheet links to inline style tags in the SSR stream. |
| Navigation | Already-inlined CSS should not become duplicate stylesheet links after client navigation. | Removes only stylesheet links covered by an inline style `data-href`. |
| Fonts | `next/font` CSS participates in the inline style surface. | Merges SSR font CSS into the first inlined stylesheet, with a fallback when no stylesheet link is rewritten. |
| RSC request selection | Next-style dynamic navigation uses `RSC: 1` plus `_rsc`, not only `.rsc`. | Normalizes that shape as an RSC request while keeping header-only/query-only HTML safe. |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| `/` HTML with `experimental.inlineCss` | Stylesheet link remained in HTML. | Matching CSS asset is inlined as a hoistable `<style>`. |
[...]

PR #1637, opened 2026-05-28
fix(app-router): implement experimental.inlineCss

### LINKED ISSUE #1499: `experimental.inlineCss` not implemented

## Problem

The `experimental.inlineCss` config inlines compiled CSS directly into a `<style>` tag in the HTML head instead of loading via `<link rel="stylesheet">`. vinext does not implement this: pages still produce a separate stylesheet `<link>`, no `<style>` tag is emitted with the page rule content.

```
Expected `<style>` block with page CSS; received `<link rel="stylesheet">`
```

## Estimated Impact

~3 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/app-dir/app-inline-css/index.test.ts` (3 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Set `experimental.inlineCss: true` and assert the rendered HTML carries the page's CSS in a `<style>` tag in the head.

2. **Implement inline CSS emission.** Collect the page's compiled CSS bundle during SSR and inline it as `<style>` in the head when the option is enabled. Skip the corresponding `<link rel="stylesheet">`.

james-elicx:
## Summary

- Implements `experimental.inlineCss` for the App Router — when enabled in `next.config`, vinext now serves CSS as inline `<style>` tags in `<head>` instead of `<link rel="stylesheet">` references, matching Next.js's behavior.
- Adds an HTML-stream transform (`server/app-inline-css.ts`) that rewrites the `<link rel="stylesheet" data-rsc-css-href="…">` tags `@vitejs/plugin-rsc` emits. User-authored stylesheet links (e.g. fonts) are left alone because they don't carry the `data-rsc-css-href` marker.
- URL→CSS-content map is built once at server startup. Node prod-server walks `dist/client/`; the Cloudflare build step inlines the map into the Worker entry alongside the existing `__VINEXT_SSR_MANIFEST__` / `__VINEXT_LAZY_CHUNKS__` globals.
- Dev is a deliberate no-op, mirroring Next.js (HMR breaks inlining; the upstream e2e test is gated to production only).

Fixes #1499

## Test plan

- [x] `vp test run tests/app-inline-css.test.ts` — unit tests for the rewrite + stream transform (entity decode, `</style>` escape, unknown URL fallback, no-map pass-through).
- [x] `vp test run tests/inline-css.test.ts` — end-to-end build + `startProdServer` smoke test: with the flag on, HTML contains `<style>…</style>` and no `<link rel="stylesheet">`; with it off, the link tag is present and CSS is not inlined.
- [x] `vp test run tests/ssr-css-assets.test.ts tests/invalid-static-asset-404.test.ts tests/app-ssr-stream.test.ts tests/next-config.test.ts tests/check.test.ts` — adjacent suites still pass.
- [x] `vp check` on all touched files (formatting + types + lint).
- [ ] CI Vitest + Playwright full suites.

## Notes / follow-ups

- Pages Router is unchanged. Next.js's `experimental.inlineCss` only affects the App Router renderer, but a follow-up could extend it to `entries/pages-server-entry.ts` if needed.
- Currently inlines every CSS file under `dist/client/` into the map. This is fine for typical apps (CSS is dwarfed by JS bundles), but a future optimisation could scope it per-route via the SSR manifest if memory becomes a concern.

Ported from Next.js: [`test/e2e/app-dir/app-inline-css/index.test.ts`](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/app-inline-css/index.test.ts). See [Next.js docs](https://nextjs.org/docs/app/api-reference/config/next-config-js/inlineCss).

### DIFF
--- a/packages/vinext/src/check.ts
+++ b/packages/vinext/src/check.ts
@@ -222,6 +222,11 @@ const CONFIG_SUPPORT: Record<string, { status: Status; detail?: string }> = {
     detail:
       "App Shell prefetching not yet implemented; requires cacheComponents and other co-flags vinext does not support",
   },
+  "experimental.inlineCss": {
+    status: "supported",
+    detail:
+      'production builds inline App Router CSS as <style> tags in <head> instead of <link rel="stylesheet">',
+  },
   "i18n.domains": {
     status: "partial",
     detail: "supported for Pages Router; App Router unchanged",
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -432,6 +432,15 @@ export type ResolvedNextConfig = {
    * Mirrors Next.js: packages/next/src/server/lib/trace/utils.ts (getTracedMetadata).
    */
   clientTraceMetadata: string[] | undefined;
+  /**
+   * Inline compiled CSS as `<style>` tags in `<head>` instead of emitting
+   * `<link rel="stylesheet">` references. Sourced from
+   * `experimental.inlineCss` in next.config. Production-only (Next.js
+   * disables inlining in dev because it breaks HMR).
+   *
+   * @see https://nextjs.org/docs/app/api-reference/config/next-config-js/inlineCss
+   */
+  inlineCss: boolean;
 };
 
 // Mirrors Next.js's accepted set in packages/next/src/shared/lib/constants.ts
@@ -1091,6 +1100,7 @@ export async function resolveNextConfig(
       compilerDefineServer: {},
       instrumentationClientInject: [],
       clientTraceMetadata: undefined,
+      inlineCss: false,
     };
     detectNextIntlConfig(root, resolved);
     return resolved;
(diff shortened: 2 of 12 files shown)


james-elicx:
Addressed the two must-fix items:
- **Nonce on inlined `<style>`** — `createInlineCssTransform` now accepts the SSR `scriptNonce` and forwards it onto each emitted `<style>` tag, with a unit test that catches regressions.
- **Absolute-URL `assetPrefix`** — when `data-rsc-css-href` is a fully-qualified URL, we now also probe the URL's `pathname` against the map, with a unit test using `https://cdn.example.com/_next/static/foo.css`.

Also:
- Updated the `escapeStyleTagBoundary` JSDoc to describe what the code actually does (insert a `\\`, not a CSS `\3C` escape).
- Added a cross-chunk split test so the partial-tag buffering behaviour has dedicated coverage.

Skipped the sync/async `walk` dedup nit since the two call sites have different async profiles and the duplication is small.

/bigbonk review

james-elicx:
Fixed the CI break — `makeResolved()` in `tests/next-config.test.ts` now sets `inlineCss: false` alongside the other defaults. `vp check` passes locally across all 1459 files.

/bigbonk review

NathanDrake2406:
ahem https://github.com/cloudflare/vinext/pull/1595

james-elicx:
Rebased on main; conflict in `packages/vinext/src/config/next-config.ts` resolved by keeping both fields (existing fields from main + `inlineCss` from this PR). Previously failing checks (Deploy app-router-cloudflare + CodeQL) should re-run with the fresh build.

james-elicx:
Rebased on `origin/main`. Resolved trivial adjacent additions in `packages/vinext/src/config/next-config.ts` and `tests/next-config.test.ts` (kept both `clientTraceMetadata` and `inlineCss` fields). `pnpm run check` passes. CodeQL fail on prior run was a 2s transient infra failure; new push triggers fresh CI.

PR #1640, opened 2026-05-28
fix(build): emit assets referenced via new URL(...,import.meta.url) in SSR

### LINKED ISSUE #1346: Build failure: CSS file missing in SSR bundle (`style.css` not emitted)

## Problem

The SSR bundle's `index.js` contains an `import './style.css'` statement, but the CSS file is not emitted to `dist/server/ssr/`. The build succeeds, but the production server crashes during prerendering when Node.js tries to resolve the `.css` import.

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
  '/tmp/.../dist/server/ssr/style.css'
  imported from /tmp/.../dist/server/ssr/index.js
```

CSS imports in SSR builds should either be stripped (since CSS is injected client-side) or the CSS file should be emitted alongside the SSR bundle.

## Estimated Impact

~5 test failures (1 runtime crash, 5 failed assertions).

## Affected Test Suites

- `test/e2e/react-version/react-version.test.ts` (5 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test that imports CSS in a component and verify the SSR build doesn't crash during prerendering. Confirm it fails.

2. **Strip or emit CSS in SSR builds.** Either configure the SSR build to strip CSS imports (replace with empty modules) or ensure CSS files are properly emitted to the SSR output directory. Study how `@vitejs/plugin-rsc` and Vite handle CSS in SSR environments.

3. **Check the Vite SSR build configuration.** The `build.rolldownOptions` for the SSR environment may need `css` handling configuration.

james-elicx:
## Summary

Fixes #1346.

Vite's built-in `vite:asset-import-meta-url` plugin only runs in the `client` environment, so server-side URL dependencies (e.g. an edge API route with `import(new URL('./style.css', import.meta.url).href)`) were left untransformed in the SSR/server bundle and the referenced file was never emitted. At runtime, `new URL('./style.css', import.meta.url)` resolved to a non-existent path under the build output and Node crashed with `ERR_MODULE_NOT_FOUND`.

Adds a `vinext:server-asset-import-meta-url` plugin that:

- Runs in non-client environments only (Pages Router SSR, App Router SSR, RSC, Cloudflare worker).
- In `transform`, emits each referenced asset via `this.emitFile` and rewrites the expression to a plugin-private placeholder wrapped in `new URL(..., import.meta.url)`.
- In `renderChunk`, resolves the placeholder to a chunk-relative URL (e.g. `./_next/static/<hash>.css`). A relative URL is required because Vite's default `toOutputFilePathInJS` for SSR returns the root-absolute path, which `new URL(..., import.meta.url)` would resolve to `file:///_next/...` at Node runtime.

Mirrors the Next.js deploy-suite fixture `test/e2e/react-version/pages/api/pages-api-edge-url-dep.js` which adds a URL dependency to an edge API route purely to validate that it does not break the build.

## Test plan

- [x] Added regression test in `tests/ssr-css-assets.test.ts` that builds a Pages Router fixture with the URL-dependency pattern, asserts the CSS file is emitted under the SSR output, and that the rewritten `new URL(...)` points at a resolvable on-disk path.
- [x] Verified the new test fails on the baseline (without the plugin registered) and passes with the fix.
- [x] Existing `tests/ssr-css-assets.test.ts`, `tests/pages-router.test.ts`, `tests/app-router.test.ts`, `tests/font-google.test.ts` all green.
- [ ] Full CI suite.

## Notes / follow-ups

- The plugin handles the literal `new URL("./X", import.meta.url)` pattern. Template-literal forms with `${...}` interpolation (which Vite handles via `import.meta.glob` for client builds) are not covered — that path is rarely useful in server code; we can extend if a fixture requires it.
- Honours `/* @vite-ignore */` to match Vite's upstream contract.


james-elicx:
The item 1 concern has now been checked against the actual runtime behavior rather than only the transformed text.

Using the repository's pinned `workerd` (`1.20260401.1`) with a `workerd test` module:

```js
import(new URL('data:text/css;base64,LmZvbyB7IGNvbG9yOiByZWQ7IH0K').href)
export default { async test() { await new Promise((resolve) => setTimeout(resolve, 10)) } }
```

passes. As a control, changing the test to `await import(...)` fails, confirming workerd sees the unsupported CSS module but tolerates the exact floating URL-dependency pattern used by the upstream fixture.

The exact affected Next.js suite also reaches assertions against the current head and passes **5/5**:

```text
test/e2e/react-version/react-version.test.ts
5 passed / 0 failed
```

So the current Cloudflare transform fixes the synchronous invalid-base failure (`import.meta.url === "worker"`) and preserves the fixture's intentionally-floating import behavior. The Node path separately needs the narrow rejection backstop because Node promotes that floating rejection to a process-level failure. No code change is needed for item 1 based on the runtime evidence.

Items 2 and 3 are useful non-blocking follow-up considerations, but neither affects this focused `react-version` parity fix.

james-elicx:
Restored the runtime-proven Cloudflare data-URL handling after the concurrent `ce8c87b` change skipped the transform entirely (which leaves the original `new URL(..., import.meta.url)` expression to synchronously fail when workerd exposes `import.meta.url` as `"worker"`).

Also addressed the review's maintainability notes:
- documented that `vinext:server-asset-import-meta-url` must remain before `vinext:og-font-patch`
- narrowed the workerd claim to URL construction validity; the consumer still determines whether an asset type is importable

Focused exact-head validation:
- `vp check` on the 3 changed files: clean
- `tests/ssr-css-assets.test.ts`
- `tests/socket-error-backstop.test.ts`
- `tests/e2e-deploy-script.test.ts`
- **16/16 passed**

The runtime implementation is the same one previously verified by the exact affected Next.js suite at **5/5** and by direct pinned-workerd probes of the floating import pattern.

james-elicx:
Addressed both actionable findings from the fresh independent review:

- replaced the raw source regex scan with Vite `parseAst` + `MagicString`, so comments and string literals are never rewritten
- AST matching now accepts ordinary comments/formatting between `new URL(` and the relative string literal while continuing to honor `/* @vite-ignore */`

Validation on exact head:
- focused format/lint/types: clean
- `ssr-css-assets`, `socket-error-backstop`, `e2e-deploy-script`, `dynamic-requests-build`: **56/56 passed**
- exact affected Next.js suite, one worker: `test/e2e/react-version/react-version.test.ts`: **5/5 passed**

james-elicx:
Addressed the final independent-review blocker on exact head:

- split relative asset URL literals into filesystem pathname plus `?query#fragment` suffix
- read/emit only the pathname
- preserve the suffix on both Cloudflare data URLs and Node chunk-relative emitted URLs
- added focused coverage for both runtime branches

Current focused validation remains clean; the asset test now has 6 passing cases, and the full focused group previously passed 56/56. The exact `react-version` suite remains 5/5 on the AST-based transform.

james-elicx:
Follow-up runtime validation found that appending `?query` to a `data:` URL makes the base64 payload unreadable. Corrected the suffix behavior:

- Node emitted asset URLs preserve `?query#fragment` exactly
- Cloudflare data URLs omit the source suffix so the inlined bytes remain valid/content-addressed
- focused tests decode and assert the inlined payload, and assert the Node URL suffix after `renderChunk`

Focused exact-head result: **57/57 passed**.

james-elicx:
Addressed the final Vite-parity blocker:

- support no-substitution template literals such as ``new URL(`./style.css`, import.meta.url)``
- retain the same AST/scope guards and suffix behavior
- Cloudflare and Node focused cases now both exercise template literals

Focused result remains **57/57 passed**.

james-elicx:
Addressed the composed Cloudflare pipeline blocker:

- dynamic-request analysis now runs before server asset URL rewriting
- it sees and preserves the original statically anchored `new URL(..., import.meta.url).href`
- the subsequent Cloudflare transform can safely inline the data URL without the guard replacing it
- added a composed guard → asset-transform regression assertion

Validation:
- focused tests: **57/57 passed**
- exact `test/e2e/react-version/react-version.test.ts`, one worker: **5/5 passed**

PR #1655, opened 2026-05-28
feat: recognize Sparkle experimental flags (varyParams, optimisticRouting, cachedNavigations)

### LINKED ISSUE #1425: Track Sparkle experimental flags flipped to defaults: varyParams, optimisticRouting, cachedNavigations
## Next.js Changes

Three sibling PRs in a short stack flipped the remaining "Sparkle" experimental flags to their new defaults:

- **`experimental.varyParams`** → `true` ([PR #94010](https://github.com/vercel/next.js/pull/94010), commit [`01c8126`](https://github.com/vercel/next.js/commit/01c81260ff351e5ec0f3defe21816b89b2e5363f))
- **`experimental.optimisticRouting`** → `true` ([PR #94011](https://github.com/vercel/next.js/pull/94011), commit [`160a7bf`](https://github.com/vercel/next.js/commit/160a7bfc3a6d8afafaa941dcf05f6acf6af71985))
- **`experimental.cachedNavigations`** → `true` *when `cacheComponents` is enabled* ([PR #94012](https://github.com/vercel/next.js/pull/94012), commit [`dd4058f`](https://github.com/vercel/next.js/commit/dd4058f0841b806ae8a867586ac54e12ff3f3773))

`cachedNavigations` is tied to `cacheComponents` rather than flipped unconditionally because it is a no-op without it (Next.js validation throws if set without `cacheComponents`).

A handful of upstream tests pin the flags back to `false` because their assertions observe pre-flip navigation timing/caching (a11y route announcer, hash scroll, search-params, segment-cache prefetch-runtime).

## What the flags do

- **`varyParams`** — Root-param-aware cache keying for the segment/prefetch cache. Changes how routes are keyed when root params are present.
- **`optimisticRouting`** — Optimistic client navigation timing: the URL/history updates and the new tree renders before the server response lands.
- **`cachedNavigations`** — Caches the result of a navigation so back/forward and repeat navigations can replay without re-rendering. Builds on the same segment/prefetch-cache machinery as `varyParams` and `prefetchInlining`.

All three are part of the same "Sparkle" workstream as `prefetchInlining` (tracked in #860) and `appShells` (tracked in #1405). The full set is a required co-flag list for `experimental.appShells` — once they all become defaults, the validation drops out and App Shell prefetching becomes a single flip.

## Impact on vinext

vinext does not currently implement any of these client-router behaviors:

- No root-param-aware cache keying in the prefetch shim
- No optimistic navigation timing — the client router waits for the server response
- No navigation result caching across back/forward

As a result, flipping these defaults in user configs does not regress vinext apps today; the features are silently no-ops. The risk is two-fold:

1. **Apps written against the new defaults may rely on the behavior** (e.g. expect that repeated navigations are instant, or that the URL flips before the response). vinext should either implement equivalents or document the divergence.
2. **Config validation** — `cachedNavigations` upstream throws if set without `cacheComponents`. vinext should at minimum accept these fields in the config schema without warning, so user configs targeting modern Next.js don't fail to load.

Behavioral implementation is a larger workstream tied to the segment cache and prefetch inlining; this issue tracks awareness and config-schema acceptance, with implementation deferred to whichever Sparkle feature we wire up first.

## Related

- #860 — `experimental.prefetchInlining` default flip (same workstream)
- #1405 — `experimental.appShells` config flag (depends on all of these as co-flags)

Divkix:
Closes #1425

Adds three Next.js Sparkle experimental flags to vinext's config schema without triggering compatibility warnings:

- `experimental.varyParams` — config recognized; vinext does not implement root-param-aware cache keying
- `experimental.optimisticRouting` — config recognized; vinext does not implement optimistic client navigation
- `experimental.cachedNavigations` — config recognized; vinext does not implement navigation result caching

Also adds a warning when `experimental.cachedNavigations` is set without `cacheComponents: true` (Next.js throws in this case; vinext warns because the feature is a no-op without it).

### Validation
- `pnpm test tests/check.test.ts` — 105 tests passed
- `pnpm test tests/next-config.test.ts` — 133 tests passed
- `pnpm run check` — format passed; no new type errors introduced

### DIFF
--- a/packages/vinext/src/check.ts
+++ b/packages/vinext/src/check.ts
@@ -227,6 +227,18 @@ const CONFIG_SUPPORT: Record<string, { status: Status; detail?: string }> = {
     detail:
       "App Router production HTML inlines stylesheet links as <style> in <head>; next/font CSS is merged into the first inline style",
   },
+  "experimental.varyParams": {
+    status: "partial",
+    detail: "config recognized; vinext does not implement root-param-aware cache keying",
+  },
+  "experimental.optimisticRouting": {
+    status: "partial",
+    detail: "config recognized; vinext does not implement optimistic client navigation",
+  },
+  "experimental.cachedNavigations": {
+    status: "partial",
+    detail: "config recognized; vinext does not implement navigation result caching",
+  },
   "i18n.domains": {
     status: "partial",
     detail: "supported for Pages Router; App Router unchanged",
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -1258,6 +1258,15 @@ export async function resolveNextConfig(
     );
   }
 
+  // Warn when experimental.cachedNavigations is set without cacheComponents.
+  // Next.js throws in this case; vinext warns because the feature is a no-op without it.
+  if (experimental?.cachedNavigations === true && !config.cacheComponents) {
+    console.warn(
+      "[vinext] `experimental.cachedNavigations` requires `cacheComponents: true` to have any effect. " +
+        "Set `cacheComponents: true` in your next.config, or remove `experimental.cachedNavigations`.",
+    );
+  }
+
   // Warn about unsupported webpack usage. We preserve alias injection and
   // extract MDX settings, but all other webpack customization is still ignored.
   if (config.webpack !== undefined) {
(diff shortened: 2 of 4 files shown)


james-elicx:
Note: the `E2E (app-router)` failure is **unrelated to this PR**. The failing test is `tests/e2e/app-router/nextjs-compat/web-vitals.spec.ts` which expected INP metric but only got CLS,FCP,FID,LCP,TTFB. INP requires sustained browser interaction timing and is a known flake — this PR only modifies config schema (`check.ts`, `next-config.ts`) and doesn't touch web-vitals plumbing. Suggest a rerun.

PR #1660, opened 2026-05-28
fix(app-router): honor experimental.staleTimes in client prefetch cache (#1490)

### LINKED ISSUE #1490: App Router: `experimental.staleTimes` config not honored by client cache

## Problem

Experimental `staleTimes: { dynamic: 0, static: 180 }` config does not change client-cache behaviour. Tests that prefetch a page expect the same cached random number on a second visit (cache reuse), but vinext fetches fresh data. Either the config is not read or the client cache always invalidates.

```
Cached random number not reused after prefetch+nav with staleTimes.static=180
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/app-dir/app-client-cache/client-cache.experimental.test.ts` (2 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Configure `experimental.staleTimes` and prefetch a page; navigate twice and assert the same cached payload is reused.

2. **Wire the config into the client cache.** Read `experimental.staleTimes` from the resolved config and use it as the TTL for the App Router client cache (segment cache).

james-elicx:
Closes #1490.

## Summary

- App Router's client-side prefetch cache used a hard-coded 30s TTL and ignored `experimental.staleTimes` from `next.config`. Users could not control how long prefetched route segments were considered fresh — failing parity with Next.js' `test/e2e/app-dir/app-client-cache/client-cache.experimental.test.ts`.
- Surface `experimental.staleTimes.{dynamic,static}` through `ResolvedNextConfig` (defaults to Next.js' `{ dynamic: 0, static: 300 }`) and inject `__NEXT_CLIENT_ROUTER_STATIC_STALETIME` into the client bundle — same variable name Next.js plumbs via `build/define-env.ts`. (`DYNAMIC_STALETIME` is not yet injected since vinext does not distinguish prefetch kinds; will land alongside the consumer.)
- `PREFETCH_CACHE_TTL` in `shims/navigation.ts` now reads the static value at module-load time (seconds → ms). When the env var is absent (e.g. unit tests outside the plugin's `define` pipeline), we keep vinext's historical 30s default so existing tests stay green.

## Behavioral change for existing users (10x TTL bump)

When the plugin is active and the user has **not** set `experimental.staleTimes`, the effective production prefetch cache TTL changes from **30s to 300s** (a 10x increase). This is the Next.js default and the intentional parity fix this PR delivers — prefetched route segments will now be served from the client cache for 5 minutes instead of 30 seconds. Users who want the old behavior can opt back in via:

```js
// next.config.js
module.exports = {
  experimental: { staleTimes: { static: 30 } },
};
```

## Test plan

- [x] `pnpm test tests/next-config.test.ts` — added cases covering defaults, explicit values, partial keys, non-numeric inputs, and negative values (all fall back to defaults).
- [x] `pnpm test tests/prefetch-cache.test.ts` — added 4 cases covering the env-driven TTL: fallback when unset, seconds-to-ms conversion, and consume-within-/-past the configured window.
- [x] `pnpm run check` — format/lint/types clean.
- [ ] CI: Vitest + Playwright E2E.

## References

- Next.js source: `.nextjs-ref/packages/next/src/build/define-env.ts` (env injection)
- Next.js source: `.nextjs-ref/packages/next/src/client/components/router-reducer/reducers/navigate-reducer.ts` (`STATIC_STALETIME_MS`)

### DIFF
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -1065,8 +1065,9 @@ const DEFAULT_STALE_TIMES = { dynamic: 0, static: 300 };
  * Parse `experimental.staleTimes` from a raw next.config object.
  *
  * Mirrors Next.js' `build/define-env.ts` parsing logic:
- *   - missing / NaN values fall back to the documented defaults
- *     (`dynamic: 0`, `static: 300`) — matching Next.js parity
+ *   - missing / NaN / negative values fall back to the documented defaults
+ *     (`dynamic: 0`, `static: 300`) — matching Next.js parity and the
+ *     non-negative guard in `resolvePrefetchCacheTtl`
  *   - all values are in seconds
  *
  * @see https://nextjs.org/docs/app/api-reference/config/next-config-js/staleTimes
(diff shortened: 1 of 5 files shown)


james-elicx:
Rebased on `origin/main`. Resolved trivial adjacent additions in `packages/vinext/src/config/next-config.ts` and `tests/next-config.test.ts` (kept both `clientTraceMetadata` and `staleTimes` fields). `pnpm run check` passes.

Issue #1543, opened 2026-05-22
Bot/crawler detection does not block fallback HTML on prerender


james-elicx:
> *This issue was created by an agent analysing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/26282973805) (vinext `main` vs Next.js `v16.2.6`, 2026-05-22).*

## Problem

When `fallback: true` static params encounter a crawler bot user-agent, Next.js renders the page synchronously instead of serving the fallback shell. vinext serves the fallback HTML containing `Loading...` to crawlers, which is bad for SEO.

```
Expected substring: not "Loading..."; received fallback shell
```

## Estimated Impact

~1 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/prerender-crawler.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Request a `fallback: true` page with a Googlebot UA and assert the body does not contain the fallback text.

2. **Switch to synchronous render for known crawler UAs.** When the UA matches the Next.js bot list and the page would otherwise serve a fallback shell, render synchronously instead.

---

Part of #1328.

james-elicx:
From a milestone CI-failure audit of the [Next.js Deploy Suite run](https://github.com/cloudflare/vinext/actions/runs/27113612207) (2026-06-08): the following currently-failing suites share this issue's root cause but aren't listed above. Suggest adding them to the Affected Test Suites:
- `test/e2e/streaming-ssr-edge/streaming-ssr-edge.test.ts` (1 — "should not stream to crawlers or google pagerender bot")

Same crawler-UA detection machinery — edge streaming SSR streams to crawlers instead of buffering.

james-elicx:
Closing as passing in the Next.js Deploy Suite run https://github.com/cloudflare/vinext/actions/runs/28692899743 (vinext main at `54497d42e7fa0effd567950534a3b6f562c14218`, July 4, 2026). The assertions tracked by this issue passed in that run. The run still has unrelated failures, which remain tracked separately.

PR #1663, opened 2026-05-28
fix(pages-router): block fallback shell for crawler UAs on prerender (#1543)

james-elicx:
## Summary

- When `getStaticPaths` returned `fallback: true` and an unlisted path was hit by a known crawler/bot UA, vinext shipped the `Loading...` shell — crawlers indexed the fallback HTML instead of the real page (bad for SEO). This mirrors the Next.js behavior in [`pages-handler.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/route-modules/pages/pages-handler.ts) where `isIsrFallback` is flipped to `false` when `isBot(req.headers['user-agent'])` is true.
- Added `isBotUserAgent` in `utils/html-limited-bots.ts` combining the DOM-bot regex (Googlebot proper) with the existing HTML-limited bots list — mirrors `.nextjs-ref/packages/next/src/shared/lib/router/utils/is-bot.ts`.
- Plumbed `userAgent` + `htmlLimitedBots` into `resolvePagesPageData` and into the dev `createSSRHandler` so the fallback flip applies on both dev and prod paths.

## Test plan

- [x] `pnpm test tests/pages-page-data.test.ts` (new unit tests assert `isFallback` flips for Googlebot UA and stays for Chrome UA)
- [x] `pnpm test tests/pages-router.test.ts -t "crawler|fallback shell"` (new integration tests cover Googlebot, Bingbot, DuckDuckBot, YandexBot, facebookexternalhit plus a Chrome control)
- [x] `pnpm test tests/pages-router.test.ts -t "fallback"` (existing fallback tests still pass — no regression to the non-bot branch)
- [x] `pnpm run check` (format, lint, types all clean)
- [ ] CI green

Refs #1543. Ported from the Next.js regression test at `test/e2e/prerender-crawler.test.ts`.


james-elicx:
Rebased on `origin/main`. Resolved adjacent parameter additions in `packages/vinext/src/server/dev-server.ts` and the matching `createSSRHandler` call site in `packages/vinext/src/index.ts` (kept both `clientTraceMetadata` and `htmlLimitedBots` params in the original order). `pnpm run check` passes.

james-elicx:
Rebased on origin/main; resolved adjacent-import conflict in `dev-server.ts` (kept both main's `loadUserDocumentInitialProps` import and this PR's `isBotUserAgent` import). `vp check` passes.

PR #1693, opened 2026-05-29
feat(cache): split CDN and data cache adapters; add Cloudflare edge adapter

james-elicx:
## Summary

Splits the single pluggable `CacheHandler` into two distinct, separately-resolved concepts and adds a concrete Cloudflare edge adapter.

### Data cache handler
The existing `CacheHandler` — backs `fetch` caching, `"use cache"`, and `unstable_cache`. Canonical `get`/`setDataCacheHandler`; `get`/`setCacheHandler` retained as deprecated aliases (fully backwards compatible).

### CDN cache adapter (new)
Owns the page-level ISR **serving strategy**, not just storage:
- `readPage` / `writePage` — where page/route artifacts live
- `buildResponseHeaders` — returns a header **map** (so an adapter can emit more than one cache header)
- `ownsBackgroundRevalidation` — whether the origin runs in-process regeneration
- `revalidate` — edge purge hook

`DefaultCdnCacheAdapter` delegates storage to the data cache and reproduces the framework's existing headers, so **default behavior is unchanged**. Page/route ISR now routes through the adapter (`isr-cache`, `app-page-cache`), and `revalidateTag`/`revalidatePath`/`updateTag` invalidate the data cache **and** ask the CDN adapter to purge (default no-op).

### CloudflareCdnCacheAdapter (edge-managed)
- `readPage` → `null`, `writePage` → no-op: the origin never serves from a store; the edge serves HIT/STALE and revalidates in the background.
- `ownsBackgroundRevalidation = false`.
- `buildResponseHeaders` carries the SWR policy on **`CDN-Cache-Control`** (edge caches + SWR) while stamping **`Cache-Control: no-store`** so no browser/private cache stores the response, plus a `Cache-Tag` header from the page's render tags.
- `revalidate(tags)` purges via the request-context `cache.purge({ tags })`. Tag values and purge tags share the `encodeCacheTag` canonical form.

**Auto-switch:** importing `vinext/cloudflare` registers a global detector that selects the Cloudflare adapter when the request context exposes a `cache` handle; an explicit `setCdnCacheAdapter` always wins. The request-context type stays runtime-agnostic (`cache: unknown`) — only the Cloudflare adapter narrows it.

### Config
Adds `cdnCacheHandler` to `next.config` (symmetric with `cacheHandler`); worker codegen now uses `setDataCacheHandler`.

## Testing
New `tests/cdn-cache.test.ts` and `tests/cloudflare-cdn-cache.test.ts`; existing cache/ISR/deploy/config suites updated and green. Typecheck and lint clean.


packages/vinext/src/config/next-config.ts
james-elicx, line 1: do not do these config changes. we'll deal with config plumbing later down the line. remove any config changes.

packages/vinext/src/cloudflare/cloudflare-cdn-cache.ts
```diff
@@ -0,0 +1,133 @@
...
+function getWorkersCache(): WorkersCacheLike | null {
+  const cache = getRequestExecutionContext()?.cache;
+  if (cache && typeof (cache as Partial<WorkersCacheLike>).purge === "function") {
+    return cache as WorkersCacheLike;
+  }
+  return null;
+}
+
+/** Don't-cache value: prevents any browser/private cache from storing the response. */
+const NO_STORE = "no-store";
```
james-elicx, line 54: must-revalidate as well? 

what about public, max-age=0, must-revalidate ?

would that work better?

what's the best option?

packages/vinext/src/cloudflare/cloudflare-cdn-cache.ts
```diff
@@ -0,0 +1,133 @@
...
+    if (!input.cacheControl) {
+      return { "Cache-Control": NO_STORE };
+    }
+
+    // Browsers/private caches must never store the response (`Cache-Control:
+    // no-store`); the SWR policy is carried on `CDN-Cache-Control`, which the
+    // edge honors to cache + stale-while-revalidate while the browser does not.
+    const headers: CdnResponseHeaders = {
+      "Cache-Control": NO_STORE,
+      "CDN-Cache-Control": input.cacheControl,
```
james-elicx: note that we will need to make sure we're using `max-age` instead of `s-maxage` for cloudflare to cache it and do swr.

we should probably be using public, for these headers as well right?

can you address where relevant

packages/vinext/src/config/next-config.ts
james-elicx: Done — removed the `cdnCacheHandler` plumbing entirely. `next-config.ts` (and its test) are back to baseline; the active adapter is wired via `setCdnCacheAdapter()` only, and config can be revisited later. (0ed36a6b)

packages/vinext/src/cloudflare/cloudflare-cdn-cache.ts
james-elicx: Switched the browser-facing header to `Cache-Control: public, max-age=0, must-revalidate` — `public` so shared caches can participate, `max-age=0, must-revalidate` so a browser always revalidates against the edge instead of serving a stored copy (still allows 304s, unlike `no-store`). Non-cacheable responses keep plain `no-store`. (0ed36a6b)

packages/vinext/src/cloudflare/cloudflare-cdn-cache.ts
james-elicx: Addressed: the adapter now converts the framework's `s-maxage=…` policy to `max-age=…` and prepends `public` for the **CDN-Cache-Control** header (e.g. `public, max-age=60, stale-while-revalidate=540`), so the edge actually caches + SWRs. Both response headers are now `public`. Done in `toEdgeCacheControl()`. (0ed36a6b)

james-elicx:
Addressed both parity gaps from the latest review (039370ca):

- **App Router route handlers** — `applyRouteHandlerRevalidateHeader` (fresh) and `buildRouteHandlerCachedResponse` (HIT/STALE) now route through the CDN adapter. `routeTags` are hoisted in `app-route-handler-execution` so the fresh response carries `Cache-Tag`, enabling edge SWR + tag purge.
- **Pages Router ISR** — the fresh ISR response routes through the adapter and emits a path-based `Cache-Tag` (`_N_T_<path>`, matching `revalidatePath`); the HIT/STALE served response routes through it too.

Supporting changes:
- Extracted a shared `applyCdnResponseHeaders()` in `cache-control.ts`; `app-page-cache` now uses it (dropped its local helper).
- Added a guard in `CloudflareCdnCacheAdapter` so non-cacheable policies (`no-store`/`no-cache`/`private`, e.g. `revalidate=0` or gssp) are never promoted to `CDN-Cache-Control`.

Default behavior is unchanged (the default adapter yields a single identical `Cache-Control`). New tests cover the shared helper, the route-handler edge path, and the non-cacheable guard. Typecheck, format, lint, and the affected unit suites are green locally.

packages/vinext/src/server/cache-control.ts
```diff
@@ -6,6 +8,23 @@ const STALE_REVALIDATE_CACHE_CONTROL = "s-maxage=0, stale-while-revalidate";
...
+ * apply the result to `headers`. The default adapter yields a single
+ * `Cache-Control` identical to `input.cacheControl` (no behavior change); edge
+ * adapters may instead emit `CDN-Cache-Control` / `Cache-Tag`. Existing cache
+ * headers are cleared first so the adapter's map is authoritative.
+ */
+export function applyCdnResponseHeaders(headers: Headers, input: CdnCacheableHeaderInput): void {
+  const map = getCdnCacheAdapter().buildResponseHeaders(input);
+  headers.delete("Cache-Control");
+  headers.delete("CDN-Cache-Control");
+  headers.delete("Cache-Tag");
```
james-elicx: there is no guarantee that these are going to be the headers the adapter sets? why do we need to do this anyway? they're gonna be overriden anyway aren't they? or maybe just delete the ones that we set internally in vinext regardless of handler?

james-elicx: Good call — simplified. `applyCdnResponseHeaders` now clears only `Cache-Control` (the header vinext stamps internally); the adapter's own headers go on via `set()`, which overrides any prior value, so pre-deleting `CDN-Cache-Control`/`Cache-Tag` was both redundant and presumptuous about what a given adapter emits. To make sure no stale edge headers can persist via the store, I also strip `cdn-cache-control`/`cache-tag` in `buildAppRouteCacheValue` so they're never baked into a cached route value. (df5d68bd)

packages/vinext/src/server/pages-page-data.ts
```diff
@@ -273,14 +274,19 @@ function buildPagesCacheResponse(
...
     ...buildCacheStateHeaders(cacheState),
-    "Cache-Control": buildCachedRevalidateCacheControl(
-      cacheState,
-      effectiveRevalidateSeconds,
-      effectiveExpireSeconds,
-    ),
+    ...getCdnCacheAdapter().buildResponseHeaders({
+      cacheControl: buildCachedRevalidateCacheControl(
+        cacheState,
+        effectiveRevalidateSeconds,
```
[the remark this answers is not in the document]
james-elicx: Fixed — `buildPagesCacheResponse` now builds a `Headers` and calls `applyCdnResponseHeaders`, consistent with the other call sites. (df5d68bd)

packages/vinext/src/server/cache-control.ts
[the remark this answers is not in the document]
james-elicx: Went with option 2 — clears only `Cache-Control`. Paired with stripping `cdn-cache-control`/`cache-tag` at store time (buildAppRouteCacheValue), so there's no stale edge header to clear on the served path. (df5d68bd)

james-elicx: Follow-up: I also reverted the `cdn-cache-control`/`cache-tag` strip I'd added to `buildAppRouteCacheValue` — same presumption (hardcoding specific adapter header names) and unnecessary anyway, since CDN policy headers never reach a stored route value (the edge adapter's `writePage` is a no-op and the default never emits them). Back to the original denylist. (7646945c)

packages/vinext/src/shims/cdn-cache.ts
```diff
@@ -0,0 +1,210 @@
...
+// (module init, background task) must not lock in the default and shadow the
+// edge adapter on a later in-request call. Both adapters are stateless, so the
+// per-isolate singletons are safe to reuse.
+// ---------------------------------------------------------------------------
+
+const _CDN_KEY = Symbol.for("vinext.cdnCacheAdapter");
+const _gCdn = globalThis as unknown as Record<PropertyKey, unknown>;
+
+let _defaultAdapter: DefaultCdnCacheAdapter | null = null;
+let _edgeAdapter: CloudflareCdnCacheAdapter | null = null;
```
james-elicx: should be a generic type

packages/vinext/src/shims/cdn-cache.ts
```diff
@@ -0,0 +1,210 @@
...
+ * explicit → edge adapter (when the request context exposes a host cache) →
+ * origin-managed {@link DefaultCdnCacheAdapter}.
+ */
+export function getCdnCacheAdapter(): CdnCacheAdapter {
+  const explicit = _gCdn[_CDN_KEY] as CdnCacheAdapter | undefined;
+  if (explicit) return explicit;
+
+  if (getRequestExecutionContext()?.cache) {
+    return (_edgeAdapter ??= new CloudflareCdnCacheAdapter());
+  }
```
james-elicx: disable auto-detection for now - needs to figure out skew protection mitigation

packages/vinext/src/shims/cdn-cache.ts
james-elicx: actually why is this even needed? shouldnt the adapter be set on the default var instead of a special edge var ??

packages/vinext/src/shims/cdn-cache.ts
```diff
@@ -0,0 +1,210 @@
...
+   * The default adapter ignores them.
+   */
+  tags?: readonly string[];
+};
+
+/**
+ * The serving strategy for page-level ISR. Implement this to delegate
+ * page/route/image caching to a CDN edge instead of the origin store.
+ */
+export type CdnCacheAdapter = {
```
james-elicx, line 77: can we align some of the field naming with the data cache adapters pls

PR #1694, opened 2026-05-29
fix(deploy): respect --env flag when invoking build

### LINKED ISSUE #1210: vinext build/deploy --env <name> ignores flag, always emits top-level wrangler.json
## Summary

`vinext build --env <name>` and `vinext deploy --env <name>` both **ignore** the `--env` flag. The generated `dist/server/wrangler.json` always reflects the **top-level** environment from `wrangler.jsonc`, regardless of which environment is requested. Subsequent `wrangler deploy --env <name>` then has nothing to override, so it deploys to the default Worker.

This breaks multi-environment workflows (HML/staging/prod) for any project using `env.<name>` sections in `wrangler.jsonc`.

## Versions

- vinext: **0.0.39**
- wrangler: **4.80.0**
- node: **v25.9.0**

## Repro

`wrangler.jsonc`:

```jsonc
{
  "name": "my-app",
  "main": "./worker/index.ts",
  "compatibility_date": "2025-04-01",
  "services": [
    { "binding": "API", "service": "my-api" }
  ],
  "env": {
    "hml": {
      "name": "my-app-hml",
      "services": [
        { "binding": "API", "service": "my-api-hml" }
      ],
      "routes": [
        { "pattern": "hml.example.com", "custom_domain": true }
      ]
    }
  }
}
```

```sh
$ pnpm exec vinext build --env hml
$ python3 -c "import json; w=json.load(open('dist/server/wrangler.json')); print('name:', w['name']); print('services:', w['services']); print('routes:', w.get('routes')); print('definedEnvironments:', w.get('definedEnvironments'))"
```

**Expected output (env.hml applied):**
```
name: my-app-hml
services: [{'binding': 'API', 'service': 'my-api-hml'}]
routes: [{'pattern': 'hml.example.com', 'custom_domain': True}]
```

**Actual output (top-level env, --env ignored):**
```
name: my-app
services: [{'binding': 'API', 'service': 'my-api'}]
routes: None
definedEnvironments: ['hml']
```

Note: `definedEnvironments: ['hml']` shows vinext *parsed* the env section, but did not **apply** it.

A subsequent `wrangler deploy --env hml` against this generated file also ignores `--env`, since the file has no `env` key — it deploys `my-app` (prod) instead of `my-app-hml`.

## Workaround

Patch the generated JSON after build:

```sh
pnpm exec vinext build
node -e "
const fs = require('fs');
const p = 'dist/server/wrangler.json';
const w = JSON.parse(fs.readFileSync(p));
Object.assign(w, {
  name: 'my-app-hml',
  services: [{ binding: 'API', service: 'my-api-hml' }],
  routes: [{ pattern: 'hml.example.com', custom_domain: true }],
});
fs.writeFileSync(p, JSON.stringify(w, null, 2));
"
pnpm exec wrangler deploy
```

Ugly but reliable. Would prefer `vinext build --env hml` to handle this natively.

## Suggested fix

When `--env <name>` is passed:
1. Read `env.<name>` from `wrangler.jsonc`.
2. Deep-merge `env.<name>` over top-level into the generated `dist/server/wrangler.json` (with shallow override semantics matching wrangler's own env handling).
3. Or, alternatively, leave the `env` key intact in the generated file so `wrangler deploy --env <name>` works downstream.

Related context: `cloudflare/workers-sdk#11741` reports a similar `--env not respected` issue in `opennextjs-cloudflare` deploy — likely same root cause across CF deployment wrappers.

## Impact

Anyone with `env.<name>` in their `wrangler.jsonc` (essentially every multi-environment vinext project) is forced to either:
- Maintain duplicate `wrangler.jsonc` files per env
- Apply manual post-build patches
- Accept that `vinext deploy --env <name>` silently deploys to the wrong Worker

The "silent wrong Worker" outcome is dangerous — I deployed to prod by accident multiple times before noticing.

aicayzer:
## Problem

Closes #1210.

`vinext deploy --env <name>` silently ignored the `--env` flag. The generated `dist/server/wrangler.json` always reflected the **top-level** environment from `wrangler.jsonc`, so the subsequent `wrangler deploy --env <name>` had nothing to override and silently deployed to the default Worker.

Root cause: the `@cloudflare/vite-plugin` reads `CLOUDFLARE_ENV` from `process.env` to drive the env-merge applied to the emitted `wrangler.json`. vinext never set `CLOUDFLARE_ENV` before invoking `builder.buildApp()` inside `runBuild`, so the plugin emitted the top-level merge regardless of the CLI flag.

## Fix

- Add a `withCloudflareEnv(env, fn)` helper in `deploy.ts` that sets `process.env.CLOUDFLARE_ENV` for the duration of `fn`, restoring the prior state (set or absent) afterwards via a `try { ... } finally { ... }`. No-op when `env` is `undefined` or empty string.
- Wrap `builder.buildApp()` inside `runBuild` (`deploy.ts`) so `vinext deploy --env <name>` propagates the env to the build step. The env is resolved the same way `buildWranglerDeployArgs` does (`options.env || (options.preview ? "preview" : undefined)`), so `--preview` continues to alias to `CLOUDFLARE_ENV=preview`.

Mutating `process.env` matches the pattern that `@cloudflare/vite-plugin` itself uses internally; the helper guarantees restoration even if `buildApp` throws.

## Tests

- `tests/deploy.test.ts` — 7 new `withCloudflareEnv` tests covering: env set during the callback, absence restored when no prior value existed, prior value restored after, no-op for `undefined` and empty string, restoration on throw, callback return value passed through.

`pnpm exec vp test run tests/deploy.test.ts` — 237 passed.

## Scope

Initially this PR also added `--env` to `vinext build`. Dropped per maintainer feedback in [#issuecomment-4586735484](https://github.com/cloudflare/vinext/pull/1694#issuecomment-4586735484): `vinext build` should remain platform-neutral, with Cloudflare-specific concerns handled in the deploy path.


james-elicx:
I would really like to avoid adding a Wrangler-specific flag to Vinext... Are you able to use the environment variable?

aicayzer:
Would you mind clarifying what you mean? I'm not so sure I understand.

`--env` already exists on `vinext deploy` (parsed in `parseDeployArgs` with `env: { type: "string" }`). The bug in #1210 is that the existing flag is silently broken, so this PR's main job is fixing that. Adding `--env` to `build` was the only new bit.

The PR also uses `CLOUDFLARE_ENV` internally. `withCloudflareEnv` sets `process.env.CLOUDFLARE_ENV` around `builder.buildApp()` so the plugin picks it up. Nothing new gets plumbed into Wrangler.

Happy to drop the `build --env` part and just fix the deploy flag. Or if you'd rather rip out the deploy `--env` entirely and document `CLOUDFLARE_ENV=hml vinext deploy` as the only path, I can do that too, though it'd be a breaking change for anyone using it. Which way works?

james-elicx:
The env flag would exist solely for the environments in the wrangler config, however, building an application with Vinext's CLI shouldn't need to know about a specific platform's configuration file.

Over the next couple of weeks, I'm intending to make changes so that all builds are the same from Vinext's perspective, without the special cases we currently have for Cloudflare-specific builds. If we still see a need for platform-specific behaviour during a build, that would later be driven through configurable deployment adapters rather than conditions in the CLI.

We can fix the deploy command, but I would prefer to avoid adding it to the build command, which this PR is also adding. For what it's worth, I am going to deprecate the current version of the Cloudflare-specific deploy command in the near future and would advise using wrangler deploy itself where possible.

aicayzer:
Makes sense. I'll drop the `build --env` parts (the `cli-args.ts` and `cli.ts` changes plus their tests) and just keep the deploy fix. Will push an update shortly.

aicayzer:
Hey @james-elicx, I completely forgot to push the update — my bad. Now pushed. Rebased onto latest main while I was at it, so you may need to kick the CI again. Lmk what you think.

aicayzer:
@james-elicx would you mind having a look at the Bonk run? looks like there was an issue?

PR #1698, opened 2026-05-30
fix: include i18n context in Pages ISR cache keys

southpolesteve:
## Summary

Pages Router ISR cache keys now include the active i18n context so domain-localized pages do not share cache entries for the same pathname.

## Details

Pages ISR entries were keyed only by pathname. With i18n domain routing, the same pathname can render different content depending on the matched domain and default locale.

This change adds an i18n cache variant when i18n is configured:
- `domain:<domain>` for matched domain-locale requests
- `locale:<locale>` otherwise

The variant is only used in the cache key. It does not change the visible route URL or page props.

## Tests

Adds a Pages i18n production fixture page with `getStaticProps` + `revalidate` and verifies:
- `example.fr /isr-about` renders and caches `fr`
- `example.com /isr-about` renders and caches `en`
- the second `example.com` request is a HIT for the `en` cache entry

### DIFF
--- a/packages/vinext/src/entries/pages-server-entry.ts
+++ b/packages/vinext/src/entries/pages-server-entry.ts
@@ -617,6 +617,14 @@ async function _renderPage(request, url, manifest, middlewareHeaders, options) {
     ? (localeInfo.domainLocale ? localeInfo.domainLocale.defaultLocale : i18nConfig.defaultLocale)
     : undefined;
   const domainLocales = i18nConfig ? i18nConfig.domains : undefined;
+  const i18nCacheVariant = i18nConfig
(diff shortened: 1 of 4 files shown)


southpolesteve:
/bigbonk Addressed the dev/prod parity point. The dev Pages Router ISR path now computes the same i18n cache variant wrapper as the production entry, so both cache read/write sites use the domain/locale-aware key. Re-ran:

- `vp test tests/entry-templates.test.ts --update`
- `vp test tests/pages-i18n-prod.test.ts -t "keys Pages ISR entries by i18n domain context" --reporter=verbose`

PR #1677, opened 2026-05-28
fix(shims/head): match Next.js charset/viewport order and merge _document.getInitialProps head

james-elicx:
## Summary

Fixes #1569.

- `next/head` shim now seeds `defaultHead()` (charset + viewport) into the SSR head collector before user tags, mirroring Next.js's `defaultHead()` in `packages/next/src/shared/lib/head.tsx`. They flow through the same dedupe pipeline so users can still override via `key="charset"` / `key="viewport"`. Output now matches the exact-string assertion in Next.js's `test/e2e/next-head/index.test.ts`:

  ```
  <meta charset="utf-8" data-next-head=""><meta name="viewport" content="width=device-width" data-next-head=""><meta name="test-head-1" content="hello" data-next-head="">
  ```

- Serialised attribute names follow HTML conventions (`charSet` -> `charset`, `httpEquiv` -> `http-equiv`, `crossOrigin` -> `crossorigin`, etc.). Previously these were emitted verbatim from the JSX prop name.

- `next/document` shim no longer hardcodes charset / viewport in its `<Head>` — the defaults now flow through `next/head`'s collector with `data-next-head=""`. Fallback HTML shells in `dev-server.ts` and `pages-page-response.ts` likewise drop the duplicates.

- New `setDocumentInitialHead()` accessor (plus ALS-backed state in `head-state.ts`) lets the SSR pipeline forward head tags returned by user `_document.getInitialProps()` into the same dedupe pipeline as `next/head` children. Both the dev server and the prod `renderPagesPageResponse` now invoke `_document.getInitialProps()` when present and pass the result through.

## Test plan

- [x] `pnpm test tests/head.test.ts`
- [x] `pnpm test tests/pages-page-response.test.ts`
- [x] `pnpm test tests/pages-router.test.ts`
- [x] `pnpm test tests/shims.test.ts`
- [x] `pnpm test tests/app-page-head.test.ts tests/script-head-ordering.test.ts`
- [x] `pnpm test tests/nextjs-compat/metadata.test.ts`
- [x] `pnpm test tests/static-export.test.ts -t "charset"`
- [x] `vp check` (lint, format, types)

PR #1317, opened 2026-05-19
fix(metadata): match Next dynamic route semantics

NathanDrake2406:
## Overview

| Field | Details |
| --- | --- |
| Goal | Bring App Router dynamic metadata routes closer to Next.js behavior for the failing `metadata-dynamic-routes` cases. |
| Core change | Align metadata route discovery, robots serialization, dynamic route cache headers, and rendered metadata URL resolution. |
| Main boundary | Metadata route scanning and response shaping stay in `server/*`; rendered head URL semantics stay in the metadata shim. |
| Primary files | `metadata-routes.ts`, `metadata-route-response.ts`, `metadata.tsx`, `file-based-metadata.ts`, app page head wiring. |
| Expected impact | Existing metadata routes keep working, with additional Next-compatible TSX/JSX discovery and corrected head URL output. |

## Why

Metadata routes are framework-owned routes, so vinext has to apply Next's route discovery, response headers, and rendered URL rules even when userland handlers return plain `Response` objects or file-based metadata contributes head entries.

| Area | Principle / invariant | What this PR changes |
| --- | --- | --- |
| Route discovery | Dynamic metadata route files use configured page extensions in Next. | Accepts `.tsx`, `.ts`, `.jsx`, and `.js` for dynamic metadata routes, including `sitemap.tsx`. |
| Route responses | The metadata route loader owns default cache headers. | Adds default `Cache-Control` to handler-returned metadata `Response`s without overriding user-set headers. |
| Robots output | Serialized route data should match Next's observable text. | Emits `Host` before `Sitemap`. |
| Head URLs | Canonical, manifest, and social image URLs are not the same class of metadata URL. | Canonical root URLs drop the trailing slash, manifest route hrefs remain relative, and file-based social images use Next's fallback metadata base rules. |

## What Changed

| Scenario | Before | After |
| --- | --- | --- |
| `app/foo/sitemap.tsx` | Not discovered, so `/foo/sitemap.xml` 404ed. | Discovered and served as a dynamic sitemap route. |
| `robots.ts` with `host` and `sitemap` | `Sitemap` appeared before `Host`. | Output matches Next's `Host` then `Sitemap` ordering. |
| Dynamic image metadata route returning `Response` | Missing framework `Cache-Control` unless handler set it. | Gets Next-style metadata route cache headers by default. |
| `metadataBase: new URL("https://mydomain.com")` and canonical `./` at `/` | Rendered `https://mydomain.com/`. | Renders `https://mydomain.com`. |
[...]

Issue #1569, opened 2026-05-22
Pages Router `next/head`: charset/viewport ordering and `_document.getInitialProps` head-tag merge order not matched


james-elicx:
Follow-up to #1475 / #1559. The attribute rename (`data-vinext-head` → `data-next-head`) was split into #1559 as a small isolated patch. Two other parts of #1475 remain and need their own work:

### 1. charset/viewport ordering
Pages Router `<Head>` emission does not match Next.js's canonical order for charset and viewport tags. The Next.js e2e suite (`test/e2e/next-head/index.test.ts`) asserts that charset appears first, followed by viewport, then user tags — in that specific order. vinext currently emits them out of order.

### 2. `_document.getInitialProps()` head-tag merge
When a custom `_document` defines `getInitialProps()` and returns head tags via `Head.rewind()` / `head: [...]`, those tags must be merged into the SSR HTML **before** the user's `next/head` tags. vinext currently does not call `_document.getInitialProps()` for head purposes at all, so downstream tag lookups in `_document` fail.

## Reference

- `.nextjs-ref/test/e2e/next-head/index.test.ts` — assertions on ordering and on `_document.getInitialProps()` merge
- `.nextjs-ref/packages/next/src/pages/_document.tsx` — canonical merge logic

## Estimated impact

~2 e2e test failures remaining after #1559.

## Recommendation

1. Reproduce in vinext's own test suite with a Pages Router fixture that imports `next/head` plus a custom `_document` with `getInitialProps()`. Assert (a) charset/viewport ordering, (b) that `_document.getInitialProps()` head tags appear ahead of user `next/head` tags.
2. Update `packages/vinext/src/shims/head.ts` (or wherever head tags are aggregated for SSR) to:
   - emit charset and viewport in Next.js's canonical position
   - call the user's `_document.getInitialProps()` (when defined) and merge its head output ahead of user `next/head` tags
3. Mirror `.nextjs-ref/packages/next/src/pages/_document.tsx` for the exact merge semantics.

Part of #1328.

james-elicx:
Reopening: the [Next.js Deploy Suite run](https://github.com/cloudflare/vinext/actions/runs/27113612207) (vinext `main` vs Next.js `v16.2.6`, 2026-06-08) shows this is not fully resolved. `next-head` fails 1: "should place charset element at the top of `<head>`" — the charset-ordering deliverable explicitly named here. (#1475 is superseded by this and stays closed.)

[...]

PR #1713, opened 2026-06-01
fix(shims/head): keep charset/viewport defaults on client hydration

james-elicx:
## Summary

Follow-up to #1677. The Pages Router charset/viewport `<meta>` tags render correctly on the server but behave incorrectly after client hydration on the live deployment (https://pages-router-cloudflare.vinext.workers.dev/). Two related problems in `next/head`'s client sync (`syncClientHead`):

1. **Defaults disappeared after hydration.** PR #1677 seeded `defaultHead()` (charset + viewport) into the **SSR** collector (`getSSRHeadHTML`) but not the client path. The client did `document.querySelectorAll("[data-next-head]").forEach(el => el.remove())` then re-added only the user `<Head>` children — so the server-rendered charset + viewport (which carry `data-next-head=""`) got stripped and never re-added.

2. **Tag order changed + DOM churn.** Even with the defaults re-added, the destructive remove-all + append-all approach moved the surviving tags to the *end* of `<head>` (after the modulepreload/script links the server placed after them) and rebuilt every node on each `<Head>` update, causing flicker.

Next.js handles both correctly:
- [`reduceComponents()`](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/head.tsx) **always** concatenates `defaultHead()` on both server and client.
- [`head-manager.ts` `updateElements()`](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/head-manager.ts) **reconciles in place**: diffs desired tags against existing `[data-next-head]` nodes via `isEqualNode()`, leaves matches untouched in their original position, appends only new tags, removes stale ones, and prepends `<meta charset>` to keep it first.

## Fix

`_syncClientHead` now:
- Seeds `defaultHead()` into the client reduce, mirroring the SSR path (charset + viewport survive hydration; users can still override via `key="charset"` / `key="viewport"` through the dedupe pipeline).
- Reconciles in place via `isEqualNode()` instead of wipe-and-rebuild — preserving SSR head ordering, avoiding flicker, and keeping `<meta charset>` first.

## Test plan

- [x] New regression tests in `tests/head.test.ts` exercise the client projection with a minimal DOM double (the repo has no jsdom). They cover: defaults surviving hydration, SSR order preservation + node reuse, appending new tags, removing stale tags, and charset-first prepend.
- [x] `vp test run tests/head.test.ts` (52 passed)
- [x] `vp test run tests/shims.test.ts tests/pages-page-response.test.ts tests/document.test.ts` (1042 passed)
- [x] `vp check` (format, lint, types)

## Out of scope

The duplicate charset/viewport meta tags on the App Router example (https://app-router-cloudflare.vinext.workers.dev/) are a **separate, pre-existing** issue — the example's `layout.tsx` manually adds `<meta charSet>`/viewport that vinext auto-injects via the metadata system (since #1317, May 19), not a regression from #1677. Being handled separately.

### DIFF
--- a/packages/vinext/src/shims/head.ts
+++ b/packages/vinext/src/shims/head.ts
@@ -417,25 +417,69 @@ export function _applyHeadPropsToElement(
   }
 }
 
-/** @internal — exported for unit tests; called from the Head client effect. */
+/**
+ * Reconcile the document <head> against the desired projection.
+ *
+ * Mirrors Next.js's client `head-manager.ts` `updateElements()`: rather than
+ * wiping every [data-next-head] node and re-appending (which reorders the
+ * SSR-emitted tags to the end of <head> and causes flicker on each update),
+ * we diff the desired tags against the existing ones with isEqualNode(). Tags
+ * that already match are left untouched in their original DOM position, only
+ * genuinely new tags are inserted, and stale tags are removed.
+ *
+ * The desired list seeds defaultHead() (charset + viewport) ahead of user
+ * tags — matching the SSR path in getSSRHeadHTML() and Next.js's
+ * reduceComponents(), which always concatenates defaultHead() on both server
+ * and client. Without it the first <Head> mount after hydration would drop the
+ * server-rendered defaults. Users can still override via key="charset" /
+ * key="viewport" through the dedupe pipeline.
+ *
+ * @internal — exported for unit tests; called from the Head client effect.
+ */
 export function _syncClientHead(): void {
-  document.querySelectorAll("[data-next-head]").forEach((el) => el.remove());
+  const headEl = document.head;
+  if (!headEl) return;
 
-  // Seed defaultHead() (charset + viewport) ahead of user tags, mirroring the
-  // SSR path in getSSRHeadHTML() and Next.js's reduceComponents() (which always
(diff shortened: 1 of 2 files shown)


james-elicx:
Addressed the non-blocking review notes in 8d591a1:

1. **Client-side user-override test** — added a test asserting a user `key="viewport"` (`content="width=500"`) wins over the default viewport on `_syncClientHead`, with the default charset still emitted alongside. Mirrors the existing SSR override test and locks in the dedupe-precedence parity claim.
2. **Charset prepend `if/else`** — added a code comment marking the deliberate divergence from Next.js's literal `head-manager.ts` (prepend with no `else` + unconditional `appendChild`, which moves a new charset node twice). Warns future readers not to "fix" it back.
3. **`<title>` reconciliation** — documented the intentional divergence: Next.js sets `document.title` directly so the node never moves; we reconcile `<title>` like any tag (reused in place on unchanged hydration; only a client-side title *change* repositions it). Position of `<title>` in `<head>` is not observable, so cosmetic only.
4. **Issue references** — reconciled the client `describe` block to reference `#1569 / #1677` (underlying issue fixed for SSR by #1677; this covers the client path).

`tests/head.test.ts` (53 passed) and `vp check` clean on both changed files.

/bigbonk review

PR #1714, opened 2026-06-01
feat(ppr): add encodePrerenderRouteParams and match kind exact payload tests

NathanDrake2406:
## Stack position

PR 2 of the original `cacheComponents` PPR root-param fallback-shell stack.

#1702 is now merged, so this PR is rebased onto current `main` and should be reviewed as the next standalone delta: **payload identity + pregenerated concrete-path identity**.

Remaining follow-ups in the stack:

1. #1714: payload identity + pregenerated concrete-path identity
2. #1715: fallback-shell render lifecycle
3. #1716: safe runtime serving

## Problem

The merged fallback-shell generation layer can now produce reusable shell artifacts such as `/en/blog/[slug]`. Runtime code must not treat those artifacts like exact pregenerated concrete routes.

A fallback-shell payload has different identity from an exact route payload:

- `/en/blog/hello-world` is an exact pregenerated concrete route.
- `/en/blog/[slug]` is a reusable fallback shell for unknown child params.
- malformed or route-incompatible fallback placeholder names must be rejected.

The later serving layer also needs route-scoped knowledge of which concrete URL paths were actually pregenerated. That prevents a known pregenerated route with a missing exact cache entry from accidentally falling through to placeholder-shell reuse.

## What changed

- Extends `encodePrerenderRouteParams(..., fallbackParamNames)` to serialize fallback-shell identity.
- Adds `matchPrerenderRouteParamsPayload`, which distinguishes:
  - `kind: "exact"`
  - `kind: "fallback-shell"`
  - invalid/no match
- Keeps `prerenderRouteParamsPayloadMatchesRoute` exact-only, so existing exact-route validation does not silently accept fallback-shell payloads.
- Rejects fallback-shell payloads with invalid or duplicate placeholder param names.
- Preserves catch-all placeholder array matching.
- Adds the pregenerated concrete-path registry helpers used by the serving PR:
  - route-scoped concrete URL path recording
  - percent-decoded pathname normalization
  - Worker global registry initialization

## Contract covered

Fallback-shell payloads are first-class but not exact-route payloads. Callers must opt into matching them via `matchPrerenderRouteParamsPayload`; the existing exact-match helper continues to reject them.

Concrete pregenerated path identity is stored by route pattern and normalized by pathname, giving the runtime serving layer a precise way to distinguish known pregenerated concrete routes from unknown child-param document requests.

## Deliberate non-goals

- No cache serving.
- No App Router dispatch order changes.
- No render lifecycle or Suspense policy changes.
- No fallback-shell stale lifecycle.

Those remain isolated in #1715 and #1716.

## Validation

After rebasing onto current `upstream/main`:

- `vp test run tests/prerender-route-params.test.ts tests/pregenerated-concrete-paths.test.ts`
- `vp check`

## References

- [Next.js fallback params in request params](https://github.com/vercel/next.js/blob/ee6e79b1792a4d401ddf2480f40a83549fe8e722/packages/next/src/server/request/params.ts#L197-L367)
- [Next.js static path fallback route params](https://github.com/vercel/next.js/blob/ee6e79b1792a4d401ddf2480f40a83549fe8e722/packages/next/src/build/static-paths/app.ts#L932-L1058)

### DIFF
--- a/packages/vinext/src/server/pregenerated-concrete-paths.ts
+++ b/packages/vinext/src/server/pregenerated-concrete-paths.ts
@@ -48,6 +48,15 @@ export function addPregeneratedConcretePath(routePattern: string, pathname: stri
   paths.add(normalized);
 }
 
+/**
+ * Returns the live backing `Set` for a route pattern (not a copy) to keep
+ * lookups allocation-free on the serving hot path. The `ReadonlySet` type
+ * forbids mutation at compile time. Callers must treat the result as
+ * point-in-time and must NOT retain it across a re-seed: each
+ * `initPregeneratedPathsFromGlobals` call runs `clearPregeneratedConcretePaths`,
+ * which empties the map, leaving any previously-returned reference stale. Read
+ * it, use it, drop it — never cache the reference.
+ */
 export function getRenderedConcreteUrlPathsForRoute(
   routePattern: string,
 ): ReadonlySet<string> | undefined {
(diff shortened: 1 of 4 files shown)


NathanDrake2406:
Replacing with stacked version

PR #1725, opened 2026-06-03
fix(css): preserve distinct media filenames for CSS url() assets

### LINKED ISSUE #1450: Build: SCSS/CSS `url(...)` references inlined as data URIs instead of emitted to `/_next/static/media/`

## Problem

Tests expect `url("/_next/static/media/<name>.<HASH>.svg")` in the rendered stylesheet for SCSS/CSS assets referenced via `url('./foo.svg')`. vinext emits the raw inline `data:image/svg+xml,...` URI instead. Vite's CSS asset pipeline is not configured to extract media references as hashed files under the `_next/static/media` layout that Next.js expects, so the test suites for `app-dir/scss/url-global` and `app-dir/scss/url-global-partial` both fail.

```
Expected pattern: /url\(".*\/_next\/static\/media\/dark.*\.svg/
Received: "url(\"data:image/svg+xml,...\")"
```

## Estimated Impact

~4 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/app-dir/scss/url-global/url-global.test.ts` (2 failures)
- `test/e2e/app-dir/scss/url-global-partial/url-global-partial.test.ts` (2 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test importing an SCSS file that uses `background: url('./image.svg')` and assert the rendered CSS references `/_next/static/media/...svg`. Confirm it fails.

2. **Study how Vite/Rolldown handles CSS `url(...)`.** By default Vite inlines small assets and emits larger ones; the inline threshold and output path need to be configured for Next.js parity.

3. **Configure the CSS asset pipeline.** Set `build.assetsInlineLimit` and the asset file-name template so `url(...)` references in CSS/SCSS get emitted as hashed files under the `_next/static/media/` prefix, not inlined as data URIs. Apply for all environments that emit client CSS.

4. **Also handle SCSS `composes` from external files.** Make sure the file-emit pipeline runs after the Sass preprocessor so that imports/composes through `.scss` files still get their `url(...)` references extracted.

james-elicx:
Closes #1690
Closes #1703
Fixes #1450

## Overview

Matches Next.js `asset/resource` behaviour for CSS `url()` dependencies: emit them as files under `/_next/static/media/` (instead of inlining small ones as `data:` URLs), and keep **byte-identical** source files (e.g. `dark.svg` / `dark2.svg`) as **distinct** output files instead of letting Vite/Rolldown's content-based asset dedupe collapse them onto one filename.

This is a slimmer reimplementation of the approach explored in #1703 — same core mechanism and public surface, roughly half the restore-side code. Co-authored with @NathanDrake2406, whose #1703 established the approach.

## Why

Next.js (webpack `asset/resource`) emits one output file per *source* file a stylesheet references, keyed by module identity. Vite/Rolldown instead dedupe emitted assets by *content*, so two byte-identical files collapse to a single output and every `url()` that referenced either source is rewritten to that one filename — the second filename disappears. There's no config switch to disable that dedupe ([vitejs/vite#8632](https://github.com/vitejs/vite/issues/8632)); the one native escape hatch is `this.emitFile({ fileName })`, whose explicit-`fileName` assets are never deduped.

## Approach

Build-only, client-scoped (dev CSS and RSC/SSR asset handling are untouched):

1. **mark** — during the client CSS transform, tag each relative asset `url()` with a private `?vinext_css_url_asset=<source-basename>` query. This is the only durable carrier of per-reference provenance: once Rolldown dedupes, the bundle can't tell which `url()` came from which source, and `originalFileNames` records the *set* of sources but not the per-reference mapping.
2. **defaults** — default the client build to `assetsInlineLimit: 0` (user config still wins) and route assets to `css/[name].[hash:8][extname]` / `media/[name].[hash:8][extname]`.
3. **restore** — at `generateBundle`, read each marked reference's source basename back out; when it differs from the deduped output, emit a sibling file under that name via the `fileName` escape hatch and rewrite the reference. The marker is then stripped (also from CSS that Vite inlined into JS chunks).

Each reference resolves from its own marker, so split CSS chunks stay correct with no shared cursor or bundle-iteration-order dependence.

## Relationship to #1703

Same mark/restore mechanism, same plugin names, same marker string, same `assetsInlineLimit`/`assetFileNames` config. The difference is the restore implementation: this drops `#1703`'s `AssetIndexes` (4 collections), `reserveBundleFileName` collision loop, `restoredFileNames` cache, and multi-fallback `findAssetForUrlPath` / `replaceUrlPathAssetFileName` in favour of a single basename map + an emitted-set + one `deriveSiblingFileName`. A review concluded those guarded scenarios are unreachable given the sibling-filename construction (`source-stem` + the deduped asset's own content-hash + ext), so the result is functionally equivalent with ~245 fewer lines. The one cheap robustness item the review flagged (defensive `decodeURIComponent`) is included.

## Tests

Ports the Next.js [`scss/url-global`](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/scss/url-global/url-global.test.ts) e2e into the committed `pages-basic` and `app-basic` fixtures (built in isolation, served via the prod server):

- **Pages Router** — one stylesheet referencing two byte-identical svgs must emit & serve both `dark.*.svg` and `dark2.*.svg`; plus a cross-chunk provenance case (two routes each referencing one of the identical svgs).
- **App Router** — client-component CSS module counterpart.
- **Unit** — marker stripping from JS-inlined CSS (escaped-quote layout, query/fragment preservation) and build-only plugin metadata.

## Validation

- `tests/css-url-assets.test.ts` — 8/8
- `tests/build-optimization.test.ts` — 80/80
- Regression sweep that builds these fixtures: `prerender` (79), `pages-router` (268; one pre-existing dev-server flake unrelated to this change, reproduces on clean `main`), `routing` + `route-sorting` (151), `check` (105).
- `tsc` clean for the changed files; formatter applied.

> [!NOTE]
> Draft — opened for comparison/review against #1703.


james-elicx:
Thanks for the thorough review! Pushed `d2bf995` addressing the feedback, plus the two real CI failures. Point-by-point:

**CI failures**
- **Check (lint)** — real, mine: `import/no-duplicates` (`vite` imported twice in the test). Merged into one import.
- **CodeQL ReDoS (point 5)** — replaced the `/\/+$/` trailing-slash strip in `client-build-config.ts` with a linear loop. Agreed it was a false positive (anchored, single-char, build-time input), but dropping the regex clears the alert and reads fine.
- **`app-router.test.ts > App Router Static export` (2 tests)** — these are a **flake**, not caused by this PR. They pass deterministically locally both in isolation and running the whole `App Router Static export` describe together (5/5). The `Element type is invalid … got: object` lines in the report are ambient stderr from the error-boundary fixture routes (e.g. `layout-metadata-error-with-boundary`), which appear in passing runs too. The likely cause is cross-shard resource/port contention during static export under CI load. The re-run on `d2bf995` should clear them.
- **`pages-router.test.ts > … generates valid JavaScript for the server entry`** — pre-existing flake, unrelated to this PR: it's a dev-server (`createServer` + `ssrLoadModule`) test, and all of this PR's plugins are `apply: "build"` (never loaded in dev). Verified it reproduces identically on clean `main` with this PR's source reverted.

**Review notes**
- **(1) shared content hash** — correct and intended; documented it's Next-parity (webpack also content-hashes byte-identical sources to one hash). No change.
- **(2) `assetsInlineLimit: 0` blast radius** — confirmed intended. Next's webpack `asset/resource` doesn't inline JS-imported client assets either, it's scoped to the **client** build only (RSC/SSR untouched), and an explicit user `build.assetsInlineLimit` still wins. The title says "CSS url()" because that's the failing parity case that motivated it, but client-asset parity generally is the goal.
- **(3) `map: null` comment** — fixed; reworded to "the marker is transient and stripped before final output" rather than claiming no column shift.
- **(4) `assetsByBase` basename collision** — added a one-line comment noting output basenames embed the content hash, so last-write-wins is safe.
- **(6) `configEnvironment` redundancy** — kept intentionally: it's the carrier for the default client environment in the single-build path (where there's no injected `environments.client` block), and applies the identical value elsewhere, so it's harmless belt-and-suspenders rather than a behavioral duplicate. Happy to consolidate if you'd prefer.

james-elicx:
CI is green now. Two follow-up commits since the review:

- `d2bf995` — lint duplicate-import fix + CodeQL regex + review-comment wording (Check & CodeQL green).
- `10fdf10b` — **moved the url() asset tests to dedicated minimal fixtures** (`tests/fixtures/css-url-assets-{app,pages}`) instead of adding routes to the shared `app-basic`/`pages-basic` and building those whole apps in isolation.

On the integration-shard failures: they were **flaky 30s timeouts** on the heaviest shard (`app-router.test.ts`, ~150s), not a defect here. Evidence:
- The failing set rotated across runs (App Router static export consistently, plus a varying third — the pages-router dev-server flake, then an unrelated ISR `revalidateTag` test). None touch CSS assets.
- `App Router Static export > exports static…` runs in **4.6s locally** (15% of the 30s budget); it only times out under CI contention.
- A plain re-run of the failed jobs went green with no code change.

The dedicated-fixtures change still reduces the blast radius the right way: the tests no longer build the full `app-basic`/`pages-basic` (≈20–30s each) in parallel, and the extra `app-basic` route no longer inflates every app-basic-wide test. Same observable contract (byte-identical `dark.svg`/`dark2.svg` from one stylesheet + the split-chunk provenance case), committed real fixtures (not written at test time), now ~1–2s to build.

james-elicx:
Thanks for the re-review! Fixed the stale header comment in `fd409cb2` — it now points at the dedicated `css-url-assets-{pages,app}` fixtures instead of the old `pages-basic`/`app-basic` paths.

(The remaining inline CodeQL annotation on `client-build-config.ts:17` is from the pre-fix commit `d43c21b`; the regex was replaced with a linear loop in `d2bf995` and CodeQL passes on the current head.)

PR #1733, opened 2026-06-04
feat(cache): configure cache adapters from vite plugin config

james-elicx:
## What

Adds a `cache` option to the `vinext()` plugin so CDN and data cache adapters can be configured declaratively in `vite.config`, instead of needing a custom worker entry that calls `setDataCacheHandler()` / `setCdnCacheAdapter()`:

```ts
import vinext from "vinext";
import { cdnAdapter } from "vinext/cloudflare/cache/cdn-adapter";
import { kvDataAdapter } from "vinext/cloudflare/cache/kv-data-adapter";

vinext({
  cache: {
    cdn:  cdnAdapter(),
    data: kvDataAdapter({ binding: "MY_KV" }), // defaults to "VINEXT_CACHE"
  },
})
```

## Design

**Config time vs. request time.** A builder runs in `vite.config` — plain Node, no Workers context. It only returns a serializable descriptor `{ adapter, options }`; it never imports the handler or reads a binding, so nothing throws at build/dev time. The builder `require.resolve`s its sibling runtime factory (`import.meta.resolve` + `fileURLToPath`, the ESM form of `require.resolve`) so the descriptor carries an absolute path, decoupled from package export wiring. On the first request the generated `virtual:vinext-cache-adapters` module imports that factory and calls it with `{ env, options }` — the binding is resolved then (`env[binding]`).

**Adapter layout.** Each adapter is a builder module + a sibling `*.runtime.ts` holding the default-export factory. Types live in one home (`vinext/shims/cache-adapter`): `CacheAdapterDescriptor`, `VinextCacheConfig`, and the `{ env, options }` factory/context types.

**Registered across every router and runtime — not just one entry:**
- **App Router**: the generated RSC entry passes `registerConfiguredCacheAdapters` into `createAppRscHandler`, which calls it per request — covering Workers, the Node server (`vinext start`), and dev through the one shared handler. The Workers entry also registers with `env` first so KV bindings resolve.
- **Pages Router**: the generated server entry registers in `renderPage` + `handleApiRoute` (Node/dev), and the generated worker registers with `env` (Workers/KV).

Registration self-guards (instantiated once per isolate; the first call with real `env` wins) and is **resilient** — a factory that can't initialize on a given runtime (e.g. a KV binding that doesn't exist on the Node server) is logged and skipped, leaving the default handler in place rather than failing requests.

## Cloudflare adapters (new `vinext/cloudflare/cache/*` subpaths)

- `kvDataAdapter({ binding?, appPrefix?, ttlSeconds?, tagCacheTtlMs? })` → `KVCacheHandler` (default binding `VINEXT_CACHE`)
- `cdnAdapter()` → `CloudflareCdnCacheAdapter` (edge-managed page ISR)

The raw `{ adapter: require.resolve("./my-adapter.js"), options }` form still works — builders are sugar over it. Explicit CDN registration takes precedence over `VINEXT_CDN_CACHE_AUTO_DETECT`; the imperative `setDataCacheHandler()` / `setCdnCacheAdapter()` API is unchanged.

## Tests

- `tests/cache-adapters-config.test.ts` — codegen permutations + inlined options; the `kvDataAdapter` / `cdnAdapter` builders (absolute-path resolution, config-time validation); runtime factories (custom-binding + missing-binding errors); and generator-level assertions that **every** router/runtime entry wires registration.
- `tests/cache-adapters-build.test.ts` — real Cloudflare build pointing `cache.data` at a local adapter by absolute path (what `require.resolve("./…")` yields), proving it resolves and bundles with no build-time Workers context.
- `tests/tsconfig-path-alias-build.test.ts` (existing) builds a worker through `app-router-entry.ts`, exercising the virtual module end-to-end.
- `vp check` clean; `app-router` (339) and `pages-router` (272) suites pass.


packages/vinext/src/cloudflare/cache/kv-data-adapter.runtime.ts
```diff
@@ -43,7 +34,15 @@ import {
...
-import { isUnknownRecord, readCacheControlNumberField } from "../utils/cache-control-metadata.js";
+import {
+  isUnknownRecord,
+  readCacheControlNumberField,
+} from "../../utils/cache-control-metadata.js";
+import type { DataCacheAdapterFactory } from "./adapter.js";
+import type { KvDataAdapterOptions } from "./kv-data-adapter.js";
+
+/** Default KV namespace binding name read from the Worker `env`. */
+const DEFAULT_BINDING = "VINEXT_CACHE";
```
james-elicx:
```suggestion
const DEFAULT_BINDING = "VINEXT_KV_CACHE";
```

packages/vinext/src/cloudflare/cache/adapter.ts
james-elicx, line 1: why is this needed and what does it have to do with cloudflare?

packages/vinext/src/cache/cache-adapters-virtual.ts
```diff
@@ -0,0 +1,146 @@
...
+ * registration module and forwarded to the adapter factory at runtime.
+ */
+export type CacheAdapterDescriptor<O extends Record<string, unknown> = Record<string, unknown>> = {
+  /**
+   * Module specifier (or absolute path, e.g. from `require.resolve(...)`) whose
+   * default export is a cache adapter factory.
+   */
+  adapter: string;
+  /** JSON-serializable options forwarded to the factory at runtime. */
+  options?: O;
```
james-elicx, line 32: use a proper type??

packages/vinext/src/cloudflare/index.ts
```diff
@@ -10,6 +10,6 @@
  * registration. An explicit `setCdnCacheAdapter(...)` always takes precedence.
  */
 
-export { KVCacheHandler } from "./kv-cache-handler.js";
-export { CloudflareCdnCacheAdapter } from "./cloudflare-cdn-cache.js";
+export { KVCacheHandler } from "./cache/kv-data-adapter.runtime.js";
+export { CloudflareCdnCacheAdapter } from "./cache/cdn-adapter.runtime.js";
```
james-elicx:
```suggestion
```

packages/vinext/src/server/app-rsc-handler.ts
```diff
@@ -782,6 +790,11 @@ export function createAppRscHandler<TRoute extends AppRscHandlerRoute>(
   options: CreateAppRscHandlerOptions<TRoute>,
 ): (request: Request, ctx: unknown) => Promise<Response> {
   return async function appRscHandler(rawRequest, ctx) {
+    // Register config-driven cache adapters before anything touches the cache.
+    // On the Cloudflare worker the entry already registered them with `env`
+    // (this guarded call is then a no-op); on Node/dev this is where they get
+    // wired, with no bindings available.
+    options.registerCacheAdapters?.();
```
james-elicx: this funcc shouldnt be optional - should always exist?

packages/vinext/src/cache/cache-adapters-virtual.ts
james-elicx, line 1: is there no other way to do this without the code injection like that or no?

packages/vinext/src/virtual-vinext-cache-adapters.d.ts
james-elicx, line 1: do we really need a separate file for this?

packages/vinext/package.json
james-elicx, line 1: can you remove this change for now

packages/vinext/src/server/app-rsc-handler.ts
james-elicx: Done — made `registerCacheAdapters` required; the generated RSC entry already passes it and the test factory now does too.

packages/vinext/src/cloudflare/cache/adapter.ts
james-elicx: Removed — deleted the separate contract file and inlined the factory param types directly into the two runtime adapters.

packages/vinext/src/cloudflare/index.ts
james-elicx: Removed.

packages/vinext/src/virtual-vinext-cache-adapters.d.ts
james-elicx: Folded the `declare module` into global.d.ts and deleted the separate file.

packages/vinext/package.json
james-elicx: Removed the `./cloudflare/cache/*` exports entry for now; updated the README example to the descriptor form and noted the builder import paths are pending.

packages/vinext/src/cloudflare/cache/kv-data-adapter.runtime.ts
james-elicx: Renamed the config-driven default to `VINEXT_KV_CACHE`. Note `deploy.ts`/`tpr.ts` still use `VINEXT_CACHE` for the imperative/auto-generated path — want those aligned to `VINEXT_KV_CACHE` too, or keep them separate?

packages/vinext/src/cache/cache-adapters-virtual.ts
james-elicx: `options` is adapter-specific JSON the plugin only inlines and forwards — it can't share one concrete interface across adapters. The concrete typing lives in each builder (e.g. `KvDataAdapterOptions` on `kvDataAdapter()`); the descriptor generic just defaults to the loose record for hand-written `{ adapter, options }`. Happy to tighten to a named per-adapter type if you'd prefer — let me know the shape you want.

packages/vinext/src/cache/cache-adapters-virtual.ts
james-elicx: It follows the same virtual-module pattern as the other generated entries (`virtual:vinext-rsc-entry`, `virtual:vinext-server-entry`): the registration module needs *static* `import` of the configured adapter modules so the bundler actually includes them in the worker — a data-only manifest read at runtime wouldn't get them bundled. Open to alternatives if you have one in mind.

packages/vinext/src/cloudflare/cache/kv-data-adapter.runtime.ts
james-elicx: Aligned everywhere now — renamed the KV binding to `VINEXT_KV_CACHE` across deploy.ts (generated wrangler + worker), tpr.ts detection, and the apps/web example, not just the config-driven default.

james-elicx:
/bigbonk review for actual issues, not pointless non-issues.

"CloudflareCdnCacheAdapter is no longer importable" is not an issue. People should be using the config.

apps/web/worker/index.ts
```diff
@@ -36,7 +36,7 @@ type Env = {
 let cacheHandlerInstalled = false;
 function installCacheHandler(env: Env): void {
   if (cacheHandlerInstalled) return;
-  setCacheHandler(new KVCacheHandler(env.VINEXT_CACHE));
+  setCacheHandler(new KVCacheHandler(env.VINEXT_KV_CACHE));
   cacheHandlerInstalled = true;
 }
```
james-elicx, line 42: remove this - add in config instead....

james-elicx: /bigbonk address this comment

PR #1738, opened 2026-06-04
fix(app-router): isolate page CSS chunks in production

NathanDrake2406:
## Overview

| Area | Detail |
| --- | --- |
| Goal | Match Next.js production App Router CSS isolation for sibling hash routes |
| Core change | Page modules in the generated RSC manifest load through cached dynamic `import()` loaders instead of static imports |
| Boundary | Production App Router page CSS chunks, prerender `generateStaticParams`, and intercepted/source-route renders |
| Primary files | `packages/vinext/src/entries/app-rsc-manifest.ts`, `packages/vinext/src/entries/app-rsc-entry.ts`, `packages/vinext/src/server/app-rsc-handler.ts`, `tests/app-router.test.ts` |
| Impact | Sibling page-level global CSS no longer leaks `scroll-padding-top` into unrelated hash pages |

## Why

Native hash scrolling uses the browser's layout state. Next.js calls `scrollIntoView()` for hash fragments, so a 20px-short scroll offset means unrelated CSS is present, not that hash math is wrong.

| --- | --- | --- |
| Route CSS | A matched page should only receive CSS for that page's route tree | Page modules are not statically imported into the shared RSC manifest chunk |
| Runtime loading | Route matching stays synchronous, rendering gets loaded modules | The handler loads the matched route once before page dispatch and action rerenders await matched/source routes |
| Prerender | Missing `generateStaticParams` must remain distinguishable from malformed return values | Lazy page sources return an internal missing-export sentinel consumed by the existing resolver |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| `/hash` with sibling `/hash-with-scroll-offset` | Linked CSS could include `scroll-padding-top:20px` from the sibling page | The no-offset route's linked CSS excludes the sibling page rule |
| Page `generateStaticParams` | Dynamic page modules were statically imported just to inspect an optional export | Page sources load lazily and report missing exports explicitly |
| Intercepted/action rerenders | Source route lookup assumed modules were already static imports | Source route lookups can await the same cached route-module loader |

<details>
<summary>Validation</summary>

- `vp test run tests/app-router.test.ts -t "keeps production route-level global CSS isolated"`
- `vp test run tests/app-router.test.ts tests/app-prerender-endpoints.test.ts tests/app-page-request.test.ts tests/app-page-dispatch.test.ts tests/app-server-action-execution.test.ts tests/app-rsc-handler.test.ts tests/entry-templates.test.ts`
- `vp check`
- Commit hook: entry-template snapshots, `vp check`, and knip
- Upstream deploy suite: `vp env exec --node 24 ./scripts/run-nextjs-deploy-suite.sh /Users/nathan/Projects/vinext/.refs/nextjs-v16.2.6 --retries 0 -c 1 --debug test/e2e/app-dir/navigation/navigation.test.ts`
  - The original `/hash-link-back-to-same-page` scroll-offset failure now passes.
  - `/hash-with-scroll-offset` passes.
  - The suite still fails on separate pending navigation parity bugs: Pages `useParams` log capture, hash `with-query-param` RSC request detection, redirect-with-loading duplicate render, nested navigation, Promise proxy navigation, and scroll restoration.

</details>

<details>

- Page modules now load at dispatch time rather than RSC manifest module initialization time.
- Static route matching remains synchronous because route metadata stays in the manifest.
- Root fallback boundaries and global-not-found behavior keep their existing loading model.
- Layout/template/error/loading modules are still static imports in this PR; this PR targets the page-level CSS leak observed in the upstream navigation fixture.

</details>

## References

| --- | --- |
| [Next.js navigation hash tests](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/navigation/navigation.test.ts) | Source of the deploy-suite failure and expected hash-scroll behavior |
| [Next.js layout-router hash scrolling](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/components/layout-router.tsx) | Shows hash navigation delegates to native `scrollIntoView()` |
| [Next.js segment-cache navigation](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/components/segment-cache/navigation.ts) | Shows same-path hash navigations are treated as hash-only scroll updates |


NathanDrake2406:
Thanks for the review! Addressed all three non-blocking notes in 1e33ba58:

1. **Unit test for the helper module** — added `tests/app-route-module-loader.test.ts` (14 tests) covering `createRouteModuleLoader` (promise dedup, rejection propagation), `loadRouteModules` (page assignment, concurrent dedup, no-op cases, missing `page` key), `loadRouteMatch` (null short-circuit, match load), and `createLazyGenerateStaticParamsSource` (delegation, missing sentinel for null / non-function exports).

2. **Rejected-load cache** — added a comment in `createRouteModuleLoader` documenting that the loaders point at built chunks whose URL is stable for a deploy, so a rejection means the chunk is genuinely missing and rolling forward (rebuild/redeploy) is the recovery path. Rejecting within the same worker would just race a known-failing chunk and confuse error responses.

3. **Per-request intercept copy in the WeakMap** — added a comment above `lazyRouteStates` explaining that the WeakMap keys on the per-request shallow copy returned by `findIntercept`, so the `loaded` flag is per-request. The actual `import()` dedup still happens in the `createRouteModuleLoader` closure promise, which is keyed on the shared manifest entry, so concurrent loads for the same intercept still share a single in-flight import.

CI is green on a9852d26 — let me know if you want any further tweaks.

NathanDrake2406:
Agreed — a positive case proving the modal CSS is emitted when the intercept is active is a good follow-up, but not blocking. The runtime already loads the intercept page before render (the `__pageLoader` resolution in `resolveAppPageInterceptState`), and the lazy page-CSS path is covered by the regular page tests, so the gap is purely a regression guard for the "intercept active" direction. Happy to add it in a follow-up if you'd like — filing nothing right now since you said it shouldn't block this PR.

james-elicx:
Pushed 669433e2 addressing the latest bonk review:

**1. Positive intercept-modal CSS guard** (`app-router-production-server.test.ts:178`) — added `emits intercepted modal CSS on RSC navigation from feed`, which fires the `/feed` → `/photos/[id]` intercept and asserts the lazy modal page's `scroll-padding-top` CSS chunk is loaded and referenced in the flight payload. Together with the existing negative test this locks both directions of the lazy-load-on-intercept path.

**Self-review finding — probe path didn't load the lazy intercept page.** While verifying, I found that `probePage()` calls `findIntercept(...)` and `buildAppPageProbes` reads `intercept.page?.default` for the dynamic-bailout probe, but — unlike the render path (`resolveAppPageInterceptState` awaits `__pageLoader`) — it never resolved the now-lazy `__pageLoader`. So the intercept probe branch was silently inspecting `undefined` and never observing the intercept page's `searchParams`/`headers` access. Fixed by resolving the lazy page before probing.

**Parallel-slot `page` leak** (`app-rsc-manifest.ts:185`) — left as the tracked follow-up, per the scope agreed across the prior reviews; generalizing lazy-loading to non-intercepting slot pages touches the render-wiring/probe/segment-config/head read sites and is a larger, separate change.

Targeted suites green locally (production-server, page-request, dispatch, rsc-handler, prerender-endpoints, dev-server intercept, route-matching, page-wiring, probe) and `vp check` is clean.

PR #1743, opened 2026-06-04
fix(app-router): restore scroll snapshots on back

### LINKED ISSUE #1367: App Router: restore popstate scroll after traversal content is restorable
> Created from the Next.js Deploy Suite failure inventory: https://github.com/cloudflare/vinext/actions/runs/26163751905
> Related foundation: https://github.com/cloudflare/vinext/issues/1366

## Problem

Vinext currently restores popstate scroll with a single `requestAnimationFrame(() => window.scrollTo(x, y))` after traversal navigation settles. If the previous route's document height or client-expanded UI has not been restored yet, the browser clamps the requested scroll position. This causes deep-list back navigation to land near the top or near a load-more boundary instead of the original row.

This is visible in real apps with long lists, intercepted detail routes, and client-side "show more" state.

## Failing Next.js tests

- `test/e2e/app-dir/navigation/navigation.test.ts`
  - `app dir - navigation > scroll restoration > should restore original scroll position when navigating back`
- `test/e2e/reload-scroll-backforward-restoration/index.test.ts`
  - `reload-scroll-back-restoration > should restore the scroll position on navigating back`
  - `reload-scroll-back-restoration > should restore the scroll position on navigating forward`

## Architectural contract

Popstate scroll restoration is a lifecycle effect that consumes a restored history entry. It must not be used as a substitute for restoring the old route/UI/cache state.

- Save scroll per browser history entry before leaving that entry.
- Restore scroll only after the traversal commit has made the destination world visible or has deliberately fallen back.
- Before applying a deep saved `scrollY`, the document must be able to represent it; otherwise the browser will clamp and silently lose the intended position.
- Any retry/wait mechanism must be bounded and tied to traversal lifecycle, not an unstructured timing loop.
- Missing or malformed scroll metadata must degrade to normal hash/default behavior.

## Relevant Vinext source

- `packages/vinext/src/server/app-browser-entry.ts` — `restorePopstateScrollPosition()`
- `packages/vinext/src/server/app-browser-popstate.ts`
- `packages/vinext/src/shims/navigation.ts` — history state scroll save before push

## Recommendation

1. Reproduce with a focused fixture: long App Router page -> scroll deep -> navigate away or open intercepted detail -> browser back -> assert original `scrollY` is restored.
2. Store scroll position as metadata for the history entry being left, not as global router state.
3. Trigger restoration from the traversal lifecycle after the target entry restore/commit path has completed.
4. Apply the saved scroll when `document.scrollingElement.scrollHeight` can represent the requested position, with a small bounded frame budget and a final best-effort attempt.
5. Keep hash restoration and explicit `scroll={false}` behavior separate from popstate saved-scroll restoration.
6. Add coverage for both success and degraded fallback: restorable long page, missing scroll state, and impossible/evicted target height.

## Dependency

This should be layered after, or at least designed with, history-entry restoration from #1366. A scroll retry without restored route/UI state only masks the race and does not preserve client-expanded list state.

NathanDrake2406:
## Summary

- Port the Next.js App Router scroll restoration regression from `navigation.test.ts` into the app-router Playwright suite.
- Stop forcing App Router documents into manual `history.scrollRestoration`; Next's App Router does not do that by default.
- Cache visible App Router state by vinext history traversal index and synchronously restore the cached source snapshot on `popstate` before the async RSC traverse finishes.

## Root Cause

The failing upstream case navigates from `/scroll-restoration` to `/scroll-restoration/other`, calls `router.back()`, and immediately reads `window.pageYOffset`. Vinext already persisted scroll coordinates in history state, but the App Router back path waited for an async RSC traverse before the source route was visible again. At the immediate assertion point the document was still rendering the short `/other` page, so the browser reported `0` instead of the original scroll offset.

Next.js stores App Router tree state in the browser history entry and restores from that state during traversal. Vinext has a different router-state representation, so this PR keeps a bounded document-local snapshot cache keyed by the existing traversal index and restores the visible snapshot synchronously while still letting the normal RSC traverse reconcile afterward.

## Next.js References

- Regression contract: https://github.com/vercel/next.js/blob/v16.2.6/test/e2e/app-dir/navigation/navigation.test.ts#L706-L735
- Upstream fixture page: https://github.com/vercel/next.js/blob/v16.2.6/test/e2e/app-dir/navigation/app/scroll-restoration/page.tsx
- Upstream fixture layout: https://github.com/vercel/next.js/blob/v16.2.6/test/e2e/app-dir/navigation/app/scroll-restoration/layout.tsx
- History state updater: https://github.com/vercel/next.js/blob/v16.2.6/packages/next/src/client/components/app-router.tsx#L56-L94
- App Router `popstate` handling: https://github.com/vercel/next.js/blob/v16.2.6/packages/next/src/client/components/app-router.tsx#L350-L371
- Restore reducer: https://github.com/vercel/next.js/blob/v16.2.6/packages/next/src/client/components/router-reducer/reducers/restore-reducer.ts#L17-L94

## Verification

Red before fix:

- `PLAYWRIGHT_PROJECT=app-router pnpm run test:e2e tests/e2e/app-router/scroll-restoration.spec.ts` failed with expected saved scroll offset and received `0`.

Green after fix:

- `vp run vinext#build`
- `PLAYWRIGHT_PROJECT=app-router pnpm run test:e2e tests/e2e/app-router/scroll-restoration.spec.ts tests/e2e/app-router/nextjs-compat/hash-popstate-scroll.spec.ts tests/e2e/app-router/nextjs-compat/router-autoscroll.spec.ts`
- `PLAYWRIGHT_PROJECT=app-router pnpm run test:e2e tests/e2e/app-router/scroll-restoration.spec.ts`
- `vp check`

Closes #1367

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -103,6 +103,12 @@ export default {
     // probed via require.resolve
     "next-intl",
 
+    // Optional peer dependency probed by tests/scss.test.ts.
+    "sass",
+
+    // Vite+ reporter name used outside CI in vite.config.ts.
+    "agent",
+
     // internal module name, not an actual dependency
     "private-next-instrumentation-client",
 
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -70,8 +70,10 @@ import {
   getVinextBrowserGlobal,
 } from "./app-browser-stream.js";
 import {
-  createAppBrowserNavigationController,
   clearHardNavigationLoopGuard,
+  createAppBrowserNavigationController,
+  createBasePathStrippedPathAndSearch,
+  createSnapshotPathAndSearch,
   type HistoryUpdateMode,
   type NavigationPayloadOutcome,
   type PendingBrowserRouterState,
@@ -223,6 +225,7 @@ function getBrowserRouteManifest(): RouteManifest | null {
 }
 
 const browserNavigationController = createAppBrowserNavigationController({
+  basePath: __basePath,
   getRouteManifest: getBrowserRouteManifest,
   syncHistoryStatePreviousNextUrl: syncCurrentHistoryStatePreviousNextUrl,
 });
@@ -1317,12 +1320,10 @@ function isSameAppRoutePopstateTarget(href: string): boolean {
 
   const target = new URL(href, window.location.origin);
   const routerState = getBrowserRouterState();
-  const targetPathname = stripBasePath(target.pathname, __basePath);
-  const targetSearch = new URLSearchParams(target.search).toString();
-  const currentSearch = routerState.navigationSnapshot.searchParams.toString();
 
   return (
-    targetPathname === routerState.navigationSnapshot.pathname && targetSearch === currentSearch
+    createBasePathStrippedPathAndSearch(target, __basePath) ===
+    createSnapshotPathAndSearch(routerState.navigationSnapshot)
   );
 }
 
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -109,10 +109,8 @@ import {
   createBfcacheSegmentStateKeyMap,
   createHistoryStateWithNavigationMetadata,
   createInitialBfcacheIdMap,
-  isHistoryStateBfcacheVersionCurrent,
   isCacheRestorableAppPayloadMetadata,
   readHistoryStateBfcacheIds,
-  readHistoryStateBfcacheVersion,
   readHistoryStatePreviousNextUrl,
   readHistoryStateTraversalIndex,
   resolveHistoryTraversalIntent,
@@ -123,6 +121,7 @@ import {
   type HistoryTraversalIntent,
   type OperationLane,
 } from "./app-browser-state.js";
+import { RestorableClientStateController } from "./app-history-state.js";
 import {
   createVisitedResponseCacheEntry,
   isVisitedResponseCacheEntryFresh,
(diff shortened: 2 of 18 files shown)


NathanDrake2406:
Addressed the latest review on the rebased head `de9c72fa`.

- The CI flake risk in the scroll regression test was valid. The test now waits for `/scroll-restoration/other`, waits for the restored `/scroll-restoration` URL and restored long content (`Item 200`), then polls `window.pageYOffset` instead of reading it immediately after the back click.
- Added focused reducer-level coverage for `reuseCurrentBfcacheIds: false`: pending commits mint fresh BFCache ids instead of reusing current ids, and approved commits do not preserve skipped layout element/flag state when the current ids are stale.
- The eager server-action invalidation concern is stale on the current head. `registerServerActionCallback()` no longer permanently calls `invalidateRestorableClientState()` before `createFromFetch`; durable invalidation remains aligned with the existing classification path through `clearClientNavigationCaches()`.
- I did not export the private snapshot/in-flight guard helpers only to unit-test them. Their observable contracts are covered through the BFCache/server-action E2E cases plus the reducer tests above.

Also rebased the branch onto current `upstream/main` and force-pushed with lease.

Local verification:
- `vp check`
- `vp run vinext#build`
- `PLAYWRIGHT_PROJECT=app-router pnpm run test:e2e tests/e2e/app-router/scroll-restoration.spec.ts`
- `PLAYWRIGHT_PROJECT=app-router pnpm run test:e2e tests/e2e/app-router/nextjs-compat/use-router-bfcache-id.spec.ts`

NathanDrake2406:
Clarifying the scope question from the latest review thread: the server-action/cache-invalidation pieces are intentional, not incidental cleanup. Once this PR adds a synchronous visible-state snapshot restore for popstate, that snapshot has to participate in the same invalidation boundary as history BFCache ids. Otherwise the scroll fix can resurrect stale client state after a server action revalidates or returns a replacement tree.

Concretely:

- The snapshot cache must be tied to the BFCache epoch, otherwise it bypasses the existing stale-history-id guard.
- The in-flight server-action guard is needed so a popstate cannot synchronously restore a snapshot while an unresolved action may still prove that the epoch should be invalidated.
- The permanent invalidation still stays aligned with the existing server-action classification point: revalidation / raw RSC payload / replacement root. Return-value-only actions with no revalidation keep restorable client state.
- The visible-commit merge change is coupled to that same invariant: when `reuseCurrentBfcacheIds` is false, we also must not preserve old visible elements/slots through reducer merge semantics, or stale client state can survive through a different path even though history ids were rejected.

So this does expand beyond the original scroll timing symptom, but it is the correctness envelope for making synchronous snapshot restore safe. The nonessential wrapper cleanup was removed; what remains is tied to the invalidation contract.

james-elicx:
> > @james-elicx Bonk workflow was cancelled.
> > [View workflow run](https://github.com/cloudflare/vinext/actions/runs/27129091259) · To retry, trigger Bonk again.
> 
> 10 mins cap broke it at the end

It died 2 minutes into its run and did nothing for the following 8 mins.

james-elicx:
> cloudflare/vinext/actions/runs/27127062967

That one died after 5 mins

NathanDrake2406:
> > cloudflare/vinext/actions/runs/27127062967
> 
> That one died after 5 mins

ah yeah you're right. that's even worse tho, hope cloudflare doesn't get billed for these failed runs

PR #1748, opened 2026-06-05
feat(cache): extract Cloudflare cache adapters into @vinext/cloudflare

james-elicx:
## What

Moves the Cloudflare cache adapters out of `vinext` into a new publishable **`@vinext/cloudflare`** package (`packages/cloudflare`).

Moved (history preserved via rename):
- `cache/kv-data-adapter.ts` + `.runtime.ts` — `KVCacheHandler`, `kvDataAdapter`, `ENTRY_PREFIX`
- `cache/cdn-adapter.ts` + `.runtime.ts` — `CloudflareCdnCacheAdapter`, `cdnAdapter`

`tpr.ts` stays in vinext (only the cache adapters moved, per scope).

## Dependency model

vinext now depends on `@vinext/cloudflare` (`workspace:*`); the new package declares `vinext` as a peer dependency. This is the intentional (approved) bidirectional edge:
- `shims/cdn-cache.ts` imports `CloudflareCdnCacheAdapter` from `@vinext/cloudflare/cache/cdn-adapter.runtime`
- `cloudflare/tpr.ts` imports `ENTRY_PREFIX` from there
- The `vinext/cloudflare` barrel still re-exports `KVCacheHandler` for backward compatibility

Neither package needs the other's `dist` to build — type imports resolve to source via tsconfig `paths`, so there's no build-order deadlock.

## Wiring

- **tsconfig.json**: added `@vinext/cloudflare` + `@vinext/cloudflare/cache/*` path mappings to source
- **vite.config.ts**: added `@vinext/cloudflare/cache` → source alias in both test projects (mirrors the existing `vinext/shims` alias) so tests run against source without a prior build
- **Build**: root `build` / `postinstall` build `@vinext/cloudflare#build` before `vinext#build`
- **Publish**: `preview-release.yml` (pkg-pr-new) publishes both packages; `publish.yml` bumps + publishes `@vinext/cloudflare` in lockstep with vinext
- **knip.ts**: added the new workspace; removed a now-unused re-export
- **Consumers updated**: `apps/web`, `examples/workers-cache`, READMEs, the `deploy.ts` error message, and the `cache` plugin-option JSDoc now reference `@vinext/cloudflare/cache/*`

## Breaking change

External users importing `vinext/cloudflare/cache/*` must switch to `@vinext/cloudflare/cache/*`. The `vinext/cloudflare` barrel still re-exports `KVCacheHandler` for compatibility.

## Verification

- `vp check` — clean (format + lint + types)
- `pnpm knip` — clean
- Targeted tests pass: `kv-cache-handler`, `cloudflare-cdn-cache`, `cache-adapters-config`, `app-route-handler-response`, `deploy`, `tpr-kv-keys`, `shims`, `isr-cache`, `fetch-cache`, `app-router`

Letting CI run the full suite + Playwright E2E.


packages/cloudflare/src/index.ts
james-elicx, line 1: dont export anything from the root

packages/cloudflare/.npmignore
james-elicx, line 1: why? isnt gitignore at the root sorting out this?

packages/cloudflare/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "dist"
+  ],
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    },
```
james-elicx: no.

packages/cloudflare/package.json
```diff
@@ -0,0 +1,56 @@
...
+      "types": "./dist/cache/kv-data-adapter.runtime.d.ts",
+      "import": "./dist/cache/kv-data-adapter.runtime.js"
+    },
+    "./cache/cdn-adapter": {
+      "types": "./dist/cache/cdn-adapter.d.ts",
+      "import": "./dist/cache/cdn-adapter.js"
+    },
+    "./cache/cdn-adapter.runtime": {
+      "types": "./dist/cache/cdn-adapter.runtime.d.ts",
+      "import": "./dist/cache/cdn-adapter.runtime.js"
```
james-elicx: cant we use a wildcard?

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "vinext-monorepo",
   "private": true,
   "scripts": {
-    "build": "vp run vinext#build",
+    "build": "vp run @vinext/cloudflare#build && vp run vinext#build",
```
james-elicx, line 5: cant we use vp task runner to run this in all packages?

README.md
```diff
@@ -274,7 +274,7 @@ Requires a custom domain (zone analytics are unavailable on `*.workers.dev`) and
 For production caching (ISR), use the built-in Cloudflare KV cache handler:
 
 ```ts
-import { KVCacheHandler } from "vinext/cloudflare";
+import { KVCacheHandler } from "@vinext/cloudflare";
 import { setCacheHandler } from "next/cache";
 
 setCacheHandler(new KVCacheHandler(env.MY_KV_NAMESPACE));
```
james-elicx, line 280: can u remove this nonsense that is no longer the supported approach

vite.config.ts
james-elicx, line 1: i wish we didnt need to do these changes... is there any way to simplify?

packages/vinext/src/cloudflare/index.ts
```diff
@@ -1,2 +1,4 @@
-export { KVCacheHandler } from "./cache/kv-data-adapter.runtime.js";
+// The Cloudflare cache adapters now live in the @vinext/cloudflare package.
+// Re-exported here for backward compatibility of the `vinext/cloudflare` entry.
+export { KVCacheHandler } from "@vinext/cloudflare";
```
james-elicx: use full path pls

packages/cloudflare/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+Or wire a handler imperatively from a worker entry:
+
+```ts
+import { KVCacheHandler } from "@vinext/cloudflare";
+import { setDataCacheHandler } from "vinext/shims/cache";
+
+setDataCacheHandler(new KVCacheHandler(env.VINEXT_KV_CACHE));
+```
```
james-elicx: absolutely not.

packages/cloudflare/src/index.ts
james-elicx: Done in b98252b1 — removed the root barrel (`src/index.ts`). The package now only exposes the `./cache/*` subpaths.

packages/cloudflare/.npmignore
james-elicx: Good call — removed it. The `files: ["dist"]` allowlist in package.json already restricts the publish to `dist` (plus the always-included package.json/README/LICENSE), so `.npmignore` was redundant.

packages/cloudflare/package.json
james-elicx: Removed the root `.` export (and `main`/`types`) in b98252b1. Subpath-only now.

packages/cloudflare/package.json
james-elicx: Yep — collapsed all four into a single wildcard:
```json
"./cache/*": { "types": "./dist/cache/*.d.ts", "import": "./dist/cache/*.js" }
```

README.md
james-elicx: Removed in b98252b1. The supported path (the `cache` plugin config via `kvDataAdapter()`) is documented in the cache-adapters section.

packages/vinext/src/cloudflare/index.ts
james-elicx: Done — now `export { KVCacheHandler } from "@vinext/cloudflare/cache/kv-data-adapter.runtime";` (matches dropping the root export).

packages/cloudflare/README.md
james-elicx: Removed the imperative `setDataCacheHandler`/`KVCacheHandler` block.

package.json
james-elicx: I tried, but the task runner can't order this one. `vinext` depends on `@vinext/cloudflare`, and `@vinext/cloudflare` has `vinext` as a peer dep, so the workspace graph is a cycle. Every graph-ordered variant fails:

```
$ vp run --filter "vinext..." build
error: Cycle dependency detected: vinext#build -> @vinext/cloudflare#build -> vinext#build
```

`-r`, `-F vinext -F @vinext/cloudflare`, and `--ignore-depends-on` all hit the same cycle (vp counts peerDependencies in the graph). The build order is actually irrelevant here — both packages compile from source via tsconfig `paths`, so neither needs the other's `dist` first — so I kept the two explicit invocations. Open to a better idea if vp has a 'select these packages, don't topo-order' mode I missed.

vite.config.ts
james-elicx: Simplified in b98252b1: deduped into a single shared `WORKSPACE_SRC_ALIAS` used by both test projects, and dropped the now-unused `@vinext/cloudflare` tsconfig path (only `@vinext/cloudflare/cache/*` remains).

The one remaining alias is the same mechanism as the existing `vinext/shims` alias: it points the `vinext` → `@vinext/cloudflare` edge at source so tests get a single module instance (e.g. `instanceof CloudflareCdnCacheAdapter` in the cdn-cache resolver) without requiring a prebuilt `dist`. Tests also import the cloudflare source directly, so without the alias we'd get two copies. If the circular dep itself is the concern, happy to discuss dropping the static `CloudflareCdnCacheAdapter` default import in favor of injection — that would remove the vinext→@vinext/cloudflare edge entirely and let all of this go away.

packages/cloudflare/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@vinext/cloudflare",
+  "version": "0.0.5",
```
james-elicx:
```suggestion
  "version": "0.0.1",
```

PR #1749, opened 2026-06-05
fix(config): guard against duplicate __dirname/__filename injection and cover node_modules

### LINKED ISSUE #1345: Build failure: `__dirname`/`__filename` shim injection conflicts and runtime errors

Divkix:
Closes #1345

## Problem

Two bugs in `cjsGlobalsInjectorPlugin` (packages/vinext/src/config/next-config.ts):

**Bug A — build crash (duplicate declaration)**
The injector unconditionally prepended `const __dirname = ...` and `const __filename = ...` without checking if the user had already declared them. A config file using a common ESM polyfill pattern:
```ts
const __dirname = dirname(fileURLToPath(import.meta.url))
```
would receive a second `const __dirname` declaration, which Rolldown rejects with a parse error.

**Bug B — runtime crash (ReferenceError in node_modules)**
The injector only ran on the config file itself. Third-party packages bundled into SSR/RSC environments that use bare `__dirname`/`__filename` got no shim, causing `ReferenceError: __dirname is not defined in ES module scope` at runtime.

## Fix

**Fix A** (`packages/vinext/src/config/next-config.ts`): Before emitting each preamble line, check for an existing declaration:
```ts
const hasOwnDirname  = /\b(?:const|let|var)\s+__dirname\b/.test(code);
const hasOwnFilename = /\b(?:const|let|var)\s+__filename\b/.test(code);
```
Only inject `const __dirname` / `const __filename` when the user hasn't declared them.

**Fix B** (`packages/vinext/src/index.ts`): Add a `vinext:cjs-globals-node-modules` plugin that injects `import.meta.dirname` / `import.meta.filename` shims (with URL fallback) into node_modules files for SSR/RSC environments, guarded by the same declaration check.

## Tests

Two new regression tests in `tests/next-config.test.ts`:
- Config file with own `const __dirname = dirname(fileURLToPath(import.meta.url))` — verifies no duplicate declaration error
- Config file with own `const __filename = fileURLToPath(import.meta.url)` — same

All 169 `next-config.test.ts` and 82 `build-optimization.test.ts` tests pass. `vp check` clean.

### DIFF
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -700,6 +700,14 @@ function cjsGlobalsInjectorPlugin(configPath: string): {
       const dirnameLiteral = JSON.stringify(dirname);
       const requireBaseLiteral = JSON.stringify(path.join(dirname, "package.json"));
 
+      // Skip injecting a shim for identifiers the user already declared.
+      // Without this guard, a config that writes its own ESM polyfill:
+      //   const __dirname = dirname(fileURLToPath(import.meta.url))
+      // would get a duplicate `const __dirname` declaration and Rolldown
+      // would reject the file with a parse error.
+      const hasOwnDirname = /\b(?:const|let|var)\s+__dirname\b/.test(code);
+      const hasOwnFilename = /\b(?:const|let|var)\s+__filename\b/.test(code);
+
       // Only wire up the wrapper `module` object — and the corresponding
       // named export read by unwrapConfig — when the source statically looks
       // like it assigns to module.exports. Pure-ESM configs avoid the extra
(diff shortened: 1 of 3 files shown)


james-elicx:
Hmm I think we might want to split these fixes into separate PRs, or maybe consider how #1740 can factor into either of them...

Divkix:
Makes sense to split these. Happy to pull Bug A (the duplicate-declaration guard in next-config.ts) out as its own small PR right now since it's independent.
For Bug B, should that go into #1740 alongside the source-file fix, or land after #1740 merges?

james-elicx:
merged 1740 now :)

i'll merge A straight away once it's separate. B has me concerned a little about performance if we're scanning or parsing every file in node_modules for the bundle... Not too sure about that part at the moment...

Divkix:
Closing this mixed PR so Bug A can be split into its own focused PR. Bug B can be revisited separately with the node_modules performance concern addressed.

PR #1753, opened 2026-06-05
feat(release): commit-driven auto-generated changesets

james-elicx:
## What & why

Replaces the manual `workflow_dispatch` (human-picks-the-bump) release with a **commit-driven Changesets** flow in **one workflow**. Changesets handles as much as possible; the only custom code is a script that **auto-creates changesets from Conventional Commits** (plus a small contributors-list step).

Auto-generated changeset files **never live on `main`** — written to the CI working tree, consumed by `changesets/action`, discarded. Manually authored changesets still work.

## One workflow: `release.yml`

On push to `main`:
1. `create-changeset.mjs` writes `.changeset/auto-*.md` to the working tree.
2. **`changesets/action` does everything else** — maintains the rolling "Version Packages" PR while changesets exist, and when none remain (the PR just merged) it **publishes** via `changeset publish` with **OIDC trusted publishing + provenance** (`id-token: write`, `NPM_CONFIG_PROVENANCE`), and creates the git tag + GitHub Release.

### The guard (why one workflow is enough)
`create-changeset.mjs` regenerates the unreleased set from `<lastTag>..HEAD` each run, but **skips when `package.json` version > the latest tag** (release merged, awaiting publish). That empty working tree is exactly what makes `changesets/action` switch from "maintain PR" to "publish" — so no separate publish workflow or guard job is needed.

## Custom surface (2 scripts)
- `scripts/create-changeset.mjs` — Conventional Commit → bump (`feat→minor`, `fix`/`perf`/`revert→patch`, `!`/`BREAKING CHANGE→major`, else skip), changed paths → affected package(s), aggregate to highest bump. Multi-package ready (independent versions).
- `scripts/version.mjs` — `changeset version`, then appends a deduped `## Contributors` list to each bumped package's `CHANGELOG.md` (idempotent; same content flows to the GitHub Release).

## Removed
- `publish.yml` entirely (guard job, manual bump, manual OIDC/tag/GitHub-release/Google-Chat steps, AI-generated notes) — `changesets/action` now owns publish + tags + releases.
- The two-workflow split and the duplicate guard job.

## Migration note
`packages/vinext/package.json` version `0.0.5 → 0.0.55` (was stale; versions were driven by `npm view`). Changesets makes `package.json` the source of truth.

> Tag format changes from `v0.0.x` (old manual scheme) to `vinext@0.0.x` (`changeset publish` default). The guard reads both, so old and new tags coexist.

## Testing
- ✅ Unit tests for both scripts: **32/32**.
- ✅ `vp check` (format + lint + types) clean.

### Needs validation in a live CI run (cannot be tested locally)
- [ ] **OIDC + `changeset publish`**: trusted publishing + provenance work through `changeset publish`/`pnpm publish` without a token (highest-risk item — the old flow used `vp pm publish --provenance` directly).
- [ ] The `auto-*.md` changeset does **not** leak into the Version PR or `main` (created → consumed → gone within the action step).
- [ ] `changeset version`'s CHANGELOG format matches what `insertContributors` expects, so the `## Contributors` block lands correctly.
- [ ] `gh api` contributor resolution authenticates with the passed `GITHUB_TOKEN` (degrades gracefully otherwise).


scripts/create-changeset.mts
```diff
@@ -0,0 +1,317 @@
...
+    stdio: ["ignore", "pipe", "pipe"],
+  }).trim();
+}
+
+type PackageJson = { name?: string; version?: string; private?: boolean };
+
+/** Publishable workspace packages (non-private, versioned): repo-relative dir → name. */
+export function discoverPublishablePackages(root: string = REPO_ROOT): Record<string, string> {
+  const map: Record<string, string> = {};
+  for (const wr of ["packages", "apps", "examples", "benchmarks"]) {
```
james-elicx: publishable only live in packages and and do not have private: true. can you simplify please

scripts/create-changeset.mts
```diff
@@ -0,0 +1,317 @@
...
+/** A version/release bump commit that must never produce a changeset. */
+export function isReleaseCommit(subject: string): boolean {
+  const s = String(subject).trim().toLowerCase();
+  return (
+    s === "version packages" ||
+    s.startsWith("chore: version packages") ||
+    s.startsWith("ci(changesets): version packages") ||
+    s.startsWith("rc:") ||
+    s.startsWith("chore(release):")
+  );
```
james-elicx: arent we hardcoding this format? why do we need 5 different conditions?

scripts/create-changeset.mts
```diff
@@ -0,0 +1,317 @@
...
+    return {
+      action: "skip",
+      reason: `package.json (${pkgVersion}) > tag (${tagVersion}); release merged, awaiting publish`,
+    };
+  }
+  return {
+    action: "generate",
+    reason: `package.json (${pkgVersion}) == tag (${tagVersion}); accumulating unreleased commits`,
+  };
+}
```
james-elicx: shouldn't we always be attending to generate a changeset unless it's a type we dont do a bump for? e.g. release ones? why do we need a version comparison? awaiting publish? why does that matter for generating a changeset?

scripts/create-changeset.mts
james-elicx: Done — simplified to scan only `packages/*` and require `private !== true`. Dropped the apps/examples/benchmarks scan and the `relative()` path juggling.

scripts/create-changeset.mts
james-elicx: Good catch — deleted the function entirely. The release commit is `chore: version packages`, and `chore` already maps to no-bump in `TYPE_BUMP`, so it is excluded by `parseBumpFromSubject` returning null. The 5 hardcoded formats were redundant.

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+    permissions:
+      contents: write # version PR branch, git tags, GitHub Releases
+      pull-requests: write # open / update the Version PR
+      id-token: write # npm OIDC trusted publishing + provenance
+    steps:
+      # zizmor: ignore[artipacked] persisted creds are required for
+      # changesets/action to push the Version PR branch, tags, and releases.
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          fetch-depth: 0 # full history + tags for the guard and contributor list
```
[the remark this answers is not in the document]
james-elicx: `changesets/action` pushes the Version PR branch, tags, and GitHub Releases using credentials persisted by `actions/checkout`, so `persist-credentials: false` would break it. The `# zizmor: ignore[artipacked]` above the step suppresses this; the job uploads no artifacts, so the credential-in-artifact leak vector does not apply.

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # changesets/action commits the Version PR branch
+      pull-requests: write # ...and opens/updates the PR
+    steps:
+      # zizmor: ignore[artipacked] persisted creds are required for
+      # changesets/action to push the Version Packages branch + open the PR.
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          fetch-depth: 0 # full history + tags for the correctness-rule guard
```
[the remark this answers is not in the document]
james-elicx: Moot — `release-pr.yml` was deleted; the two workflows were collapsed into a single `release.yml`.

scripts/create-changeset.mts
james-elicx: Fair question — the version comparison is not about *which commits* deserve a bump (you're right that 'bump-worthy type and not a release commit' decides that). It exists purely because of the design choice that **auto-changesets never get committed to `main`**.

Because they are not persisted, every run must *regenerate the full set of unreleased changesets from the last tag* so the rolling Version PR stays complete. That creates one bad window: right after the Version PR merges, `package.json` is bumped (e.g. 0.0.56) but `changeset publish` has not tagged yet (tag still v0.0.55). On that push, regenerating from v0.0.55 would re-pick-up the just-released commits → `changesets/action` would see changesets and open a *new* Version PR instead of publishing. The `package.json > tag` check skips exactly that window so the publish can happen.

So it is the cost of "never on `main`". If we instead **commit the generated changeset to `main`** like a normal changeset (the standard changesets model), there is no regeneration, no window, and the guard/`decideGeneration`/`compareVersions` all disappear — the logic becomes exactly your intuition. The trade is that changeset files live on `main` briefly until the Version PR consumes them. Bringing that choice to you before ripping it out.

james-elicx: Keeping the never-on-`main` design (your call), so the guard stays. Added a clearer inline docstring on `decideGeneration` explaining it only suppresses the post-merge/pre-publish window. Pushed.

james-elicx:
Thanks for the review — addressed:

**High — no-tag bootstrap:** This does not apply to the real repo. The premise ("`git tag -l` is empty") is false here — tags `v0.0.10`…`v0.0.55` exist, and `release.yml` checks out with `fetch-depth: 0`, which fetches tags. So `latestTagVersion("vinext")` resolves `0.0.55`, and on the post-merge push the guard sees `0.0.56 > 0.0.55 → skip → publish`. No deadlock and no all-history backfill. The bot appears to have run against a shallow/tagless clone. The `tagVersion == null` branch is only reachable for a genuinely never-released package. No change made; happy to add an npm-version fallback as defense-in-depth if preferred.

**Medium — stale `.mjs` / `release-pr.yml` refs:** Fixed in `.changeset/README.md` and the `release.yml` header comment (→ `.mts` / `release.yml`).

**Low — `git()` `fatal:` noise:** Resolved. The noisy `tagRefFor` probe already lived in `create-changeset.mts` (which pipes stderr) since `version.mts` imports it; and switching contributor resolution off `git log` (below) removed the local `git()` helper in `version.mts` entirely.

**Low — one `gh api` call per commit:** `resolveContributors` now makes a single paginated `gh api repos/…/compare/<from>...HEAD` call instead of N per-SHA calls. Verified via dry run — same contributors resolved.

**Nit — test count:** PR body updated 35 → 32 (count drifted as tests were trimmed during review).

`vp check` clean; 32 unit tests pass. Pushed.

james-elicx:
Thanks for the independent re-review and confirming the bootstrap finding was an artifact of the tagless runner. Addressed the new pass:

**Medium — dropped pre-publish CI gate (deliberate decision):** Keeping the single-workflow design and relying on the Version PR's required checks rather than re-adding an explicit gate. Rationale: publish runs on the **merged Version PR**, whose only diff is `package.json` + `CHANGELOG.md` (no source), and the repo's active `main` ruleset requires the `CI` check before merge — so the code that publishes is exactly the code that passed CI on that PR. Re-adding `needs: ci` to `release.yml` would re-run the full suite (incl. Playwright) on **every** push to `main`, duplicating `ci.yml`'s own push-to-main run. This is now a conscious tradeoff, not a silent drop. (The `prepack`-driven build remains on the first-live-run validation list — will sanity-check the published tarball's `dist` on the first real release.)

**Low — floating `@changesets/cli`:** Fixed. Both call sites now use the pinned, installed binary via `vp exec changeset …` (verified → `2.31.0`) instead of `vp dlx @changesets/cli` (floating latest), in `release.yml` publish and `version.mts`.

**Nit — dead `affectedPackages` branch:** Removed the unreachable `dir === "."` / empty-prefix branch now that discovery only returns `packages/*` keys.

`vp check` clean; 32 unit tests pass. Pushed.

james-elicx:
Thanks — addressed both new Low items:

**Malformed `@`-mentions from the name fallback:** `resolveContributors` now uses `.commits[] | .author.login // empty` (dropping the `.commit.author.name` fallback), so a commit with an unlinked email contributes no entry rather than a spaced git display name. As defense-in-depth, `dedupeSortLogins` now keeps only `^[a-zA-Z0-9-]+$` shapes — which also subsumes the previous `[bot]` filter. No more `- @Full Name`. Dry run still resolves the same valid handles.

**Floating `vp dlx` in docs:** `.changeset/README.md` now suggests `vp exec changeset` (pinned/installed) instead of `vp dlx @changesets/cli` for manual authoring.

**Pre-publish gate:** acknowledged and kept as a conscious decision — relying on the Version PR's merge-time required checks (its diff is metadata-only, so the published code is identical to what passed CI). The `prepack` `dist` build stays on the first-live-run validation list.

`vp check` clean; 32 unit tests pass. Pushed.

PR #1746, opened 2026-06-05
feat: improve dev error overlay source frames

hyoban:
## Summary
- map dev overlay stack frames to original source files via the dev server source-map endpoint, including local server/RSC frames
- show project-relative paths, code frames, and Vite open-in-editor actions in the vinext overlay
- hide ignore-listed stack frames by default with a toggle to reveal them
- surface initial dev server errors in the browser overlay by installing it before App Router and Pages Router hydration and replaying serialized server errors
- disable Vite's built-in HMR overlay so vinext owns the dev error UI
- add a copy error info action that copies the error type, message, visible stack, code frame, and component stack when available
- fix App Router dev overlay recovery when server/RSC HMR toggles a throw in a server component or layout, including the document-level `__next_error__` recovery path
- polish overlay layout, light mode, scrollbar styling, transparent backdrop, and width

## Tests
- npx vp check packages/vinext/src/server/dev-error-overlay.tsx packages/vinext/src/server/dev-error-overlay-store.ts packages/vinext/src/server/dev-stack-sourcemap.ts tests/app-browser-entry.test.ts tests/dev-stack-sourcemap.test.ts tests/dev-initial-server-error.test.ts
- npx vp test run tests/app-browser-entry.test.ts tests/dev-stack-sourcemap.test.ts tests/dev-initial-server-error.test.ts
- vp run vinext#build
- PLAYWRIGHT_PROJECT=app-router npx playwright test tests/e2e/app-router/dev-error-overlay.spec.ts -g "HMR updates"
- PLAYWRIGHT_PROJECT=app-router npx playwright test tests/e2e/app-router/dev-error-overlay.spec.ts

image

PR #1758, opened 2026-06-05
fix: client HMR dev overlay recovery

hyoban:
Follow #1746


james-elicx:
fyi @hyoban, the new overlay tests are appearing to be rather flaky - https://github.com/cloudflare/vinext/actions/runs/27077153897/job/79916376566

PR #1771, opened 2026-06-05
fix(config): avoid duplicate CJS global shims

Divkix:
## Description

Fixes bug A from #1345 by skipping vinext's injected next.config.ts `__dirname` / `__filename` shims when the config already declares those identifiers. This intentionally excludes the separate node_modules runtime shim work for bug B.

## Related Issue

Part of #1345

## Potential Risk & Impact

Low. Scope is limited to next.config CJS global injection and preserves existing shim behavior when the identifiers are not user-declared.

## How Has This Been Tested?

- `vp test run tests/next-config.test.ts -t "does not inject"`
- `vp test run tests/next-config.test.ts`
- `vp check`
- `vp test` (7145 passed, 4 skipped)

### DIFF
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -699,6 +699,8 @@ function cjsGlobalsInjectorPlugin(configPath: string): {
       const filenameLiteral = JSON.stringify(normalizedTarget);
       const dirnameLiteral = JSON.stringify(dirname);
       const requireBaseLiteral = JSON.stringify(path.join(dirname, "package.json"));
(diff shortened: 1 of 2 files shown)


james-elicx:
Pushed bc4cd50d to this branch: extended the same duplicate-declaration guard to `require`. The `createRequire` ESM polyfill (`const require = createRequire(import.meta.url)`) commonly appears alongside the `__dirname`/`__filename` polyfills and hits the identical duplicate-`const` Rolldown parse error — so when the user declares their own `require`, we now skip both the `createRequire` import and the `const require = ...` preamble line. Added a `require` regression test. All 170 next-config tests pass; `vp check` clean. (This was flagged by the /bigbonk review on #1772.)

PR #1772, opened 2026-06-05
fix(config): guard against duplicate __dirname/__filename injection

### LINKED ISSUE #1345: Build failure: `__dirname`/`__filename` shim injection conflicts and runtime errors

james-elicx:
Extracted from #1749 (by @Divkix) as a standalone, focused fix. Closes #1345.

## Problem

`cjsGlobalsInjectorPlugin` (`packages/vinext/src/config/next-config.ts`) unconditionally prepended `const __dirname = ...` and `const __filename = ...` without checking if the user had already declared them. A config file using a common ESM polyfill pattern:

```ts
const __dirname = dirname(fileURLToPath(import.meta.url));
```

would receive a second `const __dirname` declaration, which Rolldown rejects with a duplicate-declaration parse error, crashing the build.

## Fix

Before emitting each preamble line, check for an existing declaration:

```ts
const hasOwnDirname  = /\b(?:const|let|var)\s+__dirname\b/.test(code);
const hasOwnFilename = /\b(?:const|let|var)\s+__filename\b/.test(code);
```

Only inject `const __dirname` / `const __filename` when the user hasn't already declared them.

## Tests

Two new regression tests in `tests/next-config.test.ts`:

- Config file with its own `const __dirname = dirname(fileURLToPath(import.meta.url))` — verifies no duplicate-declaration error.
- Config file with its own `const __filename = fileURLToPath(import.meta.url)` — same.

All 169 `next-config.test.ts` tests pass; `vp check` is clean.

## Credit

This change is extracted from #1749 by @Divkix, which bundled this guard (Bug A) together with a separate node_modules CJS-globals shim (Bug B). This PR isolates the config-file duplicate-declaration guard so it can be reviewed and merged independently. Full credit to @Divkix for the original fix.

### DIFF
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -701,12 +701,15 @@ function cjsGlobalsInjectorPlugin(configPath: string): {
       const requireBaseLiteral = JSON.stringify(path.join(dirname, "package.json"));
 
       // Skip injecting a shim for identifiers the user already declared.
-      // Without this guard, a config that writes its own ESM polyfill:
+      // Without this guard, a config that writes its own ESM polyfills:
       //   const __dirname = dirname(fileURLToPath(import.meta.url))
-      // would get a duplicate `const __dirname` declaration and Rolldown
-      // would reject the file with a parse error.
+      //   const require = createRequire(import.meta.url)
+      // would get a duplicate `const` declaration and Rolldown would reject
+      // the file with a parse error. These polyfills commonly appear together,
+      // so all three identifiers are guarded the same way.
       const hasOwnDirname = /\b(?:const|let|var)\s+__dirname\b/.test(code);
       const hasOwnFilename = /\b(?:const|let|var)\s+__filename\b/.test(code);
+      const hasOwnRequire = /\b(?:const|let|var)\s+require\b/.test(code);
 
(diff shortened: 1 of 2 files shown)


james-elicx:
Addressed the /bigbonk review in e2483f89:

1. **`require` gap (main finding)** — Fixed. Extended the same guard to `require`: when the user declares their own `const require = createRequire(import.meta.url)`, we now skip both the `createRequire` import and the `const require = ...` preamble line, avoiding the identical duplicate-`const` Rolldown crash. Added a `require` regression test.

2. **Regex matches inside strings/comments** — Left as-is, per the review. It matches the existing precision tradeoff of the sibling `referencesCjsGlobals` / `reassignsModuleExports` heuristics in this file, and the trigger is contrived.

3. **`__dirname` test was weaker than `__filename`** — Strengthened. The test now asserts the resolved value matches the config directory (`fs.realpathSync` on both sides to normalize the macOS `/var` → `/private/var` symlink), proving the user's declaration wins over the injected shim.

All 170 `next-config.test.ts` and 82 `build-optimization.test.ts` tests pass; `vp check` clean. Original fix credit remains with @Divkix (#1749).

PR #1774, opened 2026-06-05
fix(og): lazy-load @vercel/og to keep it out of the main worker entry

james-elicx:
## Problem

The `next/og` shim (`packages/vinext/src/shims/og.tsx`) statically imported `@vercel/og`. As a result, the idiomatic Next.js usage

```ts
import { ImageResponse } from "next/og";
```

at the top of a route module pulled the **entire ~800 KB `@vercel/og` runtime** (satori + resvg + embedded wasm + base64-inlined fonts) into the always-loaded server entry (`dist/server/index.js`), where it is parsed on every cold start — even for requests that never generate an OG image.

Apps could only avoid this by manually writing `await import("next/og")` (as `examples/app-router-playground` does, with a code comment explaining the workaround). The framework shouldn't require that.

## Fix

Import `@vercel/og` via a dynamic `import()` inside the shim's already-async `ReadableStream.start()` callback. The heavy runtime is now always emitted as its own code-split chunk, **regardless of whether the app imports `next/og` statically or dynamically**. Type-only imports stay static (erased at build, no runtime pull-in).

## Impact

On `examples/app-router-cloudflare` (which uses a static `import { ImageResponse } from "next/og"`), `dist/server/index.js` drops from ~1.67 MB to ~875 KB. satori/resvg move into a separate chunk; the og chunk still contains the base64-inlined font, so OG generation still works. The already-dynamic `app-router-playground` case is unaffected.

## Tests

Verified locally with targeted runs:
- `tests/og-inline.test.ts`, `tests/og-font-patch.test.ts`, `tests/shims.test.ts` — 1040 passed
- `tests/build-optimization.test.ts` — 82 passed


james-elicx:
Thanks @ask-bonk — addressed both follow-ups in `bfee2e2b`:

1. **`vinext:og-assets` resvg.wasm copy** — confirmed the regression: with `@vercel/og` now code-split, `resvg.wasm` no longer appears in `index.js`, so the safety-net copy was skipped. The plugin now scans **all emitted chunks** via the `writeBundle` `bundle` arg instead of only `index.js`. Verified `dist/server/resvg.wasm` is copied again on the `app-router-cloudflare` build (was missing before the fix), with og still code-split (`index.js` satori/Resvg count = 0).

2. **Dead workaround in `app-router-playground`** — switched `app/api/og/route.tsx` back to the idiomatic static `import { ImageResponse } from "next/og"` and removed the stale comment. Verified the build still code-splits og (satori/Resvg in `index.js` = 0) and the resvg.wasm copy fires.

Targeted tests still green: og-inline, og-font-patch, shims, build-optimization (1122 passed).

/bigbonk review for issues

PR #1776, opened 2026-06-05
fix(check): prevent regex stack overflow / hang on very large files

james-elicx:
## Problem

`vinext check` could crash with `RangeError: Maximum call stack size exceeded` (or hang for minutes) on very large source files. Two regexes run against whole file contents are the culprits:

1. **`cjsGlobalScanRegex`** (the stack overflow) — used to detect free `__dirname` / `__filename`. Its string-body sub-pattern `(?:[^"\\]|\\.)*` is a **star over an alternation group**, which V8 cannot compile into a tight loop, so it pushes one backtrack frame *per character*. A long string literal (~10M chars; lower on smaller stacks / under load) overflows the regex stack. Reproduced reliably; unrolling the regex did **not** fix the escape-heavy variant — a backtracking regex fundamentally can't safely consume unbounded string bodies.

2. **`stringPluginRegex`** (PostCSS string-form plugin detection) — two lazy `[\s\S]*?` quantifiers around a capture group cause **quadratic catastrophic backtracking** on a config with no closing `]` (measured 25→102→402→1555ms as input doubled → multi-minute hang on a large file).

## Fix

- Replaced `cjsGlobalScanRegex` with `hasFreeCjsGlobal()` — a hand-written single-pass scanner (O(n) time, O(template-nesting) stack) that tracks code/string/comment/template state and flags `__dirname`/`__filename` only in real code, including inside `${…}` template expressions. Behavior-preserving (all existing `__dirname` tests pass).
- Replaced `stringPluginRegex` with a linear anchored regex `/plugins\s*:\s*\[\s*['"]/` that matches exactly the same configs (first array element is a bare string literal) without scanning for a closing `]`.

The two other content regexes (`viewTransitionRegex`, `importRegex`) use stars over *simple character classes*, which V8 handles iteratively — verified safe at 120MB, left unchanged.

## Tests

- New `hasFreeCjsGlobal` suite: behavior (strings/comments/plain templates ignored, `${…}` detected, nested templates, substring non-matches) **plus stack-overflow regressions** — 20MB unterminated string, 30MB escape-heavy string, real `__dirname` after a 5MB benign string, each asserting correctness + sub-2–3s completion.
- New PostCSS regressions: huge unterminated plugin array (string-form and require-form heads) asserting fast completion.

All 116 tests in `tests/check.test.ts` pass (~340ms). Scoped the run to the affected file; CI covers breadth.

### DIFF
--- a/packages/vinext/src/check.ts
+++ b/packages/vinext/src/check.ts
@@ -798,8 +798,7 @@ export function checkConventions(root: string): CheckItem[] {
         items.push({
           name: `PostCSS string-form plugins (${configFile})`,
           status: "partial",
-          detail:
-            "string-form PostCSS plugins need resolution — vinext handles this automatically",
+          detail: "string-form PostCSS plugins need resolution — vinext handles this automatically",
         });
       }
       break; // Only check the first config file found
--- a/packages/vinext/src/check.ts
+++ b/packages/vinext/src/check.ts
@@ -386,9 +386,11 @@ const REGEX_PRECEDING_KEYWORDS = new Set([
  *
  * It is a lexer-grade scanner, not a parser: it tracks just enough state (string /
  * template / comment / regex contexts, and whether a `/` is in expression position)
- * to avoid mistaking quotes inside one context for the start of another. It does not
- * attempt full JS semantics (e.g. `}` is treated as regex-allowing), which is fine for
- * an advisory check.
+ * to avoid mistaking quotes inside one context for the start of another. Where the
+ * division-vs-regex distinction is ambiguous it biases toward division, because a
+ * misread division is harmless (it never consumes a following identifier) whereas a
+ * misread regex would swallow the rest of the line and could hide a later __dirname.
+ * This is fine for an advisory check.
  */
 export function hasFreeCjsGlobal(content: string): boolean {
   const n = content.length;
@@ -505,7 +507,11 @@ export function hasFreeCjsGlobal(content: string): boolean {
         stack.pop(); // close the ${ … } and return to the template
       } else {
         if (top.depth > 0) top.depth--;
-        top.prevType = "op";
+        // Treat `}` as value-producing so a following `/` is division (the common
+        // `{ … } / x` object-literal case). A block `}` followed by a regex is rarer,
+        // and misreading that regex as division is harmless here — division never
+        // consumes a following identifier, so it cannot hide a later __dirname.
+        top.prevType = "value";
       }
       i++;
       continue;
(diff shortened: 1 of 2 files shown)


james-elicx:
Thanks for the review! Addressed in 83cee16.

**1. Regex literals masking a later `__dirname` (the regression)** — Fixed properly rather than just documenting. The scanner now tracks whether a `/` is in expression position (→ regex literal) or follows a value (→ division), via a per-frame `prevType` that's updated as tokens are consumed (identifiers/numbers/`)`/`]`/strings/templates/regexes are values; operators and regex-preceding keywords like `return`/`typeof`/`case` allow a regex). Regex-literal bodies are skipped honoring `\` escapes and `[…]` char classes (where `/` doesn't terminate), plus trailing flags. Both reported cases now return `true`:

- `hasFreeCjsGlobal("const r = /'/; const d = __dirname;")` → `true`
- `` hasFreeCjsGlobal("const r = /`/;\nconst d = __dirname;") `` → `true`

New tests cover regex-literals hiding/not-hiding `__dirname`, division not being mistaken for a regex, and `__dirname` inside a regex literal being correctly ignored.

**2. Magic-number offsets** — Removed. The scanner now consumes the full identifier and matches it against a `CJS_GLOBALS` set, so there are no `i + 9` / `i + 10` offsets.

**3. Newline/line-continuation comment** — Reworded. The code already handled it correctly (the `\` escape branch consumes a `\<newline>` continuation); only an *unescaped* newline ends the scan. The comment now says that instead of overstating the JS rule.

I also added a note to the function's doc comment that it's a lexer-grade scanner, not a full parser (e.g. `}` is treated as regex-allowing), which is fine for an advisory check.

All 119 tests in `tests/check.test.ts` pass; formatting clean.

james-elicx:
Thanks for the re-review — good catch. Fixed in 913ffaa.

The division-vs-regex misclassification is now resolved by biasing ambiguous cases toward **division**, since a misread division is harmless (it never consumes a following identifier) while a misread regex swallows the line:

- **`}`** → now value-producing, so `{ … } / x` is division.
- **postfix `++`/`--`** → consumed as a unit without changing expression position (postfix after a value keeps "value"; prefix after an operator keeps "op"), so `i++ / 2` is division.

All three reported cases now return `true`:

- `hasFreeCjsGlobal("i++ / 2; const d = __dirname;")` → `true`
- `hasFreeCjsGlobal("const x = {a:1} / 2; const d = __dirname;")` → `true`
- `hasFreeCjsGlobal("i++ / __dirname / b; z;")` → `true`

Added tests for division after `}`/postfix `++`/`--` (including your `const half = list.pop() ? a-- / 2 : 0; const root = __dirname;` example), plus one confirming a real regex after a prefix `++`/keyword still parses correctly. Updated the doc comment to state the division-biasing rationale instead of the old `}`-as-regex-allowing caveat. 121 tests pass; formatting clean.

james-elicx:
Thanks for the round-3 review. Both notes were non-blocking; addressed in 2d25423 per your recommendations.

**1. Value-position regex literal (residual false negative)** — Agreed this is the symmetric tail of the division-vs-regex ambiguity and can't be closed without real parser context (was the `}` a block or an object?). A partial brace-classifier would risk introducing new misclassifications, so I took your recommended route: pinned the limitation with tests and widened the doc caveat.

- Added a test asserting `hasFreeCjsGlobal("function f(){} /'/.test(x); const d = __dirname;")` → `false` (the known same-line gap), and one asserting the multi-line variant → `true` (string scanning bails at `\n`).
- The `hasFreeCjsGlobal` doc comment now has an explicit "Known limitation" paragraph describing the value-position-regex case, why it needs a parser, and why it's accepted for an advisory check.

**2. PostCSS "exactly the same configs"** — Softened. The comment now says it matches the same *string-form* configs but **intentionally** diverges on the `require()`-form (`plugins: [require("x")]`), which the old regex matched as a false positive and the anchored one correctly skips — consistent with the rest of the comment. Also noted the `/* comment */`-before-string case as not worth handling.

123 tests pass; formatting clean.

Issue #659, opened 2026-03-22
`rsc:scan-strip` parses MDX files as JS before `@mdx-js/rollup` transforms them, causing parse errors


keufcp:
## Environment

- vinext: 0.0.34
- @vitejs/plugin-rsc: 0.5.21
- @mdx-js/rollup: (installed, auto-detected by vinext)

## Problem

When using `import.meta.glob` to dynamically load MDX files in an RSC page,
the `rsc:scan-strip` plugin attempts to parse `.mdx` files as JavaScript
before `@mdx-js/rollup` has a chance to transform them. This causes a parse error.

## Error
```
[plugin rsc:scan-strip] src/content/posts/2025/08/20/second-post/index.mdx
Error: Parse error @:1:1
```

The MDX files contain standard frontmatter (YAML between `---`) and JSX syntax
(e.g. `<span className='text-red-500'>`), which are valid MDX but invalid JavaScript.

## Root cause

`rsc:scan-strip` runs before `@mdx-js/rollup` in the plugin pipeline,
so it sees raw MDX instead of the transformed JS output.

## Reproduction

1. Create an App Router project with MDX content files containing frontmatter
2. Use `import.meta.glob('/src/content/posts/**/*.mdx')` in a lib file
3. Run `vite build`

## Expected behavior

`@mdx-js/rollup` should transform MDX files to JS before `rsc:scan-strip` processes them,
or `rsc:scan-strip` should skip non-JS files that will be handled by other plugins.

## Workaround

None found without changing the content file structure.

PR #1777, opened 2026-06-05
perf(build): minify server build environments by default

james-elicx:
## Problem

vinext's server-side build environments ship **unminified**. Vite only applies `build.minify` to the `client` environment by default — for every non-client environment (`rsc`, `ssr`, and the Cloudflare worker env) `build.minify` defaults to `false`, and vinext never sets it. So the deployed worker (`dist/server/index.js`), the SSR renderer (`dist/server/ssr/index.js`), and the Pages-Router-on-Cloudflare worker all ship full of readable identifiers, comments, and whitespace.

Raw size drives workerd cold-start parse CPU; gzip size is counted against the Cloudflare Workers size limit. Minifying is a large, cheap win.

## Approach

Add a build-only `configEnvironment` plugin (`vinext:server-minify-defaults`) that sets `minify: true` for non-client environments. This is a **true default that yields to user configuration**, not a hard override:

- **prod-only**: `apply: "build"` scopes it to production builds (never dev/preview).
- **yields to user/plugin config**: Vite seeds each non-client environment's `build` from the top-level `config.build` before running `configEnvironment` (see `getDefaultEnvironmentOptions`), so the incoming `config.build?.minify` reflects any explicit setting — top-level, `environments.<name>.build.minify`, or set by an earlier plugin (e.g. `@cloudflare/vite-plugin`). If anyone already chose a value (including `false`), we leave it untouched; we only fill the default when it's still `undefined`.
- `minify: true` lets the rolldown/oxc toolchain pick its native minifier rather than pinning `'esbuild'`.

Using `configEnvironment` (rather than the per-environment blocks in the `config` hook) means it also covers the Cloudflare worker environment that `@cloudflare/vite-plugin` owns — which vinext's `config` hook doesn't define — so Pages-Router-on-Cloudflare apps benefit too.

Scope kept to minify only; sourcemaps are intentionally not touched (possible focused follow-up).

## Measured before/after (real builds this session)

| Example | File | raw before | raw after | gzip before | gzip after |
|---|---|---|---|---|---|
| hackernews | `dist/server/index.js` | 1342164 (1311 KB) | 620699 (606 KB), **-54%** | 338406 (330 KB) | 202564 (198 KB), **-40%** |
| hackernews | `dist/server/ssr/index.js` | 643167 (628 KB) | 275609 (269 KB), **-57%** | 136309 (133 KB) | 84575 (83 KB), **-38%** |
| app-router-cloudflare | `dist/server/index.js` | 875457 (855 KB) | 375196 (366 KB), **-57%** | 208052 (203 KB) | 107208 (105 KB), **-48%** |
| app-router-cloudflare | `dist/server/ssr/index.js` | 798913 (780 KB) | 330557 (323 KB), **-59%** | 182647 (178 KB) | 102151 (100 KB), **-44%** |
| pages-router-cloudflare | `dist/pages_router_cloudflare/index.js` (worker) | 826635 (807 KB) | 360275 (352 KB), **-56%** | 184693 (180 KB) | 106276 (104 KB), **-42%** |

"before" numbers were produced by temporarily setting `build: { minify: false }` (reproducing the old default) and rebuilding; "after" are clean builds with this change.

## User-override verification

Temporarily set `build: { minify: false }` in `examples/hackernews`, `examples/app-router-cloudflare`, and `examples/pages-router-cloudflare`'s `vite.config.ts`, rebuilt, and confirmed the output reverted to **unminified** (readable identifiers/whitespace, e.g. hackernews `dist/server/index.js` back to 1342164 raw; pages worker back to 826635 raw with `import fs from "node:fs";`). This proves the default yields to explicit user config. All temporary config edits were reverted.

## Runtime check

Ran `vp preview` on the minified `app-router-cloudflare` worker: `GET /` returned **HTTP 200** with fully rendered HTML (`<title>vinext on Cloudflare Workers</title>`), including the `counter` client component and its hydration chunk (`_next/static/counter-*.js`) — confirming the RSC/SSR/client split is intact under minification.

## Tests run (targeted only)

- `tests/build-optimization.test.ts` + `tests/shims.test.ts` — **1103 passed**
- `tests/use-client-export-all-build.test.ts` + `tests/clean-build-output.test.ts` — **6 passed** (RSC `"use client"` directive handling intact)
- `vp check` on `packages/vinext` — formatting + lint + typecheck **pass**, no new errors

(Did not run the full suite per CPU constraints.)

### DIFF
--- a/packages/vinext/src/build/layout-classification.ts
+++ b/packages/vinext/src/build/layout-classification.ts
@@ -127,7 +127,7 @@ export function isStaticModuleGraphResult(
  * Shared layouts (same file appearing in multiple routes) are classified once
  * and deduplicated by layout ID.
  *
- * @internal Not called by production code. The `generateBundle` hook in
+ * @internal Not called by production code. The `renderChunk` hook in
  * `index.ts` calls `classifyLayoutByModuleGraph` directly and composes
  * via the numeric-index manifest in `route-classification-manifest.ts`.
  * Used only by `tests/layout-classification.test.ts`.
--- a/packages/vinext/src/build/route-classification-injector.ts
+++ b/packages/vinext/src/build/route-classification-injector.ts
@@ -19,7 +19,7 @@ import {
  * those stubs with build-time decisions once the final RSC module graph exists.
  */
 
-export type RouteClassificationChunk = {
+type RouteClassificationChunk = {
   code: string;
   fileName: string;
 };
--- a/packages/vinext/src/build/route-classification-injector.ts
+++ b/packages/vinext/src/build/route-classification-injector.ts
@@ -48,9 +48,12 @@ type BuildLayer2ClassificationsOptions = Pick<
 >;
 
 // The `?` after the semicolon is intentional: Rolldown may or may not emit the
-// trailing semicolon depending on minification settings. This regex relies on
-// `__VINEXT_CLASS` retaining its name, which holds because RSC entry chunk
-// bindings are not subject to scope-hoisting renames.
+// trailing semicolon depending on codegen settings. This regex relies on the
+// stub still being in its readable, unminified form (`__VINEXT_CLASS` /
+// `routeIdx` not yet renamed) — which is why the caller patches it from a
+// `renderChunk` hook with `order: "pre"`, before rolldown's minifier mangles
+// top-level names. Patching in `generateBundle` (post-minify) would silently
+// fail to match once `build.minify` is on (the default for server envs).
 const CLASS_STUB_RE = /function __VINEXT_CLASS\(routeIdx\)\s*\{\s*return null;?\s*\}/;
 const REASONS_STUB_RE = /function __VINEXT_CLASS_REASONS\(routeIdx\)\s*\{\s*return null;?\s*\}/;
 
--- a/packages/vinext/src/build/route-classification-injector.ts
+++ b/packages/vinext/src/build/route-classification-injector.ts
@@ -81,7 +81,7 @@ function findClassificationChunk(
         .map((chunk) => chunk.fileName)
         .join(
           ", ",
-        )} but no chunk contains the stub body. The generator and generateBundle have drifted.`,
+        )} but no chunk contains the stub body. The generator and the classification injector have drifted.`,
     );
   }
 
@@ -147,7 +147,7 @@ export function planRouteClassificationInjection(
 
   if (options.enableDebugReasons && !REASONS_STUB_RE.test(target.code)) {
     throw new Error(
-      "vinext: build-time classification — __VINEXT_CLASS_REASONS stub is missing alongside __VINEXT_CLASS. The generator and generateBundle have drifted.",
+      "vinext: build-time classification — __VINEXT_CLASS_REASONS stub is missing alongside __VINEXT_CLASS. The generator and the classification injector have drifted.",
     );
   }
 
--- a/packages/vinext/src/build/route-classification-injector.ts
+++ b/packages/vinext/src/build/route-classification-injector.ts
@@ -6,7 +6,7 @@ import {
 } from "./layout-classification.js";
 import type { ModuleGraphStaticReason } from "./layout-classification-types.js";
 import {
-  buildGenerateBundleReplacement,
+  buildClassificationReplacement,
   buildReasonsReplacement,
   type RouteClassificationManifest,
 } from "./route-classification-manifest.js";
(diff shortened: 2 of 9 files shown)


james-elicx:
/bigbonk re-review for issues.

There's nothing to test in the pr - it's a self-explanatory plugin.

Sourcemaps should probs be opt-in by a user rather than on-by-default.

james-elicx:
Pushed two follow-ups since the last review — please re-review the full change:

1. **Real production bug fixed:** `build.minify` renamed the `__VINEXT_CLASS` build-time route-classification marker *before* the old `generateBundle` patch ran, so build-time classification silently no-op'd in every minified production build. Moved the injection to a `renderChunk` hook with `order: "pre"` (runs before rolldown's minifier). Verified the marker is readable at patch time and `buildTimeClassifications` is wired on a real hackernews prod build.
2. **Unit tests:** the three failing tests introspected build output for readable identifiers/markers; they now build their fixtures with `minify: false` (exercising the yields-to-user override). Audited the rest of the build-introspection tests — no others affected.
3. Dropped a now-unused exported type to satisfy knip.

/bigbonk review for issues

james-elicx:
Addressed both findings from the last review (`79f7df4a`):

**1. Regression test now exercises the bug (un-masked).** Added a "production default — minify on" suite to `build-time-classification-integration.test.ts` that builds the fixture with server minification LEFT ON and asserts the dispatch table is populated via minify-robust signals: the structural `case N: return new Map([[idx, …` shape, and (debug build) the `module-graph` reason literal — which is injected *only* by the build-time patch (note: `segment-config`/`no-classifier` also appear in bundled runtime source, so they aren't unique; `module-graph` is). 
**Verified it actually guards the bug:** I reproduced the original failure faithfully — moving the patch back to a post-minify `generateBundle` hook — and confirmed both new assertions FAIL there, then pass once the `renderChunk` fix is restored. (Note: flipping `order:"pre"`→`"post"` alone does *not* reproduce it — rolldown runs the minifier after all `renderChunk` handlers regardless of order — so the repro uses generateBundle.)
`cache-adapters-build` and `tsconfig-path-alias-build` now also run under the production default (minify on): the former asserts via a surviving string-literal marker; the latter's old `not.toContain('title: "Second Post"')` turned out to be a quote-escaping artifact (it only passed unminified by accident), replaced with a real #659 guard (compiled JSX present, no raw YAML frontmatter block).

**2. Stale `generateBundle` references fixed** — the two drift `Error` strings + comments in `route-classification-injector.ts`, the stub comments in `app-rsc-entry.ts`, and two comments in `index.ts` now name `renderChunk`. Added a comment on the per-chunk assumption (finding #3).

`vp run check` + `vp run knip` clean; affected + regression tests green (1116 passed).

/bigbonk review for issues

james-elicx:
Removed the test-only minify disabling entirely (`71e86d21`). No test builds with `minify: false` anymore — every classification suite now builds with vinext's production defaults (minify **on**), which directly addresses the masking concern from the last review.

Instead of disabling minify to keep `__VINEXT_CLASS` readable, the extractors are now **name-agnostic**: they derive the (mangled) dispatch function name from its property-keyed call site in the route table (`__buildTimeClassifications:` / `__buildTimeReasons:` — object property keys are never mangled), then eval the body (string-literal values like `"static"` / `"dynamic"` / `"module-graph"` also survive minification). So the full dispatch-correctness coverage (force-static→static, force-dyn→dynamic, next/headers omitted, reason layers) now runs against the real minified shipping output.

**Proved it actually guards the bug:** I reproduced the original failure by disabling the renderChunk patch, rebuilt minified, and the suites FAIL with `<mangled> was not patched — still returns null unconditionally` (the name came through as `$S`, confirming the name-agnostic recovery works on minified code). Restored, all 11 pass. `vp run check` + `vp run knip` clean.

/bigbonk review for issues

james-elicx:
Addressed the remaining nits (`97c29dea`):

**Finding #1 (doc/naming drift):**
- Renamed the misnamed export `buildGenerateBundleReplacement` → `buildClassificationReplacement` (parallel to its sibling `buildReasonsReplacement`), updating its import/usage in `route-classification-injector.ts` and its test in `route-classification-manifest.test.ts`.
- Fixed the two stale `generateBundle` doc comments: `route-classification-manifest.ts:203` and `layout-classification.ts:130` now say `renderChunk`.
- (The one remaining `generateBundle` mention, in `route-classification-injector.ts:55`, is intentional — it explains *why not* to patch in `generateBundle`.)

**Finding #2 (extractRouteIndexByPattern non-greedy pairing):** left as-is — it's pre-existing and, importantly, self-checking: if the call-site→`pattern:` pairing ever desynced, the dispatch-correctness assertions would fail loudly (e.g. `/force-static` would resolve to the wrong route's Map and `map.get(1)`.toBe("static") would break). So a codegen reorder can't silently mask a wrong mapping.

`vp run check` + `vp run knip` clean; classification + manifest + layout-classification tests green (37 passed).

/bigbonk review for issues

PR #1778, opened 2026-06-05
fix(check): only flag config options used as property keys

james-elicx:
## Problem

`vinext check` reported `webpack` as an unsupported config option even when the project had **no custom webpack config**.

`analyzeConfig` used the heuristic `new RegExp(\`\\b${opt}\\b\`)`, which matches the option name *anywhere* in `next.config`. So a project was flagged for `webpack` if the file merely contained a comment (`// removed our custom webpack config`) or a value with the substring (`"react-server-dom-webpack"`). `webpack` is the one users notice because it's the only `unsupported` entry in the scanned list, but the loose matching affected every option.

## Fix

Two layers:

1. **`stripCommentsAndStrings` pre-pass** — a single-pass, lexer-grade scanner (modeled on the existing `hasFreeCjsGlobal`) that runs before detection:
   - Removes line + block comments entirely.
   - Empties string/template-literal *values* (`"react-server-dom-webpack"` → `""`), preserving quoted property keys (`"webpack":`) by peeking for a trailing `:`.
   - Regex-literal aware, so a quote inside `/['"]/` can't hijack string state and swallow following code.
   - O(n) time / O(template-nesting) stack — safe on large files, consistent with #1776.

2. **Property-key regex** — match the option only as an actual key (`opt:`, method shorthand `opt(`, assignment `opt =`, quoted `"opt":`), guarding against identifier substrings.

Both the per-option and nested dot-notation checks run against the stripped content.

## Tests

`tests/check.test.ts`:
- `stripCommentsAndStrings` unit tests: comments removed, string values emptied, quoted keys preserved, `//`/`/*` inside strings not treated as comments, escaped quotes, regex literal not hijacking, template interpolation scanned as code, division not misread as regex.
- `analyzeConfig`: webpack in a comment (with and without trailing punctuation) → not flagged; webpack/headers embedded in string values → not flagged (real keys still detected); webpack as method shorthand / quoted key → flagged.

All `check` tests pass; `vp check` is clean.

## Notes

Accepted limitations (documented in the helper JSDoc, mirroring `hasFreeCjsGlobal`): a value-position regex after a block `}` is read as division, and a string in ternary key position (`cond ? "x" : y`) is preserved. Both are rare in next.config and the check is advisory.

### DIFF
--- a/packages/vinext/package.json
+++ b/packages/vinext/package.json
@@ -78,7 +78,6 @@
     "image-size": "catalog:",
     "ipaddr.js": "catalog:",
     "magic-string": "catalog:",
-    "oxc-parser": "catalog:",
     "vite-plugin-commonjs": "catalog:",
     "vite-tsconfig-paths": "catalog:",
     "web-vitals": "catalog:"
--- a/packages/vinext/package.json
+++ b/packages/vinext/package.json
@@ -78,6 +78,7 @@
     "image-size": "catalog:",
     "ipaddr.js": "catalog:",
     "magic-string": "catalog:",
+    "oxc-parser": "catalog:",
     "vite-plugin-commonjs": "catalog:",
     "vite-tsconfig-paths": "catalog:",
     "web-vitals": "catalog:"
--- a/packages/vinext/src/check.ts
+++ b/packages/vinext/src/check.ts
@@ -355,26 +355,36 @@ function escapeRegExp(s: string): string {
 }
 
 /**
- * Return the body (the text between the matching braces) of the first
- * `name: { … }` / `name = { … }` block in `content`, or null if there is none.
+ * Return the bodies (text between the matching braces) of every
+ * `name: { … }` / `name = { … }` block in `content`. Quoted parent keys
+ * (`"name": { … }`) are matched too. Returns an empty array if there are none.
+ *
+ * All matching blocks are returned, not just the first, so a child key that only
+ * appears in a later same-named block (e.g. two `experimental: {}` objects) is
+ * still found.
  *
  * Expects `content` to already be run through {@link stripCommentsAndStrings},
  * so naive `{`/`}` counting is safe — braces inside comments or string values
  * cannot throw off the balance. Used to scope nested dot-notation child lookups
  * to their parent block instead of matching the child anywhere in the file.
  */
-function extractBlockBody(content: string, name: string): string | null {
-  const opener = new RegExp(String.raw`\b${escapeRegExp(name)}\s*[:=]\s*\{`);
-  const m = opener.exec(content);
-  if (!m) return null;
-  const start = m.index + m[0].length; // first char after the opening `{`
-  let depth = 1;
-  for (let i = start; i < content.length; i++) {
-    const c = content[i];
-    if (c === "{") depth++;
-    else if (c === "}" && --depth === 0) return content.slice(start, i);
+function extractBlockBodies(content: string, name: string): string[] {
+  const opener = new RegExp(String.raw`\b${escapeRegExp(name)}["']?\s*[:=]\s*\{`, "g");
+  const bodies: string[] = [];
+  let m: RegExpExecArray | null;
+  while ((m = opener.exec(content)) !== null) {
+    const start = m.index + m[0].length; // first char after the opening `{`
+    let depth = 1;
+    let i = start;
+    for (; i < content.length; i++) {
+      const c = content[i];
+      if (c === "{") depth++;
+      else if (c === "}" && --depth === 0) break;
+    }
+    bodies.push(content.slice(start, i));
+    opener.lastIndex = i; // resume scanning after this block
   }
-  return content.slice(start); // unterminated block — return the rest
+  return bodies;
 }
 
 // The CJS globals we flag, so the identifier-match check has no magic offsets.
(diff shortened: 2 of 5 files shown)


james-elicx:
Thanks for the thorough review — both false-positive classes are now fixed properly rather than partially.

Instead of relying solely on the boundary+follower regex (which, as you noted, still matched `// webpack: ...` comments and `"webpack:1234"` string values), I added a single-pass pre-pass `stripCommentsAndStrings` that runs before the per-option regexes:

- **Comments** (line + block) are removed entirely → `// TODO: webpack: removed` and `/* old webpack(config) */` no longer match.
- **String/template-literal values** are emptied (`"react-server-dom-webpack"` → `""`) → option names embedded in values no longer match. Quoted property keys (`"webpack":`) are preserved by peeking for a trailing `:`.
- It's **modeled on the existing `hasFreeCjsGlobal` scanner** (same frame-based, regex-literal-aware approach), so a quote inside a regex literal like `/['"]/` can't hijack string state, and it's O(n) — safe on large files, consistent with #1776.

On your minor points:
- The nested dot-notation child-ref check now runs against the stripped content too, so it inherits the same protection.
- I kept the accepted-limitation note (ternary key-position strings, value-position regex after a block `}`) documented in the helper's JSDoc, mirroring `hasFreeCjsGlobal`.

Added tests for both reviewer cases (comment-with-punctuation, string-value) plus scanner edge cases (regex literal not hijacking, template interpolation scanned as code, division not misread as regex).

james-elicx:
Addressed the re-review findings:

- **Dot-notation decoupling (finding 1):** `analyzeConfig` now extracts the parent block body (via `extractBlockBody`, naive brace balancing which is safe on already-stripped content) and requires the child to appear *inside* it. So `i18n: {…}` + `images: { domains: [...] }` no longer flags `i18n.domains`. Added tests for the `images.domains` collision and a child-outside-parent case.
- **Unescaped RegExp interpolation (minor nit):** added an `escapeRegExp` helper used for both the per-option regex and the dot-notation parent/child regexes.
- **Ternary key-position limitation (finding 2):** left as the documented accepted limitation, as you suggested.

139 check tests pass, `vp check` clean.

/bigbonk re-review for issues

james-elicx:
Closed both non-blocking findings from the re-review:

- **Finding 1 (first-block-only):** `extractBlockBody` → `extractBlockBodies`, which iterates *every* matching parent block. So a child in a later same-named block (e.g. two `experimental: {}` objects) is now found. Added a test.
- **Finding 2 (quoted parent keys):** the dot-notation opener now allows an optional closing quote (`\b${name}["']?\s*[:=]\s*\{`), so `"experimental": {}` is matched, consistent with the per-option regex's quoted-key handling. Added a test.
- **Finding 3 (sub-object over-match):** left as-is — intentional conservative-correct behavior.

141 check tests pass, `vp check` clean.

/bigbonk re-review for issues

james-elicx:
Reworked the approach entirely per feedback that a hand-rolled scanner for a config file is overkill.

`analyzeConfig` now **parses next.config to an AST** with `oxc-parser` (the project's own toolchain) and reads the option keys directly off the exported config object. This is exact by construction — comments, string values, regex literals, template literals, and identifier substrings are simply not property keys, so none of the prior false-positive classes can occur.

Deleted `stripCommentsAndStrings`, `extractBlockBodies`, and `escapeRegExp` (~115 net lines removed from check.ts). The AST approach also handles cases the text scan never could:
- config wrapped in plugin calls — `withMDX(config)`, `defineConfig({…})`
- config assigned to a variable then exported
- `as` / `satisfies` annotations
- `module.exports = {…}` and `export default`

Dot-notation options (`experimental.ppr`) are now a real parent→child key relationship from the AST, not scoped text matching, so all three prior findings (decoupling, first-block-only, quoted parent keys) are moot.

`oxc-parser` added as a dependency (already present in the install tree via vite-plus). 135 check tests + 61 init tests pass, build is clean, `oxc-parser` is correctly externalized in dist.

/bigbonk re-review for issues

james-elicx:
Good catch — dropped the `oxc-parser` dependency. Vite already re-exports `parseAst` (its bundled oxc/rolldown parser), and `vite` is already a required peer dep of vinext, so there's nothing new to install.

`collectConfigKeys` now calls `parseAst(source, { lang: "ts" })` and types come from Vite's exported `ESTree` namespace. `{ lang: "ts" }` makes `.ts` configs with type annotations / `as` / `satisfies` parse identically to `.js`/`.mjs`. 135 check + 61 init tests pass, build clean, dist imports `parseAst` from `vite`.

james-elicx:
Fixed the blocking function-form regression. `resolveObject` now unwraps arrow/function configs:

- concise arrow (`(phase) => ({…})`) → the config object is the body directly
- block body (`function(phase, { defaultConfig }) { … return {…} }`) → resolves the first top-level `return` argument

So `module.exports = (phase) => ({ basePath, webpack })` and the `next/constants` `PHASE_*` block form both report correctly again, matching `main`. Added tests for both forms (137 check tests pass, `vp check` clean).

On the non-blocking notes: the parse-failure silent fallback is intentional (documented in the JSDoc), and `module.exports` only resolved at statement top level is fine to leave.

/bigbonk re-review for issues

james-elicx:
Fixed one more function-form gap found in review: `export default function (phase) { return {…} }` parses as a `FunctionDeclaration` (not `FunctionExpression` like `module.exports = function(){…}`), so it was silently reporting no options. Added `FunctionDeclaration` to the function-form branch in `resolveObject` and a regression test. 138 check tests pass, `vp check` clean.

/bigbonk re-review for issues

james-elicx:
Closed the multi-phase finding. `resolveObject` (now `resolveObjects`) returns *all* object literals a config expression can denote, and `collectConfigKeys` merges keys across them.

For function block bodies it collects every reachable `return` argument, descending through `if`/`switch`/`try` but **not** into nested functions, so the canonical `next/constants` form reports both branches:

```js
module.exports = (phase, { defaultConfig }) => {
  if (phase === PHASE_DEVELOPMENT_SERVER) return { trailingSlash: true }; // now reported
  return { webpack: (c) => c };
};
```

Also handles the ternary function form (`(phase) => phase === X ? {…} : {…}`). Nested keys merge across branches too (e.g. `experimental` from multiple returns). Added tests for both. 140 check tests pass, `vp check` clean.

Remaining notes left as documented/accepted: top-level spread sources aren't followed (statically hard), and parse-failure → empty report is intentional.

/bigbonk re-review for issues

PR #1779, opened 2026-06-05
fix(image): scan image imports via AST instead of regex

james-elicx:
## Problem

The `vinext:image-imports` transform scanned raw source text with a regex to find local image imports. A regex over raw text matches `import X from '...img'` anywhere the text appears — including inside comments, strings, and template literals.

When a **commented-out** image import points at a file that exists on disk, the transform still matched it and rewrote the span via `MagicString.overwrite()`. The first generated line stayed behind the `//`, but the synthesized meta-import and `const` declaration escaped the comment, producing:

```js
// import AbstractCat from './_images/AbstractCat.webp';
import __vinext_img_meta_AbstractCat from "/abs/.../AbstractCat.webp?vinext-meta";
const AbstractCat = { src: __vinext_img_url_AbstractCat, width: ..., height: ... };
```

`__vinext_img_url_AbstractCat` is never defined (its import is commented out), so SSR throws `ReferenceError: __vinext_img_url_AbstractCat is not defined` — page errors in dev and a 500 on every page in production.

The bug only triggers when all of these hold:
- a real image file exists at the commented-out import path
- at least one real (non-commented) image import exists in the same file
- the commented-out line matches the `import X from '...(image ext)'` pattern

## Fix

Replace the regex scan with AST-based scanning using Vite's `parseAst`. The handler now only rewrites real top-level `ImportDeclaration` nodes that:
- have a source ending in an image extension, and
- use a single default specifier (`import X from '...'`)

Comments, strings, template literals, named/namespace imports, and side-effect-only imports are all correctly ignored. The existing `code` regex filter is retained purely as a cheap pre-check for whether to invoke the handler (false positives there only cost a parse).

## Tests

Added regression cases to `tests/image-imports.test.ts`:
- commented-out image import (line comment) — the original failure
- commented-out image import (block comment)
- image-import text inside a string literal
- named/namespace image imports are left untransformed

All 22 tests in the file pass; `vp check` is clean.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -3874,22 +3874,49 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
           if (id.startsWith("\0")) return null;
           if (!id.match(/\.(tsx?|jsx?|mjs)$/)) return null;
 
-          const imageImportRe = new RegExp(
-            `import\\s+(\\w+)\\s+from\\s+['"]([^'"]+\\.(${IMAGE_EXTS}))['"];?`,
-            "g",
-          );
-          if (!imageImportRe.test(code)) return null;
+          // The `code` filter above (a regex) only decides whether to invoke
+          // this handler; it can fire on text inside comments, strings, or
+          // template literals. Scanning must therefore be AST-based so we only
+          // rewrite real `import X from '...'` declarations, not text that
+          // merely looks like one. Regex-based scanning generated phantom
+          // variables for commented-out imports, crashing SSR with
+          // `__vinext_img_url_X is not defined`.
+          const imageExtRe = new RegExp(`\\.(${IMAGE_EXTS})$`);
 
-          imageImportRe.lastIndex = 0;
+          let ast: ReturnType<typeof parseAst>;
+          try {
+            ast = parseAst(code);
+          } catch {
+            // Unparseable input (e.g. unsupported syntax) — leave untouched.
+            return null;
+          }
 
           const s = new MagicString(code);
           let hasChanges = false;
 
-          let match;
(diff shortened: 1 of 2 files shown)


james-elicx:
Confirmed and fixed. The critical issue was real: the plugin runs with `enforce: "pre"`, so the handler sees raw source, and `parseAst(code)` (plain JS) threw on JSX and TS type annotations. The handler swallowed the error and returned null, silently skipping image transforms for every `.tsx`/`.jsx`/typed-`.ts` file.

Changes in 6878e82:
- `parseAst(code, { lang: "tsx" })` so JSX + TS + JS all parse with correct node offsets (so `s.overwrite(importNode.start, importNode.end, ...)` stays accurate).
- Added two regression tests with real syntax that plain-JS parsing rejects:
  - a TSX component with JSX in the body
  - a typed `.ts` file with type annotations

I verified locally that both cases failed before the fix (handler returned null) and pass after. Full file is green (24 tests) and `vp check` is clean.

/bigbonk re-review for issues

james-elicx:
Follow-up fix in 0f9c6fd: a self-review caught that always parsing as `tsx` broke plain `.ts` files. The `tsx` grammar reads `<T>` as the start of a JSX element, so TS-only syntax — angle-bracket casts (`<Foo>bar`) and non-comma generic arrows (`<T>(x) => x`) — threw during parse. The `catch` swallowed it and returned null, silently skipping the transform.

Now the parser language is selected by extension: `.ts` → `ts`, everything else (`.tsx`/`.jsx`/`.js`/`.mjs`, which may contain JSX) → `tsx`. Added two regression tests (angle-bracket cast, generic-arrow). 26 tests pass; `vp check` is clean.

/bigbonk re-review for issues

PR #1780, opened 2026-06-05
fix(font): resolve next/font/local paths inside node_modules packages

james-elicx:
## Problem

vinext's `localFont` shim works (CSS classes + `@font-face` are generated), but for npm packages that wrap `next/font/local` the font file URL was the **raw relative path** from the package module rather than a Vite-resolvable URL.

For example, `geist/dist/mono.js` calls:

```js
export const GeistMono = localFont({
  src: "./fonts/geist-mono/GeistMono-Variable.woff2",
  variable: "--font-geist-mono",
});
```

The generated `@font-face { src: url(...) }` referenced `./fonts/geist-mono/GeistMono-Variable.woff2` (relative to the package's `dist/` dir), which 404s at request time.

## Root cause

`createLocalFontsPlugin` (`packages/vinext/src/plugins/fonts.ts`) excluded `node_modules` files — both via the filter (`filter.id.exclude: /node_modules/`) and a defensive `if (id.includes("node_modules")) return null;` guard. So the transform that promotes relative `src` paths to Vite asset imports never ran on package files. This affects **any** npm package wrapping `next/font/local`.

## Fix

Remove the `node_modules` exclusion from both the filter and the defensive guard so package files are transformed like user source — matching Next.js, whose font loader also runs on these package files. The `code: "next/font/local"` filter plus the required default-import check keep the transform from touching unrelated modules, and the `font-local` shim guard still skips vinext's own shim when it lives under `node_modules/vinext/.../shims/font-local.*`.

## Verification

- Reproduced at the unit level: the transform previously returned `null` for a `node_modules/geist/dist/mono.js` id.
- Verified end-to-end with a synthetic `node_modules` font package built through the full App Router pipeline: before the fix the relative path leaked; after the fix the font is emitted as a hashed asset (`_next/static/MyMono-Variable-<hash>.woff2`) and the raw relative path no longer appears in any bundle. (The App Router renders the package server-side via `noExternal`, so the asset import resolves rather than being bypassed by the dep optimizer.)

## Tests

- Replaced the test that encoded the buggy behavior (`returns null for node_modules files`) with a regression test asserting geist-style `node_modules` packages are transformed.
- Added a test confirming vinext's own `font-local` shim is still skipped inside `node_modules`.
- All 37 font-local tests and 1058 shim tests pass; `vp check` is clean.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -3924,7 +3924,7 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
     // Google Fonts import rewrite + self-hosting — see src/plugins/fonts.ts
     createGoogleFontsPlugin(_fontGoogleShimPath, _shimsDir),
     // Local font path resolution — see src/plugins/fonts.ts
-    createLocalFontsPlugin(),
+    createLocalFontsPlugin(_shimsDir),
     // Barrel import optimization:
     // Rewrites `import { Slot } from "radix-ui"` → `import * as Slot from "@radix-ui/react-slot"`
     // for packages listed in optimizePackageImports or DEFAULT_OPTIMIZE_PACKAGES.
--- a/packages/vinext/src/plugins/fonts.ts
+++ b/packages/vinext/src/plugins/fonts.ts
@@ -1120,21 +1120,32 @@ export function createLocalFontsPlugin(): Plugin {
     enforce: "pre",
 
     transform: {
+      // NOTE: node_modules is intentionally NOT excluded here. npm packages
+      // commonly wrap `next/font/local` and ship the font files alongside the
+      // module (e.g. `geist/dist/mono.js` calls
+      // `localFont({ src: "./fonts/geist-mono/GeistMono-Variable.woff2" })`).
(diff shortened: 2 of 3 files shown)


james-elicx:
Addressed the shim-guard concern in d0a05174.

Replaced the loose `id.includes("font-local")` substring check with a precise prefix check, mirroring `createGoogleFontsPlugin`:

- `createLocalFontsPlugin` now takes `shimsDir` (wired with `_shimsDir` at `index.ts:3927`) and guards with `id.startsWith(shimsDir)`.
- This skips only files under vinext's own shims directory, so legitimate third-party packages whose path contains `font-local` (e.g. `font-local-loader`, or fonts shipped under a `font-local/` dir) are still transformed.

Added a regression test (`transforms third-party packages whose path contains 'font-local'`) that builds a `node_modules/font-local-loader/dist/index.js` package and asserts it IS transformed, locking in against a future loosening of the guard. Updated the shim-skip test to use the real resolved shim path. All 38 font-local tests pass and `vp check` is clean.

Left the dep-optimizer / client-only `"use client"` caveat as a pre-existing shared limitation (not a regression) per the non-blocking note.

PR #1781, opened 2026-06-06
perf(router): lazy-load App Router page and route-handler modules

james-elicx:
## Summary

The App Router RSC entry statically imported every route module, so they were all evaluated at Worker startup. For typical pages this is cheap, but routes with expensive **module-level initialization** make startup scale linearly and can exceed Cloudflare's ~400ms startup CPU budget.

This makes lazy route-module loading the **default**: static-route page modules and all route-handler modules are emitted as `() => import()` thunks instead of `import * as`, so they're code-split out of the entry's top-level evaluation and loaded on demand for the matched route only.

## Measured impact (real `wrangler deploy` → "Worker Startup Time", `vinext` account)

**Typical pages** — startup is flat regardless of route count, ~0.013 ms/route (noise):

| Routes | RSC `index.js` | Startup |
|-------:|---------------:|--------:|
| 5      | 888 KB | 16–20 ms |
| 50     | 996 KB | 17–18 ms |
| 150    | 1238 KB | 20–21 ms |
| 300    | 1604 KB | 18–26 ms |

**Heavy module-level init** (each page builds a large lookup table at import) — this is the failure mode lazy loading fixes:

| Variant (150 routes) | Startup |
|----------------------|--------:|
| eager (before) | **414–467 ms** |
| lazy (this PR) | **16–19 ms** |

Typical apps pay nothing; heavy apps stop blowing the startup budget.

## How it works

- **Codegen** (`app-rsc-manifest.ts`): new `getLazyLoaderVar` emits `const load_N = () => import(path)`. Static-route pages and all route handlers use it; route table fields become `page/__loadPage` and `routeHandler/__loadRouteHandler`.
- **Helper** (`app-route-module-loader.ts`): `ensureAppRouteModulesLoaded(route)` hydrates the lazy modules onto the synchronous `page`/`routeHandler` fields. Idempotent, and dedups concurrent loads into a single import.
- **Hydration sites**: invoked at the central match point (`app-rsc-handler.ts`) and at every mid-flight route lookup that reads a module before `buildPageElement` — server-action redirect/rerender targets (`app-server-action-execution.ts`), interception + ISR revalidation source routes (`app-page-dispatch.ts`), `buildPageElements`, and `handleServerActionRequest`.

## Scope / follow-up

- **Dynamic-route pages stay eager** because their `generateStaticParams` is referenced in the module-level `generateStaticParamsMap`. Making those lazy needs a prerender static-params resolver rework — tracked as a follow-up.
- Pages Router server entry is untouched (separate change if wanted).

## Testing

- End-to-end on real Cloudflare: static (lazy), dynamic `[slug]` (eager), and `/api/hello` route handler (lazy) all render 200; heavy-150 startup 19 ms.
- Green suites: `features` (311), prod/dev server + interception + prerender + production build (316), route-handler + ISR + worker-entry + codegen (142), app-page consumers (162), action/dispatch units (81), `entry-templates` (25, updated), new `app-route-module-loader` unit tests (7). `vp check` clean.

🤖 Generated with [opencode](https://opencode.ai)

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -80,6 +80,10 @@ const appRscErrorHandlerPath = resolveEntryPath(
   import.meta.url,
 );
 const appRequestContextPath = resolveEntryPath("../server/app-request-context.js", import.meta.url);
+const appRouteModuleLoaderPath = resolveEntryPath(
+  "../server/app-route-module-loader.js",
+  import.meta.url,
+);
 const appPrerenderStaticParamsPath = resolveEntryPath(
   "../server/app-prerender-static-params.js",
   import.meta.url,
@@ -326,6 +330,7 @@ import { clearAppRequestContext as __clearRequestContext, setAppNavigationContex
 
 __configureMemoryCacheHandler({ cacheMaxMemorySize: ${JSON.stringify(cacheMaxMemorySize)} });
 import { createAppPrerenderStaticParamsResolver as __createAppPrerenderStaticParamsResolver } from ${JSON.stringify(appPrerenderStaticParamsPath)};
+import { ensureAppRouteModulesLoaded as __ensureRouteLoaded } from ${JSON.stringify(appRouteModuleLoaderPath)};
 import { seedMemoryCacheFromPrerender as __seedMemoryCacheFromPrerender } from ${JSON.stringify(seedCachePath)};
 
 const __draftModeSecret = ${JSON.stringify(draftModeSecret)};
@@ -484,6 +489,8 @@ function findIntercept(pathname, sourcePathname = null) {
 }
 
 async function buildPageElements(route, params, routePath, pageRequest, layoutParamAccess) {
+  // Hydrate lazy page/route-handler modules before any synchronous read.
+  await __ensureRouteLoaded(route);
   return __buildPageElements({
     route,
     params,
@@ -553,6 +560,7 @@ ${rootParamNameEntries.join("\n")}
 
 export default __createAppRscHandler({
   basePath: __basePath,
+  ensureRouteLoaded: __ensureRouteLoaded,
   clearRequestContext() {
     __clearRequestContext();
   },
@@ -597,6 +605,7 @@ export default __createAppRscHandler({
     const _asyncRouteParams = makeThenableParams(params);
     return __dispatchAppPage({
       basePath: __basePath,
+      ensureRouteLoaded: __ensureRouteLoaded,
       clientTraceMetadata: __clientTraceMetadata,
       buildPageElement(targetRoute, targetParams, targetOpts, targetSearchParams, layoutParamAccess) {
         return buildPageElements(targetRoute, targetParams, cleanPathname, {
@@ -771,7 +780,7 @@ export default __createAppRscHandler({
       setHeadersAccessPhase,
     });
   },
-  handleServerActionRequest({
+  async handleServerActionRequest({
     actionId,
     cleanPathname,
     contentType,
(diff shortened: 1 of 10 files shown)


james-elicx:
Thanks @ask-bonk — addressed in 61a7eb29.

**1. Sticky failed import (fixed).** `ensureAppRouteModulesLoaded` now chains a `.catch` that clears `__loading` (leaving `__loaded` false) and re-throws. The current request still observes the error, but the route is no longer wedged — the next request retries, matching the eager model's per-isolate retry semantics and avoiding the stored-rejection unhandled-rejection footgun. Added a rejection-path unit test (`does not cache a failed import: re-throws and retries on the next call`).

**2. CodeQL false positive (documented).** Added a comment at `getLazyLoaderVar` explaining the path is a trusted filesystem-scan result (`route.pagePath`/`route.routePath`), identical trust model to the eager `import * as` in `getImportVar` two lines up. Will dismiss the alert as a false positive in the UI.

**3. Double hydration on the intercept path (intentional, keeping).** The dispatch-site `await ensureRouteLoaded(interceptRoute)` in `app-page-dispatch.ts` is required, not redundant: `resolveRouteFetchCacheMode(interceptRoute)` reads `interceptRoute.page` *before* `buildPageElement` runs, so the route must be hydrated at that point. The inner `buildPageElements` await is then a cheap idempotent no-op. Leaving both for correctness.

Loader + entry-template + action/dispatch suites green (103 passing).

james-elicx:
Thanks @ask-bonk — addressed the non-blocking item in 8220e7fc.

**Item 1 (comment reworded).** Updated the `getLazyLoaderVar` justification to the accurate reason: route handlers are lazy because they're never sourced into `generateStaticParamsMap` (which reads only layouts + page, never `route.routePath`), not because they "have no generateStaticParams." Also noted that wiring route-handler `generateStaticParams` into prerendering remains a separate, pre-existing gap unaffected by this PR.

Item 2 (double hydration) is intentional/idempotent as discussed — no change. CodeQL alert will be dismissed as a false positive in the UI.

james-elicx:
Pushed `ad2d9159` — **all page modules are now lazy, including dynamic routes.**

The prior cut had a real gap (thanks for the push on this): dynamic-route pages stayed eager because their `generateStaticParams` was referenced synchronously in the module-level `generateStaticParamsMap`. Since `isDynamic` is inherited from ancestors, any app with a dynamic **root** segment (`app/[locale]/...` i18n, `app/[tenant]/...`) made *every* route dynamic → nothing was lazy, and the heavy `[slug]` page (the prime heavy-module-init candidate) stayed eager too.

Now:
- Every page module is lazy (`isDynamic` no longer factors into the decision).
- `generateStaticParamsMap` embeds lazy `{ load }` page sources; `createAppPrerenderStaticParamsResolver` imports them on demand at prerender time (already async) and returns the same `null` "no static params" sentinel the driver relies on — so SSG / `output: export` behavior is unchanged.
- Eager set is now only genuinely-shared modules: layouts, templates, boundaries, intercepts, global-error.

**Verified on real Cloudflare:** 150 heavy pages under a dynamic `[locale]` root went from eager (~450 ms startup) to **22 ms**, and SSG via the `[locale]` layout's `generateStaticParams` still prerenders correctly.

**Tests:** prerender + prerender-route-params + prerender-endpoints + features + prod/dev server + interception + actions all green (699 in the broad batch); added `app-prerender-static-params` unit tests for the lazy-source path + null sentinel; `entry-templates` updated for all-lazy pages and `{ load }` sources.

james-elicx:
Thanks @ask-bonk — addressed the non-blocking robustness note in `75165307`.

**Note 1 (source-order robustness — fixed).** The resolver no longer partitions into `[...eagerFns, ...lazyFns]`. It now resolves sources in their original declared order: each source maps to its eager function or its awaited lazy function, then non-functions are dropped. Composition order no longer depends on the codegen emitting layout (eager) sources before the page (lazy) source. Added a unit test with a lazy source ordered *before* an eager one to lock it in.

**Note 2 (sticky prerender-time load failure).** Leaving as-is per your call — it's build-time and deterministic, where an import failure is fatal anyway, so memoizing it is fine. (The runtime loader still resets-and-retries.)

Notes 3 (CodeQL false positive documented inline; double-hydration intentional/idempotent) — no change.

Prerender suites (112) + resolver unit tests (6) green.

Issue #1364, opened 2026-05-20
Interception routes: build and route resolution failures


james-elicx:
> *This issue was created by an agent analysing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/26163751905) (vinext `main` vs Next.js `v16.2.6`, 2026-05-20).*

## Problem

Interception routes (using `(.)`, `(..)`, `(...)`, and `(..)(..)` notation) fail at build time or resolve to the wrong page at runtime. The route scanner does not properly handle the interception segment notation combined with dynamic segments and parallel route slots.

All tests in the interception suites fail with build errors (`Custom deploy script failed`), indicating the build pipeline cannot process the route structure.

## Estimated Impact

~22 test failures across multiple interception route suites.

## Affected Test Suites

- `test/e2e/app-dir/interception-dynamic-segment/interception-dynamic-segment.test.ts` (14 failures)
- `test/e2e/app-dir/interception-dynamic-single-segment/interception-dynamic-single-segment.test.ts` (5 failures)
- `test/e2e/app-dir/interception-routes-multiple-catchall/interception-routes-multiple-catchall.test.ts` (5 failures)
- `test/e2e/app-dir/interception-dynamic-segment-middleware/interception-dynamic-segment-middleware.test.ts` (3 failures)
- `test/e2e/app-dir/interception-segments-two-levels-above/interception-segments-two-levels-above.test.ts` (3 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test with an interception route (e.g., `app/@modal/(.)photo/[id]/page.tsx`) and verify the build succeeds and the route resolves correctly. Confirm it fails.

2. **Study Next.js interception route handling.** Search `.nextjs-ref/packages/next/src/` for interception route parsing, the `(.)` notation, and how interception segments affect route matching.

3. **Fix `routing/app-router.ts`.** The route scanner needs to recognize `(.)`, `(..)`, `(...)`, and `(..)(..)` directory names as interception markers rather than treating them as malformed route segments. The interception notation should be stripped for route matching purposes while preserving the interception behavior for client-side navigation.

james-elicx:
Still observed in the [Next.js Deploy Suite nightly run on 2026-05-22](https://github.com/cloudflare/vinext/actions/runs/26282973805) (vinext `main` vs Next.js `v16.2.6`): **30 failures across 6 test suites**.

<details><summary>Affected suites this run</summary>

[...]

PR #1799, opened 2026-06-06
fix(dev-overlay): clear stale build error overlay on successful HMR

james-elicx:
## Problem

CI flake in `tests/e2e/app-router/dev-error-overlay.spec.ts`:

> `server component HMR surfaces Vite build errors after a clean load`
> `expect(locator).toBeHidden() failed` — the `vinext-dev-error-overlay` stays `visible` after the file is fixed.

Plus flaky `tests/e2e/app-router/interception-dynamic-segment-middleware.spec.ts` (`back/forward...` and `repeated interceptions...`).

### Build error overlay
The dev overlay's **"Build Error"** (Vite transform failure) was only cleared once App Router's `applyRscHmrUpdate` reached `dismissOverlay()`. That call sits behind several readiness guards:

- `updateId !== latestRscHmrUpdateId` (concurrent `rsc:update` races)
- `document.documentElement.id === "__next_error__"` → reload
- `browserRouterStateHasEverCommitted && !hasBrowserRouterState()` → reload / silent skip

When any guard bails, `dismissOverlay()` never runs and the stale build error lingers — exactly the observed flake.

An `rsc:update` (App Router server components) or `vite:afterUpdate` (client / Pages Router) event only fires **after** a transform succeeds, so a `"vite"`-sourced build error is unambiguously stale at that point.

### Interception tests
`interception-dynamic-segment-middleware.spec.ts` marks the unimplemented #1364 Part C behavior with `test.fail()`. Interception fires non-deterministically with middleware locale rewrites, so on some runs `test.fail()` flips to an **unexpected pass** and fails CI.

## Fix

- Add `dismissBuildErrors()` to the overlay store — clears only `"vite"`-source build errors, leaving runtime errors for the re-render to reconcile.
- Call it eagerly from the `rsc:update` listener (App Router server components) and from a new `vite:afterUpdate` listener in `installViteHmrErrorHandler` (client components + Pages Router). Recovery no longer depends on the RSC re-render's guards.
- Convert the four `test.fail()` markers in the middleware interception spec to `test.fixme()` (expected-to-fail **and** skipped), matching the repo's existing convention for known-broken tests. Revert to `test.fail()` once #1364 Part C lands.

## Testing

- `vp test run tests/app-browser-entry.test.ts tests/entry-templates.test.ts tests/pages-router.test.ts` — 464 passed (incl. new `dismissBuildErrors` unit tests).
- `dev-error-overlay.spec.ts` e2e suite run with `--repeat-each=3` — all green.
- `interception-dynamic-segment-middleware.spec.ts` — now deterministically skipped.
- `vp check` clean on all changed files.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -97,6 +97,7 @@ import { buildRequestHeadersFromMiddlewareResponse } from "./server/middleware-r
 import { detectPackageManager } from "./utils/project.js";
 import { isUnknownRecord as isRecord } from "./utils/record.js";
 import { manifestFilesWithBase } from "./utils/manifest-paths.js";
+import { resolveRscModuleUrl } from "./utils/rsc-module-url.js";
 import { hasBasePath } from "./utils/base-path.js";
 import { mergeRewriteQuery } from "./utils/query.js";
 import {
@@ -784,6 +785,11 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
   let hasNitroPlugin = false;
   let rscCompatibilityId: string | undefined;
   const draftModeSecret = randomUUID();
+
+  // App Router source files (rsc environment) that failed to transform on their
+  // last HMR pass. Used to re-emit `rsc:update` once they compile again — see
+  // the "vinext:app-dir-hmr-recovery" plugin below.
+  const rscBuildErrorFiles = new Set<string>();
 
   // Build-time layout classification manifest, captured in the RSC virtual
   // module's load hook and consumed in renderChunk to patch the generated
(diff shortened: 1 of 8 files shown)


james-elicx:
Thanks for the review. Addressing the two non-blocking observations in c759dde:

1. **Cross-module premature dismissal (theoretical flicker):** Leaving as-is. As you noted, it's a transient flicker that self-heals — the still-broken module's next failing transform re-fires `vite:error` and re-reports. Guarding against it would require per-module build-error tracking, which isn't worth the complexity for this edge case.

2. **`index` selection drift:** Fixed. `dismissBuildErrors()` now preserves the displayed error by `id` rather than slot position — it captures the selected error's id before filtering and points `index` at its new position, falling back to clamping only when the selected error was itself removed. Added a regression test (`[caught, vite, caught]` with `index=2` → selection stays on the same `caught` instance, not the other survivor).

`vp check` clean; the four `dev overlay store` unit tests pass.

james-elicx:
## Follow-up: fixed the actual root cause (server-side)

The first commit only addressed the *client* side (clearing the overlay when a recovery event arrives). But the test failed again on this PR's CI — the overlay stayed visible the full 10s, meaning **no recovery event reached the browser at all**.

### Real root cause

`@vitejs/plugin-rsc`'s `hotUpdate` hook only re-emits `rsc:update` (the event that clears the overlay + re-fetches the tree) when the changed file still resolves to a module in the rsc module graph (`ctx.modules` non-empty → `plugin-BK29Va7z.js:847/880`). After a **transform error**, the errored module can drop out of the graph, so the recovery edit produces an empty `ctx.modules`, the RSC plugin stays silent, and the overlay is stuck until a manual reload. And because `rsc:update` is a Vite *custom* event, it never triggers `vite:afterUpdate`, so the browser has no other signal.

This is why it only reproduced on CI (slower Linux I/O changes whether the errored module survives the rapid broken→clean write sequence) and never locally despite 90+ repeat-each runs.

### Fix

New `vinext:app-dir-hmr-recovery` plugin (runs before the RSC plugin) that verifies the transform itself:
- Records App Router (rsc env) source files that fail to transform.
- When a previously-failing file compiles again, emits `rsc:update` directly.

Recovery no longer depends on the errored module surviving in the graph. Vite invalidates the module cache (`moduleGraph.onFileChange`, `node.js:31988`) *before* `hotUpdate` runs (`:31989`), so the in-hook `transformRequest` reflects the new file content — throwing for broken, succeeding for clean. The browser's `rsc:update` handler is idempotent, so an extra event (when the RSC plugin also recovers) is harmless.

URL resolution extracted to `utils/rsc-module-url.ts` with unit tests (root-relative, `/@fs/` fallback, Windows separator normalization). The earlier client-side `dismissBuildErrors` clears + `index`-by-id preservation stay in place as defense-in-depth.

Letting CI run to confirm.

PR #1801, opened 2026-06-07
perf(og): dedupe resvg/yoga wasm in server bundle

james-elicx:
## Problem

Analyzing the `examples/app-router-playground` server bundle showed vinext's `next/og` (satori/resvg) handling ships each WASM module **twice** — once per loader strategy — even though only one strategy ever runs at runtime, wasting ~1.4 MiB per build.

1. **resvg.wasm duplicated (~1.3 MiB).** `import("./resvg.wasm?module")` makes the bundler emit a hashed asset `_next/static/resvg-HASH.wasm` (workerd loads this). Separately, `ogAssetsPlugin` copied a **second** root `resvg.wasm` from node_modules purely as the Node.js disk-read fallback for `new URL("./resvg.wasm", import.meta.url)`. Both were uploaded.
2. **yoga inlined as ~95 KiB base64.** The `vinext:og-font-patch` transform emitted `yoga.wasm` via `?module` **and** prepended a ~95 KiB base64 blob used only as a Node.js fallback.

## Fix (platform-agnostic — no Cloudflare/workerd detection)

The dedup decision keys off **"did the bundler already emit this asset?"**, not the deploy target — so it also helps Node/self-hosted builds.

- **`og-font-patch` (`index.ts`):** drop the `__vi_yoga_b64` base64 preamble and switch the Node.js yoga fallback to a disk read (`node:fs` + `new URL("./yoga.wasm", import.meta.url)` + `WebAssembly.instantiate`), mirroring the existing resvg handling. The `?module` import stays the primary (workerd) path.
- **`og-assets.ts`:** refactored `ogAssetsPlugin` → `createOgAssetsPlugin()`. In `generateBundle` (post, after minification), when the bundler already emitted the WASM as a hashed asset, rewrite the `new URL("./x.wasm", import.meta.url)` fallback reference in each chunk to point at that emitted sibling asset (handles `"`/`'`/`` ` `` quote styles since the minifier rewrites the literal). Only copy a single root file in `writeBundle` when **no** asset was emitted. Applied symmetrically to both `resvg.wasm` and `yoga.wasm`.

Net effect: exactly one physical copy of each WASM, every loader path resolving to it, no base64 inline — on all targets.

## Verification

`examples/app-router-playground/dist/server` WASM listing:

**Before**
```
_next/static/onig-Du5pRr7Y.wasm    466610
_next/static/resvg-Cjh1zH0p.wasm  1378357
_next/static/yoga-sbSbVeWy.wasm     71736
resvg.wasm                        1378357   <- duplicate root copy
```
`index.edge-*.js`: **527823** bytes (contains ~95 KiB yoga base64 blob)

**After**
```
_next/static/onig-Du5pRr7Y.wasm    466610
_next/static/resvg-Cjh1zH0p.wasm  1378357
_next/static/yoga-sbSbVeWy.wasm     71736
```
`index.edge-*.js`: **432359** bytes (−95464, the yoga base64 blob is gone)

In `index.edge`, both fallbacks now resolve to the single emitted asset:
- `new URL(\`./resvg-Cjh1zH0p.wasm\`, import.meta.url)`
- `new URL(\`./yoga-sbSbVeWy.wasm\`, import.meta.url)`

## Tests

- New `tests/og-assets.test.ts`: asserts (a) the fallback `new URL(...)` is rewritten to the emitted hashed asset, (b) no second root copy is written when the asset was emitted, (c) a single root copy is copied when nothing was emitted, (d) non-rsc envs are ignored.
- Updated `tests/og-font-patch.test.ts`: yoga now reads from disk (no base64 inline) and never references `new URL(yoga.wasm)` at top level.
- `vp check` (format/lint/type) clean across the repo.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -115,7 +115,7 @@ import {
 } from "./client/instrumentation-client-inject.js";
 import { createMiddlewareServerOnlyPlugin } from "./plugins/middleware-server-only.js";
 import { createOptimizeImportsPlugin } from "./plugins/optimize-imports.js";
-import { createOgInlineFetchAssetsPlugin, ogAssetsPlugin } from "./plugins/og-assets.js";
+import { createOgInlineFetchAssetsPlugin, createOgAssetsPlugin } from "./plugins/og-assets.js";
 import { generateRouteTypes } from "./typegen.js";
 import {
   mergeOptimizeDepsExclude,
@@ -4256,8 +4256,8 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
     // Inline binary assets fetched via `fetch(new URL("./asset", import.meta.url))` —
     // see src/plugins/og-assets.ts
     createOgInlineFetchAssetsPlugin(),
-    // Copy @vercel/og binary assets to the RSC output directory — see src/plugins/og-assets.ts
-    ogAssetsPlugin,
+    // Dedupe/copy @vercel/og binary WASM assets in the RSC output — see src/plugins/og-assets.ts
+    createOgAssetsPlugin(),
     // Collect SSR/RSC bundle externals and write dist/server/vinext-externals.json.
     // Used by emitStandaloneOutput to determine which packages to copy into
     // standalone/node_modules/ — uses the bundler's own import graph instead of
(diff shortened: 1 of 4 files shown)


james-elicx:
Addressed the non-blocking observations:

- **(3) Cross-directory rewrite test** — added a case with the chunk at the output root and the emitted asset under `_next/static/`, asserting the ref becomes `./_next/static/resvg-BBB.wasm`. Locks in the `path.posix.relative` behavior.
- **(1) Cross-hook state** — documented the `generateBundle` → `writeBundle` `dedupedBases` dependency and the rsc-env ordering invariant.

(2) and (4) left as-is per your note that they're fine/acceptable.

/bigbonk

PR #1785, opened 2026-06-06
test(interception): fixtures and E2E specs for 4 dynamic interception suites (#1364 Part B)

Divkix:
## What this does

Adds fixture pages and Playwright E2E specs for the 4 Next.js interception route test suites that were missing coverage, tracked in issue #1364 (Part B). The route graph logic in `app-route-graph.ts` is already correct — these are purely test additions.

**Prerequisite:** This branch should be merged after `fix/interception-compute-target` (Part A). Tests for suites 2–4 that depend on the `computeInterceptTarget` fix will fail until that branch lands. Suite 1 tests pass today.

---

### Suite 1 — `interception-dynamic-segment`
Ported from: `test/e2e/app-dir/interception-dynamic-segment/`

Fixture: `app/interception-dyn-seg/`

- `@modal/(.)[username]/[id]` — dynamic two-segment interception
- `@modal/(.)test-nested/@sidebar` — interception with a nested parallel slot

6 tests covering: intercept on navigate, refresh shows full page, back/forward, repeated navigation, sidebar slot.

### Suite 2 — `interception-dynamic-single-segment`
Ported from: `test/e2e/app-dir/interception-dynamic-single-segment/`

Fixture: `app/interception-dyn-single/`

- `@modal/(.)groups/[id]/new` — interception from a multi-segment source (`/groups/123` → `/groups/123/new`)

5 tests covering: modal opens over group page, refresh shows full page, back/forward, repeated and cross-id navigation.

### Suite 3 — `interception-dynamic-segment-middleware`
Ported from: `test/e2e/app-dir/interception-dynamic-segment-middleware/`

Fixture: `app/interception-mw/`

Middleware adds a locale prefix (`/interception-mw/foo/p/1` → `/interception-mw/en/foo/p/1`). The rewrite is scoped to `/interception-mw/*` in the shared `middleware.ts` so it can't touch anything else.

4 tests covering: intercept fires through rewrite, refresh falls back to default, back/forward, repeated navigation.

### Suite 4 — `parallel-routes-and-interception-from-root`
Ported from: `test/e2e/app-dir/parallel-routes-and-interception-from-root/`

Fixture: `app/interception-from-root/`

- `(...)[locale]/intercepted` — `(...)` modifier intercepting from the app root, with `[locale]` interpolated dynamically

3 tests covering: modal shows intercepted page, locale label stays correct, direct visit shows full page, back navigation.

---

## What passes now

[...]

PR #1804, opened 2026-06-07
feat(interception): sibling-style interception routes (#1364 Part C)

### LINKED ISSUE #1364: Interception routes: build and route resolution failures

Divkix:
## What this does

Adds support for sibling-style interception routes — interception markers like `(.)`, `(..)`, `(..)(..)`, and `(...)` that are **not** wrapped in an `@slot` directory. On soft navigation from a matching source route, the intercepting page replaces the full page response. On hard navigation (direct load or refresh), the real target page renders.

This is Part C of #1364. Parts A and B landed in #1751 and #1785 respectively.

---

## Background

Before this PR, vinext only handled slot-based interception:

```
app/@modal/(.)photo/[id]/page.tsx   ✅ worked
```

Sibling-style markers — no `@slot` wrapper — were silently dropped:

```
app/foo/bar/(..)(..)hoge/page.tsx   ❌ not supported
app/templates/(..)showcase/[...catchAll]/page.tsx  ❌ not supported
```

These map to two Next.js test suites that were failing: `interception-routes-multiple-catchall` and `interception-segments-two-levels-above`.

---

## How it works

Sibling interception uses the same `X-Vinext-Interception-Context` header gate as slot interception — the server only intercepts on soft navigation (when the header is present), never on hard loads. The difference is purely in where the result lands: a slot intercept injects the intercepting page into a parallel slot; a sibling intercept replaces the entire page response.

The implementation reuses the existing interception pipeline end-to-end (header gating, `findIntercept`, `previousNextUrl` history, cache key encoding, `resolveManifestNavigationInterceptionContext`). The changes are additive:

1. **Route graph** — `discoverSiblingInterceptingRoutes` walks the app directory and registers interception markers that sit outside `@slot` trees. Each result is attached to the owning route as `route.siblingIntercepts[]`. The existing `computeInterceptTarget` / `computeInterceptSourceMatchPattern` helpers handle the `(.)` / `(..)` / `(..)(..)` / `(...)` math unchanged.

2. **Manifest** — Sibling intercepts are serialized as `siblingIntercepts: [...]` on the route object in the generated runtime module, and as entries in `RouteManifestInterception` under `interceptionsBySlotId` (using a synthetic `slot:__page:<sourcePattern>` id so the client's existing manifest lookup works without changes).

3. **Runtime matcher** — `createInterceptLookup` gains a second pass over `route.siblingIntercepts`, pushing entries with the sentinel slot key `__vinext_page_intercept`. `findIntercept` and `matchInterceptSource` need no changes — they already gate purely on source/target patterns.

4. **Render** — In `buildPageElements`, a sibling intercept swaps `route.page` for the intercepting page module and uses the intercept's matched params. `buildSlotOverrides` skips the sentinel key so no phantom slot entry is created. The render identity receives `interceptSlotId: null` so the RSC payload carries `interception: null` — the client commits this as a normal full-page navigation rather than triggering slot-preservation validation.

5. **Intercept lookup fix** — `createInterceptLookup` now builds a `pattern → index` map and resolves `sourceRouteIndex` via the intercept's `sourceMatchPattern`. This fixes inherited slots: when `/groups/:id/new` inherits `@modal` from `/interception-dyn-single`, the lookup correctly identifies `/interception-dyn-single` as the source route, so `buildPageElement` renders the slot-owner's layout with the current route's page in `children` instead of the slot-owner's own root page.

---

## Also fixed in this PR

Three bugs from the PR #1785 comment that were attributed to "Part C":

**Bug: `(...)` root interception fixture used wrong marker** — The fixture at `interception-from-root/[locale]/example/@modal/(...)...` used `(...)` (app-root), but the vinext shared fixture is nested under `interception-from-root/`, so the computed target was `/:locale/intercepted` instead of `/interception-from-root/:locale/intercepted`. Changed the marker to `(..)(..)` which correctly climbs two visible segments to the fixture root.

**Bug: source page not preserved when intercepting from a descendant** — When a user navigated from `/interception-dyn-single/groups/123` to trigger a `@modal` intercept, `#children` showed the root page instead of "Group 123". The fix: `createInterceptLookup` now resolves `sourceRouteIndex` to the actual slot-owner route via `sourceMatchPattern`, not the inheriting descendant's index.

**Bug: interception didn't fire with middleware locale prefix** — The link href lacked the locale segment that middleware inserts, so the client couldn't match the target pattern. Fixed by including the locale in the link href via dynamic params, which also makes the path more correct (the user is already at `/en`, so the link should reflect that).

---

## Tests

**New unit tests:**
- 4 route graph tests for sibling marker discovery (`(..)`, `(..)(..)`, `(.)`, `(...)` outside `@slot`)
- 1 manifest test asserting sibling entries appear in `interceptionsBySlotId`
- 1 runtime matcher test: soft-nav hits, hard-nav returns null, wrong source returns null

**New E2E suites** (ported from Next.js):
- `interception-segments-two-levels-above` — 5 tests (soft-nav intercept, hard-nav fallback, back/forward, repeated)
- `interception-routes-multiple-catchall` — 4 tests (multiple catchall targets, hard-nav fallback)

**Previously failing tests now passing** (7 total, all `test.fail()` removed):
- `interception-dynamic-single-segment` — 1 test
- `interception-dynamic-segment-middleware` — 4 tests
- `parallel-routes-and-interception-from-root` — 2 tests

**Full results:**
- 27/27 interception E2E tests pass (0 `test.fail()` annotations remain in interception specs)
- 199/199 unit tests pass
- `vp check` clean (1703 files, no lint or type errors)

---

## Files changed

| Area | Files |
|---|---|
| Route graph (core) | `packages/vinext/src/routing/app-route-graph.ts` |
| Manifest generator | `packages/vinext/src/entries/app-rsc-manifest.ts` |
| Runtime matcher | `packages/vinext/src/server/app-rsc-route-matching.ts` |
| Render pipeline | `packages/vinext/src/server/app-page-element-builder.ts` |
| Intercept state | `packages/vinext/src/server/app-page-request.ts` |
| Unit tests | `tests/app-route-graph.test.ts`, `tests/app-rsc-route-matching.test.ts` |
| E2E fixtures | `tests/fixtures/app-basic/app/interception-segments-two-levels-above/`, `tests/fixtures/app-basic/app/interception-routes-multiple-catchall/` |
| E2E specs | `tests/e2e/app-router/interception-segments-two-levels-above.spec.ts`, `tests/e2e/app-router/interception-routes-multiple-catchall.spec.ts` |
| Bug fixes | `tests/fixtures/app-basic/app/interception-from-root/`, `tests/fixtures/app-basic/app/interception-mw/` |

---

## Related

- Part A: #1751 — `computeInterceptTarget` fix for nested slot subdirectories
- Part B: #1785 — fixtures and E2E specs for 4 dynamic interception suites
- Issue: #1364

Closes #1364


NathanDrake2406:
@codex

PR #1805, opened 2026-06-07
refactor(pages): extract render orchestrator from codegen string (#1784)

### LINKED ISSUE #1784: Move the Pages render orchestrator out of the codegen string
## Summary

`entries/pages-server-entry.ts` is supposed to be thin wiring. Most of it is. But two substantial pieces of runtime behavior live inside its generated template string:

- `_renderPage` at `:597-895` — roughly 260 lines. This is the de-facto Pages request handler: i18n resolution, route match, 404/500 fallback recursion, the `_next/data` JSON envelope, 405 method checks, ISR cache-key derivation, SSR-context setup.
- `collectAssetTags` at `:477-560` — roughly 85 lines of response shaping: SSR-manifest walking, shared-chunk injection, lazy-chunk filtering, building the `<link>` and `<script>` tags.

Because both sit inside a string that gets generated and written to disk, neither can be reached through an interface. You can't unit test them. `AGENTS.md` calls this out directly: request/response orchestration and response shaping belong in `server/*` helpers, not in generated entries.

App Router doesn't have this problem. `entries/app-rsc-entry.ts:554` delegates to `server/app-rsc-handler.ts:881` (an 881-line typed handler) and passes thin closures. Pages Router has no equivalent — `_renderPage` *is* the handler, it just happens to live in a string.

## Before vs. after

```mermaid
flowchart LR
  subgraph BEFORE["before"]
    PE["pages-server-entry.ts<br/>template string"]
    RP["_renderPage ~260 lines<br/>(:597-895)"]
    CA["collectAssetTags ~85 lines<br/>(:477-560)"]
    PE --- RP
    PE --- CA
  end
  subgraph AFTER["after — mirrors app-rsc-entry"]
    PE2["pages-server-entry.ts<br/>thin closures"]
    H["server/pages-page-handler.ts<br/>typed, unit-testable"]
    PE2 -->|delegates| H
  end
  BEFORE -.extract.-> AFTER
```

The interface is the test surface. Right now there is no interface, so there's no surface.

## Proposed change

1. Extract `_renderPage` into `server/pages-page-handler.ts` as a normal typed function, mirroring how `app-rsc-handler.ts` is structured.
2. Extract `collectAssetTags` into a small helper (its own module or alongside the handler).
3. Reduce the entry to what App Router's entry already does: imports, the manifest, and thin closures that hand off to the helper.

## Why this is the safe one

Pure code motion. The behavior doesn't change — the same logic runs, it just lives in a `.ts` file you can import instead of a string you can't. The entry keeps exporting `renderPage` with the same signature, so every caller is unaffected.

`tests/entry-templates.test.ts` is explicit about this in its header: it tests "observable behavior (structured API outputs and error paths), not on the textual shape of the generated code." That's exactly the property that makes this extraction free — moving code out of the string can't break a test that never asserted on the string.

This one also unlocks the pipeline-unification work: once Pages has a real handler module, the unified pipeline can call into it instead of re-deriving the render step in three places.

## Tests that guard it

- `tests/pages-router.test.ts` — drives `serverEntry.renderPage(webRequest, url, manifest)` (`:3539`) and `runMiddleware` through the exported interface
- `tests/entry-templates.test.ts` — behavior-focused by design
- Deploy suite: the Pages Router file set broadly

## Scope

In scope: extracting `_renderPage` and `collectAssetTags` into `server/*` and slimming the entry.

Out of scope: changing render behavior, touching App Router, the cache-decision work.

## Acceptance criteria

- [ ] `_renderPage` lives in `server/pages-page-handler.ts` with direct unit tests
- [ ] `collectAssetTags` lives in a typed helper with tests
- [ ] `pages-server-entry.ts` is thin wiring that delegates, like `app-rsc-entry.ts`
- [ ] `renderPage` keeps its current signature; existing tests pass unchanged
- [ ] Deploy-suite pass rate unchanged

## Risk

Lowest of the three. Behavior is identical by construction, and the one test that touches generated entries deliberately ignores string shape. Good first move.

---

*Filed from an architecture review. Companion issues: unify the Pages Router request pipeline and add an ISR cache-decision module.*

Divkix:
Closes #1784

## Summary

Extracts `_renderPage` (~260 lines) and `collectAssetTags` (~85 lines) from the generated template string in `entries/pages-server-entry.ts` into typed, unit-testable server modules — mirroring how `server/app-rsc-handler.ts` relates to `entries/app-rsc-entry.ts`.

## Changes

**New: `server/pages-page-handler.ts`**
`createPagesPageHandler(opts)` factory that owns the full Pages Router render lifecycle: i18n resolution, route matching, 404/500 fallback recursion, `_next/data` JSON envelope, 405 method check, ISR cache-key derivation, SSR-context setup. All `next/*`-derived values are passed as closures so the module stays importable in the test environment.

**New: `server/pages-asset-tags.ts`**
`collectAssetTags` + `resolveSsrManifest` / `getManifestFilesForModule` / `resolveClientModuleUrl` helpers, extracted from the template string so they can be unit-tested directly.

**Modified: `server/pages-data-route.ts`**
Adds `normalizePagesDataRequest(request, buildId)` so both `renderPage` and `runMiddleware` share one implementation instead of duplicating the buildId-check logic inside the template string.

**Modified: `entries/pages-server-entry.ts`**
Now thin wiring: 449 lines (down from 1064). Codegen emits route-table imports, embedded config/i18n/buildId, and closures that delegate to `createPagesPageHandler`. `renderPage` keeps its exact exported signature; all callers are unaffected.

## Bug fixed (found during extraction)

The ISR cache-key wrapper in the old template always passed `buildId` as the third arg. The extracted handler's `isrCacheKey.bind(null)` path was missing that arg on non-i18n requests, causing a cache-key mismatch (different keys in dev vs. the 3-arg form). Caught by the existing CSP-nonce ISR test.

## Tests

- **New `tests/pages-asset-tags.test.ts`** — 27 tests: manifest helpers, lazy-chunk filtering, defer toggle, shared-chunk injection, client-entry bootstrap, CSS/JS tag emission.
- **New `tests/pages-page-handler.test.ts`** — 14 tests: route miss → 404 fallback, `_next/data` envelope + wrong-buildId 404, 405 method check, i18n passthrough, internal-error guard, `renderErrorPageOnMiss: false`, `setSSRContext` called.
- **Existing `tests/pages-router.test.ts`** — 276/276 ✓ (no regressions)
- **Existing `tests/entry-templates.test.ts`** — 25/25 ✓

## Acceptance criteria from #1784

- [x] `_renderPage` lives in `server/pages-page-handler.ts` with direct unit tests
- [x] `collectAssetTags` lives in a typed helper with tests
- [x] `pages-server-entry.ts` is thin wiring that delegates, like `app-rsc-entry.ts`
- [x] `renderPage` keeps its current signature; existing tests pass unchanged
- [x] Deploy-suite pass rate unchanged

### DIFF
--- a/packages/vinext/src/entries/pages-server-entry.ts
+++ b/packages/vinext/src/entries/pages-server-entry.ts
@@ -19,20 +19,12 @@ const _requestContextShimPath = resolveEntryPath("../shims/request-context.js", 
 const _middlewareRuntimePath = resolveEntryPath("../server/middleware-runtime.js", import.meta.url);
 const _routeTriePath = resolveEntryPath("../routing/route-trie.js", import.meta.url);
 const _pagesI18nPath = resolveEntryPath("../server/pages-i18n.js", import.meta.url);
-const _pagesPageResponsePath = resolveEntryPath(
-  "../server/pages-page-response.js",
-  import.meta.url,
-);
-const _pagesPageDataPath = resolveEntryPath("../server/pages-page-data.js", import.meta.url);
-const _pagesPageMethodPath = resolveEntryPath("../server/pages-page-method.js", import.meta.url);
 const _pagesDataRoutePath = resolveEntryPath("../server/pages-data-route.js", import.meta.url);
 const _pagesDefault404Path = resolveEntryPath("../server/pages-default-404.js", import.meta.url);
-const _pagesNodeCompatPath = resolveEntryPath("../server/pages-node-compat.js", import.meta.url);
 const _pagesApiRoutePath = resolveEntryPath("../server/pages-api-route.js", import.meta.url);
-const _isrCachePath = resolveEntryPath("../server/isr-cache.js", import.meta.url);
-const _cspPath = resolveEntryPath("../server/csp.js", import.meta.url);
 const _serverGlobalsPath = resolveEntryPath("../server/server-globals.js", import.meta.url);
 const _queryUtilsPath = resolveEntryPath("../utils/query.js", import.meta.url);
+const _pagesPageHandlerPath = resolveEntryPath("../server/pages-page-handler.js", import.meta.url);
 
 /**
  * Generate the virtual SSR server entry module.
@@ -185,7 +177,7 @@ export async function runMiddleware(request, ctx, options) {
   // the normalized page path via the request URL, and the worker sees the
   // normalized URL via result.rewriteUrl (if middleware didn't already
   // rewrite to something else).
-  const __dataNorm = __normalizePagesDataRequest(request);
+  const __dataNorm = __normalizePagesDataRequest(request, buildId);
   if (__dataNorm.notFoundResponse) {
     return { continue: false, response: __dataNorm.notFoundResponse };
   }
@@ -209,7 +201,7 @@ export async function runMiddleware(request, ctx, options) {
 export async function runMiddleware(request) {
   // Even without user middleware, the data-endpoint URL must be normalized so
   // the worker pipeline sees the page path. Mismatched buildId → JSON 404.
-  const __dataNorm = __normalizePagesDataRequest(request);
+  const __dataNorm = __normalizePagesDataRequest(request, buildId);
   if (__dataNorm.notFoundResponse) {
     return { continue: false, response: __dataNorm.notFoundResponse };
   }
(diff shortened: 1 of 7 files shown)


james-elicx:
Opus says this is a clean extraction with a few behavioural improvements - thanks!

PR #1810, opened 2026-06-07
fix(build): write BUILD_ID via writeBundle so App Router builds emit it

james-elicx:
## What regressed

The nightly **Next.js Deploy Suite** regressed by ~700 tests between [run 27057467752](https://github.com/cloudflare/vinext/actions/runs/27057467752) (2143 passing) and [run 27081060115](https://github.com/cloudflare/vinext/actions/runs/27081060115) (1444 passing). Nearly every App Router fixture failed at **setup** with:

```
Failed to extract build ID from dist/server/index.js
Custom deploy script failed:  undefined (1)
```

## Root cause

Two interacting facts:

1. The `vinext:build-id` plugin wrote `dist/server/BUILD_ID` from a **`closeBundle`** hook. `closeBundle` does **not** fire during the multi-environment `createBuilder().buildApp()` pipeline used for App Router production builds, so the file was *silently never written for pure App Router apps*. Only hybrid apps (which run a second `vite.build()` pass with a fresh plugin instance) and Pages Router apps produced the file.

2. The e2e deploy harness (`scripts/e2e-deploy.sh` → `read_build_id`) falls back to regex-parsing the buildId out of `dist/server/index.js` when `BUILD_ID` is absent. That fallback worked while server bundles were unminified — until **#1777 (`perf(build): minify server build environments by default`)** landed between the two runs. Minification mangles the `get buildId() { return "..." }` / `buildId = "..."` patterns the regex relies on, so the fallback started returning no match and every App Router deploy failed at setup.

Confirmed by reproducing locally: a pure App Router `buildApp()` did not emit `dist/server/BUILD_ID`, and the regex returns `NO MATCH` against the minified `index.js`.

## Fix 1 — emit BUILD_ID reliably

Switch the `vinext:build-id` plugin from `closeBundle` to **`writeBundle`** (mirroring the already-working `vinext:image-config` plugin). `writeBundle` fires for every emitted bundle across all environments; the existing one-time write guard ensures the file is written exactly once. Verified `dist/server/BUILD_ID` is now emitted for pure App Router, hybrid, and Pages Router builds.

## Fix 2 — one build ID across all plugin instances

While reviewing Fix 1 we found a pre-existing divergence: a single `vinext build` can instantiate `vinext()` more than once (App Router `buildApp()` + the separate Pages Router `vite.build()` for hybrid apps). With no user `generateBuildId`, each instance resolved its **own random UUID**, so the App Router runtime, Pages runtime, prerender manifest, and `BUILD_ID` file could each get a different ID — risking ISR/seed-cache key mismatches for hybrid apps.

The CLI now resolves the build ID **once** (honoring the user's `generateBuildId`) and publishes it via `__VINEXT_SHARED_BUILD_ID`; the plugin adopts it unless the user supplied their own `generateBuildId`. `resolveBuildId()`'s standalone semantics are unchanged, so this build-only coordination never leaks into dev or tests.

Verified on the hybrid `app-router-cloudflare` example — the App Router RSC/SSR runtime, Pages `entry.js`, `vinext-prerender.json`, and `dist/server/BUILD_ID` now all carry the same build ID.

## Tests

`tests/app-router-production-build.test.ts` gains two assertions on the `buildApp()` path:
- `dist/server/BUILD_ID` is emitted and non-empty (the regression guard).
- With `__VINEXT_SHARED_BUILD_ID` set, the emitted file **and** the value baked into the runtime bundle both equal the shared ID.

```
✓ tests/app-router-production-build.test.ts (4 tests)
✓ tests/next-config.test.ts / prerender.test.ts (252 tests)
✓ tests/deploy / build-optimization / features / pages-router / static-export / app-router-static-export (927 tests)
```

### DIFF
--- a/packages/vinext/src/cli.ts
+++ b/packages/vinext/src/cli.ts
@@ -448,10 +448,15 @@ async function buildApp() {
   // Coordinate a single build ID across every vinext() plugin instance in this
   // build. A hybrid app+pages build runs the App Router multi-environment build
   // (buildApp) and a separate Pages Router SSR build (vite.build) as distinct
-  // plugin instances; without this, each resolves its own random UUID and the
-  // runtime, prerender manifest, and dist/server/BUILD_ID disagree. We resolve
-  // it once here (honoring the user's generateBuildId) and share it via env; the
-  // plugin adopts it unless the user supplied their own generateBuildId.
+  // plugin instances; without this, each resolves its own (potentially random)
+  // ID and the runtime, prerender manifest, and dist/server/BUILD_ID disagree.
+  // We resolve it once here — resolveNextConfig() already ran resolveBuildId()
+  // honoring the user's generateBuildId (including the null→UUID fallback) — and
+  // share that authoritative value via env so every plugin instance adopts it.
+  //
+  // Not cleaned up intentionally: `vinext build` runs once and the process
+  // exits, so there is no in-process reuse to leak into. The var is namespaced
+  // to vinext's build flow and is never read by dev or standalone resolveBuildId.
   process.env.__VINEXT_SHARED_BUILD_ID = resolvedNextConfig.buildId;
 
   const outputMode = resolvedNextConfig.output;
--- a/packages/vinext/src/cli.ts
+++ b/packages/vinext/src/cli.ts
@@ -444,6 +444,16 @@ async function buildApp() {
     await loadNextConfig(root, PHASE_PRODUCTION_BUILD),
     root,
   );
+
+  // Coordinate a single build ID across every vinext() plugin instance in this
+  // build. A hybrid app+pages build runs the App Router multi-environment build
+  // (buildApp) and a separate Pages Router SSR build (vite.build) as distinct
+  // plugin instances; without this, each resolves its own random UUID and the
+  // runtime, prerender manifest, and dist/server/BUILD_ID disagree. We resolve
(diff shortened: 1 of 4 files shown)


james-elicx:
Thanks for the review — the `generateBuildId` correctness issue is a real one, fixed in 54da50b.

**Fix:** dropped the `!rawConfig?.generateBuildId` guard. The plugin now **always** adopts `__VINEXT_SHARED_BUILD_ID` when set, exactly as suggested. The CLI's `resolvedNextConfig.buildId` is already the authoritative value (it ran `resolveBuildId(config.generateBuildId)` once, including the documented null→UUID fallback), so a per-instance re-resolution should never win. This is correct for all cases: no `generateBuildId`, `generateBuildId` returning a string, and `generateBuildId` returning `null` / being non-deterministic.

**Test gap:** added `adopts the shared build ID even when generateBuildId is set` — it builds with `generateBuildId: () => null` and a shared ID set, asserting both `dist/server/BUILD_ID` and the baked runtime value equal the shared ID. This fails under the old guard and guards against re-introduction.

**Env var lifetime:** documented at `cli.ts` that the var is intentionally not cleared — `vinext build` runs once and the process exits, so there's no in-process reuse to leak into, and it's namespaced to the build flow (never read by dev or standalone `resolveBuildId`).

`resolveBuildId()`'s standalone semantics are unchanged; `tests/next-config.test.ts` (170) and `tests/app-router-production-build.test.ts` (5/5) pass.

PR #1833, opened 2026-06-08
fix(build): inline new URL(asset, import.meta.url) blob imports for edge (#1824)

### LINKED ISSUE #1824: Edge runtime: `new URL(asset, import.meta.url)` blob/asset imports fail to fetch
> *This issue was created by an agent auditing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/27113612207) (vinext `main` vs Next.js `v16.2.6`, 2026-06-08).*

## Problem

The edge compiler cannot import static/blob assets via `new URL('./asset', import.meta.url)` and fetch them at runtime. The entire `edge-compiler-can-import-blob-assets` suite fails (0 passing), so asset-as-URL resolution in the edge build is effectively unimplemented.

## Affected Test Suites

- `test/e2e/edge-compiler-can-import-blob-assets/index.test.ts` (5 failures)
  - allows to fetch a remote URL
  - allows to fetch a remote URL with a path and basename
  - allows to fetch text assets
  - allows to fetch image assets
  - allows to fetch assets from node_modules

## Estimated impact

~5 e2e failures.

## Recommendation

1. Reproduce: an edge route that does `const url = new URL('./data.txt', import.meta.url); await fetch(url)` and assert the asset bytes come back.
2. Ensure the edge build emits referenced assets and rewrites `new URL(..., import.meta.url)` to a resolvable runtime URL (including assets pulled from `node_modules`).

james-elicx:
## Summary

Edge/worker routes that reference static assets via `new URL("./asset", import.meta.url)` and `fetch(url)` failed at runtime. Vite's built-in `vite:asset-import-meta-url` plugin only runs in the **client** environment, so the URL was left untransformed in the worker bundle. Worse, on Cloudflare Workers `import.meta.url` is the literal string `"worker"` (not a URL), so `new URL("./x", import.meta.url)` throws `TypeError: Invalid URL`. The whole upstream `edge-compiler-can-import-blob-assets` suite (5 tests) was red.

Adds a `vinext:edge-asset-import-meta-url` Vite plugin that, in non-client (server/worker) environments, rewrites the entire `new URL("<spec>", import.meta.url)` expression to an inline `data:` URL literal computed from the referenced file at build time.

A `data:` URL:
- is a valid absolute URL, so `new URL(...)` never throws and never depends on the runtime value of `import.meta.url`;
- can be bound to a variable and `fetch()`ed later — matching the fixture's `const url = new URL(...); return fetch(url)` shape, which the `fetch(new URL(...))`-only OG inliner does not cover;
- is fetchable in both workerd and Node, so no asset file needs to be emitted to / served from the worker output.

Bare specifiers (node_modules assets like `my-pkg/hello/world.json`) resolve through the bundler's `this.resolve`. Absolute/remote URLs (`new URL("https://...")`, `new URL("/", "https://...")`) and `/* @vite-ignore */`-annotated specifiers are left untouched.

This mirrors the existing `vinext:og-inline-fetch-assets` plugin, which already base64-inlines `fetch(new URL(...))` font/wasm assets for the same "`import.meta.url` is not a URL in workerd" reason. The new plugin is registered **immediately after** the OG inliner (both are `enforce: "pre"`, so array order sequences them) so the OG inliner's verbatim-pattern match is preserved.

### Relation to #1346 / #1640

PR #1640 (still open, for #1346) targets the **Node SSR** path: it emits the asset to disk and rewrites the URL to a chunk-relative `file://` path. That strategy does not work on Cloudflare Workers (no filesystem; `import.meta.url` is `"worker"`), which is why the **edge** path inlines instead. The two are complementary; #1640 is not merged, so there is no overlap on `main`.

## Test plan

- [x] New focused test `tests/edge-asset-import-meta-url.test.ts` drives the plugin's `transform` hook with the exact patterns from the upstream `edge.js` fixture (relative text asset, binary image with correct mime, bare-specifier node_modules asset) and asserts each is rewritten to the expected `data:` URL; also asserts absolute/remote URLs and missing files are left untouched, and that the plugin does not run in the client environment.
- [x] Verified end-to-end against a real Pages Router edge-API build (throwaway harness, removed before commit): the worker entry emits `new URL(\`data:text/plain;base64,...\`)` with no residual `import.meta.url`.
- [x] `tests/og-assets.test.ts` and `tests/import-meta-url.test.ts` still green (no regression from plugin ordering).
- [x] `vp check` clean on all changed files.

Closes #1824

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -116,6 +116,7 @@ import {
 } from "./client/instrumentation-client-inject.js";
 import { createMiddlewareServerOnlyPlugin } from "./plugins/middleware-server-only.js";
 import { createOptimizeImportsPlugin } from "./plugins/optimize-imports.js";
+import { createEdgeAssetImportMetaUrlPlugin } from "./plugins/edge-asset-import-meta-url.js";
 import { createOgInlineFetchAssetsPlugin, createOgAssetsPlugin } from "./plugins/og-assets.js";
 import { generateRouteTypes } from "./typegen.js";
 import {
@@ -4301,6 +4302,17 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
     // Inline binary assets fetched via `fetch(new URL("./asset", import.meta.url))` —
     // see src/plugins/og-assets.ts
     createOgInlineFetchAssetsPlugin(),
+    // Inline assets referenced via `new URL("./asset", import.meta.url)` in
+    // server/worker environments as `data:` URLs so edge routes can fetch
+    // them — see src/plugins/edge-asset-import-meta-url.ts and #1824.
+    //
+    // MUST run AFTER `vinext:og-inline-fetch-assets`: that plugin matches the
+    // verbatim `fetch(new URL(...)).then(r => r.arrayBuffer())` pattern. If we
+    // rewrote the inner `new URL(...)` to a data URL first, the OG inliner
+    // would no longer match and @vercel/og font inlining would silently
+    // change shape. Both plugins are `enforce: "pre"`, so array order here is
+    // what sequences them.
+    createEdgeAssetImportMetaUrlPlugin(),
     // Dedupe/copy @vercel/og binary WASM assets in the RSC output — see src/plugins/og-assets.ts
     createOgAssetsPlugin(),
     // Collect SSR/RSC bundle externals and write dist/server/vinext-externals.json.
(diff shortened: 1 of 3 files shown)


james-elicx:
Thanks for the review — addressed all four findings:

1. **Node SSR scope** — the plugin now takes an `isWorkerTarget` getter (Cloudflare/Nitro) and `applyToEnvironment` returns false in a plain Node SSR build, so the `fileURLToPath`/`.pathname` semantics there are untouched.
2. **End-to-end test** — added a real Pages Router edge-API build test through the full Vite pipeline (filter + applyToEnvironment + plugin ordering), with a negative case asserting no rewrite in a plain Node SSR build.
3. **Query/hash specifiers** — relative specifiers are now `?query`/`#hash`-stripped before resolving, matching the bare-specifier branch.
4. **`import.meta?.url`** — now matched (parity with `vinext:import-meta-url`).

Also fixed the trailing-accessor doc nit.

/bigbonk review for issues

PR #1834, opened 2026-06-08
fix(actions): enforce serverActions.bodySizeLimit on fetch actions (#1828)

### LINKED ISSUE #1828: Server actions: `serverActions.bodySizeLimit` not enforced
> *This issue was created by an agent auditing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/27113612207) (vinext `main` vs Next.js `v16.2.6`, 2026-06-08).*

## Problem

`serverActions.bodySizeLimit` is not enforced: server-action requests that exceed the configured limit are not rejected. Fails for both plaintext and multipart fetch actions, with and without node-runtime middleware.

## Affected Test Suites

- `test/e2e/app-dir/actions/app-action-size-limit-invalid.test.ts` (2 failures — plaintext, multipart; "should error for requests that exceed the size limit")
- `test/e2e/app-dir/actions/app-action-size-limit-invalid-node-middleware.test.ts` (2 failures — same, with node-runtime middleware)

## Estimated impact

~4 e2e failures across 2 suites.

## Notes

Not covered by the existing server-action issues #1340 / #1480.

## Recommendation

1. Reproduce: set `serverActions.bodySizeLimit` and POST an oversized action body; assert the request errors instead of succeeding.
2. Enforce the configured limit in the action-request handling path (both plaintext and multipart), including when a node-runtime middleware is present.

james-elicx:
## Problem

`serverActions.bodySizeLimit` was not enforced for **fetch (client-invoked) server actions**: oversized requests were rejected with a bare `413 Payload Too Large` plain-text response. That bypasses the React client error boundary and surfaces the wrong status/content-type, so the four upstream suites failed:

- `app-action-size-limit-invalid.test.ts` — "should error for requests that exceed the size limit" (plaintext + multipart)
- `app-action-size-limit-invalid-node-middleware.test.ts` — same two, with node-runtime middleware

## Fix

Next.js throws the body-exceeded error *before* the action runs and, for fetch actions, emits a **500 Flight response** carrying the rejected action result so the nearest client error boundary catches it (renders the route's `error.js`). vinext now mirrors that:

- In `handleServerActionRscRequest`, both oversized-body checks (content-length pre-check and stream-read limit) now call a new `renderFetchActionBodyExceededResponse`, which renders a Flight stream with `returnValue: { ok: false }` and **no page root** (the action never ran, nothing revalidated), returning **status 500** with the RSC content-type.
- The browser entry already throws `normalizeServerActionThrownValue(...)` when a fetch action result has `returnValue.ok === false` and no root, which propagates to the error boundary — no client change needed.
- The error message matches upstream (`Body exceeded {limit} limit.` + docs link) so it reads identically in logs.

The **no-JS progressive form-POST path keeps its existing 413** — there is no client error boundary there to catch a Flight error, so a 413 is the right behavior.

## Tests

Added two unit cases to `tests/app-server-action-execution.test.ts` covering the content-length and stream-limit triggers: assert status 500, RSC content-type, that the action is never loaded, and that the body-exceeded error is embedded in `returnValue`. Existing 413 assertions on the progressive path are unchanged.

`vp check` clean on both changed files; targeted suite (60 tests) green.

Closes #1828

### DIFF
--- a/packages/vinext/src/config/next-config.ts
+++ b/packages/vinext/src/config/next-config.ts
@@ -351,7 +351,7 @@ export type ResolvedNextConfig = {
   inlineCss: boolean;
   /** Parsed body size limit for server actions in bytes (from experimental.serverActions.bodySizeLimit). Defaults to 1MB. */
   serverActionsBodySizeLimit: number;
-  /** Verbatim body size limit config value (e.g. "2mb") for the "Body exceeded {limit} limit" error. Defaults to "1mb". */
+  /** Verbatim body size limit config value (e.g. "2mb") for the "Body exceeded {limit} limit" error. Defaults to "1 MB". */
   serverActionsBodySizeLimitLabel: string;
   /** Route-level expire fallback in seconds for ISR entries with numeric revalidate. */
   expireTime: number;
@@ -1186,7 +1186,7 @@ export async function resolveNextConfig(
       optimizePackageImports: [],
       inlineCss: false,
       serverActionsBodySizeLimit: 1 * 1024 * 1024,
-      serverActionsBodySizeLimitLabel: "1mb",
+      serverActionsBodySizeLimitLabel: "1 MB",
       expireTime: DEFAULT_EXPIRE_TIME,
       htmlLimitedBots: undefined,
       serverExternalPackages: [],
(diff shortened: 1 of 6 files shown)


james-elicx:
Addressed the blocking finding in 54b2ee29.

The body-exceeded error now uses the **verbatim `serverActions.bodySizeLimit` config string** instead of reconstructing it from the parsed byte count. I plumbed a new `serverActionsBodySizeLimitLabel` (the raw config value, e.g. `"2mb"`, defaulting to `"1mb"` when unset) through `next-config.ts` -> `index.ts` -> `app-rsc-entry.ts` -> `HandleServerActionRscRequestOptions`, and `createBodyExceededError` now emits `Body exceeded 2mb limit.` byte-for-byte, matching the upstream e2e log assertion. The lossy `formatBodySizeLimit` reconstructor is removed.

Tests updated: the execution unit test asserts the verbatim `Body exceeded 2mb limit` string, and `next-config.test.ts` locks in the label (string/numeric/default) preservation.

/bigbonk review for issues

james-elicx:
Addressed the non-blocking default-label nit in 1ad33872.

When `serverActions.bodySizeLimit` is unset, the label now defaults to `"1 MB"` (matching Next.js' `defaultBodySizeLimit = '1 MB'` literal) instead of `"1mb"`, so the unconfigured-case message reads `Body exceeded 1 MB limit.` byte-for-byte. Updated the `next-config` default-case assertion accordingly.

/bigbonk review for issues

PR #1840, opened 2026-06-08
fix(build): don't crash build when a PostCSS string plugin can't be resolved (#1509)

### LINKED ISSUE #1509: Build failure: CSS modules + PostCSS in the RSC environment

## Problem

When an app combines CSS modules with PostCSS configuration in the RSC environment, the build fails with `Custom deploy script failed: undefined (1)`. Likely a CSS pipeline interaction between `@vitejs/plugin-rsc` and PostCSS that does not handle CSS module class hashing through PostCSS-processed files.

```
CSS modules + PostCSS in RSC environment fails to build
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/app-dir/css-modules-rsc-postcss/css-modules-rsc-postcss.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a fixture combining a `*.module.css` import with a custom `postcss.config.js` in the RSC environment.

2. **Run PostCSS before CSS-module class hashing.** Make sure `@vitejs/plugin-rsc` runs PostCSS first, then the CSS-module class hashing pass.

---

Part of #1328.

james-elicx:
## Problem

Closes #1509.

When an app combines CSS modules with a PostCSS config in the RSC environment, the build could abort with an opaque `Custom deploy script failed: undefined (1)`.

`resolvePostcssStringPlugins` (`packages/vinext/src/plugins/postcss.ts`) runs inside the Vite `config` hook, which fires **once per Vite environment** (RSC / SSR / client). It resolves array-form PostCSS string plugin names (e.g. `plugins: ["postcss-nested", ...]`) to plugin instances via `require.resolve` + dynamic import. If any named plugin failed to resolve from the project root, the thrown error was **uncaught** — the rejected promise was cached and re-thrown for every environment, aborting the entire build.

## Reproduction

Building an App Router fixture (CSS module + `postcss.config` listing a plugin that doesn't resolve from where vinext looks) crashes during the RSC build:

```
Error: Cannot find module 'postcss-nested'
    at resolvePostcssStringPluginsUncached (packages/vinext/dist/plugins/postcss.js)
    at async ...config (packages/vinext/dist/index.js)
```

With all plugins installed and resolvable, CSS modules + PostCSS already build correctly across all environments — this fix targets the failure path.

## Fix

Wrap the plugin-resolution `Promise.all` in a `try/catch`. On failure, return `undefined` so vinext does **not** inject `css.postcss`, letting Vite's own `postcss-load-config` load and resolve the config relative to the config file instead. This mirrors the existing "let Vite handle it" fallback already used when the config file itself can't be loaded, and is strictly safer than throwing — the happy path (resolvable plugins) is unchanged.

## Tests

Added two regression tests to `tests/postcss-resolve.test.ts`:
- a string plugin that can't be resolved returns `undefined` (does not throw)
- an unresolvable plugin in tuple form returns `undefined`

Both fail on `main` (with the exact `Cannot find module` crash) and pass with the fix. `vp check` clean on both changed files; `tests/sass-options.test.ts` (same css-injection path) still passes.

### DIFF
--- a/packages/vinext/src/plugins/postcss.ts
+++ b/packages/vinext/src/plugins/postcss.ts
@@ -142,6 +142,12 @@ async function resolvePostcssStringPluginsUncached(
     // hook, so a throw here aborts the RSC/SSR/client builds with an opaque
     // error). Fall back to `undefined` so Vite's own postcss-load-config
     // loads and resolves the config relative to the config file instead.
+    //
+    // Note: a single unresolvable entry makes us defer the *entire* config to
+    // Vite (not a per-plugin skip) — the resolved list is only useful if it's
+    // complete, and Vite re-loads the whole config anyway. If the failure was a
+    // genuine plugin-factory error rather than a resolution miss, Vite's loader
+    // re-surfaces it with a clearer message than this opaque build abort.
     return undefined;
   }
 }
--- a/packages/vinext/src/plugins/postcss.ts
+++ b/packages/vinext/src/plugins/postcss.ts
@@ -112,27 +112,36 @@ async function resolvePostcssStringPluginsUncached(
 
   // Resolve string plugin names to actual plugin functions
   const req = createRequire(path.join(projectRoot, "package.json"));
-  const resolved = await Promise.all(
-    config.plugins.filter(Boolean).map(async (plugin: unknown) => {
-      if (typeof plugin === "string") {
-        const resolved = req.resolve(plugin);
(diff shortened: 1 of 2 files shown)


james-elicx:
Addressed the non-blocking notes: added a comment clarifying the whole-config (not per-plugin) fallback and that Vite's loader re-surfaces genuine plugin-factory errors with a clearer message. The broad catch is intentional per note 2.

/bigbonk review for issues

PR #1845, opened 2026-06-08
fix(routing): propagate middleware draft cookie to pages/api fallback (#1520)

### LINKED ISSUE #1520: App Router + Pages Router API endpoints: `pages/api/*` unreachable from app-dir suite

## Problem

When an App Router app coexists with Pages Router `pages/api/*` endpoints (e.g., `/api/dump-headers-serverless`, `/api/dump-headers-edge`), vinext serves an HTML `This page could not be found` page instead of executing the API route. This breaks middleware-driven request-header inspection tests for both serverless and edge runtimes and also breaks draft-mode flows that rely on the Pages API endpoint.

```
FetchError: invalid json response body at .../api/dump-headers-serverless reason: Unexpected token 'T', "This page "...
```

```
FetchError: invalid json response body at .../api/dump-headers-edge
```

## Estimated Impact

~8 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/app-dir/app-middleware/app-middleware.test.ts` (8 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a fixture with both `app/` and `pages/api/*` and assert the API routes return JSON.

2. **Make the router dispatch fall through to `pages/api/*`.** When the App Router does not match a request, the dispatcher must try Pages Router API routes before serving the not-found page.

3. **Cover the draft-mode test paths.** The same fix should restore the draft-mode flow that POSTs to the Pages API endpoint.

---

Part of #1328.

james-elicx:
## Problem

In a hybrid app (an `app/` directory alongside Pages Router `pages/api/*` routes), requests with no matching App Router route fall through to the Pages Router via `renderPagesFallback` (the app→pages bridge).

The bridge already forwards middleware-mutated **request** headers and middleware **response** headers to the Pages handler. But it dropped the `__prerender_bypass` `Set-Cookie` that `draftMode().enable()` / `disable()` emit when called **inside middleware**.

That cookie isn't part of the `NextResponse` headers — it lives in a request-scoped side channel that App Router route handlers and page renders read via `getDraftModeCookieHeader()` (see `finalizeRouteHandlerResponse` / the page response builder). The Pages fallback path never read it, so draft-mode-via-middleware was lost when the request fell through to `pages/api/*`.

This is the draft-mode flow called out in #1520 (the basic `pages/api/*` fall-through and middleware request-header propagation already work on `main`).

## Fix

Thread `getDraftModeCookieHeader` into the bridge dependencies and append the draft cookie to the Pages Router fallback response (both the `handleApiRoute` and `renderPage` branches), mirroring how App Router responses surface it. When middleware didn't touch draft mode the getter returns `null` and the response is returned unchanged.

The fix lives in the single shared `renderPagesFallback` bridge, so dev and prod (Cloudflare Workers / Node) paths stay in sync.

## Test

Added `tests/nextjs-compat/hybrid-pages-api.test.ts` — builds the `cf-app-basic` hybrid fixture as a Cloudflare Workers bundle and serves it through the Node prod server (the same dispatch path the deploy suite uses). Covers:

- `pages/api/*` returns JSON in a hybrid app
- App Router route handler still served in the same app
- middleware-mutated request headers reach `pages/api/*` (serverless + edge runtime)
- **draft mode enabled in middleware sets `__prerender_bypass` on a `pages/api/*` request** (the case that failed before this change)

Extended the `cf-app-basic` fixture with a header-mutating + draft-enabling `middleware.ts` and `pages/api/dump-headers{,-edge}.ts`, mirroring the upstream `app-dir/app-middleware` fixture.

Closes #1520

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -936,6 +936,7 @@ export default __createAppRscHandler({
         buildRequestHeaders: __buildRequestHeadersFromMiddlewareResponse,
         decodePathParams: __decodePathParams,
         applyRouteHandlerMiddlewareContext: __applyRouteHandlerMiddlewareContext,
+        getDraftModeCookieHeader,
       }
     );
   },`
--- a/packages/vinext/src/server/app-pages-bridge.ts
+++ b/packages/vinext/src/server/app-pages-bridge.ts
@@ -28,6 +28,14 @@ type RenderPagesFallbackDependencies = {
    * it clears it. Mirrors how App Router route handlers and page renders surface
    * the middleware-enabled draft cookie so the same flow works when the request
    * falls through to a Pages Router route.
+   *
+   * Note: this closes the draft-mode flow for production (Cloudflare Workers /
+   * Node), where middleware runs inline in the same RSC handler context that
+   * builds this fallback. In hybrid *dev*, middleware runs in a separate Vite
+   * Pages SSR runner and `draftMode()` inside middleware is not yet permitted
+   * there (it throws a scope error before any cookie is set), so this getter
+   * returns `null` and no cookie is appended. That dev limitation is pre-existing
+   * and tracked separately from #1520.
    */
   getDraftModeCookieHeader: () => string | null | undefined;
 };
--- a/packages/vinext/src/server/app-pages-bridge.ts
+++ b/packages/vinext/src/server/app-pages-bridge.ts
@@ -22,6 +22,14 @@ type RenderPagesFallbackDependencies = {
     response: Response,
     middlewareContext: AppMiddlewareContext,
   ) => Response;
+  /**
+   * Returns the `__prerender_bypass` Set-Cookie header emitted by a
+   * `draftMode().enable()`/`disable()` call inside middleware, if any. Reading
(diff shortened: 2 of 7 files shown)


james-elicx:
Thanks for the review. Addressed both findings:

**🔴 Broken existing test** — Fixed in ae5978a4: added `getDraftModeCookieHeader` to `defaultDeps` in `tests/app-pages-bridge.test.ts` and added 3 focused unit tests for the draft-cookie behavior (API + renderPage branches, plus the no-draft no-op case). Required Vitest checks are now green.

**🟡 Dev parity gap** — Investigated. In hybrid dev, middleware runs in the Vite Pages SSR runner; calling `draftMode().enable()` inside middleware throws `draftMode() can only be called from a Server Component, Route Handler, or Server Action` *before* any cookie is set — i.e. draft-mode-via-middleware doesn't work in dev at all, independent of the fallthrough. Wiring up the dev runner's request-scoped headers context to permit `draftMode()` is an architectural change well beyond #1520's scope, so I've left it as a pre-existing, documented limitation (note added in `app-pages-bridge.ts`) rather than an accidental divergence. Header-mutation forwarding to `pages/api/*` already works in dev. The production path (what the deploy suite exercises) is fully fixed and covered.

/bigbonk review for issues

PR #1847, opened 2026-06-08
test(actions): cover server action POST under node-runtime middleware (#1480)

### LINKED ISSUE #1480: Server actions: request tracker never intercepts action POST when node-runtime middleware is configured

## Problem

Tests that invoke a server action with a `serverActions.bodySizeLimit` config under a node-runtime middleware never see the expected action POST request fire (`Did not intercept a request within 5000ms`). The action endpoint appears to be short-circuited by the node middleware instead of being dispatched.

```
Did not intercept a request within 5000ms (server action POST under node middleware)
```

## Estimated Impact

~5 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/app-dir/actions/app-action-size-limit-invalid-node-middleware.test.ts` (5 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a server action inside an app that also defines a node-runtime `middleware.ts`. Submit the action and assert the POST request is observed by the test runner.

2. **Route action POSTs through middleware properly.** Ensure that, after node middleware decides to fall through, the server-action POST is dispatched to the action handler rather than being swallowed.

3. **Also check the `bodySizeLimit` enforcement path.** The original suite is `app-action-size-limit-invalid-node-middleware`; verify the body-size-limit enforcement still runs after middleware.

james-elicx:
## Summary

Adds regression coverage for #1480 — *"Server actions: request tracker never intercepts action POST when node-runtime middleware is configured"*.

The issue (raised by a CI-analysis agent from the Next.js deploy suite) hypothesised that when a node-runtime middleware is configured, the server-action POST is short-circuited by the middleware — likely a dispatch-ordering / body-consumption bug.

## Investigation

I reproduced the exact scenario at three layers and **the bug does not reproduce on `main`**:

- **Dev** (inline middleware path) — `tests/nextjs-compat/action-node-middleware.test.ts`
- **Production** (RSC entry path the deploy suite exercises) — added to `tests/app-router-production-server.test.ts`
- **Unit** — confirmed `executeMiddleware` runs and continues with `config: { runtime: "nodejs" }`, and that `applyAppMiddleware` leaves the original request body intact for the action handler even when the middleware reads the body.

Root cause analysis:
- vinext treats node-runtime and edge middleware identically (it has no `config.runtime` branch), so a `runtime: "nodejs"` export does not change dispatch.
- `applyAppMiddleware` (`requestWithoutFlightHeaders`) and `createNextRequest` already give the middleware an **isolated body branch** (`request.clone()`), so a middleware that reads `req.text()` / `req.json()` does not consume the body the downstream server-action handler reads.
- The body-size-limit check reads `content-length` from the original request headers, which survive the middleware clone.

So the action POST is intercepted and executed correctly even when a node-runtime middleware matches the path and consumes the body. This change locks that behavior in with regression tests across the dev and production request paths (which the repo keeps in sync).

## Changes (tests only)

- `tests/fixtures/app-basic/app/nextjs-compat/action-node-mw/{page.tsx,actions.ts}` — a page with a bound `<form>` server action.
- `tests/fixtures/app-basic/middleware.ts` — match `/nextjs-compat/action-node-mw` and read the POST body before `NextResponse.next()` (mirrors Next.js' `middleware-node.js`).
- `tests/nextjs-compat/action-node-middleware.test.ts` — dev-path coverage.
- `tests/app-router-production-server.test.ts` — production-path coverage.

No `packages/vinext/src` changes were needed.

## Verification

- `tests/nextjs-compat/action-node-middleware.test.ts` — 2 passed
- `tests/app-router-production-server.test.ts` — 44 passed (incl. the new case; no regressions to the existing 43)
- `tests/nextjs-compat/action-headers.test.ts`, `tests/app-router-middleware-next-request.test.ts` — passed (no regression to actions-without-middleware or middleware-without-actions)
- `vp check` clean on all changed files.

Closes #1480

### DIFF
--- a/tests/app-router-production-server.test.ts
+++ b/tests/app-router-production-server.test.ts
@@ -886,4 +886,39 @@ describe("App Router Production server (startProdServer)", () => {
     expect(html).toContain('id="middleware-header">hello-from-middleware<');
     expect(html).toContain('id="cookie-count">0<');
   });
+
+  // Regression for cloudflare/vinext#1480: a node-runtime middleware that
+  // matches the action path and reads the request body on POST (then falls
+  // through with `NextResponse.next()`) must not prevent the server-action
+  // POST from being intercepted and executed. The app-basic middleware matches
+  // `/nextjs-compat/action-node-mw` and consumes the body for POSTs, mirroring
+  // Next.js' `middleware-node.js`. This runs against the production server
+  // because the upstream failure surfaced in the deploy suite.
+  it("dispatches a server action POST under a body-reading node middleware", async () => {
+    const html = await (await fetch(`${baseUrl}/nextjs-compat/action-node-mw`)).text();
+    // Production action ids are hashed; extract the id from the bound form's
+    // `$ACTION_1:0` reference payload rather than hardcoding it.
+    const refValue = html.match(/name="\$ACTION_[^"]*:0"\s+value="([^"]+)"/)?.[1];
+    expect(refValue).toBeDefined();
+    const decoded = refValue!
+      .replace(/&quot;/g, '"')
+      .replace(/&amp;/g, "&");
+    const actionId = JSON.parse(decoded).id as string;
+    expect(actionId).toBeTruthy();
+
+    const res = await fetch(`${baseUrl}/nextjs-compat/action-node-mw.rsc`, {
+      method: "POST",
+      headers: {
+        "Content-Type": "text/plain",
+        "x-rsc-action": actionId,
+      },
+      body: JSON.stringify(["world"]),
+    });
+    const text = await res.text();
+
+    expect(res.status).toBe(200);
+    expect(res.headers.get("x-nextjs-action-not-found")).toBeNull();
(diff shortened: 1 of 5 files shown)


james-elicx:
Addressed the review: the fixture middleware now exports `config.runtime = "nodejs"` so the `/nextjs-compat/action-node-mw` regression faithfully mirrors Next.js' `middleware-node.js`. As noted, vinext runs all middleware on Node regardless (no `config.runtime` branch), so it's a behavioural no-op for the other matcher entries but makes this a true guard for the named node-runtime scenario. Verified the full prod-server suite (44 passed) and worker-entry test still build/pass with the runtime export.

/bigbonk review for issues

james-elicx:
Addressed the doc nit: the fixture comment now points at `packages/vinext/src/server/middleware-runtime.ts` (where the matcher-only config logic lives and where a future `runtime` branch would land), instead of `middleware.ts`.

/bigbonk review for issues

PR #1849, opened 2026-06-08
fix(build): drop browser export condition for Workers SSR rendering

james-elicx:
## Problem

A `"use client"` component that imports a library which registers a custom web component (extends `HTMLElement`) at module-init time crashes on **Cloudflare Workers** with:

```
ReferenceError: HTMLElement is not defined
```

Affected library in the report: `@number-flow/react` (animated numbers). Any library that does `class … extends HTMLElement` at module load triggers it.

## Root cause

The crash hinges on this pattern inside `number-flow` (via `esm-env`):

```js
const Base = BROWSER ? HTMLElement : class {};   // BROWSER from esm-env
class NumberFlowElement extends Base {}          // evaluated at import time
```

`esm-env`'s `BROWSER` resolves to `true` **only** under the `browser` export condition, and `false` otherwise. So the crash requires the `browser` condition to be active during server rendering.

`@cloudflare/vite-plugin` sets `conditions: ["workerd", "worker", "module", "browser", …]` for every worker environment — including the `rsc`/`ssr` environments vinext renders `"use client"` components through. Because `browser` precedes `development`/`production` in `esm-env`'s export map, `BROWSER` resolves to `true`, and the SSR client-reference chunk bakes `const i = HTMLElement, a = class extends i{}`, which throws at module-init in workerd (no DOM globals).

Plain Node SSR already resolves with server conditions (no `browser`), which is exactly why `next build && next start` works — and why this only reproduces on Workers, not on `vinext build && vinext start` (verified: the Node SSR bundle resolves `BROWSER=false` and renders fine).

## Fix

In `configResolved` (after the Cloudflare plugin has merged its conditions), strip the `browser` export condition from the `rsc` and `ssr` environments' `resolve.conditions` and dep-optimizer conditions. SSR/RSC never run in a real browser, so isomorphic libraries now resolve their server-safe entry — matching Node SSR. The `client` environment is left untouched, so the browser bundle is unaffected. It is a no-op on plain Node (those environments never carry `browser`).

- `plugins/rsc-client-shim-excludes.ts`: add `withoutBrowserCondition()` + `stripBrowserConditionFromServerEnv()` (pure, typed helpers).
- `index.ts`: call the helper for the `rsc`/`ssr` environments.

## Testing

`tests/use-client-html-element-build.test.ts`:

- Unit tests for `withoutBrowserCondition` / `stripBrowserConditionFromServerEnv`.
- A Cloudflare build regression test using local `esm-env`/`number-flow`/`@number-flow/react` look-alikes that share the exact resolution shape. It asserts the worker server bundle resolves the **server-safe** entry (`BROWSER=false`) while the client bundle keeps the browser entry. **Fails before this change, passes after.**

No regressions in the affected build/SSR suites: `tsconfig-path-alias-build`, `prerender`, `build-optimization`, `features`, `app-router-production-build`, `app-router-production-server`, `app-router-worker-entry`, `app-router-client-preloading`, `app-router-rsc-plugin`. `vp check` (format/lint/types) passes.

## Notes / scope

- Scoped to App Router (`rsc`/`ssr`), which is the reported and reproduced case. A pure Pages-Router-on-Workers app importing such a library could in theory hit the same condition — that's a separate environment-setup path and would warrant its own reproduction + follow-up.
- The report also mentions `vinext build && vinext start` (Node) crashing; that part does **not** reproduce with current vinext (Node SSR resolves `BROWSER=false`). The Workers crash is the real, verifiable bug this PR fixes.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -121,6 +121,7 @@ import { generateRouteTypes } from "./typegen.js";
 import {
   mergeOptimizeDepsExclude,
   SSR_EXTERNAL_REACT_ENTRIES,
+  stripBrowserConditionFromServerEnv,
   VINEXT_OPTIMIZE_DEPS_EXCLUDE,
 } from "./plugins/rsc-client-shim-excludes.js";
 import { createServerExternalsManifestPlugin } from "./plugins/server-externals-manifest.js";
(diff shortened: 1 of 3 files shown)


james-elicx:
/bigbonk adversarial review for edge cases / side effects / unexpected bugs

PR #1852, opened 2026-06-08
fix(app-router): resolve explicit parallel slot with no page (#1535)

### LINKED ISSUE #1535: Parallel routes: explicit `@children` slot/page should win over catchall and resolve when no `default`

## Problem

Parallel-routes catchall behaviour is incorrect in two places: (1) `@children` slot for a specific page should win over the catchall slot, but vinext keeps the children hidden; (2) defining an explicit slot but no page times out instead of resolving. Distinct from #1339 — this is specifically about slot/page selection priority within parallel routes.

```
@children slot hidden when it should win over catchall
```

```
explicit slot but no page times out
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/app-dir/parallel-routes-catchall-children-slot/parallel-routes-catchall-children-slot.test.ts`
- `test/e2e/app-dir/parallel-routes-catchall/parallel-routes-catchall.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a parallel-routes fixture with `@children/[page]` and `@children/catchall` and assert the specific page wins over the catchall.

2. **Fix slot priority ordering.** When resolving parallel route slots, prefer the exact match over the catchall. Make sure explicit slots with no `default` still resolve to a usable subtree.

---

Part of #1328.

james-elicx:
Finishes the explicit-slot-but-no-page case of #1535 (children-slot priority landed in #1659).

## Problem

When a parallel slot defines an explicit sub-page (e.g. `@slot/baz/page.tsx`) but there is **no** corresponding children page (`baz/page.tsx`) and **no** `default.tsx` for the children slot, vinext minted a synthetic `/baz` route with a `null` children page. That static route shadowed the sibling catch-all (`[...catchAll]`) and rendered an empty children prop, so the request hung — the upstream `should match correctly when defining an explicit slot but no page` test timed out.

## Fix

Next.js serves such a slot-only sub-route's children from the nearest sibling catch-all. `discoverSlotSubRoutes` now falls the children prop through to a sibling catch-all page (`findCatchAllPage`) when no children `default.tsx` exists. For the upstream fixture, `/baz` renders `[...catchAll]/page.tsx` ("main catchall") for children while `@slot/baz/page.tsx` ("baz slot") fills the slot. A children `default.tsx` still wins over the catch-all when both are present.

The change is localized to `discoverSlotSubRoutes`; the children-slot priority case (`/` and `/nested`) from #1659 is untouched and still passes.

## Tests

Added a regression `describe("explicit slot but no page (issue #1535)")` in `tests/app-route-graph.test.ts`, ported from Next.js' `test/e2e/app-dir/parallel-routes-catchall`:
- slot-only sub-route resolves children to the sibling catch-all and the slot to `@slot/baz/page.tsx`
- a children `default.tsx` still wins over the catch-all when both exist

Closes #1535

### DIFF
--- a/packages/vinext/src/routing/app-route-graph.ts
+++ b/packages/vinext/src/routing/app-route-graph.ts
@@ -1102,6 +1102,13 @@ function discoverSlotSubRoutes(
     // `@slot/baz/page.tsx` for the slot. Without this, the synthetic sub-route
     // shadows the catch-all with an empty children prop and the request hangs.
     // See test/e2e/app-dir/parallel-routes-catchall ("explicit slot but no page").
+    //
+    // Known limitation: the synthetic route's URL pattern is static (e.g.
+    // `/baz`), so the catch-all children page receives empty params — Next.js
+    // would pass `params.catchAll = ["baz"]`. The route still renders correctly;
+    // only a catch-all children page that *reads* its catch-all param diverges.
+    // Populating that param needs the slot-override style request-time pattern
+    // matching threaded to the children prop; tracked as a follow-up.
     const childrenCatchAll = childrenDefault ? null : findCatchAllPage(parentPageDir, matcher);
     const childrenFallback = childrenDefault ?? childrenCatchAll;
 
--- a/packages/vinext/src/routing/app-route-graph.ts
+++ b/packages/vinext/src/routing/app-route-graph.ts
@@ -1094,6 +1094,17 @@ function discoverSlotSubRoutes(
     const childrenDefault = findFile(parentPageDir, "default", matcher);
     if (parentRoute.pagePath && !childrenDefault) continue;
 
+    // When a slot sub-route has no children page of its own (no page.tsx for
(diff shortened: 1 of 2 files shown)


james-elicx:
Thanks for the review. Addressed the catch-all-children params gap in-PR:

- Documented the divergence inline in `discoverSlotSubRoutes` (the synthetic sub-route's URL pattern is static, so a catch-all children page reading `params.catchAll` gets empty params where Next.js passes `["baz"]`).
- Added a graph-level test that locks in the current static route shape (`isDynamic: false`, `params: []`, `patternParts: ["baz"]`), so a future fix that populates the catch-all param must update that assertion rather than diverge silently.

The full param-population fix needs the slot-override-style request-time pattern matching threaded to the children prop (graph → manifest → wiring → element builder), which is broader than this hang fix; I've spun it off as a dedicated follow-up so this PR stays scoped to resolving the timeout.

PR #1853, opened 2026-06-08
fix(actions): return 404 for MPA action on app with no server actions (#1340)

### LINKED ISSUE #1340: Server actions: incorrect error handling (500 instead of 404, error boundary propagation)

## Problem

Several server action error handling behaviors diverge from Next.js:

1. **Apps with no server actions return 500 instead of 404.** When triggering a server action on an app that has no server actions defined, Next.js returns 404 with `x-nextjs-action-not-found: 1` header. vinext returns 500 (unhandled error).

2. **Error boundary propagation for action responses.** Errors from `text/plain` responses and responses with invalid content types should propagate to error boundaries. This propagation appears broken.

3. **Unrecognized action handling incorrect.** When server actions receive unrecognized action IDs, the error response format does not match Next.js expectations.

```
Expected: 404
Received: 500  // app with no server actions

Expected: error boundary to catch action error
Received: unhandled error
```

## Estimated Impact

~22 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/app-dir/actions/app-action-node-middleware.test.ts` (13 failures)
- `test/e2e/app-dir/actions/app-action.test.ts` (13 failures)
- `test/e2e/app-dir/actions-unrecognized/actions-unrecognized.test.ts` (9 failures)
- `test/e2e/app-dir/no-server-actions/no-server-actions.test.ts` (2 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test that sends a server action request to an app with no actions defined and asserts a 404 response with the `x-nextjs-action-not-found: 1` header. Add a test for error boundary propagation. Confirm they fail.

2. **Return 404 for missing actions.** When the server action handler cannot find the action ID, return `new Response(null, { status: 404, headers: { "x-nextjs-action-not-found": "1" } })` instead of throwing.

3. **Fix error propagation to error boundaries.** Study how Next.js serializes action errors for the client-side error boundary mechanism, particularly for `text/plain` and invalid content-type responses.

james-elicx:
Finishes the no-server-actions MPA case of #1340 (error-boundary/content-type/unrecognized landed in #1386/#1668).

## Problem

When an app has **no server actions** and a non-fetch **MPA / no-JS form POST** (`multipart/form-data`) is submitted to a page, vinext rendered the page with **200** instead of Next.js' **404 + `x-nextjs-action-not-found: 1`**. The fetch-action variant (via the `Next-Action` header) already returned 404; only the MPA/form-POST variant on a no-server-actions app was wrong.

Root cause: a multipart form POST to an App Router page is always a server-action attempt. When the form body decodes to no action at all, React's `decodeAction` returns `null` (it does not throw), so the progressive handler fell through to a normal page render.

## Fix

In the progressive (no-JS) action path, when `decodeAction` yields no action **and the posted-to route is a page**, return the Next.js action-not-found response (404 + `x-nextjs-action-not-found: 1`) and log `Failed to find Server Action. This request might be from an older or newer deployment.`, mirroring Next.js.

The 404 is gated on the route being a page: route handlers (`route.ts`) are dispatched *after* action handling in vinext and legitimately accept raw multipart POSTs, so they still fall through. The page-vs-route distinction uses the static `__loadPage` / `__loadRouteHandler` markers, available before lazy module hydration.

## Tests

Ported from upstream `test/e2e/app-dir/no-server-actions/no-server-actions.test.ts` ("should error when triggering an MPA action on an app with no server actions") into `tests/app-server-action-execution.test.ts`:
- 404 + action-not-found header + log when an MPA action targets a page with no server actions (fails without the fix).
- Regression guard: multipart POSTs that decode to no action on a non-page route still fall through.

Closes #1340

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -781,6 +781,17 @@ export default __createAppRscHandler({
     middlewareContext,
     request,
   }) {
+    // A multipart form POST to a page is always a server-action attempt, so a
+    // body that decodes to no action must surface as 404 action-not-found
+    // (#1340). Route handlers run after this dispatch and accept raw multipart
+    // POSTs, so only flag actual page routes. The __loadPage / __loadRouteHandler
+    // markers are static and available before lazy module hydration.
+    const __progressiveActionMatch = matchRoute(cleanPathname);
+    const __hasPageRoute = Boolean(
+      __progressiveActionMatch &&
+        __progressiveActionMatch.route.__loadPage &&
+        !__progressiveActionMatch.route.__loadRouteHandler,
+    );
     return __handleProgressiveServerActionRequest({
       actionId,
       allowedOrigins: __allowedOrigins,
@@ -794,6 +805,7 @@ export default __createAppRscHandler({
       decodeFormState,
       getAndClearPendingCookies,
       getDraftModeCookieHeader,
+      hasPageRoute: __hasPageRoute,
       maxActionBodySize: __MAX_ACTION_BODY_SIZE,
       middlewareHeaders: middlewareContext.headers,
       readFormDataWithLimit: __readFormDataWithLimit,
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -266,6 +266,7 @@ import {
 import {
   handleProgressiveServerActionRequest as __handleProgressiveServerActionRequest,
   handleServerActionRscRequest as __handleServerActionRscRequest,
+  isProgressiveServerActionRequest as __isProgressiveServerActionRequest,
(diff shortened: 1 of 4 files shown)


james-elicx:
Thanks for the review. Addressed both non-blocking observations in b62af5d:

1. **Integration coverage for the entry-side wiring** — added a dev-server test (`tests/app-router-dev-server.test.ts`) that POSTs `multipart/form-data` to `/about` (a page) and asserts 404 + `x-nextjs-action-not-found: 1`, exercising the entry's `matchRoute` + `__loadPage`/`__loadRouteHandler` classification end-to-end (not just `hasPageRoute` passed directly).

2. **Redundant `matchRoute` on every request** — gated the progressive route match on `isProgressiveServerActionRequest(request, contentType, actionId)`, so `matchRoute` now runs only for genuine multipart/no-actionId POST candidates instead of on every App Router request.

PR #1857, opened 2026-06-08
fix(app-router): scroll to top with hoisted children and loading.js (#1368)

### LINKED ISSUE #1368: App Router: move navigation scroll, hash, and focus into committed route lifecycle
> Created from the Next.js Deploy Suite failure inventory: https://github.com/cloudflare/vinext/actions/runs/26163751905
> Related: https://github.com/cloudflare/vinext/issues/1366 and https://github.com/cloudflare/vinext/issues/1367

## Problem

Vinext's client navigation scroll behavior is mostly a global side effect after navigation work completes. Next.js handles scroll/focus through the layout-router lifecycle using segment/cache-node scroll refs, so the scroll target is tied to the committed route tree rather than a raw `window.scrollTo()` after a promise.

This causes failures around hash navigation, autoscroll to newly navigated segments, horizontal scroll preservation, and focus movement.

## Failing Next.js tests

- `test/e2e/app-dir/router-autoscroll/router-autoscroll.test.ts`
  - vertical scroll top/down cases
  - horizontal scroll preservation
  - `router.refresh() should not stop router.push() from scrolling`
  - `should scroll to top even if React hoists children`
- `test/e2e/app-dir/navigation/navigation.test.ts`
  - `hash > should scroll to the specified hash`
  - `hash-link-back-to-same-page > should scroll to the specified hash`
- `test/e2e/app-dir/navigation-focus/navigation-focus.test.ts`
  - `navigation to an interactive segment`
  - `navigation to a scrollable segment`

## Architectural contract

Normal navigation scroll/focus is a committed route effect, not navigation-shim cleanup.

- Push/replace/hard refresh, hash navigation, and history traversal need separate scroll policies.
- The chosen scroll/focus policy should be part of the pending/approved navigation commit data.
- The effect should run from the rendered route/slot lifecycle once the target DOM node exists.
- Only one segment should consume a scroll/focus intent.
- `scroll={false}` must suppress normal navigation scroll without disabling unrelated traversal restoration.

## Next.js reference

- `.nextjs-ref/packages/next/src/client/components/layout-router.tsx`
  - scroll/focus is handled by layout-router after the relevant element is mounted.
- `.nextjs-ref/packages/next/src/client/components/router-reducer/ppr-navigations.ts`
  - navigation accumulates scroll refs for normal navigation, not history traversal.

## Recommendation

1. Add focused vinext e2e coverage for App Router push/replace scroll-to-top, hash scroll, horizontal scroll preservation, and focus target behavior.
2. Introduce a typed scroll/focus intent in browser router state or approved commit metadata.
3. Apply scroll/focus from the route/slot render lifecycle, not from the navigation shim after a promise resolves.
4. Separate policies for: default push/replace autoscroll, hash target scroll, `scroll={false}`, refresh, and popstate saved-scroll restoration.
5. Ensure refresh cannot consume or cancel a later push/replace scroll intent.
6. Preserve horizontal scroll unless the matching Next.js behavior requires a reset.

## Acceptance criteria

- The listed `router-autoscroll`, hash, and `navigation-focus` tests pass or are represented by equivalent vinext e2e coverage.
- Scroll/focus effects are traceable to a committed navigation intent.
- No raw normal-navigation `window.scrollTo(0, 0)` remains as the source of truth for App Router push/replace behavior.

## Non-goals

- BFCache/history-entry restoration itself. That is tracked by #1366.
- Popstate saved-scroll timing. That is tracked by #1367.

james-elicx:
Finishes the scroll-to-top-with-hoisted-children / loading.js cases of #1368 (autoscroll parity landed in #1599-1602). Closes #1368

## Problem

Two deploy-suite cases of `test/e2e/app-dir/router-autoscroll/router-autoscroll.test.ts` still failed:

- `should scroll to top even if React hoists children`
- `bugs > Should apply scroll when loading.js is used`

Both relied on the post-navigation `scrollTo` fallback in `navigateClientSide` rather than a committed route effect:

- **loading.js**: the `AppRouterScrollTarget` was mounted *inside* the route loading Suspense, so it suspended with the page content and could not scroll while the `loading.js` fallback was on screen. Scroll only happened via the fallback after the final streamed content resolved.
- **hoisted children**: when the route's first DOM node is hoisted into `<head>` (e.g. `<style precedence>`), `findDOMNode` resolves a head element with no in-body sibling, so the committed effect found no target and bailed — again leaving the raw fallback as the source of truth.

## Fix

- Mount `AppRouterScrollTarget` *outside* the route loading Suspense, matching Next.js placing `ScrollAndMaybeFocusHandler` above the `LoadingBoundary`. The handler now stays committed while the loading fallback renders, so default-navigation scroll fires against the loading boundary DOM and again when final content commits.
- When `findNextScrollTarget` resolves a node hoisted into `<head>`, scroll the document to the top as a committed effect — the equivalent of Next.js deferring the document scroll to a higher layout-router level.

This makes the scroll a traceable committed navigation effect instead of a raw post-promise `scrollTo`, per the issue's architectural contract.

## Verification

- All 23 `router-autoscroll.spec.ts` cases pass (incl. the ported hoisted + loading.js cases).
- Verified the committed effect now owns the scroll by temporarily disabling the `scrollTo` fallback — both cases still scroll to top, confirming the effect (not the fallback) is now the source of truth.
- No regressions in `navigation.spec.ts`, `hash-popstate-scroll.spec.ts`, `loading.spec.ts`, `streaming.spec.ts`, `router-push-pending.spec.ts`, and `shims.test.ts` (1033 unit tests).
- `vp check` clean on both changed files.

### DIFF
--- a/packages/vinext/src/server/app-page-route-wiring.tsx
+++ b/packages/vinext/src/server/app-page-route-wiring.tsx
@@ -817,9 +817,7 @@ export function buildAppPageElements<
 
   let routeChildren: ReactNode = (
     <LayoutSegmentProvider segmentMap={{ children: [] }}>
-      <AppRouterScrollTarget>
-        <Slot id={pageId} />
-      </AppRouterScrollTarget>
+      <Slot id={pageId} />
     </LayoutSegmentProvider>
   );
 
@@ -863,6 +861,16 @@ export function buildAppPageElements<
         </Suspense>
       );
     }
+
+    // Mount the scroll/focus target *outside* the loading Suspense so it does
+    // not suspend with the page content. Next.js places ScrollAndMaybeFocusHandler
+    // above the LoadingBoundary for the same reason: the handler must stay
+    // committed while the loading.js fallback renders, so the default-navigation
+    // scroll fires against the loading boundary's DOM (`should apply scroll when
+    // loading.js is used`) and again when the final content commits — rather than
+    // relying on a raw post-navigation scrollTo fallback that only runs after the
+    // streamed content resolves.
+    routeChildren = <AppRouterScrollTarget>{routeChildren}</AppRouterScrollTarget>;
   }
 
   const lastLayoutErrorModule =
--- a/packages/vinext/src/server/app-page-route-wiring.tsx
+++ b/packages/vinext/src/server/app-page-route-wiring.tsx
@@ -822,6 +822,10 @@ export function buildAppPageElements<
   );
 
   if (isPrefetchLoadingShell) {
+    // A prefetch loading shell is a cached payload, not a committed navigation,
+    // so it intentionally does not mount AppRouterScrollTarget — the scroll/focus
+    // effect belongs to the real render that replaces this shell (handled in the
+    // else branch below).
     if (routeLoadingComponent === null) {
       routeChildren = null;
     } else {
--- a/packages/vinext/src/shims/app-router-scroll.tsx
+++ b/packages/vinext/src/shims/app-router-scroll.tsx
@@ -68,9 +68,29 @@ function getHashFragmentDomNode(hash: string): HTMLElement | null {
   return element instanceof HTMLElement ? element : null;
 }
 
-function findNextScrollTarget(node: Element | Text | null): HTMLElement | null {
+function isInDocumentHead(node: Element | Text): boolean {
+  const head = node.ownerDocument?.head;
+  return head != null && head.contains(node);
+}
+
+type NextScrollTarget =
+  | { kind: "element"; element: HTMLElement }
+  // The route content's first DOM node was hoisted into <head> (e.g. a
+  // `<style precedence>` or metadata element) so there is no in-body element to
+  // scroll into view. Next.js skips this layout-router level and lets a higher
+  // one scroll the document to the top; vinext owns the whole page branch in a
+  // single scroll target, so it performs that document-top scroll here instead
+  // of leaving the navigation without a committed scroll.
+  | { kind: "document-top" }
+  | null;
+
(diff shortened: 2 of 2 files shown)


james-elicx:
Thanks for the review. Addressing each point:

1. **`document-top` divergence** — confirmed intentional. vinext's flat single scroll target has no higher layout-router to defer to, so scrolling the document to top here is the committed-effect equivalent of Next.js `return`ing and letting a higher level own the scroll. The comment documents this.

2. **Head-check ordering is load-bearing** — agreed, and it holds: a head-resident first node's `nextElementSibling` stays within `<head>` and can never reach in-body content, so short-circuiting before the sibling loop is correct (the loop could never find a body target anyway).

3. **Prefetch shell no longer wraps the scroll target** — confirmed intended. A prefetch loading shell is a cached payload, not a committed navigation; the scroll/focus effect belongs to the real render that replaces it. Added an explicit comment on the `isPrefetchLoadingShell` branch (10eb4c6c) so this is no longer an implicit behavioral change.

4. **Unit test for `document-top`** — `findNextScrollTarget` depends on real DOM (`instanceof Element`, `getComputedStyle`, `ownerDocument.head`), and `tests/shims.test.ts` runs in a Node environment with no `window`/`document`, so it can't exercise this path without introducing a new jsdom environment. The e2e `scrolls to top even if React hoists children` covers the `document-top` branch and the other 22 `router-autoscroll.spec.ts` cases cover the `element`/hash branches, which is the project's established mechanism for browser scroll behavior. Keeping it as e2e-only to avoid scope creep, matching how this surface is already tested.

Nits (direct `scrollTop` vs local var; no focus in the document-top branch) — left as-is; both are correct and consistent with the existing code.

PR #1860, opened 2026-06-08
refactor(pages-router): unify triplicated request pipeline into runPagesRequest (#1782)

### LINKED ISSUE #1782: Unify the triplicated Pages Router request pipeline
## Summary

The Pages Router request sequence is written three times. Same steps, same order, three copies kept in step by hand. `AGENTS.md` already names this as a maintenance hazard ("Request handling logic exists in multiple places that must stay in sync"), and the App Router solved the same problem by funneling every runtime through one handler. Pages Router never got that treatment.

The proof is hard to argue with: all three copies carry the *identical* numbered step comments.

| Step | `prod-server.ts` (Web `Response`) | `index.ts` dev (Node `res`) | `deploy.ts` (worker template) |
| --- | --- | --- | --- |
| redirects | `:1696` | `:3338` | `:701` |
| middleware | `:1721` | `:3392` | `:726` |
| config headers | `:1835` | `:3542` | `:800` |
| beforeFiles rewrites | `:1877` | `:3548` | `:823` |
| afterFiles rewrites | `:1959` | `:3613` | `:872` |
| fallback rewrites | `:1997` | `:3659` | `:907` |

Two of these (`prod-server.ts`, `deploy.ts`) operate on Web `Response`. The third (`index.ts` dev) mutates a Node `res`, so it has a parallel set of helpers — `applyRedirects` / `applyRewrites` / `applyHeaders` at `index.ts:4792-4954` — that mirror the `matchRedirect` / `matchRewrite` / `applyConfigHeadersToHeaderRecord` calls the other two use. Same semantics, two shapes, three call sites.

`server/request-pipeline.ts` looks like it should own this, but it doesn't. It holds the per-step primitives (`normalizeTrailingSlash`, `filterInternalHeaders`, `applyConfigHeadersToHeaderRecord`), not the sequence. Its own header comment admits the extraction is unfinished.

## What it looks like now vs. after

```mermaid
flowchart TB
  subgraph BEFORE["before — ordering authored 3x"]
    direction LR
    P["prod-server.ts<br/>1690-1997<br/>Web Response"]
    I["index.ts dev<br/>3338-3659<br/>Node res"]
    D["deploy.ts<br/>699-907<br/>worker template"]
  end
  subgraph AFTER["after — one owner, two adapters"]
    direction TB
    R["runPagesRequest()<br/>owns the 9-step order once"]
    R --> AW["adapter: Web Response<br/>(prod + worker)"]
    R --> AN["adapter: Node res<br/>(dev)"]
  end
  BEFORE -.collapse.-> AFTER
```

The model already exists for App Router: `entries/app-rsc-entry.ts:554` exports `__createAppRscHandler({...})`, and prod, dev, and the worker all delegate to it. One place to fix an ordering bug. That's the target shape.

## Proposed change

1. Extract the ordering into `server/pages-request-pipeline.ts` exposing one entry — roughly `runPagesRequest(req, deps): Promise<PipelineResult>` — where `deps` carries the config (redirects, headers, rewrites), the route matchers, and the render/api callbacks.
2. Express the two response models as thin adapters at the seam. The Web `Response` adapter serves `prod-server.ts` and the `deploy.ts` template; the Node `res` adapter serves `index.ts` dev. The adapter is the only part that differs.
3. Delete the duplicated sequences and the parallel `applyRedirects` / `applyRewrites` / `applyHeaders` helpers once the dev path calls through the shared owner.

Suggested order of work: merge `deploy.ts` and `prod-server.ts` first, since they share a response model and both feed the compatibility number. Fold in the dev adapter second.

## Why this is safe for compatibility

The deploy suite runs in `NEXT_TEST_MODE: deploy`, which exercises `deploy.ts` + `prod-server.ts`. It never runs the `index.ts` dev pipeline. The two prod copies are behaviorally identical today (that's the whole problem), so collapsing them is a no-op for the suite. The dev adapter is the third copy and isn't measured by the page at all, which lowers the stakes on the trickiest part.

This is mechanical reshaping, not a semantics change. No step moves, no header changes, no reordering.

## Tests that guard it

- `tests/request-pipeline.test.ts` (36 KB) — the per-step primitives
- `tests/pages-router.test.ts` (240 KB) — calls `renderPage` / `runMiddleware` through the exported interface
- `tests/deploy.test.ts` — the worker template
- `tests/routing.test.ts` (80 KB) — redirects, rewrites, trailing slash
- Deploy suite: `trailing-slashes`, `redirects`, `rewrites`, middleware suites

## Scope

In scope: the three Pages Router copies and their shared owner.

Out of scope: App Router (already unified), the documented "config headers run after middleware instead of before" parity gap (track separately — don't fold a behavior change into a refactor).

## Acceptance criteria

- [ ] One `server/pages-request-pipeline.ts` owns the step order
- [ ] `prod-server.ts`, `deploy.ts`, and `index.ts` dev call through it via adapters
- [ ] The parallel Node-`res` helpers at `index.ts:4792-4954` are gone
- [ ] Full Vitest suite green; deploy-suite pass rate unchanged
- [ ] New focused tests cover the pipeline owner directly

## Risk

Low to moderate. The two prod copies are identical, so risk is concentrated in the dev adapter — which is the one path the compatibility number doesn't watch.

---

*Filed from an architecture review. Companion issues: extract the Pages render orchestrator (unblocks this one) and add an ISR cache-decision module.*

Divkix:
Closes #1782.

The Pages Router request sequence was copy-pasted three times — `prod-server.ts` (Node), `deploy.ts` (Cloudflare Worker template), and `index.ts` (Vite dev middleware). This PR collapses all three into one owner that the adapters delegate to.

## What changed

**New:** `server/pages-request-pipeline.ts` exports `runPagesRequest(request, deps)` which owns the canonical Next.js 9-step ordering once. Returns a `PagesPipelineResult` discriminated union: `{type:"response"}` for every early return + prod/worker render, `{type:"render"|"api"}` intents when render callbacks are absent (dev). Mirrors `createAppRscHandler`'s thin-closure injection pattern.

**Migrated:**
- `prod-server.ts` `startPagesRouterServer` — keeps Node-specific concerns: decode/400, `_next/data` normalization, static assets, image opt, compression sink. Delegates the 9-step ordering.
- `deploy.ts` worker template — keeps worker-specific concerns: `registerConfiguredCacheAdapters`, static-asset 404, image bindings. Delegates the 9-step ordering.
- `index.ts` dev middleware — keeps Vite-specific concerns: Vite skips, cross-origin guard, image 302, basePath strip, `_next/data` normalization. Uses a `devRunMiddlewareAdapter` wrapper to bridge the dev runner API to `MiddlewareResult`.

**Deleted:**
- `applyRedirects`, `applyRewrites`, `applyHeaders` (parallel Node-`res` helpers at `index.ts:4829-5015`) — these were the exact three functions the issue called out.
- `proxyExternalRewriteNode` — external rewrites now return `{type:"response"}` from the pipeline.

## Behavioral decisions

**Preserved as-is (out of scope):** Config-headers-after-middleware timing gap; external-proxy merge asymmetry (merge only on post-middleware rewrite path, not on beforeFiles/afterFiles/fallback external paths — both prod and worker already agreed on this).

**Reconciled (latent worker drift fixed):** Two subtle differences between the prod and worker copies were unified to the superset:
- `middlewareStatus = result.status ?? result.rewriteStatus` (worker previously dropped `result.status`)
- `preserveCredentialHeaders: isExternalUrl(resolvedUrl)` passed to `applyMiddlewareRequestHeaders` (worker previously omitted this)

**`renderPage` signature extension:** Added optional `stagedHeaders?: Headers` parameter so the prod adapter can pass pre-staged middleware headers to the SSR renderer for CSP nonce injection — same information that was previously passed as `middlewareResponseHeaders` in the old inline code.

## Tests

- `tests/pages-request-pipeline.test.ts` (new, 28 tests): drives `runPagesRequest` directly. Covers step order, redirect 307/308 + query preservation, middleware short-circuits + multi-value Set-Cookie, before/after/fallback rewrites + external proxy, out-of-basePath 404, trailing-slash, staged-header precedence, reconciled deltas, and render/api intent emission.
- All existing tests pass: 276 pages-router, 112 routing, 245 deploy, 28 pipeline, 36 request-pipeline.

## Line counts (net)

| File | Before | After | Delta |
|---|---|---|---|
| `prod-server.ts` | 2124 | ~1680 | −444 |
| `deploy.ts` | 1630 | ~1280 | −350 |
| `index.ts` | 5157 | ~4460 | −697 |
| `pages-request-pipeline.ts` | — | 478 | +478 |
| `pages-request-pipeline.test.ts` | — | 563 | +563 |

Total: net ~−450 lines, with clear ownership and focused tests.


james-elicx:
Addressed the bonk findings plus two further regressions found while reviewing the extraction against all three original copies. Pushed in 568deadc.

**Bonk findings**
1. **Content-type regression (prod-server.ts)** — restored. The pipeline now tags API results with `isApiResponse`, and the Node adapter defaults a missing content-type to `application/octet-stream` (arbitrary data) instead of `text/html`. Page renders still default to `text/html`.
2. **Unused `{type:"next"}`** — added a comment noting it's reserved forward-compat surface; the dev passthrough is decided in the adapter via `hasAppDir`.
3. **External-proxy merge asymmetry** — added inline comments on each bare `proxyExternal(...)` return pointing to the Step 8 asymmetry note, so it isn't "fixed" into an accidental divergence.

**Two additional regressions fixed**
4. **Public-directory static serving dropped (prod-server.ts, Node)** — the original ran a post-middleware `tryServeStatic` (step 5b) for `public/` files (`favicon.ico`, `robots.txt`, …); the migrated adapter lost it, so those requests fell through to the renderer and 404'd. Restored via a `serveStaticFile` dep the pipeline calls post-middleware/pre-rewrites, with a new `{type:"handled"}` result for callbacks that write their own output.
5. **Hybrid app+pages dev: `app/api/*` lost middleware context** — the dev api intent ran `flushStagedHeaders()`/`flushRequestHeaders()` unconditionally before discovering no pages-api matched, wiping `req.headers` (incl. `VINEXT_MW_CTX_HEADER`) on the `next()` fallthrough. Gated both behind the pages-api match, mirroring the original `if (apiMatch)` guard.

Tests: +5 pipeline cases (serveStaticFile/handled ordering, isApiResponse tag); 33 pipeline, 276 pages-router, 245 deploy, 9 after-deploy, 112 routing all pass locally. `vp check` clean (pre-existing env module-resolution errors only).

james-elicx:
Addressed the remaining code-review findings in dba28a68 (the two worker divergences were already fixed in 448764d5):

3. **Content-type on passthrough responses (prod)** — middleware short-circuits and external-proxy responses with a body but no Content-Type no longer get a `text/html` default injected. Generalized the response tag to `defaultContentType` (render → `text/html`, api → `application/octet-stream`, passthrough → unset); the Node adapter sends untagged responses verbatim via `sendWebResponse`, matching pre-refactor behavior.

4. **Redirect query re-encoding (prod + dev)** — config-redirect `Location` now uses the raw `req.url` query (new `rawSearch` dep) instead of `new URL().search`, which re-encoded `"`/`<`/`>` and truncated at a literal `#`. Confirmed the original prod *and* dev both used the raw query (`applyRedirects(..., url.slice(indexOf('?')))`); the worker keeps `url.search` (it only ever had a Web Request).

5. **matchPageRoute computed twice per request** — now reuses the Step 12 match unless an afterFiles rewrite actually changed the pathname.

6. **mergeWebResponse / mergeHeaders duplication** — `mergeWebResponse` was dead on the request path (the pipeline uses `mergeHeaders`); collapsed it into a thin delegation to the canonical `mergeHeaders` and removed the now-orphaned `hasHeader`/`stripHeaders` helpers. The existing `mergeWebResponse` tests validate the delegation is behavior-identical.

3 review candidates were verified and **dropped as refuted**: dev `req.url` staleness on app-passthrough (the app router reads the rewrite from the `x-vinext-mw-ctx` `r` field, not `req.url`), dev `next.config` headers on app-passthrough (the App Router's own finalizer re-applies them), and dev `proxyExternal` header/query forwarding (it correctly uses the pipeline's post-middleware request).

Tests: 36 pipeline + 309 features + 276 pages-router + 245 deploy + 9 after-deploy + 112 routing pass; `vp check` clean.

PR #1865, opened 2026-06-09
fix(app-router): discover parallel-slot pages inside route groups (catchall-parallel-routes-group)

james-elicx:
## What failed

E2E test `catchall-parallel-routes-group > should work without throwing any errors about invalid pages`.

The fixture has this structure:
```
app/[...catchAll]/layout.tsx       (layout with `slot` prop)
app/[...catchAll]/page.tsx         (catch-all page)
app/[...catchAll]/@slot/layout.tsx (slot layout)
app/[...catchAll]/@slot/(group)/page.tsx  ← page inside route group
```

The route renders without the `@slot` content, causing the navigation to `/foobar` to not show "Catch-all Slot Group Page".

## Root cause

Two bugs in `discoverParallelSlots` / `discoverSlotSubRoutes` in `packages/vinext/src/routing/app-route-graph.ts`:

**Bug 1** — `discoverParallelSlots` only looked for `page.tsx` directly at the slot root via `findFile(slotDir, "page", matcher)`. When the page lives inside a route-group subdirectory (`(group)/page.tsx`), `findFile` returns `null`, the "has content" guard fires, and the entire slot is excluded from the route graph.

**Bug 2** — Even if the slot were included, `discoverSlotSubRoutes` processes the ghost-parent layout's sub-path. `findSlotSubPages` returns `(group)/page.tsx` as a sub-page with `relativePath = "(group)"`. Since `(group)` is URL-invisible, `urlParts = []` and the pattern matches the existing route. `applySlotSubPages` is called with `rawSegments = ["(group)"]`, overwriting `routeSegments` to `["(group)"]`. At render time, `resolveAppPageChildSegments` pushes `"(group)"` into the child-segment list, so `useSelectedLayoutSegment("slot")` would surface the route-group name rather than `null`.

## Fix

1. **`findSlotRootPage`** — new helper that finds a page at the slot's root URL level, descending through transparent route-group directories (`(name)`). `discoverParallelSlots` now uses this instead of a bare `findFile`.

2. **`discoverSlotSubRoutes`** — when `urlParts.length === 0` (all sub-segments are URL-invisible), skip `applySlotSubPages`. The existing route's slot was already populated with the correct empty `routeSegments` by `discoverParallelSlots`; calling `applySlotSubPages` here would corrupt `routeSegments` with raw filesystem segments.

## Mapping to test

`catchall-parallel-routes-group > should work without throwing any errors about invalid pages` navigates to `/foobar`, expects both "Catch-all Page" and "Catch-all Slot Group Page" to appear. With the fix, `/:catchAll+` has the `@slot` correctly populated with `pagePath` pointing to `@slot/(group)/page.tsx`, so the slot renders correctly.

## Verification

- `vp test run tests/app-route-graph.test.ts` — 49/49 pass (includes new regression test)
- `vp test run tests/app-page-route-wiring.test.ts` — 34/34 pass
- `vp test run tests/app-rsc-route-matching.test.ts` — 17/17 pass
- `vp check` — 22 pre-existing env errors, no new errors

### DIFF
--- a/packages/vinext/src/routing/app-route-graph.ts
+++ b/packages/vinext/src/routing/app-route-graph.ts
@@ -1173,7 +1173,16 @@ function discoverSlotSubRoutes(
             `You cannot have two routes that resolve to the same path ("${pattern}").`,
           );
         }
-        applySlotSubPages(existingRoute, slotPages, rawSegments);
+        // When urlParts is empty, all sub-segments are URL-invisible (e.g. route
+        // groups like "(group)"). The slot page is at the same URL level as the
+        // parent route, so discoverParallelSlots already assigned it with the
+        // correct empty routeSegments. Calling applySlotSubPages here would
+        // overwrite routeSegments with the raw filesystem segments (e.g.
+        // ["(group)"]), making useSelectedLayoutSegment return the route-group
+        // name rather than null.
+        if (urlParts.length > 0) {
+          applySlotSubPages(existingRoute, slotPages, rawSegments);
+        }
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


james-elicx:
Re: the non-blocking collision-detection observation — I verified against the Next.js reference (`next-app-loader/index.ts` `resolveParallelSegments`, ~L749-784): Next.js does **not** error when two route groups under the same slot each have a root page (`@slot/(a)/page` + `@slot/(b)/page`); it silently overwrites. The "two parallel pages resolve to the same path" error only applies to non-slot sibling-group *children* pages. Adding collision detection here would diverge from Next.js, so no change is made (mirroring upstream is correct). The other note (e2e test not in-repo) is covered by the deploy suite. Marking ready.

PR #1870, opened 2026-06-09
fix(build): support Babel flowtype page compilation (#1506)

### LINKED ISSUE #1506: Build failure: Babel flowtype page compilation not supported

## Problem

Pages using flowtype syntax (compiled via Babel + `@babel/preset-flow`) fail the vinext build with `Custom deploy script failed: undefined (1)`. The Vite/Rolldown pipeline does not run user Babel configuration for these files.

```
Babel + preset-flow page fails to build
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/babel/index.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a page using flow syntax with a `.babelrc` enabling `@babel/preset-flow`, confirm the build fails.

2. **Decide on Babel support.** Either add an opt-in Babel-compatibility loader for files matched by user `.babelrc`, or document that Babel is not portable and convert the fixture to TS.

james-elicx:
Fixes #1506.

**Failing test:** `test/e2e/babel/index.test.ts` — "Babel > Should compile a page with flowtype correctly".

Adds Babel flowtype page compilation support (changes in `packages/vinext/src/index.ts` + `tests/babel-flowtype.test.ts`).

_Note: recovered from a worker that completed the edit but was cut off before committing; relying on CI (Check/Vitest/E2E) + bonk for verification._

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -1,5 +1,5 @@
 import type { Plugin, PluginOption, UserConfig, ViteDevServer } from "vite";
-import { loadEnv, parseAst, transformWithOxc } from "vite";
+import { loadEnv, parseAst } from "vite";
 import {
   pagesRouter,
   apiRouter,
@@ -178,8 +178,7 @@ import fs from "node:fs";
 import { randomBytes, randomUUID } from "node:crypto";
 import commonjs from "vite-plugin-commonjs";
 import { normalizePathSeparators } from "./utils/path.js";
-import { hasFlowPragma } from "./utils/flow-pragma.js";
-import { transformWithFlowBabel } from "./utils/flow-babel.js";
+import { transformJsxInJs } from "./utils/flow-babel.js";
 
 // Install the process-level peer-disconnect backstop at module load.
 // Vite plugin lifecycle hooks (config / configureServer) proved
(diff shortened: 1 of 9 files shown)


james-elicx:
Review addressed in `6479cee` — both remaining findings from the latest /bigbonk review:

**#1 (PR-time coverage of the actual Babel transform):**
- Extracted `transformWithFlowBabel` into `packages/vinext/src/utils/flow-babel.ts` so it can be tested in isolation.
- Added a workspace fixture `tests/fixtures/babel-flowtype/` that declares `@babel/core` + `@babel/preset-flow` + `@babel/plugin-syntax-jsx` as its own deps (the way a real user project would) with a `.babelrc` and the Next.js-style `mycomponent.js` Flow fixture.
- `tests/babel-flowtype.test.ts` now drives the real two-stage transform: asserts Flow annotations are stripped via the fixture's `.babelrc`, the OXC re-pass compiles the leftover JSX (automatic runtime), and the composed sourcemap's `sources`/`sourcesContent` trace back to the original Flow source — so the sourcemap chaining from `a6df73a` is now verified on every PR, not just nightly. Also covers the memoized `@babel/core`-not-found path. 26 tests pass.
- Interesting wrinkle surfaced by the new test: `@babel/preset-flow` alone cannot *parse* JSX, so a Flow+JSX project's Babel config necessarily either compiles JSX (`next/babel`) or parses-but-keeps it (`@babel/plugin-syntax-jsx`). The fixture uses the latter so the OXC re-pass is meaningfully exercised; doc comments corrected to match.

**#2 (babel-missing path / inaccurate "OXC will surface Flow is not supported" comment):**
- When a leading `@flow` pragma is detected but `@babel/core` is not resolvable, the plugin now falls through to the regular OXC pass instead of skipping the transform entirely. Plain JS that merely mentions `@flow` in a leading comment still compiles; genuine Flow syntax fails OXC's parse and is rethrown as an actionable error ("install @babel/core and @babel/preset-flow…") with the original parse error as `cause` — no more confusing downstream rolldown/RSC analysis error.

Also fixed while in there: Babel's `root`/`cwd` are now pinned to the project root so `.babelrc` resolution works when the Vite process is launched from a different directory (Babel's default `babelrcRoots` only honors `.babelrc` files in its root package — previously the Babel path would silently load no config in monorepo setups).

Local verification: `vp test run tests/babel-flowtype.test.ts` → 26 passed; `vp check` → clean; `knip` → clean.

james-elicx:
Addressed all three minor notes from the last review in df5774c:

**1. Tests skip-by-null on a non-installed checkout → now fail loud.** The `transformWithFlowBabel` describe block gains a `beforeAll` that resolves `@babel/core` from the fixture root (`tests/fixtures/babel-flowtype`) and, when it isn't materialized, throws an actionable error telling you to run `vp install` — instead of three opaque "expected null not to be null" failures.

**2. `_babelCoreCache` no longer pins a not-found result for the process lifetime.** Resolution misses are no longer cached, so installing `@babel/core` mid-dev-session takes effect on the next transform without a restart. Successful resolutions stay memoized, so the hot path (Babel installed, per-module transform) is unchanged; the re-attempt cost is a failed `require.resolve` and only applies to `@flow`-pragma files in a project without Babel — which either fail the build anyway (genuine Flow syntax) or are rare pragma-comment-only plain-JS files. Locked in by a new test that "installs" `@babel/core` after a miss and asserts the next call picks it up.

**3. `cleanId` vs `id` mismatch fixed.** The fallback `transformWithOxc` call in `vinext:jsx-in-js` now receives `cleanId` (query params stripped), matching the filename given to `transformWithFlowBabel`, so sourcemap filenames and parse errors are consistent across both paths.

**Bonus robustness fix found while verifying:** `vp test` launches workers with `NODE_PATH` pointing at pnpm's hoisted `.pnpm/node_modules` store. On a fully-hoisted install that makes `@babel/core` resolvable from *any* directory, so the babel-missing tests (which rely on an empty temp dir) were environment-sensitive — green in CI, red on a local `pnpm install`. The two babel-missing tests now neutralize the `NODE_PATH` fallback (via `Module._initPaths()`, restored in `finally`) so they are deterministic regardless of installer hoisting.

Verification: `vp test run tests/babel-flowtype.test.ts` 27/27, `vp test run tests/jsx-in-js.test.ts tests/jsx-in-js-node-modules.test.ts` 3/3, `vp check` clean.

james-elicx:
Round-2 notes addressed in 1569302:

**1. Actionable-rethrow path now has unit coverage.** The `vinext:jsx-in-js` transform body (Flow pragma detection → Babel attempt → OXC fallback → actionable rethrow) was extracted into `transformJsxInJs` in `packages/vinext/src/utils/flow-babel.ts`; the plugin in index.ts now delegates to it after its extension/node_modules filtering. Three new tests cover the previously untested paths: Flow syntax + missing `@babel/core` → wrapped "install @babel/core and @babel/preset-flow" error with the OXC parse error preserved as `cause`; pragma-comment-only plain JS + missing Babel → compiles via the OXC fall-through; and the Babel-installed happy path through the full pipeline.

**2. `hasFlowPragma` on the hot path** — keeping as-is: it is a bounded short-circuiting scan of the leading comment trivia only (bails at the first real token), runs once per `.js`/`.mjs` transform immediately before a full OXC parse of the same source, so its cost is noise relative to the work that follows. node_modules files are additionally pre-filtered by the cheap directive check before reaching it.

**3. `withoutNodePathFallback` global mutation** — kept, but hoisted to module scope with its safety contract documented in the JSDoc: the mutation is restored in `finally`, tests in this file run sequentially (no `it.concurrent`), and vitest isolates other test files in separate worker processes. There is no per-require isolation hook in Node for NODE_PATH, so this is the minimal deterministic option.

Verification: `vp test run tests/babel-flowtype.test.ts` 30/30, `vp test run tests/jsx-in-js.test.ts tests/jsx-in-js-node-modules.test.ts` 3/3, `vp check` clean.

PR #1871, opened 2026-06-09
fix(use-cache): support nested cache functions passed as props

james-elicx:
**Failing test:** `test/e2e/app-dir/use-cache-with-server-function-props` — "should be able to use nested cache functions as props".

Change in `packages/vinext/src/index.ts`.

_Recovered from a cut-off worker; CI + bonk to verify. (No unit test added by the worker — may add per bonk.)_


james-elicx:
Review addressed (commit 1d3c8336):

**Build-mode key formula (blocking-ish):** The build key is now derived via plugin-rsc's own `manager.toRelativeId(id)` (against `manager.config.root`), so the hash input is byte-for-byte identical to the plugin's `hashString(toRelativeId(id))`. The string-slice is gone; the `path.relative` fallback only applies when the manager is unavailable (in which case the manifest can't be populated anyway).

**Real round-trip test (blocking, process):** Replaced `tests/use-cache-server-ref-key.test.ts` (self-referential formula test) with:
- `tests/fixtures/app-basic/app/use-cache-nested-fn-props/` — ported from the Next.js `use-cache-with-server-function-props/nested-cache` fixture (cached component passing inline `"use cache"` fns as props to a client `useActionState` form).
- A dev-mode Playwright round-trip test in `tests/e2e/app-router/use-cache.spec.ts` (click → assert ISO date / random render).
- A production integration test in `tests/app-router-production-server.test.ts` that builds the fixture, extracts the serialized reference ids from the flight payload, invokes them via action POSTs against the prod server, and asserts the values resolve — i.e. exactly the prod-manifest path that can't be covered in dev.

**Uncached-export semantics (note):** Confirmed against Next.js — its use-cache transform exports the cache wrapper itself (`$$RSC_SERVER_CACHE_n`) as the server reference, so direct action invocation goes through the cache. Matched that: the hoisted export is now reassigned at module level to `registerServerReference(registerCachedFunction(fn))` (prepended, so hoisted function declarations are wrapped before any call site or top-level initializer reads the binding). The new prod test asserts cached-invoke semantics (two identical POSTs return the same cached date).

**Found while adding the prod test — a real bug in the previous revision:** plugin-rsc's `rsc:use-server` transform runs after `vinext:use-cache` and *deletes* `manager.serverReferenceMetaMap[id]` for every module whose code lacks `"use server"`. The entry written during the use-cache transform was wiped immediately, so prod action POSTs failed with `server reference not found` (verified: the new prod test fails on the previous revision). Registration now happens in a small `vinext:use-cache-server-references` plugin placed after the plugin-rsc plugins, consuming entries queued by the transform; it also unions export names if a module legitimately has both directives.

**Minor items:**
- Deduplicated the RSC/non-RSC branches into a single `transformHoistInlineDirective` call with a shared variant parser; only the `runtime` callback differs.
- `@vitejs/plugin-rsc/react/rsc` resolution hoisted out of the per-module transform (memoised once; resolves to the same file as the cache-runtime shim's bare import, so no duplicate module instance).
- metaMap collision handling now documented + unions export names in the both-directives case.
- Rebased onto current `main`.

**Known limitation (documented in code):** closure-captured bind args of inline cached fns passed as props are serialized unencrypted (plugin-rsc encrypts `"use server"` bound args by default). Encrypting them interacts with cache-key determinism (AES-GCM ciphertext differs per call, which would break cache keying) and needs its own design pass; the ported Next.js fixture has no closures.

james-elicx:
Review addressed (commit e8a434b9 — comment-only):

**Dev key normalisation (non-blocking #1):** Documented the scope directly at the dev-key derivation. Two of the divergent shapes are structurally unreachable rather than latent bugs: the transform's filter excludes `node_modules` entirely (`exclude: /node_modules/` plus the handler guard), and the extension-anchored id regex (`/\.(tsx?|jsx?|mjs)$/`) rejects ids carrying a `?query`, so the `cleanUrl()` and query-stripping branches of the plugin's `getNormalizedId()` can't apply here. Ids outside the project root (e.g. linked packages) keep the raw absolute path instead of the plugin's `/@fs/` URL — intentionally, and self-consistently: the same key is registered in `serverReferenceMetaMap` (which dev validation checks) and handed to the dev loader's `import(id)`, which accepts absolute paths.

**Unencrypted bound args (non-blocking #2):** Tracked as a follow-up work item with full context (design constraint: cache keys are computed from the wrapper's arguments, so AES-GCM ciphertext per-call nondeterminism must not leak into the key — likely needs decrypt-inside-the-hoisted-fn before keying, mirroring Next.js' `$$cache__`/`boundArgsLength` handling). The in-code "Known limitation" comment remains as the pointer.

james-elicx:
Addressed both non-blocking follow-ups from the latest review (commits f6002305, eeda6cfa):

**1. Unencrypted closure-captured bound args — coverage + documentation (eeda6cfa)**

Full encryption parity is out of scope here (as noted in the code, encrypting bound args interacts with cache-key determinism — AES-GCM ciphertext differs per render — and needs its own design pass), so the divergence is now pinned and prominently documented instead of silently untested:

- The `use-cache-nested-fn-props` fixture gained a third prop, `getMessage`, which closes over a value from the cached component's scope — exercising the `.bind(null, ...)` bound-arg path that previously had zero coverage.
- The production round-trip test now asserts the captured value appears **verbatim (plaintext) in the flight payload** — pinning the divergence from Next.js explicitly (with a note to invert the assertion if encryption lands) — and that invoking the bound reference with the bound arg resolves and observes the captured value.
- The Playwright test clicks the new button, covering the real flight-client round-trip ($$bound serialization → `encodeReply` → server decode) in dev.
- A new transform-level test (`tests/use-cache-transform.test.ts`) pins that captures are emitted as plain bind args with no encrypt wrapper.
- User-facing documentation added to the README's **Known limitations** section ("don't close over secrets in cached functions you pass to the client"), alongside the existing code comment.

**2. Manager-less build fallback → fail loud (f6002305)**

The transform now throws when the rsc environment is active but the plugin-rsc manager is unavailable, instead of computing a fallback key and emitting a serializable-but-unresolvable reference. The throw is placed before the hoist `try` block (so the swallow-on-hoist-failure `catch` can't eat it) and covers dev as well as build — dev-mode reference validation reads the same `serverReferenceMetaMap` the registration plugin can't populate without the manager. Unit tests cover the build and dev throw paths, a non-rsc-environment control (no manager needed), and build reference-key parity (`sha256(toRelativeId(id))[:12]`) with a fake manager.

**3. Dev key cleanUrl//@fs/ normalization** — no change, per the review's agreement that the divergent shapes are structurally unreachable (documented at the dev-key computation site).

Verified: `vp check` clean; new unit tests, the prod-server nested-fn-props test, and the Playwright nested-cache spec all pass locally.

james-elicx:
Both optional polish items from the latest review are now addressed in-PR (`a9cd049f` + `67d7fb27`):

**1. `file://` vs bare-specifier coupling for `@vitejs/plugin-rsc/react/rsc` — decoupled (a9cd049f).** Rather than documenting the latent coupling, I removed it. The transform's prepend no longer imports `registerServerReference` from a `file://` URL of the plugin-rsc package entry; it imports it from a new vinext-owned shim, `packages/vinext/src/shims/cache-server-reference.ts`, which re-exports it via the **same bare `@vitejs/plugin-rsc/react/rsc` specifier the cache runtime uses, resolved from the same importer location (vinext's shims directory)**. That guarantees a single react/rsc module instance by construction — no reliance on Vite normalising a `file://` URL and a bare import to one module id, so a future change to plugin-rsc's module-id normalisation can't split the instance. The shim itself is only ever imported via the transform's `file://` URL (one specifier shape → one id), the `file://` indirection is still needed because transformed modules can live outside the project root where the bare specifier may not resolve (rationale in the shim header), and `getRscReactRscUrl()` is deleted. The transform unit test now pins that the emitted import targets the shim and never a plugin-rsc file URL.

**2. Cross-path cache assertion for the closure-bound `getMessage` (67d7fb27).** `getMessage` was deterministic, so a repeat-invocation equality check would have been vacuous. The fixture now appends a `Math.random()` suffix (making hits observable, same trick as `getDate`'s timestamp), and the prod-server round-trip test mirrors the `getDate` assertion on the bound path: two POSTs with the same bound arg must return the identical cached value (a recompute would change the suffix), and a POST with a **different** bound arg must miss with a fresh value observing the new arg — pinning both that the bound arg produces a stable cache key and that it participates in the key. The Playwright assertion now matches the suffixed message via an anchored regex.

**Verification:** `vp test run tests/use-cache-transform.test.ts` (5 passed), `vp test run tests/app-router-production-server.test.ts -t nested` (3 passed), `playwright test tests/e2e/app-router/use-cache.spec.ts` (8 passed), `vp check` clean.

README.md
```diff
@@ -599,6 +599,7 @@ These are gaps we'd like to close — distinct from the [intentional exclusions]
...
 - **Google Fonts are loaded from the CDN, not self-hosted.** No `size-adjust` fallback font metrics. Local fonts work but `@font-face` CSS is injected at runtime, not extracted at build time.
 - **Route segment config** — `runtime` and `preferredRegion` are ignored (everything runs in the same environment).
+- **Closure-captured arguments of inline `"use cache"` functions are not encrypted.** When an inline `"use cache"` function that closes over server-scope variables is passed to a client component (e.g. as a `formAction` / `useActionState` prop), the captured values are serialized into the RSC payload as plain, unencrypted bound arguments — Next.js encrypts these by default. Until this gap is closed, don't close over secrets in cached functions you pass to the client; pass an identifier and re-read the secret on the server instead. (Encrypting bound args interacts with cache-key determinism — ciphertext differs per render — and needs its own design pass.)
```
james-elicx: WTF?! fix this

packages/vinext/src/shims/cache-server-reference.ts
james-elicx, line 1: would it help at all if we exported some extra stuff from that library? there's a potential to make more of their functionality available for us to use if needed. let me know.

Issue #1830, opened 2026-06-08
App Router: misc correctness stragglers (root-layout redirect, global-not-found default, esm `.js`, react@experimental, middleware-base-path)


james-elicx:
> *This issue was created by an agent auditing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/27113612207) (vinext `main` vs Next.js `v16.2.6`, 2026-06-08).*

## Problem

A handful of small, independent App Router correctness failures with no shared root cause and no existing tracking issue. Grouped here for triage; split out if any turns out to be substantial.

## Affected Test Suites

- `test/e2e/app-dir/root-layout-redirect/root-layout-redirect.test.ts` (1 — "should work using browser"; redirect issued from the root layout)
- `test/e2e/app-dir/global-not-found/not-present/not-present.test.ts` (1 — "should render default 404 when global-not-found is not defined but enabled")
- `test/e2e/app-dir/app-esm-js/index.test.ts` (1 — "should be able to render nextjs api in app router"; `.js` ESM extension)
- `test/e2e/app-dir/rsc-basic/rsc-basic-react-experimental.test.ts` (1 — "should opt into the react@experimental when enabling $flag")
- `test/e2e/middleware-base-path/test/index.test.ts` (1 — "should execute from absolute paths")

## Estimated impact

~5 e2e failures across 5 suites.

## Recommendation

Reproduce and triage each independently; promote any non-trivial one to its own issue. `middleware-base-path` is adjacent to the closed basePath issues (#1329 / #1333).

PR #1872, opened 2026-06-09
fix(middleware): clear nextUrl.basePath for absolute paths outside basePath (part of #1830)

james-elicx:
Part of #1830.

**Failing test:** `test/e2e/middleware-base-path/test/index.test.ts` — "Middleware base tests > should execute from absolute paths".

Mirrors Next.js `getNextPathnameInfo`: `NextURL` only treats basePath as active when the pathname actually starts with the configured prefix, so middleware receiving a request without the basePath prefix sees `nextUrl.basePath === ""` and can redirect it into the basePath. Strips basePath before matcher evaluation and from rewrite targets; passes the original (un-stripped) URL to middleware from both prod-server and deploy adapters. Includes `tests/shims.test.ts` coverage.

_Recovered from a cut-off worker; CI + bonk to verify._


james-elicx:
## Review addressed (`0f10ff1`)

Both blockers from the last bonk review, plus two regressions found in self-review:

1. **Fixed the missed import assertion** at `tests/deploy.test.ts:845` — now asserts the widened `import { addBasePathToPathname, hasBasePath, stripBasePath }` emitted by the generated worker entry. `vp test run tests/deploy.test.ts` → 245 passed.
2. **Added unit coverage for the App Router `nextUrl.basePath` re-add** — new `middleware nextUrl basePath` suite in `tests/middleware-runtime.test.ts` covering: the App Router stripped-pathname re-add (`normalizedPathname: "/dashboard"` + `basePath: "/app"` → `nextUrl.basePath === "/app"`, prefixed `req.url`), the Pages in-basePath flow, the out-of-basePath absolute-path case (#1830 — `nextUrl.basePath === ""`), and matcher evaluation against the stripped path.
3. **Self-review: the unconditional re-add in `createNextRequest` regressed the Pages out-of-basePath flow** — the very case this PR fixes. The adapters deliberately pass a bare URL for absolute-path requests, but `createNextRequest` re-prefixed it via `addBasePathToPathname`, so middleware saw `nextUrl.basePath === "/root"` again. Now gated on a new `hadBasePath` option: derived from the request URL by default (correct for the prod/deploy Pages adapters, which pass the original URL), and asserted `true` by callers that pass pre-stripped URLs (App Router `applyAppMiddleware`, dev-server `runMiddleware`).
4. **Self-review: same regression class in App Route handlers** — `createTrackedAppRouteRequest` receives basePath-stripped URLs, so `request.url`/`nextUrl.href` lost the prefix and `nextUrl.basePath` reported `""` (the earlier test edit had codified this). Route handlers only run for in-basePath requests, so the prefix is now re-added before constructing the NextRequest, restoring Next.js parity; the test asserts the prefixed URL + active basePath again.
5. **Self-review: `NextURL` href reassignment parity** — `_stripBasePath` now re-derives the active basePath from the configured value on every parse (new `_configBasePath`), so setting `href` across the basePath boundary toggles `basePath` like Next.js `NextURL.analyze()` instead of clearing it permanently. Covered by a new shims test.

Verified locally: `vp test run` on `tests/middleware-runtime.test.ts`, `tests/shims.test.ts`, `tests/deploy.test.ts`, `tests/app-route-handler-runtime.test.ts`, `tests/middleware-runtime-trailing-slash.test.ts`, `tests/app-router-middleware-next-request.test.ts`, `tests/app-post-middleware-context.test.ts`, `tests/pages-api-with-middleware.test.ts` — all green; `vp check` clean.

The webkit E2E failure on `be242b4` (`hash-rsc-requests.browser.spec.ts` module-load SyntaxError) touches no file in this diff, passed on both prior commits of this branch, and main is green — treating as flake pending this run.

james-elicx:
Addressed the remaining nit from the last review: extracted the duplicated `runMiddleware` basePath re-add closure from `prod-server.ts` and the generated worker entry in `deploy.ts` into a shared `wrapMiddlewareWithBasePath` helper in `server/pages-request-pipeline.ts` (both adapters already import from that module, so the generated entry needs no new import source). Updated the `tests/deploy.test.ts` assertions to match the generated entry. `vp test run tests/deploy.test.ts tests/middleware-runtime.test.ts tests/pages-request-pipeline.test.ts` → 291 passed; `vp check` clean. (fd71a747)

james-elicx:
Addressed the follow-on nit: added a focused `wrapMiddlewareWithBasePath` suite to `tests/pages-request-pipeline.test.ts` covering the gating contract directly — pass-through when `hadBasePath === false`, pass-through when `basePath` is empty, prefix re-add (preserving query/headers/ctx/opts), and idempotent re-add for an already-prefixed URL. `vp test run tests/pages-request-pipeline.test.ts` → 38 passed; `vp check` clean. (bed17e10)

PR #1873, opened 2026-06-09
feat(images): configure image optimization via vinext({ images }) adapter

james-elicx:
## Summary

Moves server-side image optimization to a declarative, config-driven **adapter**, mirroring the cache adapter configuration. Image optimization is handled by vinext across supported targets, honors the Next.js `images` config, and no longer requires users to hand-wire the Cloudflare Images binding.

```ts
// vite.config.ts
import { imageAdapter } from "@vinext/cloudflare/images/images-optimizer";

export default defineConfig({
  plugins: [vinext({ images: { optimizer: imageAdapter() } }), cloudflare()],
});
```

## How it works

- **Registry** — `server/image-optimization.ts` exposes the optimizer registry and configured image handler, with unoptimized passthrough when no optimizer is registered.
- **Virtual module** — `virtual:vinext-image-adapters` registers the configured optimizer from the runtime environment and safely degrades when its binding is unavailable.
- **Cloudflare adapter** — `@vinext/cloudflare/images/images-optimizer` resolves its runtime at config time and uses the configured Cloudflare Images binding at request time.
- **Default entries** — generated App and Pages worker entries register the adapter and forward `request`, `env`, and `ctx`. The runtime also preserves the existing entrypoint calling convention, so pre-existing custom workers that call `handler.fetch(request)` remain compatible and receive the unoptimized fallback; forwarding `env` enables binding-backed optimization.
- **Next.js config** — allowed widths, qualities, and image security settings are embedded in the generated entries. `vinext start` reads the generated module export and retains the old `image-config.json` read fallback for pre-existing builds.
- **Deploy compatibility** — existing custom worker entrypoints remain selected by generated Wrangler config. App Router deploys without a custom worker use the current generated thin delegating worker architecture from `main`.

## Behavior / parity

- Image optimization is opt-in through `vinext({ images: { optimizer } })`.
- A missing runtime binding degrades to the existing unoptimized behavior rather than breaking requests.
- Next.js `images` options continue to control validation, allowed widths and qualities, SVG handling, content security policy, and content disposition.
- Image handling remains before middleware in both App and Pages worker paths.

## Example

`examples/app-router-cloudflare` and `examples/app-router-playground` now declare/configure the adapter and use thin workers that forward `request`, `env`, and `ctx`. `apps/web` configures the same adapter while retaining its custom API routes and scheduled maintenance worker logic.

## Validation

- Targeted image adapter, deploy, after-deploy, entry-template, image parity, and App Router production tests pass.
- The Cloudflare package builds and emits the adapter runtime sibling used by `import.meta.resolve`.
- Frozen filtered install and the `app-router-cloudflare` example build pass.
- Full PR CI, deploy previews, smoke tests, CodeQL, and performance checks are green after merging the latest `main`.
- Final `/bigbonk review for issues` reports no actionable correctness or security findings.


james-elicx:
Pushed 061cffc2 to fix the failing **Check** job and one self-review finding:

- **CI (knip)**: `viteConfigHasImageAdapter` and `formatImageOptimizationHint` were exported from `deploy.ts` but never imported anywhere, so knip failed the Check job. Added unit-test coverage for both in `tests/deploy.test.ts` (mirroring the existing `viteConfigHasCacheAdapter` / `formatMissingCacheAdapterError` suites), which also closes the coverage gap for the new deploy hint path.
- **Self-review**: `handleConfiguredImageOptimization` passed `optimizer?.transformImage` detached from its receiver, which would break an `ImageOptimizer` implemented as a class instance relying on `this`. It now wraps the call instead of detaching the method.

`vp check` and `vp run knip` are clean locally; targeted suites (`deploy`, `after-deploy`, `image-adapters-config`) pass.

james-elicx:
Addressed all findings from the review in 16b99410:

- **App Router Node prod server parity (main finding)**: `startAppRouterServer` now reads the `__imageAllowedWidths` constant inlined into the RSC entry (with the Next.js defaults as fallback for older builds), so `vinext start` accepts the same configured `deviceSizes`/`imageSizes` widths as the Cloudflare App Router worker and the Pages prod path — no more 400s for custom-width optimizer URLs.
- **`this`-binding test gap (nit)**: added a class-instance `ImageOptimizer` case to `tests/image-adapters-config.test.ts` that fails if `transformImage` is ever invoked detached again.
- **App Router passthrough wording (nit)**: updated the PR description — the "passthrough with security headers" phrasing now correctly distinguishes the Pages/Node passthrough from the App Router worker's pre-existing 302-redirect fallthrough, and the parity bullet now covers all three targets.
- **Registration-flag note**: agreed, intentional and consistent with the cache-adapter prior art — no change.

`vp check` clean; targeted suites pass (`image-adapters-config`: 20, `app-router-production-server`: 46).

james-elicx:
Addressed the remaining awareness note in 20a23a87: the Node App Router width fallback no longer conflates a missing `__imageAllowedWidths` export (old build → Next.js defaults) with an explicit empty `deviceSizes`/`imageSizes` config, which now passes through as-is — identical to the Cloudflare worker. The other note (App Router 302-redirect passthrough skipping CSP headers) is pre-existing behavior, now documented in the PR description as the review suggests, so no code change there.

james-elicx:
Addressed the drift-surface finding in 83076a6c (in this PR rather than as a follow-up): the App Router now has a single source for the next.config `images` security/header settings — the `__imageConfig` constant inlined into the RSC entry. `vinext start` reads `rscModule.__imageConfig` (same seam as the Cloudflare worker entry), the `vinext:image-config` sidecar-writer plugin is removed, and `image-config.json` remains only as a read-side fallback for dist outputs built by older vinext versions. `vp check` + knip clean; targeted suites pass (`app-router-production-server`: 46, `image-optimization-parity`: 6, `deploy`/`image-adapters-config`: 268).

james-elicx:
Addressed the custom-worker finding in 1bad0e4a: `generateWranglerConfig` now respects `hasWorkerEntry`, so an App Router app with a user-authored `worker/index.ts` but no `wrangler.jsonc` keeps `main: ./worker/index.ts` instead of having its custom worker silently dropped. Covered by a new regression test in `tests/deploy.test.ts`.

packages/cloudflare/src/image/image-adapter.ts
```diff
@@ -0,0 +1,34 @@
...
+ * absolute path to the runtime factory. Safe to call from vite.config — it
+ * never instantiates the optimizer or reads a binding.
+ *
+ * Wires `next/image` optimization (`/_next/image`) to the Cloudflare Images
+ * binding for on-the-fly resize, format negotiation (AVIF/WebP), and quality
+ * transforms at the edge — no custom worker entry required.
+ *
+ * @example
+ * import { vinext } from "vinext";
+ * import { imageAdapter } from "@vinext/cloudflare/image/image-adapter";
```
james-elicx: can we change this image adapter to be available via @vinext/cloudflare/images/images-optimizer instead

packages/cloudflare/src/images/images-optimizer.ts
james-elicx, line 1: can we use this new config in the app router cloudflare example app please?

PR #1874, opened 2026-06-09
refactor(cli): spec-driven command framework (dev/build/start/init migrated)

james-elicx:
## Summary

Introduces a **zero-dependency, spec-driven CLI command framework** under `packages/vinext/src/cli/` where a single `CommandSpec` is the source of truth for **both** argument parsing and help text. This eliminates the long-standing drift between the hand-rolled parser (`cli-args.ts`), the `node:util` parser (`parseDeployArgs`), and the separate inline `printHelp` templates.

This is the **first slice** of a stacked refactor (per AGENTS.md "prefer small stacked PRs"). It builds the framework and migrates the **`dev`, `build`, `start`, and `init`** commands end-to-end (`dev` in the first commit as the proof, the other three after review feedback). The remaining commands (`lint`, `check`, `typegen`, `deploy`) keep using the existing `parseArgs`/`printHelp` path and are untouched.

## Why

`dev`'s flags were parsed in `cli-args.ts` but documented in a separate template string in `printHelp("dev")`. Adding a flag meant editing two disconnected places. The drift was already real (e.g. `--experimental-https` is parsed but documented nowhere; global flags only appear under `build`'s help).

## What's in this slice

| File | Role |
| --- | --- |
| `cli/types.ts` | `CommandSpec` / `ArgSpec` + `InferValues` type inference |
| `cli/parse.ts` | `node:util` parseArgs wrapper; typed coercion (`port`/`int`/`positiveInt`/`string`/`bool`) + the value-validation guards ported from `cli-args.ts` (`CliUsageError`); unknown flags are a hard error by default with a per-command `passthroughUnknown` opt-in |
| `cli/help.ts` | Help rendered from the same spec (TTY-aware styling, `hidden` flags omitted) |
| `cli/command.ts` | `defineCommand` + `runCommand` (handles `--help`, prints clean usage errors with a help hint) |
| `cli/runtime.ts` | Extracted shared Vite helpers (`loadVite`, `getViteVersion`, `buildViteConfig`, `applyViteConfigCompatibility`) so commands no longer depend on `cli.ts` |
| `cli/commands/{dev,build,start,init}.ts` | The migrated command specs + `run` bodies |
| `tests/cli-framework.test.ts` | Parse + help-generation coverage (49 tests) |

`cli.ts` now routes `dev`/`build`/`start`/`init` through `runCommand(...)`; the old command bodies and their hand-written help templates are removed.

## Design notes

- **Built-in only**, matching AGENTS.md's "prefer Node built-ins" rule — wraps `node:util` parseArgs (already used by `deploy`). No new dependencies.
- **Strict by default**: unknown flags error (`Unknown option "--foo".` + a help hint, exit 1) instead of being silently ignored, per review feedback. Commands that must tolerate arbitrary flags can opt in via `passthroughUnknown`. Known `next` CLI flags that vinext doesn't need (`--turbopack` on dev/build, `--experimental-https` on dev) are declared as **hidden no-ops** so drop-in usage keeps working.
- **Validation parity**: the well-tested guards from `cli-args.ts` are preserved — missing values, empty values, "looks like another flag", strict integer/port/positive-int coercion — with messages that reference the exact form the user typed (`-p` vs `--port`).

## Verification

- `vp test run tests/cli-framework.test.ts tests/cli-args.test.ts` → **101 passed** (legacy `cli-args` suite still green; `cli-args.ts` intentionally untouched this slice).
- `vp check` + `knip` → format + lint + types + unused-exports clean.
- `vp run vinext#build` → builds successfully.
- Manual smoke test against the built `dist/cli.js`:
  - `vinext dev --help` / `-h`, `vinext build --help`, `vinext start --help`, `vinext init --help` → generated help (description + Examples; hidden compat flags omitted).
  - `vinext --help` (root) and legacy-path commands (`lint`, `check`, `typegen`, `deploy`) → unchanged.
  - Error paths: `dev -p abc`, `dev --port` (missing), `dev --port 70000` (range), `dev --port --hostname x` (looks-like-flag), `build --bogus` (unknown option) → clean message + `Run \`vinext <cmd> --help\`` hint + exit 1.
  - `vinext dev -p 31847` in `tests/fixtures/app-basic` → boots, prints banner + Local URL.

## Next steps (follow-up PRs)

1. Migrate the remaining commands (`lint`, `check`, `typegen`, `deploy`) onto the framework, then move root help/version/dispatch into a `runCli(registry)` and delete `cli-args.ts` + the `printHelp` templates.
2. Add interactive prompting (`@clack/prompts`) to `init`/`deploy` via a TTY-guarded `cli/prompt.ts` with `--yes`/non-interactive fallbacks. (Adding that dependency will require a maintainer to run the install, since the lockfile is frozen.)

### DIFF
--- a/packages/vinext/src/cli/commands/dev.ts
+++ b/packages/vinext/src/cli/commands/dev.ts
@@ -6,7 +6,18 @@
  */
 
 import { defineCommand } from "../command.js";
-import type { DevLockfile } from "../../server/dev-lockfile.js";
+import {
+  applyViteConfigCompatibility,
+  buildViteConfig,
+  getViteVersion,
+  loadVite,
+} from "../runtime.js";
+import { loadDotenv } from "../../config/dotenv.js";
+import {
+  type DevLockfile,
+  formatAlreadyRunningError,
+  tryAcquireLockfile,
+} from "../../server/dev-lockfile.js";
 
 export const devCommand = defineCommand({
   name: "dev",
@@ -29,7 +40,13 @@ export const devCommand = defineCommand({
     },
     turbopack: {
       type: "boolean",
+      hidden: true,
       description: "Accepted for compatibility (no-op, Vite is always used)",
+    },
+    "experimental-https": {
+      type: "boolean",
+      hidden: true,
+      description: "Accepted for `next dev` compatibility (no-op)",
     },
   },
   examples: [
@@ -39,15 +56,6 @@ export const devCommand = defineCommand({
   async run({ values }) {
     const port = values.port;
     const host = values.hostname;
-
-    // Lazy-load the heavy runtime helpers (which pull in the full vinext Vite
-    // plugin) so importing this command module — e.g. for unit-testing its
-    // spec and help output — stays cheap and side-effect-free.
-    const { applyViteConfigCompatibility, buildViteConfig, getViteVersion, loadVite } =
-      await import("../runtime.js");
-    const { loadDotenv } = await import("../../config/dotenv.js");
-    const { formatAlreadyRunningError, tryAcquireLockfile } =
-      await import("../../server/dev-lockfile.js");
 
     loadDotenv({
       root: process.cwd(),
--- a/packages/vinext/src/cli/commands/dev.ts
+++ b/packages/vinext/src/cli/commands/dev.ts
@@ -1,0 +1,165 @@
+/**
+ * `vinext dev` — start the development server (Vite).
+ *
+ * The command's flags and help text are both derived from the {@link CommandSpec}
+ * below; there is no separate help template to keep in sync.
+ */
+
+import { defineCommand } from "../command.js";
+import type { DevLockfile } from "../../server/dev-lockfile.js";
+
+export const devCommand = defineCommand({
+  name: "dev",
+  summary: "Start development server",
+  description: "Start the Vite-powered development server for your Next.js app.",
+  args: {
+    port: {
+      type: "port",
+      short: "p",
+      valueHint: "port",
+      description: "Port to listen on",
+      default: 3000,
+    },
+    hostname: {
+      type: "string",
+      short: "H",
+      valueHint: "host",
+      description: "Hostname to bind to",
+      default: "localhost",
+    },
+    turbopack: {
+      type: "boolean",
+      description: "Accepted for compatibility (no-op, Vite is always used)",
+    },
+  },
+  examples: [
+    { command: "vinext dev", description: "Start dev server on port 3000" },
+    { command: "vinext dev -p 4000", description: "Start dev server on port 4000" },
+  ],
+  async run({ values }) {
+    const port = values.port;
+    const host = values.hostname;
+
+    // Lazy-load the heavy runtime helpers (which pull in the full vinext Vite
+    // plugin) so importing this command module — e.g. for unit-testing its
+    // spec and help output — stays cheap and side-effect-free.
+    const { applyViteConfigCompatibility, buildViteConfig, getViteVersion, loadVite } =
+      await import("../runtime.js");
+    const { loadDotenv } = await import("../../config/dotenv.js");
+    const { formatAlreadyRunningError, tryAcquireLockfile } =
+      await import("../../server/dev-lockfile.js");
+
+    loadDotenv({
+      root: process.cwd(),
+      mode: "development",
+    });
+
+    // Ensure "type": "module" in package.json before Vite loads vite.config.ts.
+    // Without this, Vite bundles the config as CJS and tries require() on pure-ESM
+    // packages like @cloudflare/vite-plugin, which fails on Node 22.
+    applyViteConfigCompatibility(process.cwd());
+
+    const vite = await loadVite();
+
+    // Acquire the dev lock file. If another live `vinext dev` is running in this
+    // directory, print an actionable error (PID + URL) and exit. This is
+    // especially useful for AI coding agents, which frequently attempt to start
+    // a dev server without knowing one is already running.
+    //
+    // Disabled when VINEXT_NO_DEV_LOCK is set (escape hatch for unusual setups).
+    let lockfile: DevLockfile | undefined;
+    // Capture the acquisition timestamp so we can preserve it across the
+    // post-listen update(). `startedAt` is meant to reflect when this process
+    // started, not when the URL was resolved.
+    const startedAt = Date.now();
+    if (process.env.VINEXT_NO_DEV_LOCK !== "1") {
+      const root = process.cwd();
+      // Substitute "localhost" for wildcard binds so the URL is actually
+      // clickable when surfaced in the lock file before server.listen() has
+      // had a chance to resolve the real URL.
+      const initialDisplayHost = host === "0.0.0.0" ? "localhost" : host;
+      const acquired = tryAcquireLockfile({
+        root,
+        info: {
+          pid: process.pid,
+          port,
+          hostname: host,
+          appUrl: `http://${initialDisplayHost}:${port}`,
+          startedAt,
+          cwd: root,
+        },
+      });
+      if (!acquired.ok) {
+        console.error(
+          "\n  " +
+            formatAlreadyRunningError({
+              existing: acquired.existing,
+              cwd: root,
+              lockfilePath: acquired.lockfilePath,
+            }).replace(/\n/g, "\n  ") +
+            "\n",
+        );
+        process.exit(1);
+      }
+      lockfile = acquired.lockfile;
+    }
+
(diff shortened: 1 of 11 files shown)


packages/vinext/src/cli/commands/dev.ts
```diff
@@ -0,0 +1,165 @@
...
+      type: "string",
+      short: "H",
+      valueHint: "host",
+      description: "Hostname to bind to",
+      default: "localhost",
+    },
+    turbopack: {
+      type: "boolean",
+      description: "Accepted for compatibility (no-op, Vite is always used)",
+    },
```
james-elicx, line 45: can we add an option to hide a flag from the help menu so that we can hide this turbopack one

packages/vinext/src/cli/parse.ts
```diff
@@ -0,0 +1,180 @@
...
+ * Parse a command's argv into typed, validated values.
+ *
+ * A `help` boolean flag (with `-h` short alias) is always recognized, even if
+ * the command does not declare it, so `--help` works uniformly everywhere.
+ *
+ * @throws {CliUsageError} when a value-taking flag is missing its value, is
+ *   given an empty value, is followed by something that looks like another
+ *   flag, or fails type coercion (e.g. a non-integer `--port`).
+ */
+export function parseCommand<A extends Record<string, ArgSpec>>(
```
james-elicx, line 97: can we error for unknown flags

packages/vinext/src/cli/commands/dev.ts
```diff
@@ -0,0 +1,165 @@
...
+    const host = values.hostname;
+
+    // Lazy-load the heavy runtime helpers (which pull in the full vinext Vite
+    // plugin) so importing this command module — e.g. for unit-testing its
+    // spec and help output — stays cheap and side-effect-free.
+    const { applyViteConfigCompatibility, buildViteConfig, getViteVersion, loadVite } =
+      await import("../runtime.js");
+    const { loadDotenv } = await import("../../config/dotenv.js");
+    const { formatAlreadyRunningError, tryAcquireLockfile } =
+      await import("../../server/dev-lockfile.js");
```
james-elicx: kinda pointless, we can hoist these up

packages/vinext/src/cli/commands/dev.ts
james-elicx: Done in a5d3a520. Added a `hidden?: boolean` option to `ArgSpec`; `--turbopack` is now `hidden: true` (and I added a hidden `--experimental-https` no-op too). Hidden flags are still parsed/accepted but omitted from `--help`.

packages/vinext/src/cli/parse.ts
james-elicx: Done in a5d3a520. `parseCommand` now errors on undeclared flags by default (`Unknown option "--foo".` + a `Run \`vinext <cmd> --help\`` hint, exit 1). For commands that must tolerate arbitrary pass-through flags there is a per-command `passthroughUnknown?: boolean` opt-in (off by default). To keep `next dev` drop-in usage working under the stricter policy, I declared the legacy `--turbopack` and `--experimental-https` flags on `dev` as hidden no-ops.

packages/vinext/src/cli/commands/dev.ts
james-elicx: Done in a5d3a520 — hoisted back to static module-level imports.

james-elicx:
## Review addressed + `build`/`start`/`init` migrated (a5d3a520)

### Inline comments (@james-elicx)
- **Hide a flag from help** → added `hidden?: boolean` to `ArgSpec`; `--turbopack` is now `hidden: true`.
- **Error for unknown flags** → unknown flags now error by default (`Unknown option "--foo".`), with a per-command `passthroughUnknown` opt-in.
- **Hoist the dynamic imports** → `dev`'s lazy imports are back to static module-level imports.

### `/bigbonk` review items
1. **`boolean` + `multiple` collapse (latent)** — fixed at the type level: `ArgSpec` is now a discriminated union (`BooleanArgSpec | ValueArgSpec`) where `multiple` only exists on value specs, so `boolean` + `multiple` is a compile error. Added a runtime defensive check too (a repeated boolean still resolves to `true`).
2. **`boolean` `default: true` footgun** — fixed at the type level: `BooleanArgSpec.default?: false`, so a truthy boolean default is not representable.
3. **Unknown-flag policy was global** — now strict by default with a per-command `passthroughUnknown?: boolean` opt-in, exactly as suggested (also resolves the inline comment above).

Nits:
- `valueHint` on integer flags — `--prerender-concurrency` now sets `valueHint: "count"` → renders `--prerender-concurrency <count>`.
- `--experimental-https` — rather than relying on silent passthrough, it's now a **declared hidden no-op** on `dev`, so it's explicitly accepted and doesn't trip the new unknown-flag error.

### Command migrations
- Migrated `build`, `start`, and `init` onto the framework (spec-driven parse + help).
- Moved `createBuildLogger` / `hasPagesDir` / `loadBuildEmptyOutDir` into `cli/commands/build.ts`; dropped their `printHelp` templates and legacy `parseArgs` usage.
- `cli.ts` shrank further; only `deploy`/`check`/`typegen`/`lint` remain on the legacy path (next slices).

### Verification
- `vp check` on all changed files → format/lint/types clean.
- `vp test run tests/cli-framework.test.ts tests/cli-args.test.ts` → **100 passed** (48 framework incl. unknown-flag/hidden/passthrough/compat-flag coverage + 52 legacy).
- Built `dist/cli.js` and smoke-tested `build`/`start`/`init --help`, unknown-flag errors, and hidden-flag omission.
- Ran a **real `vinext build`** in `tests/fixtures/pages-basic` → full build report + exit 0, no fixture mutation.

Remaining for follow-ups: migrate `deploy`/`check`/`typegen`/`lint`, then move root help/version/dispatch into `runCli(registry)` and delete `cli-args.ts` + `printHelp`; wire `@clack/prompts` into `init`/`deploy`.

james-elicx:
## CI fixed + bonk review verified at HEAD (fa866782)

### Check job failure
- **knip — 6 unused exported types** in `cli/types.ts` (`ArgType`, `BooleanArgSpec`, `ValueArgSpec`, `PositionalSpec`, `ExampleSpec`, `CommandContext`): these are only referenced within the module, so they're no longer exported. `ArgSpec`, `CommandSpec`, and `InferValues` stay exported (consumed by `parse`/`help`/`command` and the tests). The two knip configuration hints (`sass`/`agent` in `ignoreDependencies`) are pre-existing on `main` and don't affect the exit code.

### Bonk review (2026-06-09 14:16) — verified addressed at HEAD
1. **`boolean` + `multiple` collapses to default** → fixed both ways in a5d3a520: the `BooleanArgSpec`/`ValueArgSpec` discriminated union makes `multiple` unrepresentable on booleans, and `parse.ts` still defensively resolves a repeated boolean array to `true`.
2. **`boolean` `default: true` footgun** → `BooleanArgSpec` now constrains `default?: false`, so a truthy boolean default is a compile error.
3. **Unknown-flag policy global** → unknown flags now hard-error by default with a per-command `passthroughUnknown` opt-in (also requested inline).
- **Nit: `valueHint` for int flags** → `--prerender-concurrency` declares `valueHint: "count"`; help reads `--prerender-concurrency <count>`.
- **Nit: `--experimental-https` silently dropped** → now an explicitly declared hidden no-op on `dev` (conscious, not accidental).

### Self-review
- **`vinext build --turbopack` regression**: the legacy parser accepted `--turbopack` on every command and it's a real `next build` flag, but the migrated `build` spec hard-errored on it under the new strict policy. Declared it a hidden no-op on `build` (same treatment `dev` got) + test.
- Refreshed the stale test-file header and the PR description (it still said only `dev` was migrated and that unknown flags are ignored).

### Verification
`vp test run tests/cli-framework.test.ts tests/cli-args.test.ts` → 101 passed; `vp check` + `knip` clean; `vp run vinext#build` + `dist/cli.js` smoke tests (help, unknown-flag error paths, hidden compat flags) all good.

james-elicx:
## Bonk round 2 addressed (0664f8c6)

1. **Stale `parse.ts` docstring** → rewritten to describe the strict-by-default unknown-flag policy with the per-command `passthroughUnknown` opt-in (matches `parseCommand`'s behavior and the `CommandSpec` doc).
2. **Boolean flag with an inline value silently resolves to `false`** → `--verbose=true` (stored by `node:util` as the string `"true"`) now raises a clean `CliUsageError` — `--verbose does not take a value, but got "true".` — instead of silently reading as not-set. Covered by tests for `=true`, `=false`, and the empty `=` form.

`vp test run tests/cli-framework.test.ts tests/cli-args.test.ts` → 102 passed; `vp check` + `knip` clean.

PR #1876, opened 2026-06-09
fix(pages-router): buffer SSR response for crawler/bot User-Agents

james-elicx:
### Failure

The Next.js compat test `test/e2e/streaming-ssr-edge/streaming-ssr-edge.test.ts — should not stream to crawlers or google pagerender bot` was failing. This test fetches a Pages Router page (with `runtime: 'experimental-edge'`) using Googlebot and Google-PageRenderer User-Agents and asserts:

1. The full resolved Suspense content (`next_streaming_data`) is present in `res1.text()` (Googlebot).
2. `flushCount === 1` for `res2` — the Google-PageRenderer response arrives in a **single chunk**, not multiple flushes.
3. Both `res1` and `res2` carry an `ETag` response header.

Without this fix, vinext streamed all Pages Router SSR responses identically regardless of UA, so bots received multi-chunk incremental responses without an ETag.

### Root Cause

`renderPagesPageResponse` in `packages/vinext/src/server/pages-page-response.ts` assembled the HTML as a `compositeStream` and returned it directly as a `Response(compositeStream, …)` for every request. There was no bot/crawler detection at any layer in the Pages Router SSR path.

By contrast, Next.js's Node.js path in `server/send-payload.ts` checks `!result.isDynamic` (which is `true` for static/buffered renders) and calls `generateETag(payload)` before sending. The edge template (`build/templates/edge-ssr.ts`) and the app-page template similarly call `getBotType(userAgent)` to decide whether to buffer.

### Fix

Three files changed:

- **`packages/vinext/src/server/pages-page-response.ts`** — Adds:
  - `isPagesStreamingBot(ua)`: detects crawler UAs using the same regexes as Next.js's `html-bots.ts` + `is-bot.ts` (`HTML_LIMITED_BOT_UA_RE` for HTML-limited crawlers, `HEADLESS_BROWSER_BOT_UA_RE` for Googlebot).
  - `generatePagesETag(payload)`: FNV-1a 52-bit ETag, byte-for-byte matching Next.js's `lib/etag.ts`.
  - `userAgent?: string` field added to `RenderPagesPageResponseOptions`.
  - After the `compositeStream` is built, when `userAgent` matches a bot, the stream is drained to a string (`readStreamAsText`), an ETag is computed, and a single-chunk `Response` is returned with the `ETag` header set.

- **`packages/vinext/src/server/pages-page-handler.ts`** — Passes `request.headers.get("user-agent") ?? undefined` as `userAgent` to `renderPagesPageResponse`.

- **`tests/pages-page-response.test.ts`** — Adds `isPagesStreamingBot` unit tests and four `renderPagesPageResponse` integration tests covering Googlebot, Google-PageRenderer, a normal browser UA, and the case where `userAgent` is omitted.

### Test Mapping

| Next.js assertion | This fix |
|---|---|
| `res1.text()` contains `next_streaming_data` (resolved Suspense) | Buffering awaits `readStreamAsText(compositeStream)` which drains the Suspense body fully |
| `flushCount === 1` for Google-PageRenderer | Single-chunk `Response(fullHtml)` emits exactly one flush |
| `res1.headers.get('etag')` is defined | `generatePagesETag(fullHtml)` → `ETag` header set |
| `res2.headers.get('etag')` is defined | Same |

## Test plan

- [x] `vp test run tests/pages-page-response.test.ts` — 28 tests pass (14 pre-existing + 14 new)
- [x] `vp test run tests/app-ssr-stream.test.ts` — 39 tests pass (no regression)
- [x] `vp check` — 0 new TS/lint errors (22 pre-existing env errors unchanged)
- [ ] CI: `streaming-ssr-edge/should not stream to crawlers or google pagerender bot`

### DIFF
--- a/packages/vinext/src/server/pages-page-response.ts
+++ b/packages/vinext/src/server/pages-page-response.ts
@@ -14,6 +14,7 @@ import {
   type RenderPageEnhancers,
   runDocumentRenderPage,
 } from "./pages-document-initial-props.js";
+import { fnv1a52 } from "../utils/hash.js";
 import { readStreamAsText } from "../utils/text-stream.js";
 import { callDocumentGetInitialProps } from "./document-initial-head.js";
 
@@ -54,37 +55,6 @@ export function isPagesStreamingBot(userAgent: string): boolean {
 // packages/next/src/server/lib/etag.ts. Both produce a strong ETag (no W/
 // prefix) when weak=false, which is the default Next.js uses at runtime.
 // ---------------------------------------------------------------------------
-
-function fnv1a52(str: string): number {
-  const len = str.length;
-  let i = 0,
-    t0 = 0,
-    v0 = 0x2325,
-    t1 = 0,
-    v1 = 0x8422,
-    t2 = 0,
-    v2 = 0x9ce4,
-    t3 = 0,
-    v3 = 0xcbf2;
-
-  while (i < len) {
-    v0 ^= str.charCodeAt(i++);
-    t0 = v0 * 435;
-    t1 = v1 * 435;
-    t2 = v2 * 435;
-    t3 = v3 * 435;
-    t2 += v0 << 8;
-    t3 += v1 << 8;
-    t1 += t0 >>> 16;
-    v0 = t0 & 65535;
-    t2 += t1 >>> 16;
-    v1 = t1 & 65535;
-    v3 = (t3 + (t2 >>> 16)) & 65535;
-    v2 = t2 & 65535;
-  }
-
-  return (v3 & 15) * 281474976710656 + v2 * 4294967296 + v1 * 65536 + (v0 ^ (v3 >> 4));
-}
 
 export function generatePagesETag(payload: string): string {
   return '"' + fnv1a52(payload).toString(36) + payload.length.toString(36) + '"';
--- a/packages/vinext/src/server/pages-page-response.ts
+++ b/packages/vinext/src/server/pages-page-response.ts
@@ -688,6 +688,14 @@ export async function renderPagesPageResponse(
   // cache-HIT path in `pages-page-data.ts` which applies the same UA gate for
   // consistent ETag coverage on cached responses.
   //
+  // A consequence of UA-gating is that the ETag/304 path below can also fire
+  // for dynamic (GSSP) responses, which Next.js never 304s (`isDynamic` renders
+  // skip ETag generation entirely). The risk is minimal in practice: GSSP
+  // responses carry `Cache-Control: private, no-cache, no-store,
+  // must-revalidate`, so a conformant client won't revalidate them with
+  // `If-None-Match` — but a dynamic page may legitimately differ between the
+  // ETag-generating render and a revalidation request.
+  //
   // When the incoming `If-None-Match` header matches the computed ETag, return
   // `304 Not Modified` with no body (mirrors Next.js `sendEtagResponse`
   // semantics in packages/next/src/server/send-payload.ts, with weak-ETag
--- a/packages/vinext/src/server/pages-page-response.ts
+++ b/packages/vinext/src/server/pages-page-response.ts
@@ -112,6 +112,17 @@ export function etagMatches(etag: string, ifNoneMatch: string): boolean {
   return false;
 }
 
+/**
+ * Returns true when a request `Cache-Control` header asks to bypass the 304
+ * short-circuit. Mirrors the `fresh` package's check used by Next.js's
+ * `sendEtagResponse` (`/(?:^|,)\s*?no-cache\s*?(?:,|$)/`). Shared by the
+ * fresh-MISS bot path here and the ISR HIT/STALE paths in
+ * `pages-page-data.ts` so the two cannot drift.
+ */
+export function requestsNoCache(cacheControl: string | undefined): boolean {
+  return /(?:^|,)\s*no-cache\s*(?:,|$)/.test(cacheControl ?? "");
+}
+
 type PagesFontPreload = {
   href: string;
   type: string;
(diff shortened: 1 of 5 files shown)


james-elicx:
Flipped to ready — the targeted failing test (no-stream-to-crawlers) is fixed and bonk confirmed merge-ready. The 4 non-blocking observations (If-None-Match→304 handling, ETag on ISR cache-HIT responses, documenting the UA-gated-vs-isDynamic divergence) are tracked as follow-ups beyond this PR's scope.

james-elicx:
Addressed the non-blocking observations from the two latest reviews (both on `64e65f8`):

1. **RFC citation nit** — fixed in `9bca2f18`: both comments in `pages-page-response.ts` now cite RFC 7232 §2.3.2 (weak comparison) instead of §3.2 (`If-None-Match` field definition).

2. **Naive `.split(",")` in `etagMatches`** — leaving as-is, per the review's own assessment. Vinext's generated ETags (`base36(fnv1a52) + base36(length)`) can never contain a comma, so a comma-bearing client token could never equal one of our ETags anyway; the naive split is behaviorally identical to `fresh`'s `parseTokenList` for every token we could possibly match.

3. **MISS-vs-HIT ETag divergence** (per-request `clientTraceMetadata` `<meta>` tags) — not changing: this matches Next.js semantics. `generateETag(payload)` in Next.js's `send-payload.ts` hashes whatever payload is actually sent — the cached HTML on a cache hit, the fresh render on a miss — so Next.js has the identical property when per-request markup differs across renders.

4. **HEAD requests on the buffered bot path** — not changing: the body is already dropped at the send layer, not in the render path. The Node server explicitly sends headers-only for HEAD (`prod-server.ts:773`, "HEAD requests: send headers only, skip the body"), and workerd strips HEAD response bodies at the runtime level. The only delta vs `sendRenderResult`'s early HEAD exit is wasted render work, which is identical to the pre-existing streaming path's behavior for HEAD.

No functional changes — `9bca2f18` is comment-only; `tests/pages-page-response.test.ts` (40 tests) still passes.

james-elicx:
Addressed the latest review (on `9bca2f1`):

1. **304-on-dynamic divergence note** — added in `019c61ea`: the divergence comment in `pages-page-response.ts` now explicitly calls out that UA-gating means the ETag/304 path can also fire for dynamic (GSSP) responses, which Next.js never 304s, and why the practical risk is minimal (GSSP responses carry `Cache-Control: private, no-cache, no-store, must-revalidate`).

2. **MISS-vs-HIT ETag** — no change needed per the review's own assessment (Next.js's `generateETag(payload)` has the same property).

Comment-only change; `tests/pages-page-response.test.ts` (40 tests) still passes.

james-elicx:
Addressed the latest review (on `019c61e`):

**1. HEAD requests on the bot path (claimed Node prod-server regression) — rebutting with evidence; no code change.** The claim that a bot issuing `HEAD` "receives a full response body" via `sendCompressed` is not accurate: Node's HTTP layer itself suppresses response bodies for HEAD requests. `_http_server` sets `res._hasBody = false` when `req.method === 'HEAD'`, and `OutgoingMessage.write_` drops all body chunks for such responses ("This type of response MUST NOT have a body. Ignoring write() calls."). I verified empirically against both `sendCompressed` branches with raw-socket HEAD requests:

- `res.writeHead(...); res.end(buf)` (uncompressed branch) → wire response is headers-only, with correct `Content-Length: 47` and zero body bytes.
- `pipeline(compressor, res)` (compressed branch) → headers-only as well; every compressed chunk written to `res` is dropped by Node core.

So on every runtime the wire behavior is headers-only: workerd strips HEAD bodies, and the Node prod/dev servers drop them in Node core. Note also that buffered API-route responses (unmarked, `defaultContentType: application/octet-stream`) have always flowed through `sendCompressed`, so HEAD-through-`sendCompressed` is a long-standing, working pattern — not something this PR routes for the first time. `sendWebResponse`'s explicit HEAD branch exists to cancel the source stream early (resource cleanup for streamed bodies), not because `sendCompressed` would otherwise leak a body. The only true delta vs Next.js's `sendRenderResult` HEAD early-exit is wasted render/compress CPU, identical in kind to the pre-existing streaming path's behavior for HEAD.

**2. 304 carries `Content-Type` on the Node path** — matches Next.js, no change: `sendRenderResult` sets `Content-Type` on the response *before* calling `sendEtagResponse`, so Next.js's own 304s carry `Content-Type: text/html` too (and RFC 9110 §15.4.5 permits representation metadata on 304). The Worker path additionally strips it via `worker-utils.ts`; harmless asymmetry, same as upstream.

**3. Duplicated `no-cache` regex** — fixed in `a1d5918e`: extracted a shared `requestsNoCache()` helper in `pages-page-response.ts`, now used by both the fresh-MISS bot path and the ISR HIT/STALE `applyBotETagAndCheck` helper, so the two cannot drift.

`vp check` clean (format, lint, types); `tests/pages-page-response.test.ts` (40 tests) passes.

packages/vinext/src/server/pages-page-response.ts
```diff
@@ -17,6 +17,112 @@ import {
...
+  return HEADLESS_BROWSER_BOT_UA_RE.test(userAgent) || HTML_LIMITED_BOT_UA_RE.test(userAgent);
+}
+
+// ---------------------------------------------------------------------------
+// ETag generation — FNV-1a 52-bit, matching Next.js's generateETag() in
+// packages/next/src/server/lib/etag.ts. Both produce a strong ETag (no W/
+// prefix) when weak=false, which is the default Next.js uses at runtime.
+// ---------------------------------------------------------------------------
+
+function fnv1a52(str: string): number {
```
james-elicx: what about the fnv1a64 utility?

PR #1877, opened 2026-06-09
fix(edge-wasm): handle `*.wasm?module` imports in non-Cloudflare builds

### LINKED ISSUE #1351: Build failure: WASM `?module` import not supported by Rolldown

## Problem

Rolldown cannot handle `.wasm?module` imports. This is a Cloudflare Workers-specific import convention (`?module` suffix tells the runtime to instantiate the WASM module). Neither Rolldown nor Vite have built-in support for this pattern.

```
Build failed with 1 error:
[UNLOADABLE_DEPENDENCY] Could not load src/add.wasm?module
   ╭─[ src/add.js:1:18 ]
 1 │ import wasm from "./add.wasm?module";
```

Next.js handles this through Webpack's experimental WASM support.

## Estimated Impact

~4 test failures (build failure in all 3 Vite environments: RSC, client, SSR).

## Affected Test Suites

- `test/e2e/edge-can-use-wasm-files/index.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test that imports a `.wasm?module` file and verify the build fails.

2. **Implement a Vite plugin for WASM module imports.** Create a plugin that intercepts `.wasm?module` imports and handles them appropriately for the target environment. For Cloudflare Workers, the `?module` import should be preserved in the output. For Node.js, it should use `WebAssembly.compile()` with a file read.

3. **Note:** This may require upstream Rolldown support. Check if there is an existing Rolldown issue or plugin for WASM imports.

james-elicx:
## Failure

`test/e2e/edge-can-use-wasm-files` — four tests fail in the deploy-suite:

- edge api endpoints can use wasm files > uses the wasm file
- middleware can use wasm files > uses the wasm file
- middleware can use wasm files > can be called twice
- middleware can use wasm files with the experimental modes on > uses the wasm file

Bundle error: Rolldown cannot resolve `import wasm from './add.wasm?module'` in non-Cloudflare builds.

## Root cause

The `?module` query on `.wasm` imports is a Cloudflare Workers / workerd convention. It is handled by `@cloudflare/vite-plugin`'s `additionalModulesPlugin` (enforce:"pre") — but only when that plugin is present.

`vinext init` emits a plain `vite.config.ts` with only `plugins: [vinext()]` (no `@cloudflare/vite-plugin`). In this configuration — used by the deploy-suite and standalone `vinext start` — Rolldown/Vite 8 encounters `*.wasm?module` imports and has no handler for them.

## Fix

New `vinext:wasm-module-import` Vite plugin (enforce:"pre") in `packages/vinext/src/index.ts`:

1. **resolveId** — intercepts any `*.wasm?module` import specifier, bails out early when `hasCloudflarePlugin` is true (checked at call time, not array-construction time), resolves the real `.wasm` path via `this.resolve(..., { skipSelf: true })`, and returns a virtual module ID `\0vinext-wasm-module:<path>`.

2. **load** — reads the `.wasm` binary synchronously, inlines it as a base64 string, and emits:
   ```js
   const _b64 = "<base64>";
   const _buf = Uint8Array.from(atob(_b64), c => c.charCodeAt(0));
   export default await WebAssembly.compile(_buf.buffer);
   ```
   `atob()` is available on Node 16+, browsers, and workerd. Top-level await is valid because Vite/Rolldown always emits ESM for SSR/server builds.

workerd forbids compiling WASM from bytes at runtime — the `hasCloudflarePlugin` guard ensures this path never runs in workerd builds.

## Test mapping

`tests/wasm-module-import.test.ts` (new unit test):

| Test | What it covers |
|---|---|
| plugin is present in the vinext() array | registration / enforce:"pre" |
| load hook emits correct base64+compile code | correct virtual-module output |
| uses atob() rather than Buffer | universal compatibility |
| throws on missing file | error handling |
| reads actual bytes faithfully | byte-level round-trip |
| end-to-end WebAssembly compile produces working module | full compile→instantiate→call |

`vp check`: 22 errors / 3 warnings — identical to pre-existing count, no new errors.

Fixes #1351.


james-elicx:
Addressed the three minor items from the last review in a5cb2e6a:

- **`bytes!` non-null assertion**: the `catch` block now does `return this.error(...)` so the control flow is explicit and the assertion is gone.
- **Top-level await scope**: expanded the comment to state this output is intended for server/edge (SSR) imports only, where ESM output is guaranteed — not client bundles.
- **Exact-match `?module` filter**: kept the exact match deliberately and documented why — `?module` as the entire query is the documented Cloudflare/Next.js convention; variants like `?module&v=1` are intentionally left to the bundler rather than silently claimed.

`tests/wasm-module-import.test.ts` (12/12) and `vp check` are clean.

james-elicx:
Addressed the latest review's items in 784935d3:

- **Guard keys on plugin, not target runtime**: added a NOTE in the plugin header spelling out that the gating assumption is "no `@cloudflare/vite-plugin`", not "Node.js target", with the Nitro-edge divergence (Cloudflare/Deno presets) called out explicitly and a pointer to gate on target runtime if that combination ever needs support.
- **Build-level test coverage**: added a build-level test to `tests/wasm-module-import.test.ts` that reproduces the reported failure mode through the real pipeline — a tmpdir App Router fixture with `import addWasm from "./add.wasm?module"` in a route handler, built via `createBuilder().buildApp()` (no Cloudflare plugin). It asserts the build succeeds (pre-fix this throws on resolve), the emitted server chunk carries the byte-exact base64 payload plus `WebAssembly.compile`, and the payload compiles/instantiates with `add_one(41) === 42`. Same pattern as `tests/inline-css-build.test.ts`, which already runs `createBuilder` in the unit project.
- **Substring match in the og guard**: kept as-is, deliberately — the predicate must mirror the `vinext:og-font-patch` transform filter (`id.includes("@vercel/og")`, `index.ts` transform at the og patch plugin). Tightening this guard without tightening that filter would let an id that font-patch transforms slip past the guard and reintroduce the double-shipped-WASM bug. Added a code comment stating the two predicates must never diverge.

13/13 tests pass; `vp check` clean.

james-elicx:
Addressed all three items from the latest review in aa48075b:

- **TLA/client-environment assumption is now enforced, not assumed**: `resolveId` bails for the `client` environment (`this.environment?.name === "client"`), so the top-level-await output can only land in server/edge (SSR) bundles — a client-side `.wasm?module` import is left to the bundler to error on loudly. The TLA comment now describes the enforced guard instead of an assumption, and there's a new test covering the client-env bail.
- **Dev HMR**: the load hook now calls `this.addWatchFile(filePath)` so editing the underlying `.wasm` invalidates and reloads the importing module in dev; covered by a new test asserting the watch-file registration.
- **Extracted to `plugins/`**: the plugin now lives in `packages/vinext/src/plugins/wasm-module-import.ts` as `createWasmModuleImportPlugin({ getHasCloudflarePlugin })`, following the existing factory-with-getter pattern (`hasCloudflarePlugin` is resolved in the `config` hook, so it's read at hook call-time). `index.ts` keeps a short pointer comment.

15/15 wasm tests pass (two new), `tests/og-font-patch.test.ts` (12) + `tests/og-assets.test.ts` (8) still pass, `vp check` clean.

packages/vinext/src/plugins/wasm-module-import.ts
```diff
@@ -0,0 +1,141 @@
...
+      // we want to silently claim — leave it to the bundler to error on.
+      filter: { id: /\.wasm\?module$/ },
+      async handler(source: string, importer: string | undefined) {
+        // Defer to @cloudflare/vite-plugin when it's present — it handles
+        // ?module imports for all worker environments via its own
+        // `additionalModulesPlugin` (also enforce:"pre").  Both plugins have
+        // the same enforce level; by checking this flag at call-time we let
+        // Cloudflare's plugin "win" by returning null here and allowing it to
+        // intercept first in the registration order.
+        if (getHasCloudflarePlugin()) return null;
```
james-elicx: i dont want a cloudflare plugin check in here - it should work the same across all targets

tests/wasm-module-import.test.ts
james-elicx, line 1: add proper e2e test in one of our e2es as well

PR #1878, opened 2026-06-09
fix(app-router): suppress redirect console errors in production onCaughtError (root-layout-redirect)

james-elicx:
## Failure

The Next.js compat deploy-suite test **root-layout-redirect > should work using browser** was failing. The test:

1. Navigates to `/`
2. Clicks a button in the `'use client'` root layout that calls `redirect('/result')`
3. Waits for `#result` element with text "Result Page"
4. Asserts **no browser console errors** (`log.source === 'error'`)

The test was failing on step 4 — a spurious console error was being emitted.

## Root Cause

When `redirect()` is called from a `'use client'` component during a state update, React catches the `NEXT_REDIRECT` sentinel error via `RedirectErrorBoundary`. After an error boundary catches an error, React invokes the `onCaughtError` callback registered on `hydrateRoot`.

In **dev mode**, vinext already provided `devOnCaughtError` which filters navigation signal errors with `if (isNavigationSignalError(error)) return;` — so no console error in dev.

In **production mode**, vinext did NOT provide an `onCaughtError` at all. React's built-in fallback behaviour is to call `console.error` with the caught error. This produced a browser console error for the `NEXT_REDIRECT` sentinel, causing the test assertion `expect(foundErrors).toBe(false)` to fail.

The root layout is a particularly visible place for this because:
- It renders `<html>/<body>` and the redirect throws during a button-click state update
- `HandleRedirect` renders `null`, replacing the subtree
- React then calls `onCaughtError` with the redirect error before triggering navigation

## Fix

Added `prodOnCaughtError` to `packages/vinext/src/server/app-browser-error.ts`:
- Navigation sentinel errors (`NEXT_REDIRECT;*`, `NEXT_NOT_FOUND`, `NEXT_HTTP_ERROR_FALLBACK;*`) are silently dropped — the dedicated boundary has already consumed them and triggered navigation
- All other caught errors are forwarded to `console.error`, preserving React's default behaviour

Wired into the production `hydrateRoot` options in `app-browser-entry.ts` (the dev path already used `devOnCaughtError`).

## Test Mapping

| Next.js test | vinext test |
|---|---|
| `test/e2e/app-dir/root-layout-redirect/root-layout-redirect.test.ts` > `should work using browser` | `tests/e2e/root-layout-redirect/root-layout-redirect.spec.ts` > `should work using browser` |

New fixture at `tests/fixtures/root-layout-redirect/`: a dedicated app with a `'use client'` root layout (matching the Next.js reference fixture exactly) and port 4184 Playwright project.

## Notes

- `tests/fixtures/root-layout-redirect/app/result/page.tsx` is force-tracked because `.gitignore` contains a global `result` pattern for Playwright output dirs; this is intentional app code.
- The `prodOnCaughtError` inlines the navigation-signal digest check to keep `app-browser-error.ts` dependency-free in production bundles.

Part of #1830

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -326,6 +326,10 @@ jobs:
             label: standalone-output
             shardIndex: 1
             shardTotal: 1
+          - project: root-layout-redirect
+            label: root-layout-redirect
+            shardIndex: 1
+            shardTotal: 1
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,9 @@ reports/
 # Nix
 result
 result-*
+# The Nix `result` pattern above also matches Next.js compat fixture routes
+# named `result`; re-include them so they don't silently fall out of tracking.
+!tests/fixtures/root-layout-redirect/app/result/
 .direnv/
 .corepack/
 
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -141,7 +141,7 @@ import { AppRouterContext } from "vinext/shims/internal/app-router-context";
 import { BfcacheStateKeyMapContext, ElementsContext, Slot } from "vinext/shims/slot";
 import type { RouteManifest } from "../routing/app-route-graph.js";
 import { stripBasePath } from "../utils/base-path.js";
-import { createOnUncaughtError } from "./app-browser-error.js";
+import { createOnUncaughtError, prodOnCaughtError } from "./app-browser-error.js";
 import { createClientReuseManifestHeaderFromVisibleAppState } from "./app-browser-client-reuse-manifest.js";
 import {
   devOnCaughtError,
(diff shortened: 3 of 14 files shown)


james-elicx:
Addressed the remaining findings from the latest review in 6271850f:

- **Unit tests for `prodOnCaughtError`**: added a `describe` block in `tests/app-browser-entry.test.ts` mirroring the existing `devOnCaughtError` coverage — ignores redirect/notFound/HTTP-fallback sentinels, forwards real errors to `console.error`, includes the component stack, logs only the error when no stack is provided, and a digest-vs-message classification guard (an error whose *message* mentions `NEXT_REDIRECT` but has no digest is still logged). All 186 tests in the file pass.
- **Gitignore nit**: added a scoped negation `!tests/fixtures/root-layout-redirect/app/result/` under the Nix `result` pattern so the fixture's result page can't silently fall out of tracking (verified with `git check-ignore`).
- **Self-review finding — the new e2e test never ran in CI**: the `root-layout-redirect` Playwright project was missing from the E2E matrix in `.github/workflows/ci.yml` (CI selects projects via `PLAYWRIGHT_PROJECT`), so the regression test was defined but never executed. Added the matrix entry and verified the spec passes locally against the production server (`1 passed`).

`vp check` passes (format, lint, types).

PR #1879, opened 2026-06-09
fix(pages-router): hard-navigate to App Router destinations from Pages Router links

james-elicx:
## Summary

- In a hybrid (app/ + pages/) build, clicking a Next.js `<Link>` on a Pages Router page that targets an App Router route was attempting a Pages Router SPA fetch cycle instead of performing a hard navigation. The Pages Router client cannot render App Router pages so the navigation would fail or stall.
- Added an `__appRouter` marker check to `performNavigation` in `packages/vinext/src/shims/router.ts`. When the destination was detected as an App Router route during prefetch (recorded as `components[path] = { __appRouter: true }` via `markAppRouteDetectedOnPrefetch`), the function now immediately calls `window.location.assign/replace` and returns a never-resolving Promise — exactly mirroring Next.js `router.ts:1448-1453`.
- Added regression fixture (`tests/fixtures/app-basic/pages/pages-to-app/[slug].tsx`) and Playwright test (`tests/e2e/app-router/pages-to-app-routing.spec.ts`) ported from the deploy-suite test `pages-to-app-routing > should work using browser`.

Fixes deploy-suite failure: `test/e2e/app-dir/pages-to-app-routing/pages-to-app-routing.test.ts` → `should work using browser`.

## Test plan

- [ ] `vp check` passes (no new TS errors beyond the ~23 pre-existing env errors)
- [ ] New Playwright test `tests/e2e/app-router/pages-to-app-routing.spec.ts` passes under `app-router` project
- [ ] Existing `pages-router-use-params.spec.ts` tests unaffected
- [ ] Deploy-suite `pages-to-app-routing > should work using browser` passes in CI

### DIFF
--- a/packages/vinext/src/shims/internal/app-route-detection.ts
+++ b/packages/vinext/src/shims/internal/app-route-detection.ts
@@ -36,7 +36,7 @@
  */
 import type { VinextLinkPrefetchRoute } from "../../client/vinext-next-data.js";
 import { createRouteTrieCache, matchRouteWithTrie } from "../../routing/route-matching.js";
-import { stripBasePath } from "../../utils/base-path.js";
+import { stripBasePath, removeTrailingSlash } from "../../utils/base-path.js";
 
 const appRouteTrieCache = createRouteTrieCache<VinextLinkPrefetchRoute>();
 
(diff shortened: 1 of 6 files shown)


james-elicx:
Addressed both non-blocking nits from the latest review round in `8bf0252`:

- **e2e** (`tests/e2e/app-router/pages-to-app-routing.spec.ts`): after clicking the link, the test now calls `waitForAppRouterHydration(page)` so the App Router landing is asserted explicitly rather than only matching the new DOM.
- **unit** (`tests/pages-router-app-prefetch-detection.test.ts`): removed the unnecessary `await Promise.resolve()` tick and its misleading comment — since there is no `await` between `performNavigation` entry and the fast-path return, the test now asserts `window.location.assign` synchronously after `push()`, which also pins that the hard navigation cannot lose a race against the SPA path.

Verified locally: `vp test run tests/pages-router-app-prefetch-detection.test.ts` → 6/6 pass; `npx playwright test tests/e2e/app-router/pages-to-app-routing.spec.ts --project app-router` → 2/2 pass; `vp check` clean on both files.

Issue #1825, opened 2026-06-08
Build: SCSS imports from `node_modules` (tilde `~` and CSS modules) not resolved


james-elicx:
> *This issue was created by an agent auditing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/27113612207) (vinext `main` vs Next.js `v16.2.6`, 2026-06-08).*

## Problem

SCSS imports that resolve into `node_modules` are not handled: both the tilde (`~package`) import form and CSS-Module imports from within `node_modules` fail to build/render. Both suites are fully red (0 passing) across `sass` and `sass-embedded`.

## Affected Test Suites

- `test/e2e/app-dir/scss/npm-import-tilde/npm-import-tilde.test.ts` (2 failures — `sass` 1.54.0, `sass-embedded` 1.75.0; "should render the page")
- `test/e2e/app-dir/scss/nm-module/nm-module.test.ts` (2 failures — `sass` 1.54.0, `sass-embedded` 1.75.0; "should render the page")

## Estimated impact

~4 e2e failures across 2 suites.

## Notes

Distinct from #1343 (SCSS variables leaking to LightningCSS) and the closed #1450 (`url()` emission). This is specifically `node_modules`/tilde **import resolution** in the SCSS pipeline.

## Recommendation

1. Reproduce with an app importing `@import '~some-pkg/styles'` and a `.module.scss` that lives in `node_modules`.
2. Wire the SCSS resolver to resolve `~`-prefixed and bare `node_modules` specifiers the way Next.js/webpack's sass-loader does.

PR #1882, opened 2026-06-09
fix(scss): preprocess SCSS CSS-module composes deps via SassAwareFileSystemLoader

james-elicx:
## Failures

Both `{"sass":"1.54.0"}` and `{"sass-embedded":"1.75.0"}` variants of these three tests crash in production builds:

- `test/e2e/app-dir/scss/nm-module/nm-module.test.ts`
- `test/e2e/app-dir/scss/nm-module-nested/nm-module-nested.test.ts`
- `test/e2e/app-dir/scss/composes-external/composes-external.test.ts`

CI crash messages (from run 27200990128):

```
SyntaxError: [lightningcss minify] Invalid empty selector.
  1 | @import 'other3.scss';$var: red;._className_1344s_4 {
```

```
SyntaxError: [lightningcss minify] Invalid empty selector.
  1 | $var: red;._className_10j3d_2 {
```

## Root Cause

`postcss-modules`' built-in `FileSystemLoader` handles `composes: className from './other.module.scss'` by reading `other.module.scss` as **raw text** and piping it straight into PostCSS's CSS-module scoping plugins — **without Sass preprocessing**. For `.scss`/`.sass` files this leaves SCSS variables (`$var: red;`) and bare `@import 'file.scss'` directives verbatim in the composed-class CSS output. LightningCSS then fails during production minification with "Invalid empty selector" because it cannot parse raw SCSS syntax.

Vite exposes a `css.modules.Loader` option (undocumented in type definitions but honoured since Vite 5) that lets callers replace the default `FileSystemLoader`. The fix uses that hook.

### `packages/vinext/src/plugins/sass.ts`

Adds `SassAwareFileSystemLoader` — a drop-in replacement for postcss-modules' `FileSystemLoader` that, instead of reading files raw, calls Vite's `preprocessCSS(code, filename, config)` API for each `composes:` dependency. `preprocessCSS` runs Sass compilation **and** postcss-modules scoping in one shot using the same resolved config as the main Vite build, producing valid CSS that LightningCSS can safely minify.

- Uses Vite's exported `preprocessCSS` / `PreprocessCSSResult` directly (top-level import, no casts).
- When a `composes` dependency is not named `*.module.*` (e.g. `composes: x from './plain.css'`), the loader hands `preprocessCSS` a virtual `*.module.*` filename so the dependency is still scoped and its tokens extracted — postcss-modules treats every `composes` dependency as a CSS module regardless of filename, so this preserves the built-in loader's behaviour.
- Preprocessing failures are surfaced via `config.logger.warn` (composed classes would otherwise silently disappear); a missing/unreadable dependency file rejects, matching the original loader.

Also exports `setSassLoaderResolvedConfig(config)` so the `configResolved` hook can hand off the `ResolvedConfig` (which is needed by `preprocessCSS`) to the loader class.

### `packages/vinext/src/index.ts`

- Imports `SassAwareFileSystemLoader` and `setSassLoaderResolvedConfig` from `./plugins/sass.js`.
- In the `configResolved` hook: calls `setSassLoaderResolvedConfig(config)`.
- In the `config` hook return value: adds `css.modules.Loader: SassAwareFileSystemLoader` so Vite registers our custom loader for CSS-module files. The injection is skipped when the user set `css.modules: false` (a plugin's object return would otherwise override `false` in `mergeConfigRecursively` and silently re-enable CSS Modules) or supplied their own custom `Loader`.

## Tests

`tests/scss-composes.test.ts` ports the three failing Next.js fixtures as production-build regression tests (the LightningCSS minification step runs in these builds):

| Test | Next.js fixture | SCSS content |
|---|---|---|
| composes from an external `.module.scss` | `composes-external` | `$var: red;` |
| composes from a `.module.scss` in `node_modules` | `nm-module` | `node_modules/example/index.module.scss` with `$var: red;` |
| nested Sass `@import` inside a composed `node_modules` module | `nm-module-nested` | `@import 'other3.scss'; $var: red;` |

All three reproduce the exact `[lightningcss minify] Invalid empty selector` crash when the Loader injection is disabled. A fourth test proves parity for plain `.module.css` → `.module.css` composes and for composing from a non-`*.module.*` `.css` file (the loader now runs for every CSS module, so the pure-CSS path must keep working).

`sass` is added as a root devDependency so these tests — and the previously always-skipped `tests/scss.test.ts` — actually run in CI.

---

Part of #1825 (non-tilde CSS-module/composes cases; complements #1881)


james-elicx:
Addressed all five items from the review in c87a5a0d:

- **Tests (blocking)**: ported the three failing Next.js fixtures (`composes-external`, `nm-module`, `nm-module-nested`) as production-build regression tests in `tests/scss-composes.test.ts`, modeled on the `tests/css-modules-data-urls.test.ts` harness. Each builds a tmpdir Pages Router fixture through the real pipeline (LightningCSS minify included); all three reproduce the exact `[lightningcss minify] Invalid empty selector` crash when the Loader injection is disabled. Since `sass` was not installed anywhere in the workspace (the existing `tests/scss.test.ts` silently skipped everywhere, including CI), `sass` is now a root devDependency so these tests — and `tests/scss.test.ts` — actually run; the corresponding redundant `knip.ts` ignore entry is dropped.
- **`css.modules` re-enabling / Loader clobbering**: the `config` hook now skips the `css.modules.Loader` injection when the user set `css.modules: false` or provided their own custom `Loader`, with a comment explaining the `mergeConfigRecursively` object-over-`false` behaviour.
- **Plain `.module.css` composes parity**: covered by a fourth test composing from `./other.module.css` (no Sass involved). It also covers `composes: x from './plain.css'` — postcss-modules treats every `composes` dependency as a CSS module regardless of filename, but Vite's pipeline only scopes `*.module.*` names, so the loader now hands `preprocessCSS` a virtual `*.module.*` filename for such deps (same directory, so relative/Sass resolution is unaffected). This parity gap existed in the previous revision; the test passes against the default `FileSystemLoader` too, proving baseline parity.
- **Silent failure mode**: the bare `catch {}` is gone — preprocessing failures now log `config.logger.warn` with the dependency path and the underlying error, and a missing/unreadable dependency file rejects (failing the build) exactly like postcss-modules' built-in loader.
- **Minor items**: switched to a top-level `import { preprocessCSS, type PreprocessCSSResult } from "vite"` (the lazy dynamic-import tri-state and the `Record<string, unknown>` casts are gone), and the callback-style `fs.readFile` fallback is now `fs.promises`.

`tests/scss-composes.test.ts` (4/4), `tests/scss.test.ts` (2/2, no longer skipped), `tests/css-modules-data-urls.test.ts` (2/2), `tests/sass-options.test.ts` (16/16), `tests/postcss-resolve.test.ts` (13/13), `vp check`, and `vp run knip` are all clean.

james-elicx:
Rebased onto main to resolve conflicts with #1881 (tilde importer now lives alongside the composes Loader in the same `css` block). The review items above are now in 05001fbd; the `css.modules` guard moved to a typed `cssModulesOverride` const so the conditional spread doesn't disturb the config literal's contextual typing.

james-elicx:
Follow-up in 951373ee: the first full CI run (the PR was previously conflicting with main, so pull_request CI had never actually executed) surfaced that installing `sass` makes Vite's `SassPreprocessorOptions` resolve to the concrete sass types — previously `any` — which broke `npx tsc` on the tilde-importer's `importers: unknown[]` array from #1881 (E2E webServer steps run `tsc` first). `baseOpts` is now typed `SassPreprocessorOptions` with an explicit `importers` cast; `npx tsc -p packages/vinext/tsconfig.json` and `vp check` are clean.

james-elicx:
Addressed the four non-blocking observations from the latest review in c3bff5d6:

**1. Module-level `_resolvedConfig` singleton → per-instance binding.** Replaced the module-level singleton with `createSassAwareFileSystemLoader()`, a factory called once per vinext plugin instance. postcss-modules constructs the Loader itself with a fixed `(root, plugins, fileResolve)` signature, so the resolved config can't be a constructor argument — the factory instead returns a `Loader` subclass with the config captured in the factory closure, plus a `setResolvedConfig` bound to that same closure (called from `configResolved`). Multiple builds in one process now each preprocess `composes` deps with their own root/sass options/scoped-name generator. The base class is no longer exported; `index.ts` injects `sassComposesLoader.Loader` per instance.

**2. Extensionless `composes` deps now scoped.** The virtual rename now appends `.module.css` when the dep has no extension (the built-in `FileSystemLoader` piped extensionless deps through the scoping plugins as CSS, so plain CSS is the parity-preserving interpretation). Covered by a new `composes: extlessClass from './extless'` case in the parity test — it fails without the fix (tokens silently dropped) and passes with it.

**3. `url()` in composed deps pinned by test.** Added a `url('./dot.svg')` reference to the composed `.module.css` dep in the parity fixture. Empirically, `preprocessCSS` does **not** rewrite or emit the asset (it has no plugin context for `fileToUrl`), and the composed-dep CSS also bypasses the main pipeline's url rewriter — the reference passes through verbatim. I verified the built-in `FileSystemLoader` behaves identically by building the same fixture with vanilla Vite (no vinext): both emit `url(./dot.svg)` unchanged and emit no asset. The test pins exact parity: the url survives verbatim and is never mangled into an unresolved `__VITE_ASSET__` placeholder or dropped, so any pipeline change here surfaces as a test failure.

**4. Hard-fail when neither `sass` nor `sass-embedded` resolves.** Leaving as-is intentionally: `sass` is now a pinned root devDependency, so a fresh install always provides it — a stale `node_modules` is an environment problem that breaks plenty of other tests too, and CI always installs fresh. The old skip-when-absent behaviour would silently drop coverage for exactly the regression this PR fixes.

Targeted runs are green locally: `tests/scss-composes.test.ts` (4/4, including the new extensionless + url() assertions), `tests/scss.test.ts` (2/2), `tests/sass-options.test.ts` (23/23); `vp check` and `knip` clean.

james-elicx:
Addressed the two new observations in 08075ad0 with the suggested doc comment: the `SassAwareFileSystemLoader` docblock now documents the per-subtree recursion boundary — sibling-branch deps are inlined once per subtree (identical duplicates collapsed by LightningCSS during minification), and circular `composes` chains are not short-circuited across the `preprocessCSS` boundary (invalid input; webpack/Next.js errors on them too, so no cycle bookkeeping is layered on). No behaviour change — comment only; targeted tests (29/29) and `vp check` still green.

james-elicx:
Addressed the latest round in 10242fb3:

1. **Docstring overstating the null-config warn** — fixed: the class doc and `getResolvedConfig` doc now state the no-config path silently returns empty tokens (only reachable outside vinext's `configResolved` wiring).
2. **Catch-all swallowing real errors** — implemented the suggested narrowing: only a missing Sass implementation ("Preprocessor dependency ... not found") is downgraded to a logged warning; any other `preprocessCSS` error (e.g. a Sass syntax error in a composed dep) now re-throws and fails the build, matching the built-in `FileSystemLoader`. Added a regression test (`fails the build when a composed dependency has a Sass error") that asserts the build rejects.
3. **`config` hook can't see other plugins' `config` returns** — acknowledged as an inherent Vite limitation, not actionable; no change.

Targeted tests 30/30, `vp check` clean.

package.json
```diff
@@ -38,6 +38,7 @@
     "playwright": "catalog:",
     "react": "catalog:",
     "react-dom": "catalog:",
+    "sass": "1.100.0",
```
james-elicx: huh? why do you need to add this dep here?

you said "tests/scss.test.ts" was always skipped which is why this needed to be added, but there's no unskip in this pr ??

james-elicx: if this stays, it should at least be added to the workspace catalog instead

james-elicx: The dep is needed by the new `tests/scss-composes.test.ts` — its fixtures are real SCSS, so the production builds in those tests require `sass` to be installed.

On the "no unskip" point: the skip in `tests/scss.test.ts` was runtime-conditional (`await import("sass")` → `describe` vs `describe.skip`), so installing the dep *was* the unskip — nothing in the test source needed to change, which is why it didn't show in the diff. I've now removed that dynamic-skip machinery in 8ed84756 since `sass` is guaranteed to be present, so the suite unconditionally runs and the unskip is visible.

Also moved `sass` into the workspace catalog in the same commit.

PR #1883, opened 2026-06-09
refactor(isr): centralise HIT/STALE/MISS + Cache-Control decision in isr-decision.ts

### LINKED ISSUE #1783: Add a single ISR cache-decision module above the handler seam
## Summary

The low-level cache seams in this repo are good. `CacheHandler` (`shims/cache.ts:144`) is a clean key-value store, `CdnCacheAdapter` (`shims/cdn-cache.ts:78`) is a clean serving-strategy seam, and `isr-cache.ts` sits above both and talks only through `getCdnCacheAdapter()`. That layering is right and this issue doesn't touch it.

The problem is one layer up. The policy decision — "is this HIT, STALE, or MISS, should I schedule a regeneration, and which `Cache-Control` do I stamp" — gets re-derived in four places:

- `server/app-page-cache.ts:321`
- `server/app-route-handler-cache.ts:81`
- `server/pages-page-data.ts:537-605`
- `server/dev-server.ts:818-1418` (fully inline; doesn't reuse the app-page version)

And the `Cache-Control` strings that `cache-control.ts` exists to own leak out as hardcoded literals:

- `dev-server.ts:789` — duplicates `NEVER_CACHE_CONTROL` (`cache-control.ts:3`)
- `dev-server.ts:837`, `:994`, `:1309` — unbounded SWR string by hand
- `pages-page-response.ts:528` — duplicates `NEVER_CACHE_CONTROL` (and this file already imports `cache-control.ts`)
- `entries/pages-server-entry.ts:908` — same literal in generated code

## Where the decision lives now vs. after

```mermaid
flowchart TB
  subgraph BEFORE["before — policy re-derived, strings leak"]
    AP["app-page-cache<br/>:321"]
    AR["route-handler-cache<br/>:81"]
    PP["pages-page-data<br/>:537"]
    DV["dev-server inline<br/>:818"]
    CC["cache-control.ts<br/>(canonical strings)"]
    ISR["isr-cache · CacheHandler"]
    AP --> ISR
    AR --> ISR
    PP --> ISR
    DV --> ISR
    DV -. hardcoded literal .-> CC
    PP -. hardcoded literal .-> CC
  end
  subgraph AFTER["after — one decision, strings centralized"]
    A2["app-page dispatch"]
    R2["route-handler dispatch"]
    P2["pages dispatch"]
    D2["dev-server"]
    DEC["decideIsr()<br/>HIT/STALE/MISS · regen · Cache-Control"]
    CC2["cache-control.ts"]
    ISR2["isr-cache · CacheHandler"]
    A2 --> DEC
    R2 --> DEC
    P2 --> DEC
    D2 --> DEC
    DEC --> CC2
    DEC --> ISR2
  end
```

## Proposed change

1. Add `server/isr-decision.ts` (name open) exposing one entry that takes the request, the route's cache policy, and the current cache read, and returns a decision: disposition (HIT/STALE/MISS), whether to schedule a regeneration, and the `Cache-Control` to apply.
2. Route every `Cache-Control` string through `cache-control.ts`. No literals in `dev-server.ts`, `pages-page-response.ts`, or `pages-server-entry.ts`.
3. Point the four call sites at the new module instead of each re-deriving the branch logic.

## Why this one needs care — it is not a guaranteed no-op

This is the candidate that can actually move the compatibility number, and I want to be honest about why rather than wave it through.

The four call sites feed different inputs today. Take `dev-server.ts:837`: it hardcodes `s-maxage=${revalidateSecs}, stale-while-revalidate` — an unbounded SWR window. If you route that through `buildRevalidateCacheControl(revalidateSecs, expire)` and `expire` is known, the helper adds a bound and you get `s-maxage=60, stale-while-revalidate=240`. That is a different header on the wire. The same trap applies to any leaked `NEVER_CACHE`/`NO_STORE` literal whose call site policy doesn't match the helper's output.

So the work here isn't "centralize and done." Each substitution needs a small equivalence proof: the literal a call site emits today must equal what the helper produces for that site's inputs, or the change is deliberate and documented. Skip that step and you ship a behavioral drift that surfaces three days later in the nightly run.

Cache-sensitive deploy suites that would catch it: `app-static` (currently 66/94), `prerender` (41/63), `revalidate-reason`, `on-request-error/isr`, `custom-cache-control`.

## Tests that guard it

- `tests/cache-control.test.ts` — pins exact output strings, e.g. `buildRevalidateCacheControl(60, 300)` -> `"s-maxage=60, stale-while-revalidate=240"`. This is your equivalence oracle.
- `tests/isr-cache.test.ts`
- `tests/app-page-cache.test.ts` (34 KB)
- `tests/app-route-handler-cache.test.ts`
- `tests/pages-page-data.test.ts`

## Scope

In scope: the four ISR policy call sites and the leaked `Cache-Control` literals.

Out of scope: the `CacheHandler` and `CdnCacheAdapter` seams (already clean), the `cache-proof.*` family (a separate artifact-reuse subsystem, no runtime ISR coupling — different problem, different issue).

## Acceptance criteria

- [ ] One module owns the HIT/STALE/MISS + regen + `Cache-Control` decision
- [ ] No hardcoded `Cache-Control` literals outside `cache-control.ts`
- [ ] A documented per-call-site note showing each migrated path emits the same header it did before (or a deliberate, justified change)
- [ ] Full Vitest suite green; cache-sensitive deploy suites no worse than baseline

## Risk

Highest of the three. The centralization is sound, but it only stays a no-op if the equivalence check is done call site by call site. Treat any header change as a real change, not a cleanup.

---

*Filed from an architecture review. Companion issues: unify the Pages Router request pipeline and extract the Pages render orchestrator.*

Divkix:
Closes #1783

## What

Adds `server/isr-decision.ts` as the single owner of the ISR cache policy decision. All four call sites that previously re-derived HIT/STALE/MISS disposition, background-regen scheduling, and `Cache-Control` string selection now delegate to `decideIsr()` or its MISS helpers.

Also eliminates all hardcoded `Cache-Control` literals outside `cache-control.ts` — raw strings like `"private, no-cache, no-store, max-age=0, must-revalidate"` that were copy-pasted across `dev-server.ts`, `pages-page-response.ts`, and `pages-page-handler.ts`.

## Call sites migrated

- `server/app-page-cache.ts` — removed local `buildAppPageCacheControl` wrapper
- `server/app-route-handler-response.ts` — removed local `buildRouteHandlerCacheControl`; cached responses use `decideIsr(kind:"app-route")`, fresh MISS responses use `buildAppRouteMissIsrCacheControl` (preserves the `revalidate=0`→NEVER and `Infinity`→STATIC gates)
- `server/pages-page-data.ts` — `buildPagesCacheResponse` delegates to `decideIsr(kind:"pages")`
- `server/dev-server.ts` — all four inline literals replaced

## Deliberate behaviour change

Dev STALE responses now emit `s-maxage=0, stale-while-revalidate` (matching prod Pages Router and the canonical `buildCachedRevalidateCacheControl` helper) instead of `s-maxage=<revalidate>, stale-while-revalidate`. The old value incorrectly told downstream caches that a stale-served payload was freshly cacheable — a dev/prod parity gap that was masked by the duplication.

## Equivalence

Every other migrated path emits the same header as before. The full per-call-site equivalence table is in the `isr-decision.ts` module doc comment.

## Tests

- `tests/isr-decision.test.ts` — 33 new unit tests covering all dispositions, router kinds, `0`/`Infinity` special cases, metadata fallback semantics, and the MISS helpers
- All existing cache-related tests pass unchanged: `app-page-cache`, `app-route-handler-cache`, `app-route-handler-response`, `pages-page-data`, `isr-cache`, `cache-control`, `features`

### DIFF
--- a/packages/vinext/src/server/app-route-handler-response.ts
+++ b/packages/vinext/src/server/app-route-handler-response.ts
@@ -1,10 +1,6 @@
 import type { CachedRouteValue, CacheControlMetadata } from "vinext/shims/cache";
-import {
-  applyCdnResponseHeaders,
-  buildCachedRevalidateCacheControl,
-  NEVER_CACHE_CONTROL,
-  STATIC_CACHE_CONTROL,
-} from "./cache-control.js";
+import { applyCdnResponseHeaders } from "./cache-control.js";
+import { decideIsr, buildAppRouteMissIsrCacheControl } from "./isr-decision.js";
 import {
   MIDDLEWARE_HEADER_PREFIX,
   MIDDLEWARE_NEXT_HEADER,
@@ -46,28 +42,6 @@ function hasMiddlewareHeader(headers: Headers): boolean {
     if (key.startsWith(MIDDLEWARE_HEADER_PREFIX)) return true;
   }
   return false;
-}
-
-function buildRouteHandlerCacheControl(
-  cacheState: BuildRouteHandlerCachedResponseOptions["cacheState"],
-  revalidateSeconds: number,
-  expireSeconds?: number,
-): string {
-  if (revalidateSeconds === 0) {
-    // A cached response is never produced for revalidate = 0 (the ISR write
-    // path skips it), so only the HIT/STALE->fresh rewrite can arrive here
-    // with a 0 value, via applyRouteHandlerRevalidateHeader. In all such
-    // cases the author opted out of caching entirely.
-    return NEVER_CACHE_CONTROL;
-  }
-
-  if (revalidateSeconds === Infinity) {
-    // revalidate = false / Infinity means "cache indefinitely" — emit the
-    // same static Cache-Control used by pages, not a dynamic SWR value.
-    return STATIC_CACHE_CONTROL;
-  }
-
-  return buildCachedRevalidateCacheControl(cacheState, revalidateSeconds, expireSeconds);
 }
 
 export function applyRouteHandlerMiddlewareContext(
@@ -115,21 +89,18 @@ export function buildRouteHandlerCachedResponse(
     }
   }
   setCacheStateHeaders(headers, options.cacheState);
-  const revalidateSeconds = options.cacheControl?.revalidate ?? options.revalidateSeconds;
-  const expireSeconds =
-    options.cacheControl === undefined
-      ? undefined
-      : (options.cacheControl.expire ?? options.expireSeconds);
   // HIT/STALE served from the origin store: route the cache header through the
   // CDN adapter (default: identical single Cache-Control). Edge adapters never
   // reach this path because their get() returns null.
-  applyCdnResponseHeaders(headers, {
-    cacheControl: buildRouteHandlerCacheControl(
-      options.cacheState,
-      revalidateSeconds,
-      expireSeconds,
-    ),
+  const { cacheControl } = decideIsr({
+    hasUsableValue: true,
+    isStale: options.cacheState === "STALE",
+    kind: "app-route",
+    revalidateSeconds: options.revalidateSeconds,
+    expireSeconds: options.expireSeconds,
+    cacheControlMeta: options.cacheControl,
   });
+  applyCdnResponseHeaders(headers, { cacheControl });
 
   return new Response(options.isHead ? null : cachedValue.body, {
     status: cachedValue.status,
@@ -145,8 +116,11 @@ export function applyRouteHandlerRevalidateHeader(
 ): void {
   // Fresh (MISS) response: route through the CDN adapter so edge adapters emit
   // CDN-Cache-Control + Cache-Tag while the default emits a single Cache-Control.
+  // Uses buildAppRouteMissIsrCacheControl so the revalidate=0→NEVER and
+  // Infinity→STATIC gates apply, and expireSeconds is used as the direct route
+  // config ceiling (not a per-entry metadata fallback).
   applyCdnResponseHeaders(response.headers, {
-    cacheControl: buildRouteHandlerCacheControl("HIT", revalidateSeconds, expireSeconds),
+    cacheControl: buildAppRouteMissIsrCacheControl(revalidateSeconds, expireSeconds),
     tags,
   });
 }
--- a/packages/vinext/src/server/app-route-handler-response.ts
+++ b/packages/vinext/src/server/app-route-handler-response.ts
@@ -93,8 +93,7 @@ export function buildRouteHandlerCachedResponse(
   // CDN adapter (default: identical single Cache-Control). Edge adapters never
   // reach this path because their get() returns null.
   const { cacheControl } = decideIsr({
-    hasUsableValue: true,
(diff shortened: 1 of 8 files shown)


packages/vinext/src/server/app-route-handler-response.ts
```diff
@@ -115,21 +89,18 @@ export function buildRouteHandlerCachedResponse(
...
   // reach this path because their get() returns null.
-  applyCdnResponseHeaders(headers, {
-    cacheControl: buildRouteHandlerCacheControl(
-      options.cacheState,
-      revalidateSeconds,
-      expireSeconds,
-    ),
+  const { cacheControl } = decideIsr({
+    hasUsableValue: true,
+    isStale: options.cacheState === "STALE",
```
james-elicx: we could just pass the cache state through to this function and check it internally instead of doing the check out here

packages/vinext/src/server/isr-decision.ts
```diff
@@ -0,0 +1,220 @@
...
+ */
+type IsrPolicyKind = "app-page" | "app-route" | "pages" | "dev";
+
+type DecideIsrOptions = {
+  /**
+   * True when the cache returned a value that can be forwarded to the client
+   * (the content guards — kind-mismatch, empty body, query-variant-unproven —
+   * have already passed). MISS = false; HIT or STALE = true.
+   */
+  hasUsableValue: boolean;
```
james-elicx: does that mean this can be inferred from the state if we pass the state instead?

packages/vinext/src/server/isr-decision.ts
```diff
@@ -0,0 +1,220 @@
...
+ * | pages STALE (no expire)               | buildCachedRevalidateCacheControl("STALE", r)    | same → s-maxage=0  | no       |
+ * | dev HIT (getStaticProps)               | s-maxage=${secs}, stale-while-revalidate         | same               | no       |
+ * | dev STALE (getStaticProps)             | s-maxage=${secs}, stale-while-revalidate         | s-maxage=0, ...    | **yes**  |
+ * | dev MISS/regen (getStaticProps)        | s-maxage=${secs}, stale-while-revalidate         | same               | no       |
+ * | dev gssp default (no-store)            | private, no-cache, no-store, max-age=0...        | same (NEVER_CACHE) | no       |
+ * | dev nonce (no-store)                   | no-store, must-revalidate                        | same (NO_STORE)    | no       |
+ * | pages-page-response scriptNonce        | no-store, must-revalidate                        | same (NO_STORE)    | no       |
+ * | pages-page-response gssp default       | private, no-cache, no-store...                   | same (NEVER_CACHE) | no       |
+ * | pages-page-handler _next/data default  | private, no-cache, no-store...                   | same (NEVER_CACHE) | no       |
```
james-elicx: this seems accidentally included? a before/after/changed comparison becomes irrelevant as soon as this is merged

Issue #1237, opened 2026-05-15
Error: react-dom/server is not supported in React Server Components.


eashish93:
I can reproduce this on vinext 0.0.50 with `@react-email/render@1.4.0` and React 19.2.6.

Root cause appears to be that `@react-email/render`'s workerd/edge entry dynamically imports `react-dom/server.edge`, but vinext resolves that import under the `react-server` condition inside an App Route handler. React then resolves `react-dom/server.edge` to `react-dom/server.react-server.js`, which throws:

`react-dom/server is not supported in React Server Components.`

A local workaround is:

```ts
// vite.config.ts
resolve: {
  alias: [
    {
      find: 'react-dom/server.edge',
      replacement: path.resolve(__dirname, 'node_modules/react-dom/server.edge.js'),
    },
  ],
}

**Comment by eashish93** [comment]
**Update — the alias workaround above is incomplete.** I verified this against a fresh repro on `vinext@0.0.50`.

Aliasing `react-dom/server.edge` → absolute path gets past the *outer* `server.react-server.js` throw, but the inner `require('react')` inside `react-dom/cjs/react-dom-server.edge.development.js:8817` is still resolved by the RSC env's resolver with the `react-server` condition. It picks the `react-server` variant of `react`, which exports `__SERVER_INTERNALS_…` instead of `__CLIENT_INTERNALS_…`. Then at line 9533 `ReactSharedInternals` is `undefined`, and the next write throws:

```
TypeError: Cannot set properties of undefined (setting 'recentlyCreatedOwnerStacks')
    at resetOwnerStackLimit (react-dom/cjs/react-dom-server.edge.development.js:4534:32)
    at render (@react-email/render/src/edge/render.tsx:18:22)
    at sendTemplatedEmail (lib/email.ts:23:26)
```

(Same root cause, different symptom than the original `server is not supported in React Server Components` error — it surfaces depending on whether the optimizer happened to already cache an optimized `server.edge` or not.)

---

## Actual working workaround

The clean fix is to load App Route handlers in the **SSR** env instead of the RSC env, so `react-dom/server.edge` resolution doesn't pick up the `react-server` condition. The SSR entry statically imports each route module and exposes a `loadAppRouteHandler(pattern)` lookup; the patched RSC dispatcher fetches the handler via `import.meta.viteRsc.loadModule(\"ssr\", \"index\")` before invoking it.

Verified end-to-end against `vinext@0.0.50`:

[...]

Issue #1246, opened 2026-05-15
App Route handlers can't both render react-dom/server.edge and call 'use server' functions (RSC vs SSR env conflict)


eashish93:
## Summary

Calling a `'use server'` function directly from an App Route handler (`app/api/**/route.ts`) and rendering `react-dom/server.edge` from inside the same route handler are mutually exclusive in vinext today. Whichever Vite environment the route handler loads in breaks one or the other.

This is a follow-up to #882 / #1237 (react-dom/server in route handlers) — fixing those bugs by loading route handlers in the SSR env introduces a new bug: server functions throw `noServerCall`.

## Repro

A typical sign-in flow that does both — call a server action to provision the account, then send a welcome email:

```ts
// app/api/auth/session/route.ts
import { createAccount } from '@/lib/actions/accounts/createAccount';
import { sendPaidWelcomeEmail } from '@/lib/email';

export async function POST(req: NextRequest) {
  const decoded = await firebaseAdmin.verifyIdToken(...);
  await createAccount(decoded);          // <- 'use server' function
  await sendPaidWelcomeEmail({...});     // <- uses @react-email/render -> react-dom/server.edge
  ...
}
```

```ts
// lib/actions/accounts/createAccount.ts
'use server';
export async function createAccount(decoded: DecodedIdToken) { ... }
```

```ts
// lib/email.ts
import { render } from '@react-email/render';
// render() does: import("react-dom/server.edge").catch(() => import("react-dom/server"))
```

Reproduces on `vinext@0.0.50`, `@vitejs/plugin-rsc@0.5.23`, `@react-email/render@1.4.0`, `react@19.2.6`, with the Cloudflare plugin (`viteEnvironment: { name: 'rsc', childEnvironments: ['ssr'] }`).

### Path A — route handler loads in the RSC env (current vinext default)

`react-dom/server.edge` resolution under `react-server` condition picks `server.react-server.js`, which throws:

```
Error: react-dom/server is not supported in React Server Components.
    at runInRunnerObject (workers/runner-worker/index.js:107:3)
    at render (@react-email/render/src/edge/render.tsx:13:26)
    ...
    at Object.POST [as handlerFn] (app/api/auth/session/route.ts:30:11)
    at runAppRouteHandler (vinext/src/server/app-route-handler-execution.ts:127:20)
```

[...]

PR #1890, opened 2026-06-10
fix(entries): load app route handlers via SSR loadAppRouteHandler

digwis:
## Problem

App Router route handlers (e.g. `app/api/<segment>/route.ts`) are imported and registered by the RSC entry's manifest, which runs under the `react-server` condition. Modules that need `react-dom/server` to construct `NextResponse` / `Response` bodies (notably any handler that returns a streaming or DOM-typed body) fail to evaluate in that environment.

The current code eagerly pulls `route.routePath` modules into the RSC entry via `getLazyLoaderVar`, but those modules still get parsed under `react-server` and crash at import time when they reach for `react-dom/server`.

## Fix

Move route handler loading to the SSR environment:

- **`packages/vinext/src/entries/app-ssr-entry.ts`** — `generateSsrEntry` now accepts the resolved `AppRoute[]`, statically imports every `route.routePath` into a `__appRouteHandlers` pattern-indexed map, and exports a `loadAppRouteHandler(pattern)` resolver.
- **`packages/vinext/src/entries/app-rsc-manifest.ts`** — drops the eager `getLazyLoaderVar(route.routePath)` registration in `registerRouteModules`; the emitted `__loadRouteHandler` thunk now calls `import.meta.viteRsc.loadModule("ssr", "index").loadAppRouteHandler(pattern)` on demand.
- **`packages/vinext/src/index.ts`** — passes the resolved routes through to `generateSsrEntry`.

Result: route handler modules are evaluated in the SSR environment (where `react-dom/server` is available) while still being loaded lazily on first request from the RSC entry.

## Verification

- `npx tsc --noEmit -p packages/vinext/tsconfig.json` — clean
- `npx vp test run --project unit` on the affected suites (`app-route-handler-response`, `app-route-handler-dispatch`, `app-route-handler-policy`, `app-route-handler-trailing-slash`, `app-route-handler-execution`, `app-route-handler-runtime`, `app-route-handler-cache`, `route-classification-injector`, `app-rsc-route-matching`) — 86/86 pass

## Notes

- The repro / exact failing handlers are downstream of the issue: any `route.ts` that imports `react-dom/server` (directly or transitively via `next/server`) under an App Router tree.
- Mirrors what Next.js does in production: route handler modules live in the SSR environment and the RSC layer only knows their patterns.

### DIFF
--- a/packages/vinext/src/entries/app-rsc-manifest.ts
+++ b/packages/vinext/src/entries/app-rsc-manifest.ts
@@ -142,13 +142,14 @@ function registerRouteModules(routes: AppRoute[], imports: ImportAllocator): voi
     // reached via lazy `{ load }` sources in generateStaticParamsMap, resolved
     // on demand at prerender time.
     if (route.pagePath) imports.getLazyLoaderVar(route.pagePath);
-    // Route handlers are always lazy: they are never referenced by
-    // generateStaticParamsMap (buildGenerateStaticParamsEntries sources only
-    // from layouts + page, never route.routePath), so unlike dynamic-route
-    // pages they have no module-load-time consumer. (Next.js route handlers can
-    // export generateStaticParams for prerendering, but vinext does not wire
-    // that into the map yet — a separate gap, unaffected by lazy loading.)
-    if (route.routePath) imports.getLazyLoaderVar(route.routePath);
+    // Route handlers (e.g. `app/api/<segment>/route.ts`) are NOT pre-registered
+    // as eager imports or lazy loaders here. Instead, the SSR entry generated
+    // by `app-ssr-entry.ts` statically imports every route handler and exposes
+    // a `loadAppRouteHandler(pattern)` resolver. The RSC manifest emits a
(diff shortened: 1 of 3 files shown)


digwis:
Closing in favor of a discussion issue — see the linked issue for the architectural discussion.

This PR moves route handler loading to the SSR env, which fixes #1237 (`react-dom/server.edge` in route handlers) but reintroduces the regression tracked in #1246 (calling `'use server'` from a route handler now hits `noServerCall`).

Rather than land a partial fix and ship the conflict, I'd like to align with the maintainers on the right architecture (likely a dedicated `route-handler` Vite env per #1246's option 3) before pushing more code.

I'll keep the branch `fix/route-handler-lazy-loader` in [digwis/vinext](https://github.com/digwis/vinext/tree/fix/route-handler-lazy-loader) as a reference for the implementation detail of how an SSR-side handler registry + `import.meta.viteRsc.loadModule("ssr","index").loadAppRouteHandler(pattern)` thunk would look.

PR #1906, opened 2026-06-11
fix(app-router): preserve unstable_catchError boundaries

NathanDrake2406:
## Summary

Ports the Next.js `unstable_catchError` E2E contract and fixes the vinext runtime paths that were masking userland error boundaries.

The upstream deploy-suite initially failed three cases in `test/e2e/app-dir/catch-error/catch-error.test.ts`: server-component `unstable_retry`, Pages Router reset, and Pages Router retry. After this change the same upstream file passes all five cases.

## What changed

- Added a faithful Playwright port of the upstream catch-error suite, including client component reset/retry, server component retry, Pages Router reset, and the exact Pages Router `unstable_retry()` error message.
- Reworked `next/error`'s `unstable_catchError` shim to follow Next.js's boundary shape: fallback is rendered as a component, Pages Router detection is context-based, App Router retry refreshes through `AppRouterContext`, and pathname changes reset caught state.
- Made speculative page/layout probes respect `connection()` semantics: code after `await connection()` no longer executes while classifying or probing a route, matching Next.js's prerender behavior.
- Stopped rerendering App Router HTML with `global-error` after an RSC shell exists. React may serialize errors in Flight that are still caught by client boundaries such as `unstable_catchError`; shell-level SSR failures still go through the existing recovery path.

## References

- Upstream test contract: https://github.com/vercel/next.js/blob/v16.2.6/test/e2e/app-dir/catch-error/catch-error.test.ts
- Next.js `unstable_catchError` implementation: https://github.com/vercel/next.js/blob/v16.2.6/packages/next/src/client/components/catch-error.tsx
- Next.js error boundary context implementation: https://github.com/vercel/next.js/blob/v16.2.6/packages/next/src/client/components/error-boundary.tsx
- Next.js `connection()` implementation: https://github.com/vercel/next.js/blob/v16.2.6/packages/next/src/server/request/connection.ts
- Next.js `connection()` docs: https://github.com/vercel/next.js/blob/canary/docs/01-app/03-api-reference/04-functions/connection.mdx

## Verification

- `vp check`
- `vp test run tests/app-page-stream.test.ts tests/app-page-render.test.ts tests/shims.test.ts tests/app-page-execution.test.ts tests/app-page-probe.test.ts` — 1173 passed
- `PLAYWRIGHT_PROJECT=catch-error pnpm run test:e2e tests/e2e/catch-error/catch-error.spec.ts` — 5 passed
- `PLAYWRIGHT_PROJECT=app-router pnpm run test:e2e tests/e2e/app-router/error-handling.spec.ts` — 10 passed
- `vp env exec --node 24 ./scripts/run-nextjs-deploy-suite.sh /Users/nathan/Projects/vinext/.refs/nextjs-v16.2.6 --retries 0 -c 1 --debug test/e2e/app-dir/catch-error/catch-error.test.ts` — upstream file passed, 5/5


NathanDrake2406:
**Assessment: `usePathname` vs upstream `useUntrackedPathname`** (ask-bonk review, non-blocking)

The reviewer correctly notes that Next.js upstream uses `useUntrackedPathname()` (which returns `null` during missing-params shell prerender and does not register a tracked render dependency), while the `error.tsx` shim in this PR uses `usePathname()` (navigation.ts:1478, returns server-context pathname or `"/"`).

No code change is needed in this PR because:

1. **`useUntrackedPathname` does not exist in the vinext codebase** — implementing it would require adding a new hook to `navigation.ts`, which is scope-creep beyond this PRs goal of porting the `unstable_catchError` boundary shape and the E2E contract.
2. **The divergence is pre-existing**, not introduced by this PR. The `usePathname` import was already present in the prior `error.tsx` implementation.
3. **The reviewer explicitly confirms** "for normal navigation the pathname-reset behavior works correctly, so this is fine for the cases this PR targets" and calls it "worth a follow-up."

This should be tracked as a separate issue/PR: add a `useUntrackedPathname` hook to `navigation.ts` that mirrors Next.jss behavior (returns `null` during missing-params prerender, avoids tracked render dependency), then switch `error.tsx` to use it.

NathanDrake2406:
**Assessment: `playwright.config.ts` build command inconsistency** (ask-bonk review, non-blocking)

The reviewer notes that the `catch-error` project uses `npx vp run vinext#build` while sibling prod-build projects (`pages-router-prod`, `static-export`, `standalone-output`, `root-layout-redirect`) use `npx tsc -p ../../../packages/vinext/tsconfig.json`. Both produce `dist/cli.js`.

No code change is needed in this PR because:

1. **`vp run vinext#build` is the correct/preferred approach** per the AGENTS.md convention (Production Builds section: "Prefer `vp` for day-to-day local work"). Standardizing the sibling projects to use `vp` instead of `tsc` would be the right direction, but that change is outside the scope of this PR and would need verification across CI.
2. **The reviewer acknowledges this**: "`vp run vinext#build` is arguably more correct per AGENTS.md."
3. **Functional equivalence**: Both commands produce the same `dist/cli.js` output. The catch-error project uses the correct approach; the inconsistency is cosmetic.

Standardizing all prod-build projects to `vp run vinext#build` is a separate cleanup task, not a fix for this PR.

NathanDrake2406:
**Assessment: module-level `let hasThrown` in server-component/page.tsx** (ask-bonk review, non-blocking)

The reviewer correctly notes that `let hasThrown = false` at `tests/fixtures/global-not-found-basic/app/server-component/page.tsx:12` is cross-request shared state that persists across requests within a single worker/process.

No code change is needed because:

1. **This mirrors the upstream fixture** — Next.jss `catch-error.test.ts` fixture uses the same pattern intentionally: the retry test relies on throw-once-then-recover behavior within a single E2E test run.
2. **Fixture-only code** — This is a test fixture, not a reusable pattern. The flag is precisely to prevent misuse, as the reviewer intends.
3. **The catch-error E2E spec starts a fresh server per test** (Playwright `webServer` with `reuseExistingServer: false` at `playwright.config.ts:80`), so each test gets an isolated module instance anyway.

This is correctly flagged as a caution for anyone copying the fixture pattern, not as something to fix.

NathanDrake2406:
**Correction: `usePathname` vs `useUntrackedPathname`**

The reviewer correctly flags that the prior `error.tsx` on `main` imported `appRouterInstance`/`isNextRouterError` and the wrapper called no hooks at all (`git show origin/main:packages/vinext/src/shims/error.tsx`). The `usePathname` call in `error.tsx:243` is introduced by this PR, not pre-existing.

Acknowledged — tracked as a follow-up to add a `useUntrackedPathname` hook (that returns `null` during the missing-params shell prerender and does not register a tracked render dependency) and switch to it for full parity.

PR #1908, opened 2026-06-11
fix(app-router): recover SSR shell render errors via __next_error__ document

NathanDrake2406:
## What this changes

Scoped down from the original next/dynamic CSS PR after review (see comments): this branch now carries only the SSR shell-error recovery and the shim resolver fix, as two clean commits. The dynamic-CSS ordering mechanism was removed and will return redesigned (follow-up issue linked below).

1. **SSR shell-error recovery.** When the HTML (Fizz) render rejects during SSR with an error that did not originate in the RSC render (no `digest`), and the app has no custom `global-error.tsx`, `handleSsr` serves the default `__next_error__` error document carrying the original flight payload and bootstrap module. The browser detects the marker and re-renders the real tree with `createRoot` instead of hydrating — Next.js shell-error semantics. Local `error.tsx` boundaries still win: they ship inside the flight payload and catch re-thrown errors client-side.
2. **`vinext/shims/*` resolver fix.** Package-subpath shim imports resolve to the same local shim files as `next/*` aliases in source checkouts, so request-scoped singleton state is not split between module copies. The recovery fixture exercises this.
3. **Error response parity required by recovery.** App Router error documents use the same client-render fallback trigger as Next.js (`__next_error__` alone), and uncaught/global error responses return HTTP 500 with `private, no-cache, no-store, max-age=0, must-revalidate` while local `error.tsx` boundaries remain 200.

## Why

Previously a failed SSR shell rendered a server-side global-error page whose flight payload encodes the error tree. Without a custom `global-error.tsx` that meant the default error card with no path back to the real page — an SSR-phase-only client throw (the "throw to opt out of server rendering" pattern, used by the Next.js `next-dynamic-css` fixture) left the browser permanently on the card even though the client render would succeed.

## Approach

`fallbackToErrorDocumentOnShellError` is threaded from the generated entry (which knows `hasCustomGlobalError` at build time) through dispatch/render options into `handleSsr`. Rejection semantics are preserved whenever:
- the error carries an RSC `digest` (flight errors, `redirect()`/`notFound()`, server-component throws) — the existing rejection-driven boundary machinery in `renderAppPageHtmlStreamWithRecovery` is untouched, or
- the app defines a custom `global-error.tsx` — the server-rendered boundary re-render path is kept.

Non-goals: dynamic CSS ordering (removed from this PR; follow-up issue), unrelated router behavior, or broader module-filter expansion. Custom `global-error.tsx` apps retain their existing server-rendered boundary path, but uncaught errors now use Next.js-compatible 500 and never-cache response semantics.

## Upstream References

- Shell-error recovery: https://github.com/vercel/next.js/blob/v16.2.6/packages/next/src/server/app-render/app-render.tsx
- SSR-only throw pattern: https://github.com/vercel/next.js/blob/v16.2.6/test/e2e/app-dir/next-dynamic-css/next-dynamic-css.test.ts

## Validation

- New browser spec `tests/e2e/app-router/nextjs-compat/ssr-error-shell-recovery.browser.spec.ts`: recovery to real content for an SSR-only throw without a boundary; local `error.tsx` semantics for SSR-only (content renders, boundary absent) and unconditional (boundary renders) client throws; strict console-error checking apart from the intentional throws.
- `tests/nextjs-compat/global-error.test.ts` (24 tests): all local/nested/global boundary semantics unchanged.
- `tests/entry-templates.test.ts`, `tests/build-optimization.test.ts`, `tests/deploy.test.ts` green.

## Risks / follow-ups

- Apps with a custom `global-error.tsx` and an SSR-only client throw still use vinext's existing server-rendered custom boundary path. The PR only aligns the visible error response status/cache semantics and the `__next_error__` client bootstrap behavior.
- Dynamic CSS ordering returns in a follow-up redesigned on React stylesheet precedence instead of DOM insertion interception.


packages/vinext/src/server/app-browser-entry.ts
```diff
@@ -1603,16 +1603,36 @@ function bootstrapHydration(rscStream: ReadableStream<Uint8Array>): void {
...
+  const children = createElement(BrowserRoot, {
+    initialElements: root,
+    initialNavigationSnapshot,
   });
+  const errorShellStyles = document.querySelectorAll("style[data-vinext-error-shell-style]");
+  // Only the SSR shell-error recovery document carries the style marker.
+  // Legacy server-rendered default-error pages also use `__next_error__`, but
+  // keep their existing hydration path so the serialized server error digest
+  // remains available to DefaultGlobalError.
+  if (document.documentElement.id === "__next_error__" && errorShellStyles.length > 0) {
```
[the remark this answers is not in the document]
NathanDrake2406: Checked this on the current PR head and it does not leave a blank page. The shell-recovery path already lands on the default `This page couldn’t load` card when there is no local boundary. I pushed `a2f481b9` to lock that behavior in with a dedicated no-boundary regression case in `ssr-error-shell-recovery.browser.spec.ts`, then re-ran `PLAYWRIGHT_PROJECT=app-router-chrome-browser-specific pnpm run test:e2e -- tests/e2e/app-router/nextjs-compat/ssr-error-shell-recovery.browser.spec.ts`.

packages/vinext/src/server/app-page-render.ts
```diff
@@ -975,6 +977,25 @@ export async function renderAppPageLifecycle(
...
+        headers: options.middlewareContext.headers,
+        status: 500,
+      },
+      policy: { cacheControl: NEVER_CACHE_CONTROL },
+      timing: htmlResponseTiming,
+    });
+    response.headers.set("Cache-Control", NEVER_CACHE_CONTROL);
+    response.headers.delete("CDN-Cache-Control");
+    response.headers.delete("Cloudflare-CDN-Cache-Control");
+    response.headers.delete("Cache-Tag");
```
james-elicx: this should not need to exist here - vinext shouldnt know about these 3 headers, that's why the cdn cache adapters exist.

james-elicx: Addressed in 158e0827. `app-page-render` no longer names or deletes Cloudflare/CDN-specific headers. It routes the final non-cacheable policy through `applyCdnResponseHeaders`, and the active CDN adapter can now return `null` deletion directives for headers it owns. `CloudflareCdnCacheAdapter` clears `CDN-Cache-Control`, `Cloudflare-CDN-Cache-Control`, and `Cache-Tag` for non-cacheable responses. Added focused adapter and shell-recovery coverage.

PR #1945, opened 2026-06-12
fix(cache): prevent app route cookie replay

james-elicx:
## Summary

Fixes a shared-cache vulnerability where a production App Router `GET` Route Handler with ISR enabled could persist and replay `Set-Cookie` headers across users.

## Reproduction / impact

Before this fix, a handler such as:

```ts
export const revalidate = 300;

export function GET(request: Request) {
  const user = new URL(request.url).searchParams.get("user");
  return new Response(`issued for ${user}`, {
    headers: {
      "Set-Cookie": `session=${user}-secret; Path=/; HttpOnly`,
    },
  });
}
```

could cache the first request's response under the pathname-only App Route ISR key. A later user could receive the cached victim session cookie and body, enabling session fixation, credential disclosure, or cross-user state leakage.

The regression test issues independent victim and attacker requests and verifies no shared cache entry is written and each receives only its own cookie.

## Fix

- Treat direct `Set-Cookie`, pending mutable cookies, draft-mode cookies, and middleware cookies as shared-cache opt-outs.
- Force cookie-bearing responses to `private, no-cache, no-store, max-age=0, must-revalidate` and remove edge cache headers, including explicit user-provided public cache policies.
- Skip stale background-regeneration writes when cookies are produced.
- Never serialize `Set-Cookie` into new App Route ISR values.
- Strip `Set-Cookie` when reading legacy cached values.
- Clone responses before changing cache headers so immutable responses returned from `fetch()` remain valid.

## Next.js research

Reviewed the current Next.js canary App Route implementation in `packages/next/src/server/route-modules/app-route/module.ts`. Next.js merges mutable cookies after producing the prerender result. The focused test includes a source link documenting the parity reference; no direct upstream test for this exact security regression was found.

## Tests

- `vp test run tests/app-route-handler-cache.test.ts tests/app-route-handler-execution.test.ts tests/app-route-handler-response.test.ts`
  - 3 files passed
  - 39 tests passed
- `vp check` on all seven changed files
- `vp fmt` on all changed files

## Review

Independent review passes found and prompted fixes for:

1. stale background regeneration caching cookie-bearing responses;
2. explicit public cache headers allowing downstream CDN replay;
3. immutable `fetch()` response headers throwing during no-store enforcement.

All findings are addressed in this PR.


james-elicx:
Diverges from nextjs

PR #1947, opened 2026-06-12
fix(pages): harden cookie parsing

james-elicx:
## Summary

- consolidate Pages API and config-condition cookie parsing into one shared implementation
- preserve the first duplicate cookie, matching Next.js's compiled `cookie` parser
- return a null-prototype dictionary so inherited names cannot satisfy authentication checks
- decode percent-encoded values with malformed-value fallback and unwrap quoted values

## Security impact

Before this change, vinext selected the last duplicate cookie while Next.js selects the first. In cookie-tossing scenarios, a sibling subdomain, path-scoped cookie, or appending proxy could therefore change which session token a migrated Pages API route authenticated.

The old normal-object dictionaries also exposed inherited properties such as `constructor` and `toString` even when no cookie with that name was supplied. Presence-based authorization using a configurable cookie name could consequently treat an inherited property as an authenticated cookie.

## Reproduction

```bash
curl -i \
  -H 'Cookie: session=trusted; session=attacker' \
  https://TARGET/api/account
```

Before: `req.cookies.session === "attacker"`.

After: `req.cookies.session === "trusted"`, matching Next.js.

Special-name reproduction:

```bash
curl -i \
  -H 'Cookie: __proto__=p; constructor=c; toString=t' \
  https://TARGET/api/account
```

After this change, all three values are explicit own properties on a null-prototype dictionary; absent inherited names no longer appear present.

## Next.js reference

- `packages/next/src/server/api-utils/get-cookie-parser.ts`
- `packages/next/src/compiled/cookie/index.js`

## Validation

- `vp test run tests/api-handler.test.ts tests/pages-bodyparser-config.test.ts`
- `vp test run tests/shims.test.ts -t "parseCookies|config cookie conditions"`
- `vp check packages/vinext/src/utils/parse-cookie.ts packages/vinext/src/config/config-matchers.ts packages/vinext/src/server/api-handler.ts packages/vinext/src/server/pages-node-compat.ts tests/api-handler.test.ts tests/pages-bodyparser-config.test.ts tests/shims.test.ts`

The isolated worktree reused an existing frozen workspace dependency installation through temporary symlinks. No install or lockfile modification was performed.


james-elicx:
Addressed the independent review findings in commit `"$(git rev-parse --short HEAD)"`:

- restored normal object semantics for `req.cookies`, including `hasOwnProperty()` and inherited `toString()`
- matched Next.js compiled `cookie` behavior for `__proto__`, `constructor`, and `toString` cookie names
- preserved empty-name cookies for parser parity
- hardened config `has`/`missing` cookie conditions with `Object.hasOwn()` so inherited names cannot satisfy authentication/routing checks
- added Node, Workers, parser, and config-condition regressions

Focused API tests, config tests, formatting, lint, and type checks pass.

james-elicx:
Resolved the required `vinext-utils/prefer-shared-utils` failure in commit `"$(git rev-parse --short HEAD)"`.

The two parsers intentionally implement different Next.js contracts, so they cannot be safely consolidated:

- `utils/parse-cookie.ts`: Pages API / compiled `cookie` semantics (first duplicate, trimming/quotes, malformed decode fallback, normal object)
- `shims/internal/parse-cookie-header.ts`: `@edge-runtime/cookies` RequestCookies semantics (last duplicate, bare tokens, malformed values skipped, Map)

Renamed the latter to `parseEdgeRequestCookieHeader` and updated all consumers, making the semantic distinction explicit and avoiding the protected shared-helper name collision.

Validation:
- touched-file `vp lint` and `vp check`: pass
- `tests/oxlint-prefer-shared-utils.test.ts`: 3/3 pass
- Pages API focused tests: 75/75 pass
- RequestCookies/config focused tests: 47/47 pass

A full-workspace `vp check` was also attempted; formatting and the shared-utils lint stage passed, but the isolated worktree lacks complete app/benchmark dependency links (`@cloudflare/workers-types`, workspace package resolution). No install or lockfile changes were made.

PR #1950, opened 2026-06-12
fix(cloudflare): prevent auth-dependent edge caching

james-elicx:
## Summary

Prevent Cloudflare's outer edge cache from publishing App Router or Pages Router responses whose contents or routing decision depended on request credentials or user middleware.

## Security impact

With `cdnAdapter()` and Cloudflare platform caching enabled, vinext previously emitted public `CDN-Cache-Control` for otherwise-static/ISR responses after middleware ran inside the Worker. Cloudflare can serve a later edge hit without invoking the Worker, so middleware authentication/authorization would not execute for that hit.

An authenticated request could therefore fill a shared outer-edge cache entry that was reusable by an unauthenticated request.

## Prerequisites

- Cloudflare CDN adapter enabled.
- Platform-level cache enabled for the Worker route.
- A static/ISR page or route protected only by middleware or request credentials.
- The authenticated response is cacheable under the framework policy.

## Reproduction before this fix

```ts
// middleware.ts
export function middleware(request: Request) {
  if (request.headers.get("cookie") !== "session=admin") {
    return new Response("Unauthorized", { status: 401 });
  }
}
```

```tsx
// app/admin/page.tsx
export const revalidate = 300;

export default function Admin() {
  return <pre>Private administration data</pre>;
}
```

```bash
# Authenticated request reaches the Worker and fills the outer cache.
curl -i -H 'Cookie: session=admin' https://TARGET/admin

# A later request can be answered by the edge before middleware runs.
curl -i https://TARGET/admin
```

The first response previously carried a public policy such as:

```http
CDN-Cache-Control: public, max-age=300, stale-while-revalidate=...
```

## Fix

- Add a request-scoped `cdnCacheRequestDependent` safety signal.
- Mark requests dependent when they carry `Cookie`, `Authorization`, or any `CF-Access-*` identity header.
- Mark every request that executes user middleware, including middleware request-header and response-header/rewrite decisions.
- Propagate the signal across nested App Router render/route-handler contexts and the Pages Router pipeline without using a spoofable HTTP header.
- Make the Cloudflare CDN adapter emit only `Cache-Control: no-store` for request-dependent responses.
- Remove any stale `CDN-Cache-Control` and `Cache-Tag` headers when applying the no-store decision.
- Preserve public outer-edge caching for credential-free routes without middleware.

## Tests

```bash
vp check tests/cdn-cache.test.ts tests/pages-request-pipeline.test.ts \
  packages/vinext/src/shims/cdn-cache.ts \
  packages/vinext/src/shims/unified-request-context.ts \
  packages/vinext/src/server/cache-control.ts \
  packages/vinext/src/server/app-rsc-handler.ts \
  packages/vinext/src/server/pages-request-pipeline.ts \
  packages/vinext/src/server/pages-page-handler.ts \
  packages/cloudflare/src/cache/cdn-adapter.runtime.ts

vp test run tests/cdn-cache.test.ts \
  tests/pages-request-pipeline.test.ts \
  tests/pages-page-handler.test.ts \
  tests/app-rsc-handler.test.ts \
  tests/cloudflare-cdn-cache.test.ts
```

Result: 137 tests passed.

## Review notes

An adversarial review found that nested App Router request contexts initially reset the safety bit after middleware. The final patch inherits the bit into nested contexts and includes a regression for that path.

## Limitations

- Local Miniflare does not implement Cloudflare's outer cache, so tests assert the response-policy boundary (`no-store`, no `CDN-Cache-Control`, no `Cache-Tag`) rather than a deployed `CF-Cache-Status` HIT.
- This intentionally disables outer-edge caching for all routes traversing user middleware because middleware may make authorization decisions without leaving observable response headers.
- Framework/origin ISR behavior is unchanged; this patch only prevents unsafe shared outer-edge reuse.

### DIFF
--- a/packages/vinext/src/server/cache-control.ts
+++ b/packages/vinext/src/server/cache-control.ts
@@ -1,4 +1,8 @@
-import { getCdnCacheAdapter, type CdnCacheableHeaderInput } from "vinext/shims/cdn-cache";
+import {
+  getCdnCacheAdapter,
+  isCdnCacheRequestDependent,
+  type CdnCacheableHeaderInput,
+} from "vinext/shims/cdn-cache";
 
 export const NEVER_CACHE_CONTROL = "private, no-cache, no-store, max-age=0, must-revalidate";
 
@@ -21,7 +25,12 @@ export const NO_STORE_CACHE_CONTROL = "no-store, must-revalidate";
  */
 export function applyCdnResponseHeaders(headers: Headers, input: CdnCacheableHeaderInput): void {
   headers.delete("Cache-Control");
-  const map = getCdnCacheAdapter().buildResponseHeaders(input);
+  headers.delete("CDN-Cache-Control");
+  headers.delete("Cache-Tag");
+  const map = getCdnCacheAdapter().buildResponseHeaders({
+    ...input,
+    requestDependent: input.requestDependent ?? isCdnCacheRequestDependent(),
+  });
   for (const [name, value] of Object.entries(map)) {
     // Never stamp an empty header. An adapter returns an empty `Cache-Control`
     // only when it has no default for an empty policy (e.g. the default
--- a/packages/vinext/src/server/cache-control.ts
+++ b/packages/vinext/src/server/cache-control.ts
@@ -52,17 +52,42 @@ export function finalizeCdnResponsePolicy(
   response: Response,
   requestDependent = isCdnCacheRequestDependent(),
 ): Response {
-  if (!requestDependent) return response;
+  if (!requestDependent || process.env.VINEXT_PRERENDER === "1") return response;
+
+  const currentCacheControl = response.headers.get("Cache-Control");
+  const cacheControl =
+    currentCacheControl && /\b(?:no-store|no-cache|private)\b/i.test(currentCacheControl)
+      ? currentCacheControl
+      : "no-store";
+  const needsCleanup = CDN_RESPONSE_POLICY_HEADERS.some((name) => response.headers.has(name));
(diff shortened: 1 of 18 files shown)


packages/vinext/src/server/cache-control.ts
```diff
@@ -21,7 +25,12 @@ export const NO_STORE_CACHE_CONTROL = "no-store, must-revalidate";
  */
 export function applyCdnResponseHeaders(headers: Headers, input: CdnCacheableHeaderInput): void {
   headers.delete("Cache-Control");
-  const map = getCdnCacheAdapter().buildResponseHeaders(input);
+  headers.delete("CDN-Cache-Control");
+  headers.delete("Cache-Tag");
```
james-elicx, line 29: this assumption is unacceptable to make inside vinext itself. adapters control their own headers.

james-elicx: feel free to add an OPTIONAL function on the cdn cache to wipe them if you really must. i dont really see why though. doesnt building overwrite?

packages/vinext/src/server/cache-control.ts
```diff
@@ -32,6 +41,55 @@ export function applyCdnResponseHeaders(headers: Headers, input: CdnCacheableHea
   }
 }
 
+const CDN_RESPONSE_POLICY_HEADERS = [
+  "CDN-Cache-Control",
+  "Cloudflare-CDN-Cache-Control",
+  "Cache-Tag",
```
james-elicx, line 47: ABSOLUTELY NOT.

PR #1956, opened 2026-06-12
fix(app-router): fold typeof window before resolution

james-elicx:
## Summary
- fold `typeof window` per Vite environment before Rolldown resolves imports
- remove statically unreachable `if (typeof window ...)` branches so server builds do not resolve browser-only conditional exports
- add a focused App Router production-build regression based on Next.js `test/e2e/app-dir/typeof-window/typeof-window.test.ts`

## Investigation
The requested path `test/e2e/app-dir/type-of-window/type-of-window.test.ts` does not exist in Next.js canary or the CI-pinned `v16.2.6`; the matching suite is `test/e2e/app-dir/typeof-window/typeof-window.test.ts`.

That upstream suite has `skipDeployment: true`, so the exact deploy manifest result is `outputSuites: 0` and no assertion runs in deploy mode. A clean equivalent fixture without the suite's corrupted tarball reproduced the vinext failure under Node 24:

```
"./browser" is not exported under the conditions ["module", "node", "production", "import"]
```

The failure occurred during the SSR build because Rolldown resolved a browser-only dynamic import guarded by `if (typeof window !== 'undefined')` before dead-code elimination.

## Tests
- `vp test run tests/type-of-window.test.ts tests/compiler-define.test.ts`
- `vp check tests/type-of-window.test.ts tests/compiler-define.test.ts packages/vinext/src/plugins/typeof-window.ts packages/vinext/src/index.ts`
- `vp run vinext#build`

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -173,6 +173,7 @@ import { removeConsoleCalls } from "./plugins/remove-console.js";
 import { createImportMetaUrlPlugin } from "./plugins/import-meta-url.js";
 import { createRequireContextPlugin } from "./plugins/require-context.js";
 import { createWasmModuleImportPlugin } from "./plugins/wasm-module-import.js";
+import { replaceTypeofWindow } from "./plugins/typeof-window.js";
 import { hasMdxFiles } from "./utils/mdx-scan.js";
 import { scanPublicFileRoutes } from "./utils/public-routes.js";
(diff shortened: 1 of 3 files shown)


james-elicx:
Independent review found and fixed shadowed-window constant-folding cases and improved dead-branch removal/source-map fidelity. CI is rerunning on commit 0c071958.

PR #1961, opened 2026-06-12
fix(css): load local CommonJS PostCSS plugins

james-elicx:
## Summary
- preserve CommonJS semantics for local `.js` PostCSS plugins after `vinext init` adds `"type": "module"`
- retry those plugins through a temporary sibling `.cjs` module when ESM loading fails on `module`/`exports`/`require`
- port focused coverage from Next.js `css-modules-rsc-postcss` for CSS and SCSS modules

## Root cause
The upstream fixture uses `postcss.config.js` with `require.resolve('./plugin')`, where `plugin.js` is CommonJS. The deploy adapter runs `vinext init`, which adds `"type": "module"` and renames the config to `postcss.config.cjs`, but the referenced `plugin.js` remains unchanged. vinext's string-plugin resolver dynamically imported that file as ESM, failed on `module.exports`, returned no override, and Vite then rejected both string plugin entries (one failure for each CSS module import).

## Validation
- `vp check packages/vinext/src/plugins/postcss.ts tests/postcss-resolve.test.ts`
- `vp test run tests/postcss-resolve.test.ts` (15 passed)
- `vp run vinext#build`
- exact deploy fixture from `test/e2e/app-dir/css-modules-rsc-postcss/` builds and serves successfully
- local Chrome computed styles: `<p>` and `<span>` both `rgb(0, 128, 0)`

## Harness note
A fresh full Next.js canary preparation was attempted, but the unrelated monorepo build was OOM-killed with exit 137. The exact fixture was therefore exercised directly through `scripts/e2e-deploy.sh`; before the fix it reproduced the two `Invalid PostCSS Plugin found at: plugins[0]` failures, and after the fix both upstream browser assertions pass.

PPR/cache paths are intentionally untouched.

### DIFF
--- a/packages/vinext/src/plugins/postcss.ts
+++ b/packages/vinext/src/plugins/postcss.ts
@@ -1,27 +1,53 @@
 import path from "node:path";
 import fs from "node:fs";
 import { pathToFileURL } from "node:url";
-import { createRequire } from "node:module";
-import { randomUUID } from "node:crypto";
+import { createRequire, Module } from "node:module";
+
+type CommonJsModule = Module & {
+  _compile(content: string, filename: string): void;
+};
+
+const CommonJsModule = Module as typeof Module & {
+  _nodeModulePaths(from: string): string[];
+};
 
 function shouldRetryAsCommonJs(error: unknown, resolvedPath: string): boolean {
   return (
     resolvedPath.endsWith(".js") &&
-    error instanceof ReferenceError &&
-    /(?:module|exports|require) is not defined in ES module scope/.test(error.message)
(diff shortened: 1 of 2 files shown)


james-elicx:
Independent review replaced unsafe temporary plugin copies with in-memory CommonJS compilation and added symlink/read-only/cleanup coverage. CI is rerunning on a709b6a7.

james-elicx:
The smoke failure was isolated to Hacker News preview availability (HTTP 000000); all other 12 checks passed. Failed deployment/smoke jobs rerun requested.

PR #1997, opened 2026-06-13
fix(router): honor hybrid pages route priority

NathanDrake2406:
## Summary
- Port the upstream App Router useParams behavior across App + Pages fixtures, including the Pages Router case where `/pages-dir/foobar` must render `pages/pages-dir/[dynamic]` instead of `app/[...path]`.
- Add a hybrid route priority helper that compares matched Pages and App dynamic route patterns through the existing Next-compatible route sorter.
- Export Pages route matchers through the generated bridge so production can decide ownership before render-probing, and align dev middleware ownership with the same ordering.
- Preserve Vinext static App ownership for identical static App/Page routes, which keeps Cloudflare Worker dev from falling back to the Pages host stub at `/`.

## Root Cause
Vinext kept App and Pages route ownership separate. In production, the App RSC handler only delegated to Pages when no App route matched, so `app/[...path]` captured `/pages-dir/foobar` before the Pages Router route could render. Next.js registers Pages route matcher providers before App providers and sorts dynamic route pathnames together, so `pages/pages-dir/[dynamic]` has priority over an App root catch-all for this request.

The implementation now uses the same sorted dynamic-route comparison only when dynamic route ownership is ambiguous. Static App/Page duplicates remain App-owned in Vinext, matching the existing Cloudflare Worker dev contract for the app-router-cloudflare fixture.

## References
- [Next.js upstream use-params test](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/use-params/use-params.test.ts#L37-L40)
- [Next.js matcher provider registration order](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/base-server.ts#L725-L740)
- [Next.js dynamic route sorting in DefaultRouteMatcherManager](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/route-matcher-managers/default-route-matcher-manager.ts#L118-L165)
- [Next.js Pages useParams adapter](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/router/adapters.tsx#L52-L65)

## Verification
- `vp test run tests/hybrid-route-priority.test.ts tests/app-router-next-config-codegen.test.ts tests/app-pages-bridge.test.ts`
- `CI=1 PLAYWRIGHT_PROJECT=use-params-app-pages vp run test:e2e`
- `PLAYWRIGHT_PROJECT=cloudflare-dev vp run test:e2e tests/e2e/cloudflare-dev/pages-router.spec.ts tests/e2e/cloudflare-dev/middleware.spec.ts`
- `PLAYWRIGHT_PROJECT=app-router vp run test:e2e tests/e2e/app-router/pages-router-use-params.spec.ts`
- `vp env exec --node 24 ./scripts/run-nextjs-deploy-suite.sh /Users/nathan/Projects/vinext/.refs/nextjs-v16.2.6 --retries 0 -c 1 --debug test/e2e/app-dir/use-params/use-params.test.ts`
- `vp check` on the changed runtime, fixture, and test files
- pre-commit full check, staged unit/integration tests, and knip


NathanDrake2406:
Addressed the review feedback:

- **`should work for single dynamic param`** (direct-load): Changed navigation URL from `/a/b` to `/a` so it actually exercises the `[id]` route instead of the `[id]/[id2]` nested route.
- Added an extra assertion that `#param-id2` has count 0 on this page, confirming the single-param route (not the nested one) is serving the request.

Rebased onto latest `upstream/main` — no conflicts.

PR #2019, opened 2026-06-14
fix(script): register src beforeInteractive scripts and mark hoisted output

### LINKED ISSUE #2016: `next/script` `beforeInteractive`: hydration duplicate, `src` bypasses server registry, attr-map gaps
**vinext:** 0.1.2

**Summary.** Three related defects in the `next/script` shim around `beforeInteractive`:
1. **Hydration duplication.** SSR splices the beforeInteractive script into the pre-head, but the emitted node carries no marker. On the client, `<Script>` re-renders the element because it can't tell the server already handled it (the nav runtime may not be registered yet — dev loads it as an async ESM module). Result: a duplicate script element and a hydration mismatch on essentially every page.
2. **`src` scripts bypass the server registry.** Only inline `beforeInteractive` content is registered into the pre-head splice path. A `beforeInteractive` script with `src` falls through to the normal React element path, so it does not get the same pre-head ordering, marker, or client dedupe treatment as inline scripts.
3. **Attribute mapping gaps.** `REACT_TO_HTML_ATTR` maps `className`/`htmlFor`/`httpEquiv` but not `crossOrigin` → `crossorigin` or `referrerPolicy` → `referrerpolicy`, so server-spliced script attrs are not normalized to the same HTML spelling Next uses.

**Reproduce.**
1. Add `<Script id="x" strategy="beforeInteractive">…</Script>` (and separately one with `src=…`).
2. Load/hydrate the page.
3. Observe: two copies of the inline script + a hydration warning (defect 1); the `src` script is not emitted through the same server splice/marker/dedupe path as inline `beforeInteractive` scripts (defect 2); `crossOrigin`/`referrerPolicy` are not normalized in the server-spliced string attrs (defect 3).

**Where (0.1.2 dist).**
- `dist/server/app-ssr-entry.js:80` — `renderBeforeInteractiveInlineScripts` emits scripts with no identifying marker.
- `dist/shims/script.js:113` — `REACT_TO_HTML_ATTR` missing `crossOrigin`/`referrerPolicy`.
- `dist/shims/script.js` beforeInteractive branch (~L269 / ~L353) — only handles inline content (`inlineContent !== null`) and dedups only via `hasAppNavigationRuntimeBootstrap()`.

**Suggested fix.**
- Emit a marker on SSR scripts: `data-vinext-before-interactive=""` (in `renderBeforeInteractiveInlineScripts`).
- On the client, also treat a matching spliced DOM node (by marker + `id`/`src`) as "server already handled it" → return `null` instead of rendering a duplicate.
- Register `src` beforeInteractive scripts too (`inlineContent !== null || src`), passing `src` through the registered attributes.
- Add `crossOrigin: "crossorigin"` and `referrerPolicy: "referrerpolicy"` to `REACT_TO_HTML_ATTR`.

Related existing issues: #1518 and #1557 cover earlier `beforeInteractive` head-placement/order failures; #1517 covers nonce/stylesheets/bootstrap preinit. This report is narrower: server/client duplicate-dedupe marker, `src` parity with the beforeInteractive server registry, and the remaining attr-map gaps. Defect 3 is independent and tiny enough to split if maintainers prefer, but keeping it here is reasonable because the fix is in the same `next/script` attr collection/emission path.

Xplod13:
## Summary

- Register `src` `beforeInteractive` `next/script`s (not just inline ones) so they are hoisted into `<head>` server-side, matching inline behavior.
- Mark every hoisted `beforeInteractive` script with `data-nscript="beforeInteractive"` and suppress the duplicate client render for both inline and `src` scripts, preventing double execution / hydration mismatch.
- Map `crossOrigin` → `crossorigin` and `referrerPolicy` → `referrerpolicy` when serializing hoisted script attributes.

## Root Cause

Three defects in App Router `beforeInteractive` handling:

1. **`src` scripts bypassed the server registry.** In `shims/script.tsx` the SSR branch computed `inlineContent = src ? null : …` and only registered when `inlineContent !== null`, so external `src` scripts were never captured and never hoisted into `<head>` before interactive — they stayed in source order behind React's resource hints. The condition now registers when `src` is present too.
2. **No hydration marker → duplicate execution.** Hoisted scripts carried no marker, so the client React render of the same `<Script>` produced a second `<script>` node (double execution / hydration mismatch). Next.js marks server-rendered `beforeInteractive` scripts with `data-nscript="beforeInteractive"` and dedupes on the client; we now emit the same marker and return `null` on the client for both inline and `src` `beforeInteractive` scripts in App Router (afterInteractive/lazyOnload paths are unchanged).
3. **Attribute name mismatch.** `REACT_TO_HTML_ATTR` lacked `crossOrigin`/`referrerPolicy`, so they round-tripped as camelCase; added the lowercase mappings to match Next.js' attribute serialization.

For testability the pure emit logic (`renderBeforeInteractiveInlineScripts`) was extracted from the virtual-import-heavy `app-ssr-entry.ts` into a new `server/before-interactive-head.ts`; attribute values are HTML-escaped at that boundary.

## References

- Fixes #2016
- Next.js parity: `packages/next/src/client/script.tsx` (`data-nscript` marker + `addBeforeInteractiveToCache` dedupe on `id || src`)
- Ported test: `test/e2e/app-dir/script-before-interactive/script-before-interactive.test.ts` (the multiple-scripts case)

## Verification

- `CI=true pnpm test tests/script.test.ts tests/script-head-ordering.test.ts` — new tests cover: `src` `beforeInteractive` hoisted into `<head>`; `crossOrigin`/`referrerPolicy` lowercased on hoisted tags; client does not re-render a hoisted `src` script; every hoisted script carries `data-nscript`; multiple hoisted scripts (ported from Next.js). All green (red first).
- `CI=true pnpm test` — full battery green (only the pre-existing env flakes `deploy.test.ts > resolveWranglerBin` and `oxlint-prefer-shared-utils`, reproduced identically on `origin/main`).
- `CI=true npx vp check` — clean on all changed files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,7 @@ result-*
 
 .vinext
 next-env.d.ts
+
+# Leaked temp fixtures from the oxlint prefer-shared-utils rule test
+# (normally cleaned up in afterEach; ignore so a crashed run can't commit them).
+__lint_rule_fixtures__-*
--- a/packages/vinext/src/server/app-ssr-entry.ts
+++ b/packages/vinext/src/server/app-ssr-entry.ts
@@ -33,6 +33,7 @@ import {
   escapeHtmlAttr,
   safeJsonStringify,
 } from "./html.js";
+import { renderBeforeInteractiveInlineScripts } from "./before-interactive-head.js";
 import {
   createNavigationRuntimeRscMetadataScript,
   createRscEmbedTransform,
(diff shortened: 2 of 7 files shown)


Xplod13:
Good catch — those two `__lint_rule_fixtures__-*` entries were leaked temp fixtures from `tests/oxlint-prefer-shared-utils.test.ts` (broken symlinks into a machine-local macOS temp path) that got swept into the commit by an over-broad `git add`. They have nothing to do with the fix.

Removed both with `git rm` and added `__lint_rule_fixtures__-*` to `.gitignore` so a crashed/interrupted run of that test can't commit them again. Pushed in `f7eba352`. The remaining diff is just the script-shim fix.

Thanks for verifying parity and the streaming path (`app-ssr-stream.ts`) too.

PR #2036, opened 2026-06-15
fix(router): match basePath router event semantics

james-elicx:
## Summary

- include `basePath` in Pages Router event URL payloads
- align normal, hash, cancellation, failure, popstate, and redirect event semantics with Next.js
- emit redirected completion and hash scrolling for the committed destination URL
- preserve the cancelled transition's original `shallow` event options

## Next.js parity

Validated against pinned Next.js commit `ee6e79b1792a4d401ddf2480f40a83549fe8e722` and the upstream coverage in:

- `test/e2e/basepath/router-events.test.ts`
- `packages/next/src/shared/lib/router/router.ts`

The vinext fixture links directly to the upstream E2E test and reproduces the behavior locally without running the raw Next.js suite.

## Targeted local validation

- focused Pages Router basePath Playwright fixture passed with one worker
- focused router unit assertions passed for normal, hash, cancelled, failed, and redirected navigations
- scoped formatting, lint, types, and `git diff --check` passed

No broad local suite or raw Next.js E2E was run.

PR #2037, opened 2026-06-15
fix(router): ignore superseded page loader failures

james-elicx:
## Summary

- keep the current page mounted when a Pages route-data request returns an HTTP 5xx
- emit `routeChangeError` instead of hard-navigating for route-data server failures
- preserve hard navigation for deploy-skew 404s, malformed payloads, missing modules, and active chunk-loader failures
- ignore late loader failures from superseded navigations so they cannot override a winning route

## Stack

Stacked on #2036, which is stacked on #2033.

## Next.js parity

Completes the failed-route assertion in `test/e2e/basepath/router-events.test.ts`.

## Validation

- exact Next success, hash, cancellation, and failed-route assertions: 4/4 passed
- focused stale-loader and active-failure controls passed
- targeted `vp check` passed
- independent cumulative review: clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -270,6 +270,10 @@ jobs:
             label: pages-router
             shardIndex: 1
             shardTotal: 1
+          - project: pages-router-basepath-dev
+            label: pages-router-basepath-dev
+            shardIndex: 1
+            shardTotal: 1
           - project: app-router
             label: app-router 1/3
             shardIndex: 1
--- a/packages/vinext/src/server/dev-server.ts
+++ b/packages/vinext/src/server/dev-server.ts
@@ -1209,11 +1209,7 @@ export function createSSRHandler(
                         viteBase,
                       );
                       const regenAppUrl = RegenApp
-                        ? createPagesDevModuleUrl(
-                            viteRoot,
-                            path.join(pagesDir, "_app"),
-                            viteBase,
-                          )
+                        ? createPagesDevModuleUrl(viteRoot, path.join(pagesDir, "_app"), viteBase)
                         : null;
                       const freshPagesNextData = {
                         ...pagesNextData,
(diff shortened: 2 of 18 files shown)


james-elicx:
Addressed the exact-head review: removed the incorrect 5xx page-retention behavior and restored Next.js hard-navigation semantics. The branch now contains only the independently valid superseded-loader cancellation fix, plus the upstream basePath router-event fixture without a no-reload assertion.

/bigbonk review for issues

james-elicx:
Deploy-suite run 28478866791 / job 84413308650 triage: I think this PR already covers the basePath failed-route event failure: test/e2e/basepath/router-events.test.ts includes the failed route-change assertion covered by this stacked PR. I am skipping a duplicate backlog fix for that item and leaving this note so the overlap is visible.

james-elicx:
Run 28938793088 overlap (authoritative oracle: Next.js v16.3.0-canary.80, commit 9dd6d677b63b249584c8ff0bccffcc6a65288683): this stacked PR owns the superseded page-load behavior exercised by:

- `test/e2e/route-load-cancel/route-load-cancel.test.ts` — `should cancel slow page loads on re-navigation`
- `test/e2e/basepath/router-events.test.ts` — `should use urls with basepath in router events for failed route change` (failed-route handling portion)

I reproduced the route-load cancellation failure with the targeted Next.js deploy E2E wrapper and am dropping my duplicate lane.

james-elicx:
Run 29871622126 ownership note\n\nThe four failures in test/e2e/basepath/router-events.test.ts are now all fixed by refreshed #2036 at head 1636cbe2e250079d4bee9c5b1860d35bb58d84cc, including the failed-route assertion. Its exact Next.js v16.2.6 proof passes 4/4.\n\nThis PR was originally opened for the failed-route row, but the current title/code evolved into the separate superseded page-loader failure race. It therefore has no unique assertion from run 29871622126 after #2036 is refreshed. It is also no longer a descendant of #2036 even though the GitHub base is still configured as that branch, so it will need a separate current-main refresh if we continue the loader-race fix.\n\nThe PR body is stale: it still describes retaining the page on HTTP 5xx, while the reviewed code restored hard-navigation semantics and only keeps the superseded-loader guard.

PR #2047, opened 2026-06-15
fix(server): q-value-aware Accept-Encoding negotiation

### LINKED ISSUE #1981: Prod server: Accept-Encoding negotiation ignores q-values (q=0 refusals served anyway)
### Problem
A client that advertises support but disables a codec via q=0 (a legitimate, spec-defined signal) can be served content in that very codec; a client unable to decode it gets a broken response. Also, substring matching is brittle (any future/uncommon token containing these substrings would false-match). Impact is bounded because the standard browser header `gzip, deflate, br` has no q-values, but it is a real content-negotiation correctness deviation.

### Evidence
- `packages/vinext/src/server/prod-server.ts:214` — negotiateEncoding does `lower.includes("zstd"|"br"|"gzip"|"deflate")` — pure substring tests that ignore q-values, so `Accept-Encoding: gzip, br;q=0` selects br even though the client set q=0 (explicit refusal).
- `packages/vinext/src/server/prod-server.ts:546` — The static fast path repeats the same pattern: `ae.includes("zstd")`/`includes("br")`/`includes("gzip")` to pick a precompressed variant, again ignoring q=0 refusals.

Next.js's production server negotiates content-encoding entirely through the vendored `compression` middleware (router-server.ts wires setupCompression and runs compress(req,res) on every response). That middleware resolves the encoding via accepts -> negotiator -> preferredEncodings, which fully parses Accept-Encoding q-values: parseEncoding extracts q with parseFloat, and preferredEncodings filters candidates through isQuality (a.q>0), so an explicitly-refused codec (q=0) is excluded and the server falls back to identity. Thus for `gzip, br;q=0` Next.js would never emit the refused codec — the opposite of vinext's substring `includes()` checks that ignore q-values. For vinext's second cited site (the static fast-path picking a precompressed .br/.gz/.zst variant by substring), Next.js has no equivalent: serve-static.ts uses the `send` library, which does no Accept-Encoding negotiation and never serves precompressed disk variants. Both vinext code paths are therefore either spec-correct in Next.js (on-the-fly compression) or nonexistent in Next.js (precompressed static selection), making this a vinext-only defect.

Citations: packages/next/src/server/lib/router-server.ts:118-119 (setupCompression) and :285-287 (compress(req,res,...) wraps every response); packages/next/src/compiled/compression/index.js:48 — accepts.encoding -> negotiator.encodings -> preferredEncodings: parseEncoding reads q via `if(t[0]==="q"){o=parseFloat(t[1])}` (~col 2420) and preferredEncodings applies `function isQuality(a){return a.q>0}` (~col 3241), filtering out any q=0 encoding; the middleware calls `.encoding(["gzip","deflate","identity"])`. packages/next/src/server/serve-static.ts:13-31 and packages/next/src/compiled/send/index.js — sta

### Suggested fix
Parse Accept-Encoding into (token, q) pairs, drop tokens with q=0, and match exact tokens by descending q (and the documented zstd>br>gzip>deflate preference) instead of substring includes(). Apply the same parsing to the static fast-path variant selection.

### Test plan
Unit tests: `gzip, br;q=0` → gzip; `*;q=0, gzip` → gzip; `identity;q=0` handling.

### Related / notes
Next.js delegates to the `compression` middleware which uses negotiator (q-aware). Both negotiateEncoding (prod-server.ts:214) and the static precompressed fast path (:546) need the q-aware parse.

---

MaxtuneLee:
# Summary

Fix Accept-Encoding negotiation to honor RFC 9110 q-values instead of brittle substring matching. A new parseAcceptedEncodings() helper parses the header into a set of accepted encoding tokens, dropping any token explicitly refused with q=0 and matching tokens exactly. Both negotiation sites and the static precompressed now use this parser.

# Why

The previous implementation negotiated content encoding with pure substring tests (lower.includes("br"), ae.includes("gzip"), etc.), which ignore q-values entirely.

This caused two correctness problems:

1. q=0 refusals were ignored. q=0 is a spec-defined, explicit refusal of a codec. With Accept-Encoding: gzip, br;q=0, the old code still selected br because includes("br") matched — serving a client a codec it explicitly said it cannot decode, producing a broken response.
2. Substring matching is brittle. Any current or future token containing br, gzip, etc. as a substring would false-match.

# Files touched

- `packages/vinext/src/server/prod-server.ts`
  - Added parseAcceptedEncodings(accept): parses the header into a Set of accepted lowercase tokens, honoring q-values (drops q=0, rejects malformed q-values, defaults missing q to 1, matches tokens exactly, preserves the * wildcard token).
  - Rewrote negotiateEncoding() to select via exact token matching over the parsed set, preserving the zstd > br > gzip > deflate preference order.
  - Updated the static precompressed-variant fast path to use parseAcceptedEncodings instead of ae.includes(...), so q=0 refusals are honored when picking .zst / .br / .gz variants.
  - Exported parseAcceptedEncodings for testing.
- tests/accept-encoding-negotiation.test.ts (new)
  - 32 unit tests covering the parser and negotiateEncoding.

# Testing plan

- New unit tests (tests/accept-encoding-negotiation.test.ts) — 32/32 passing. Coverage includes:
  - gzip, br;q=0 → gzip
  - *;q=0, gzip → gzip
  - identity;q=0 handling
  - q=0.0 / q=0.000 treated as refusals
  - exact-token matching (no substring false-match, e.g. brotli-future does not imply br)
  - malformed q-values dropped (q=abc, q=1.5, q=0.1.2)
  - preference order and fallback past a refused higher-preference codec
  - all-refused → null
  - zstd cases gated behind runtime zstd support

# Reference

- Fixes #1981
- RFC 9110 §12.5.3 (Accept-Encoding, q-values, q=0 as refusal)


james-elicx:
Addressed the valid empty-`Vary` edge case from the latest Bonk review in `fix(server): normalize empty vary headers`.

The first-vs-last repeated `q` note is not actionable: direct `negotiator@1.0.0` verification gives `br;q=0;q=0.5` → identity and `br;q=0.5;q=0` → br, so the current first-`q` behavior matches the library Next.js uses.

PR #2052, opened 2026-06-15
fix(build): preserve sibling bindings when stripping server exports

### LINKED ISSUE #1972: Build: stripServerExports drops sibling bindings in combined `export const` declarations
### Problem
A page that combines a data-fetching export with any other export in a single `export const a, getStaticProps` declaration loses the sibling binding `a` from the client bundle, producing a runtime `a is not defined` / undefined-export error during hydration. Next.js's next-ssg-transform removes only the specific binding, preserving siblings. Niche pattern but a clear correctness/parity deviation that silently corrupts the client module.

### Evidence
- `packages/vinext/src/plugins/strip-server-exports.ts:47` — When a server-export name (e.g. getStaticProps) is one declarator inside a VariableDeclaration, the code overwrites the ENTIRE export statement (node.start..node.end) with a single stub, discarding every other declarator in the same `export const ... , ... ;`.
- `packages/vinext/src/plugins/strip-server-exports.ts:44` — The loop iterates all declarators but each match overwrites the full node range; for `export const myData = 42, getStaticProps = ...` the result becomes only `export const getStaticProps = undefined;`, deleting `myData` from the client bundle.
- `packages/vinext/src/plugins/strip-server-exports.ts:45` — If two server exports share one declaration (`export const getStaticProps = ..., getStaticPaths = ...`), s.overwrite is called twice on the identical range; MagicString 0.30.21 silently keeps only the last, emitting just one stub (verified locally).

vinext's strip-server-exports.ts:44-49 iterates declarators but, on a match, overwrites the entire VariableDeclaration range (node.start..node.end) with a single stub, so a combined `export const a = 1, getStaticProps = ...` loses the sibling `a`, and two server exports in one declaration collapse to one stub due to MagicString overwriting the same range twice. I traced both Next.js counterparts. The default SWC transform (next_ssg.rs) marks only the server-export declarator's name as Pat::Invalid (line 616) and then prunes it with `decls.retain(|d| !d.name.is_invalid())` (line 625), leaving every sibling declarator intact; the legacy Babel plugin (next-ssg-transform.ts:286-300) similarly calls `d.remove()` per matching declarator. Next.js even ships a dedicated fixture (should-not-remove-extra-named-export-variable-declarations) whose input mixes a server export with a sibling `foo = 2` in one `export const`, and whose committed output is `export const foo = 2;` — the sibling is preserved and the server export removed entirely. This proves Next.js handles the exact triggering scenario correctly, so the defect is vinext-only.

Citations: crates/next-custom-transforms/src/transforms/next_ssg.rs:623-625 (visit_mut_var_declarators uses `decls.retain(|d| !d.name.is_invalid())`, removing only the matched declarator and keeping siblings); next_ssg.rs:610-621 (visit_mut_var_declarator marks only the server-export name as Pat::Invalid); packages/next/src/build/babel/plugins/next-ssg-transform.ts:286-300 (VariableDeclaration case loops declarators and calls per-declarator `d.remove()`, preserving siblings); test fixture crates/next-custom-transforms/tests/fixture/ssg/getStaticProps/should-not-remove-extra-named-export-variable-declarat

### Suggested fix
Rewrite only the matching declarator (overwrite the declarator's own start..end to `<name> = undefined`, or drop it and rely on tree-shaking) instead of the whole node, and accumulate matches so multiple server-export declarators in one statement are each handled without overlapping overwrites.

### Test plan
Port Next.js `babel-plugin-next-ssg-transform.test.ts:167` ('should not remove extra named export variable declarations').

### Related / notes
MagicString double-overwrite on identical range also collapses two server exports in one declaration to a single stub.

---

Xplod13:
## Summary

- Preserve sibling bindings when stripping server-only exports from a combined `export const a = …, getStaticProps = …` declaration.
- Stub each matched server export independently instead of overwriting the whole statement.

## Root Cause

`stripServerExports` (`plugins/strip-server-exports.ts`) handled a `VariableDeclaration` by overwriting the entire statement range (`node.start..node.end`) with a single `export const <name> = undefined;` when any declarator matched a server export. Two failures followed:

1. Sibling bindings sharing the declaration were dropped — `export const keep = 1, getStaticProps = () => {}` became `export const getStaticProps = undefined;`, so `keep` vanished and threw `ReferenceError` at runtime.
2. When several server exports shared one declaration, the same range was overwritten twice (MagicString), collapsing them to a single stub.

The fix overwrites **per declarator** (`declarator.start..declarator.end` → `<name> = undefined`), which preserves the `const`/`let`/`var` keyword, the commas, and every sibling, and touches only disjoint ranges. Non-`Identifier` declarators (destructuring) are still skipped. This matches Next.js' `next-ssg` transforms, which mark/remove only the matched declarator and retain the rest.

## References

- Fixes #1972
- Next.js parity: `crates/next-custom-transforms/src/transforms/next_ssg.rs` and `packages/next/src/build/babel/plugins/next-ssg-transform.ts` (per-declarator removal)
- Ported test: `test/unit/babel-plugin-next-ssg-transform.test.ts` ("should not remove extra named export variable declarations")

## Verification

- `CI=true pnpm test tests/build-optimization.test.ts` — new cases: a sibling binding sharing a declaration with a server export is preserved; multiple server exports in one declaration are each stubbed. Red before, green after; all pre-existing `stripServerExports` tests still pass (115 passed).
- `CI=true pnpm test` — full battery green (only the pre-existing env flakes `deploy.test.ts > resolveWranglerBin` and `oxlint-prefer-shared-utils`, reproduced identically on `origin/main`).
- `CI=true npx vp check` — clean on both changed files.

### DIFF
--- a/packages/vinext/src/plugins/strip-server-exports.ts
+++ b/packages/vinext/src/plugins/strip-server-exports.ts
@@ -42,9 +42,15 @@ export function stripServerExports(code: string): string | null {
         );
         changed = true;
       } else if (decl.type === "VariableDeclaration") {
+        // Stub each matched declarator in place rather than overwriting the
+        // whole statement. This preserves any sibling bindings sharing the
+        // declaration (`export const a = 1, getStaticProps = ...`) and avoids
+        // a double-overwrite collapse when two server exports share one
+        // declaration. Matches Next.js's next-ssg, which removes only the
+        // matched declarator and retains the rest.
         for (const declarator of decl.declarations) {
           if (declarator.id?.type === "Identifier" && SERVER_EXPORTS.has(declarator.id.name)) {
-            s.overwrite(node.start, node.end, `export const ${declarator.id.name} = undefined;`);
+            s.overwrite(declarator.start, declarator.end, `${declarator.id.name} = undefined`);
             changed = true;
           }
         }
--- a/tests/build-optimization.test.ts
+++ b/tests/build-optimization.test.ts
@@ -2772,6 +2772,48 @@ export const getStaticPaths = () => [
     expect(result).toContain("export const getStaticPaths = undefined;");
     expect(result).not.toContain("a;b");
   });
+
+  // Regression test for #1972: a server export sharing one `const` statement
+  // with a sibling binding must not drop the sibling.
+  // Ported from Next.js: test/unit/babel-plugin-next-ssg-transform.test.ts
+  // "should not remove extra named export variable declarations"
+  // https://github.com/vercel/next.js/blob/canary/test/unit/babel-plugin-next-ssg-transform.test.ts
+  it("preserves sibling bindings sharing a declaration with a server export", () => {
+    const code = `
+export const keep = 1, getStaticProps = () => ({ props: {} });
+
+export default function Page() {
+  return null;
+}
+`;
+    const result = _stripServerExports(code);
+    expect(result).not.toBeNull();
+    // The non-server sibling binding must survive untouched.
+    expect(result).toContain("keep = 1");
+    // The server export is stubbed, not its initializer kept.
+    expect(result).toContain("getStaticProps = undefined");
+    expect(result).not.toContain("props: {}");
+    expect(() => parseAst(result!)).not.toThrow();
+  });
+
+  // Regression test for #1972: two server exports sharing one declaration must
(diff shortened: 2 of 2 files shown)


james-elicx:
Superseded by #2055

PR #2061, opened 2026-06-16
fix(cache): tag ISR/TPR-seeded entries with the live-render tag set

### LINKED ISSUE #1984: ISR: prerender-seeded cache tags use concrete paths, so typed revalidatePath('/posts/[slug]') can't reach seeded entries
### Problem
A seeded/TPR entry for a dynamic route (e.g. /posts/hello under /posts/[slug]) carries `_N_T_/posts/hello/layout|page`, but a runtime re-render and Next.js-style revalidatePath('/posts/[slug]','layout') target `_N_T_/posts/[slug]/...`. The seeded entry is unreachable by layout/page-typed path revalidation until it is naturally regenerated, leaving stale prerendered HTML/RSC after content updates. Concrete-path revalidatePath('/posts/hello') still works; the gap is specifically the typed/segment-pattern invalidation.

### Evidence
- `packages/vinext/src/server/app-page-cache.ts:184` — buildAppPageCacheTags derives layout/page tags purely from the concrete pathname segments (e.g. `_N_T_/posts/hello/layout`, `_N_T_/posts/hello/page`) with no knowledge of dynamic segments.
- `packages/vinext/src/server/implicit-tags.ts:57` — The live-render tag builder buildPageCacheTags(pathname, extraTags, routeSegments, leafKind) uses routeSegments to emit bracketed tags (`_N_T_/posts/[slug]/layout`, `_N_T_/posts/[slug]/page`) — a different set than the seeded concrete-path tags for the same route.
- `packages/vinext/src/server/seed-cache.ts:130` — seedMemoryCacheFromPrerender tags seeded entries via buildAppPageCacheTags(cachePathname, []) (concrete-path form).
- `packages/vinext/src/cloudflare/tpr.ts:706` — TPR tags uploaded entries via buildAppPageCacheTags(routePath, []) (concrete-path form), inheriting the same mismatch.

In Next.js, every cache entry (both build-time prerender via exportAppPage and runtime re-render via renderToHTMLOrFlightImpl) is tagged through a single function, getImplicitTags(page, pathname). The layout/page derived tags come from getDerivedTags(page), where `page` is the bracketed route pattern (e.g. /articles/[slug]) — work-store.ts:16-22 documents page as "the path to the page file" and export/worker.ts:405-406 shows page is the route pattern while path is the concrete pathname. So a build-prerendered entry for /articles/works carries the bracketed tags _N_T_/articles/[slug]/layout and _N_T_/articles/[slug]/page (plus the concrete _N_T_/articles/works), exactly the same set a runtime re-render produces, and revalidatePath('/articles/[slug]','layout') (revalidate.ts:104-107) emits _N_T_/articles/[slug]/layout which matches. vinext's bug is that its seed/TPR path uses buildAppPageCacheTags(concretePathname) (app-page-cache.ts:180-202), which derives layout/page tags from the concrete segments and has no notion of the bracketed pattern, while its runtime path uses a separate routeSegments-aware builder — a divergence Next.js does not have because it always derives layout/page tags from the bracketed `page` regardless of whether it is prerendering or live-rendering. This is therefore a vinext-only defect.

Citations: .nextjs-ref/packages/next/src/server/lib/implicit-tags.ts:25-48 (getDerivedTags(page)) and :74-93 (getImplicitTags(page, pathname, ...)); .nextjs-ref/packages/next/src/server/app-render/app-render.tsx:2762-2766 (getImplicitTags(workStore.page, resolvedPathname, fallbackRouteParams)) and :1767/:1930/:5731 (tags: [...implicitTags.tags] written onto cache entries); .nextjs-ref/packages/next/src/server/async-storage/work-store.ts:16-22 (page = "path to the page file", i.e. bracketed route pattern); .nextjs-ref/packages/next/src/export/worker.ts:405-406 (page = route pattern, path = concrete pathna

### Suggested fix
Have seed-cache and TPR pass the route's segment pattern into the implicit-tags buildPageCacheTags (the same builder the live render uses) instead of the concrete-path buildAppPageCacheTags, so seeded and rendered entries share an identical tag set.

### Test plan
Seed + `revalidatePath('/posts/[slug]','layout')` test: seeded entry is invalidated; concrete-path revalidation keeps working.

### Related / notes
Live renders tag with bracketed segments (implicit-tags.ts:57); seed-cache.ts:130 and tpr.ts:706 tag with concrete paths — same route, two tag sets. Distinct from #921 (nested layout implicit tags) and #1487.

---

Divkix:
## Problem

Prerender-seeded entries for **dynamic** routes are tagged from the concrete pathname (e.g. `_N_T_/posts/hello/page`), while runtime re-renders tag from the bracketed route pattern (`_N_T_/posts/[slug]/page`). A typed `revalidatePath('/posts/[slug]','layout')` emits `_N_T_/posts/[slug]/layout`, which can never reach a seeded entry — so stale prerendered HTML/RSC survives content updates until natural regeneration. Concrete-path `revalidatePath('/posts/hello')` still works; the gap is the typed/segment-pattern invalidation.

Next.js derives layout/page tags from the bracketed `page` for **both** build-time prerender and runtime re-render (single `getImplicitTags(page, pathname)` path), so seeded and rendered entries share an identical tag set. vinext's seed/TPR path used the concrete-path builder while its live render used a `routeSegments`-aware builder — a vinext-only divergence.

Closes #1984.

## Fix

Make seeded and TPR-uploaded entries carry the **identical** tag set the live render produces.

- **Build prerender** (`prerender.ts`) now records each app route's `routeSegments` — the raw filesystem segments (incl. route groups `(main)` and brackets `[slug]`), the same field the live render reads — into `vinext-prerender.json`.
- **seed-cache** builds tags via `buildAppPageTags(pathname, [], routeSegments)` instead of the concrete-path `buildAppPageCacheTags`. Falls back to the concrete builder when `routeSegments` is absent (legacy manifest / root route) — byte-identical for static routes. _(field-absent, not empty-array, is the fallback trigger: `[]` is valid for `/`)._
- **TPR** has only Cloudflare-analytics URLs (no route pattern), and its local server is the Node prod server (popular pages are served as cache **HITs** with no tags on the response). So the prod server now emits the live-computed implicit tags as an internal-only `x-vinext-implicit-tags` response header — gated strictly to the TPR `User-Agent`, on **both** the cache-hit and fresh-render paths, and stripped from inbound requests. `buildTprKVPairs` reads it verbatim, falling back to the concrete builder when absent.

## Behavior notes

- The exact-path tag (`_N_T_/posts/hello`) is still emitted, so concrete `revalidatePath('/posts/hello')` (no type) keeps working.
- `revalidatePath(concrete, 'page'|'layout')` for a dynamic route no longer matches seeded entries — **aligning with Next.js**, where dynamic-route page/layout revalidation requires the route pattern (`/posts/[slug]`), not a concrete URL.
- Route groups, catch-all / optional catch-all, `@slots`, and i18n are correct by construction (same `routeSegments` source as the live render). Pages Router is untouched (`router === "app"` only).

## Tests

- `tests/seed-cache.test.ts` (+5): dynamic-route bracketed tags; typed `revalidatePath('/posts/[slug]','layout')` invalidation; concrete `revalidatePath('/posts/hello')` still invalidates; route-group `(main)` case; legacy-manifest concrete fallback.
- `tests/tpr-kv-keys.test.ts` (+3): `buildTprKVPairs` uses the `x-vinext-implicit-tags` header verbatim; route-group round-trip; header-absent fallback.
- `tests/request-pipeline.test.ts`: updated for the new internal header (also asserts a forged inbound value is stripped).

Full unit suite green except a pre-existing `better-sqlite3` native-build issue and a `ppr-fallback-shell` timing flake (passes in isolation), both unrelated. Typecheck/format clean for all changed files.


Divkix:
Addressed all three items:
1. (Security): Replaced the spoofable User-Agent gate with a per-build random UUID token. prerenderRoutes now calls randomUUID(), passes the value to the local prod-server subprocess via VINEXT_TPR_SECRET env, and sends it as x-vinext-tpr-secret in every pre-render request. The server checks process.env.VINEXT_TPR_SECRET and the header must match exactly — the env var is never set in a deployed Worker so the response header can never reach a public client. The secret header is also added to VINEXT_INTERNAL_HEADERS so it gets stripped from inbound requests.
2. (Root route): prerender.ts now always writes routeSegments (even []) into the manifest — the length > 0 guard was the bug. With routeSegments: [] present, seed-cache takes the buildAppPageTags branch for root and correctly emits _N_T_/index. Added a focused regression test.
3. (Fresh-render timing): Updated the comment to acknowledge the gap (streaming-collected fetch tags may be missing from the header on the fresh-render path). Left the behavior as-is — TPR is mostly HITs and the path/layout/page implicit tags are always present; fixing it properly would require teeing the stream and seems out of scope.

Issue #1891, opened 2026-06-10
Discussion: route handlers need a dedicated Vite env to fix #1237 + #1246


digwis:
## Background

This is a follow-up discussion to #1246 (and its parent #1237).

#1237 reports that App Route handlers running in the RSC env break `react-dom/server.edge` because the `react-server` export condition is applied transitively to `require('react')` inside `react-dom-server.edge.development.js`.

#1246 documents the full architectural conflict — the only known mitigation (loading route handlers in the SSR env) fixes #1237 cleanly but breaks `'use server'` calls from inside route handlers, because plugin-rsc's SSR-env transform replaces the function body with a server-reference stub whose `callServer` is `null` in that env.

#1246's option 3 ("add a dedicated `route-handler` Vite env") feels like the architecturally correct answer. I'd like to align with the maintainers before pushing more code.

## What's already been tried

I had a branch (`fix/route-handler-lazy-loader` in [digwis/vinext](https://github.com/digwis/vinext/tree/fix/route-handler-lazy-loader), now closed via #1890) that implemented the partial SSR-env fix from #1246's Path B:

- `app-ssr-entry.ts` — accept `AppRoute[]`, statically import every `route.routePath` into a `__appRouteHandlers` map, expose `loadAppRouteHandler(pattern)`.
- `app-rsc-manifest.ts` — drop the eager `getLazyLoaderVar(route.routePath)` registration, emit a thunk per route that calls `import.meta.viteRsc.loadModule("ssr", "index").loadAppRouteHandler(pattern)`.
- `index.ts` — pass resolved routes into `generateSsrEntry`.

Verified end-to-end with `@react-email/render` sending via Cloudflare Email Sending — works for the #1237 case, breaks for the `'use server'` case (exactly as #1246 predicts).

Local verification: 9 route-handler unit test files (86/86) pass; `tsc --noEmit` is clean.

## What I think option 3 looks like

A new Vite environment, e.g. `route-handler`, configured roughly as:

- `resolve.conditions`: `['edge-light', 'workerd', 'browser', 'import', 'module', 'default']` (no `react-server`)
- `optimizeDeps.exclude`: route handler entrypoints that statically import `react-dom/server.edge`
- **Not** transformed by plugin-rsc's SSR `'use server'` transform — so `'use server'` functions imported by route handlers keep their real body and can be called locally. (Server references, where the *client* needs to invoke a server function, stay on the existing path through plugin-rsc.)

The flow becomes:

[...]

PR #2074, opened 2026-06-16
fix(build): preserve lazy RSC framework chunks

### LINKED ISSUE #2073: vinext 0.1.1 regression: RSC framework-chunk split makes a lazy `react-dom/server` import eager → Cloudflare Worker deploy fails (`code: 10021`)
## Summary

After upgrading vinext to **≥ 0.1.1**, `vinext build` succeeds but **`wrangler deploy` / `vinext deploy` fails during Cloudflare Worker validation** with:

```txt
✘ [ERROR] A request to the Cloudflare API (/accounts/<acct>/workers/scripts/<name>/versions) failed.

  Uncaught Error: react-dom/server is not supported in React Server Components.
    at null.<anonymous> (_next/static/framework-<hash>.js:…)
  [code: 10021]
```

The build is clean; the failure happens only when Cloudflare instantiates the uploaded Worker (top-level module evaluation).

Bisected to **0.1.1**, which introduced RSC **"framework chunk"** splitting. The split relocates a module that was reached only through a lazy `import("react-dom/server.edge")` into the **statically-loaded** `framework` chunk. Because `react-dom/server.edge` resolves to React's throwing `react-server` stub under the RSC env's export condition, that throw now executes at **worker startup** instead of never running.

## Versions

```txt
broken on:   vinext 0.1.1, 0.1.2, 0.1.3
last good:   vinext 0.1.0
vite:        8.x (Rolldown)
@vitejs/plugin-rsc: 0.5.26 / 0.5.27
@cloudflare/vite-plugin: 1.3x–1.40
react / react-dom: 19.2.6 / 19.2.7
react-server-dom-webpack: 19.2.6 / 19.2.7
runtime target: Cloudflare Workers (App Router / RSC)
```

## Version boundary (bisect evidence)

The regression is the addition of `createRscFrameworkChunkOutputConfig` to the RSC build output config. Confirmed by unpacking the published tarballs:

| version | `createRscFrameworkChunkOutputConfig` present | RSC build passes `output:` chunk config | deploy |
| --- | --- | --- | --- |
| 0.0.55 | no | no (`input: { index: VIRTUAL_RSC_ENTRY }` only) | ✅ |
| 0.1.0  | no | no (`input: { index: VIRTUAL_RSC_ENTRY }` only) | ✅ |
| 0.1.1  | **yes** | **yes** (`output: createRscFrameworkChunkOutputConfig(viteMajorVersion)`) | ❌ |
| 0.1.2  | yes | yes | ❌ |
| 0.1.3  | yes | yes | ❌ |

In 0.1.1+, on Vite 8 / Rolldown, the function returns:

```js
// dist/build/client-build-config.js
function createRscFrameworkChunkOutputConfig(viteMajorVersion) {
  if (viteMajorVersion >= 8) return { codeSplitting: { groups: [{
    name: "framework",
    test: RSC_FRAMEWORK_CHUNK_TEST   // /[\\/]node_modules[\\/](react|react-dom|…)[\\/]/
  }] } };
  return { manualChunks(id) { return isRscFrameworkModule(id) ? "framework" : void 0; } };
}
```

## Reproduction (public packages, no app-specific code needed)

An App Router app targeting Cloudflare Workers that uses **fumadocs** triggers it out of the box:

```ts
// lib/source.ts  — runs at module scope, reachable from the RSC graph
import { loader } from 'fumadocs-core/source';
export const source = loader({ /* … */ });
```

`fumadocs-core/source`'s `loader()` contains, for menu-icon rendering:

```js
// node_modules/fumadocs-core/dist/source/index.js
const { renderToString } = await import("react-dom/server.edge");
```

Then:

```sh
vinext build        # ✅ succeeds
wrangler deploy     # ❌ fails: react-dom/server is not supported … [code: 10021]
```

More generally: **any module-scope code in the RSC graph that (even lazily) imports `react-dom/server` or `react-dom/server.edge`** will crash on deploy under 0.1.1+. Other common sources of such an import: `@react-email/render`, `@react-aria/ssr`.

## Root cause

1. In the RSC environment, the `react-server` export condition is active, so **`react-dom/server.edge` resolves to `react-dom/server.react-server.js`**, whose entire body is:

   ```js
   throw new Error('react-dom/server is not supported in React Server Components.');
   ```

   (This stub resolution by itself is the subject of #1237 / #1246 / #1891.)

2. **Before 0.1.1:** that module was reached only via a *dynamic* `import("react-dom/server.edge")`, so it lived in its own on-demand chunk that is **never loaded at worker startup**. The throwing stub was inert unless icon rendering actually invoked it at request time (which doesn't happen on these RSC paths).

3. **From 0.1.1:** `RSC_FRAMEWORK_CHUNK_TEST` matches `node_modules/react-dom/…` and **hoists `react-dom/server.edge` out of its lazy chunk into the shared `framework` chunk**. The RSC entry imports the `framework` chunk **statically** (it holds the react-dom core the entry needs synchronously). So the stub's top-level `throw` now evaluates when the Worker is instantiated → Cloudflare deploy validation fails (`code: 10021`).

Confirmed by inspecting the emitted bundle:

- **0.1.1+ (broken):** `dist/server/_next/static/framework-<hash>.js` is statically imported by `dist/server/index.js`, and contains a **top-level** `…;throw Error(\`react-dom/server is not supported…\`)`. Importing that chunk in node throws immediately.
- **With the split disabled (`codeSplitting: false`):** no `framework-*.js` chunk is emitted; the stub reverts to a **lazy thunk** `wrapper((()=>{throw Error(…)}))` inside `index.js` that only runs if actually accessed. Importing `index.js` no longer throws the react-dom/server error.

## Expected behavior

The RSC framework-chunk split must not move a lazily-imported module onto the eager worker-startup path, and must not emit an RSC server chunk graph that fails Cloudflare Worker startup validation.

## Workaround

Patch the Vite-8 branch of `createRscFrameworkChunkOutputConfig` (`dist/build/client-build-config.js`) to disable the split:

```diff
- if (viteMajorVersion >= 8) return { codeSplitting: { groups: [{
-   name: "framework",
-   test: RSC_FRAMEWORK_CHUNK_TEST
- }] } };
+ if (viteMajorVersion >= 8) return { codeSplitting: false };
```

Result: `vinext build` still succeeds, no `framework-*.js` RSC chunk is emitted, and `wrangler deploy` passes. (Equivalently, pin vinext to **0.1.0**, the last release before the split.)

## Related issues

- **#1237**, **#882** — same `react-dom/server is not supported` error, but triggered by `@react-email/render` inside **App Route handlers**. Those are about the underlying `react-server`-condition stub resolution, not the chunk split. This report's trigger is module-scope (fumadocs `loader()`), and the *new* failure mode is the **0.1.1 framework-chunk split** making the otherwise-inert stub fatal at startup.
- **#1246**, **#1891** — architectural discussion of a dedicated route-handler Vite env to fix the stub resolution. Even if that lands, the framework-chunk split independently produces an unsafe eager server chunk (see next point).
- A **second manifestation of the same 0.1.1 split** (no react-dom/server involved): a `db-*.js` RSC chunk that back-imports the RSC entry (`../../index.js`) produces a cyclic eager graph → `Uncaught TypeError: a is not a function [code: 10021]` at startup. Harder to reduce to a public repro (needs a particular server dependency graph), but disappears with the same `codeSplitting: false` workaround — suggesting the framework-chunk split itself emits unsafe eager/cyclic RSC server chunks under Vite 8 / Rolldown + Cloudflare Workers.

james-elicx:
## Summary

- preserve dynamic entry boundaries within the Vite 8/Rolldown RSC framework chunk group
- prevent lazy `react-dom/server.edge` stubs from being evaluated at Worker startup
- retain the framework splitting needed for `global-not-found` CSS ordering
- add production regression coverage for importing the built RSC entry

## Root cause

Rolldown's framework group merged every matching React module into one chunk by default. A lazy `import("react-dom/server.edge")` could therefore place React's throwing `server.react-server.js` stub in the same framework chunk statically imported by the RSC entry.

Setting `entriesAware: true` keeps matching framework modules grouped by the entry points that actually use them. Lazy-only React modules remain behind their dynamic import instead of moving onto the Worker startup path.

## Next.js comparison

Next.js does not create its client-style framework chunk for App Router server builds. It also aliases legacy `react-dom/server.edge` imports in server-component builds. This change addresses vinext's additional eager-chunk regression without attempting to resolve the separate RSC export-condition behavior tracked in #1237 and #1891.

## Validation

- `vp test run tests/build-optimization.test.ts tests/app-router-global-not-found-css-order.test.ts` — 116 tests passed
- `vp check` — formatting, lint, and type checks passed
- rebuilt `vinext` and `examples/fumadocs-docs-template`
- verified the throwing React stub is emitted in a separate lazy framework subgroup
- verified importing `dist/server/index.js` does not throw
- verified `wrangler deploy --dry-run --no-bundle` accepts the Fumadocs artifact

Fixes #2073

### DIFF
--- a/packages/vinext/src/build/client-build-config.ts
+++ b/packages/vinext/src/build/client-build-config.ts
@@ -219,7 +219,13 @@ export function createRscFrameworkChunkOutputConfig(viteMajorVersion: number) {
   if (viteMajorVersion >= 8) {
     return {
       codeSplitting: {
-        groups: [{ name: "framework", test: RSC_FRAMEWORK_CHUNK_TEST }],
+        groups: [
+          {
+            name: "framework",
+            test: RSC_FRAMEWORK_CHUNK_TEST,
+            entriesAware: true,
+          },
+        ],
       },
     };
   }
--- a/packages/vinext/src/build/client-build-config.ts
+++ b/packages/vinext/src/build/client-build-config.ts
@@ -7,6 +7,15 @@ type ClientAssetFileNameInfo = {
   readonly originalFileNames?: readonly string[];
 };
 
+type RscFrameworkModuleInfo = {
+  importers: string[];
+  isEntry: boolean;
+};
+
+type RscFrameworkManualChunksMeta = {
+  getModuleInfo(id: string): RscFrameworkModuleInfo | null;
+};
+
 // Next.js emits CSS under `static/css/` and CSS url() dependencies (images,
 // fonts, …) under `static/media/`, both with an 8-char content hash. Mirror
 // that layout so migrated apps keep stable, Next-shaped asset URLs.
@@ -208,12 +217,29 @@ export function isRscFrameworkModule(id: string): boolean {
   return pkg !== null && (FRAMEWORK_PACKAGES as readonly string[]).includes(pkg);
 }
 
+function isStaticallyReachableFromEntry(
+  id: string,
+  meta: RscFrameworkManualChunksMeta,
+  visited = new Set<string>(),
+): boolean {
+  if (visited.has(id)) return false;
+  visited.add(id);
+
+  const moduleInfo = meta.getModuleInfo(id);
+  if (!moduleInfo) return false;
+  if (moduleInfo.isEntry) return true;
+  return moduleInfo.importers.some((importer) =>
+    isStaticallyReachableFromEntry(importer, meta, visited),
+  );
+}
+
 /**
  * Output config that isolates React (and the RSC flight runtime) into a
  * dedicated "framework" chunk in the RSC server build. Returns the bundler-
  * appropriate shape: rolldown's `codeSplitting` for Vite 8+, Rollup's
  * `manualChunks` for Vite 7. See {@link RSC_FRAMEWORK_CHUNK_TEST} for the
- * motivation (issue #1549).
+ * motivation (issue #1549). Framework modules that are only reachable through
+ * dynamic imports must stay out of the eager framework chunk (issue #2073).
  */
 export function createRscFrameworkChunkOutputConfig(viteMajorVersion: number) {
   if (viteMajorVersion >= 8) {
@@ -223,6 +249,8 @@ export function createRscFrameworkChunkOutputConfig(viteMajorVersion: number) {
           {
             name: "framework",
             test: RSC_FRAMEWORK_CHUNK_TEST,
+            // Split by the entries that use each module so lazy framework
+            // imports cannot become eager Worker-startup dependencies (#2073).
             entriesAware: true,
           },
         ],
@@ -230,8 +258,10 @@ export function createRscFrameworkChunkOutputConfig(viteMajorVersion: number) {
     };
   }
   return {
-    manualChunks(id: string): string | undefined {
-      return isRscFrameworkModule(id) ? "framework" : undefined;
+    manualChunks(id: string, meta: RscFrameworkManualChunksMeta): string | undefined {
+      return isRscFrameworkModule(id) && isStaticallyReachableFromEntry(id, meta)
+        ? "framework"
+        : undefined;
     },
   };
 }
--- a/packages/vinext/src/build/client-build-config.ts
+++ b/packages/vinext/src/build/client-build-config.ts
@@ -64,9 +64,10 @@ export function createClientAssetFileNames(assetsDir: string) {
  * (node_modules/.pnpm/pkg@ver/node_modules/pkg).
  */
 function getPackageName(id: string): string | null {
-  const nmIdx = id.lastIndexOf("node_modules/");
+  const normalizedId = id.replaceAll("\\", "/");
+  const nmIdx = normalizedId.lastIndexOf("node_modules/");
   if (nmIdx === -1) return null;
-  const rest = id.slice(nmIdx + "node_modules/".length);
+  const rest = normalizedId.slice(nmIdx + "node_modules/".length);
   if (rest.startsWith("@")) {
     // Scoped package: @org/pkg
     const parts = rest.split("/");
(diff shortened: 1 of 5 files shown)


james-elicx:
Addressed the test-design finding in a655ef58. The production build now completes in `beforeAll`, but preview startup is lazy and only happens inside the CSS tests. This guarantees the direct `dist/server/index.js` import test runs independently before any preview server evaluation. I also temporarily removed `entriesAware: true` to verify the negative case: the named `does not eagerly evaluate dynamically imported React server stubs` test now fails directly with `react-dom/server is not supported in React Server Components` instead of being skipped during setup. Restored the fix; focused tests and `vp check` pass.

james-elicx:
Addressed both findings in 4e5d25c8. The Vite 7 `manualChunks` path now uses Rollup’s `getModuleInfo()` graph to assign a framework module only when it is statically reachable from a real entry. Framework modules reachable only through dynamic imports are left to automatic chunking, which mirrors the relevant `entriesAware` behavior instead of merely documenting the gap. Added Vite 7 unit coverage for eager React/RSC modules versus a lazy `react-dom/server.react-server.js` module. Also documented the #2073 reason next to `entriesAware: true` and in the function comment. Full `tests/build-optimization.test.ts`, the production startup/CSS regression suite, and `vp check` pass.

james-elicx:
Addressed the remaining coverage finding in e0335f52. The Vite 7 `manualChunks` unit graph now includes a multi-hop static path (`entry -> middleman -> react/index.js`), so the recursive `isStaticallyReachableFromEntry` branch is exercised directly while the lazy-only React DOM stub remains excluded. Focused chunking tests and targeted `vp check` pass.

james-elicx:
Addressed both minor findings in adc12877. `getPackageName()` now normalizes Windows separators before extracting package names, with Windows-path framework matching coverage. The production regression suite now explicitly guards preview startup behind successful completion of the direct RSC-entry import assertion, so the test-order dependency is enforced rather than implicit. Focused suites (116 tests) and targeted `vp check` pass.

PR #2075, opened 2026-06-16
perf(dev): prune NODE_ENV during dependency optimization

TheAlexLichter:
This PR improves cold start time by replacing `NODE_ENV` during Rolldown's prebundling step, resulting in less JavaScript dependencies during dev.

Over 10x10 runs:

- main: 1417.5 ms
- new branch: 1309.3 ms
- Difference: **108.2 ms faster**


TheAlexLichter:
Decoupled the test and added comments 👍

PR #2077, opened 2026-06-16
perf(rsc): prebundle static renderer entry

TheAlexLichter:
This PR prebundles `react-server-dom-webpack/static.edge` during RSC dependency optimization when an RSC env is detected

This avoids transforming its large CommonJS renderer on the first request.
In turn, cold start is improved by ± 90ms (mainly the first request is sped up through this change).


TheAlexLichter:
Addressed both non-blocking issues ☺️

PR #2080, opened 2026-06-16
perf(rsc): avoid client decoder in server graph

TheAlexLichter:
This PR adds a lightweight development RSC runtime that avoids loading the client Flight decoder. The full serialization exports for "use cache" functions is preserved. It also adds a resolver and private-cache deduplication regression coverage.

This reduces the cold start time.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -568,6 +568,18 @@ function createStaticImageAsset(imagePath: string): { fileName: string; source: 
  */
 const _shimsDir = normalizePathSeparators(path.resolve(__dirname, "shims")) + "/";
 const _fontGoogleShimPath = resolveShimModulePath(_shimsDir, "font-google");
+const _cacheRuntimeShimPath = normalizePathSeparators(
+  resolveShimModulePath(_shimsDir, "cache-runtime"),
+);
+
+function isCacheRuntimeImporter(importer: string | undefined): boolean {
+  if (!importer) return false;
+  const cleanImporter = normalizePathSeparators(stripViteModuleQuery(importer));
+  return (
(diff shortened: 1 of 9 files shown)


packages/vinext/src/index.ts
```diff
@@ -2624,6 +2636,15 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
...
           const cleanId = id.startsWith("\0") ? id.slice(1) : id;
 
+          if (
+            this.environment?.name === "rsc" &&
+            this.environment.config?.command === "serve" &&
+            !isCacheRuntimeImporter(importer) &&
+            stripViteModuleQuery(cleanId).endsWith("/@vitejs/plugin-rsc/dist/react/rsc.js")
+          ) {
+            return resolveShimModulePath(path.resolve(__dirname, "server"), "app-rsc-runtime");
+          }
```
james-elicx, line 2646: i would prefer to avoid filtering on a package's build directory if possible...

I can't help but feel like this is more of a temporary workaround for development mode, rather than something that makes sense long-term.

the need for the cache-runtime exclusion suggests to me that this could throw up additional edge cases or unintended behaviour in the future.

what sort of perf benefit were you seeing with this? is there something we may be able to make available in the rsc plugin for this rather than putting it together ourselves?

TheAlexLichter: Yeah thats totally fair. It is quite brittle and rather something for upstream I think (cc @hi-ogawa). Maybe the plugin-rsc can expose a server-only runtime.

Perf wins were ~60-90ms, so not too shabby. Let me see if I can bring that upstream somehow.

TheAlexLichter:
(Closing due to https://github.com/cloudflare/vinext/pull/2080#discussion_r3430612348)

PR #2114, opened 2026-06-17
fix(css): keep module hashes stable with url assets

### LINKED ISSUE #2111: App Router CSS Modules hash mismatch when module contains CSS url() assets
## Summary

App Router production builds can generate different CSS Module class names for SSR/server output vs the client JS/CSS when a CSS Module contains a relative `url(...)` asset reference.

This breaks SSR-rendered markup: the server-side class map points at one hash family, but the browser only loads CSS for a different hash family, so the SSR HTML is effectively unstyled until/unless client rendering replaces it. In our real app this made a hero carousel's absolutely-positioned layers fall back to normal flow and stack vertically.

## Reproduction

Minimal app shape:

```tsx
// app/page.tsx
import { Hero } from '../components/hero'
export default function Page() { return <Hero /> }
```

```tsx
// components/hero.tsx
'use client'
import styles from './hero.module.scss'
export function Hero() {
  return <main className={styles.hero}><span className={styles.icon}>icon</span></main>
}
```

```scss
/* components/hero.module.scss */
.hero { position: relative; color: red; }
.icon { mask-image: url('../assets/icon.svg'); width: 24px; height: 24px; display: block; }
```

```ts
// vite.config.ts
import { cloudflare } from '@cloudflare/vite-plugin'
import vinext from 'vinext'
import { defineConfig } from 'vite'

export default defineConfig({
  plugins: [
    vinext(),
    cloudflare({ viteEnvironment: { name: 'rsc', childEnvironments: ['ssr'] } }),
  ],
})
```

Build and inspect class tokens in `dist/server/**/*.js`, `dist/client/**/*.js`, and `dist/client/**/*.css`:

```sh
npm run build
node check-css-module-hashes.mjs
```

Observed on `vinext@0.1.4`:

```json
{
  "server": [
    "_hero_ois0b_1",
    "_icon_ois0b_6"
  ],
  "client": [
    "_hero_50rj6_1",
    "_icon_50rj6_6"
  ],
  "css": [
    "_hero_50rj6_1",
    "_icon_50rj6_6"
  ]
}
```

Expected: server, client JS, and client CSS should all use the same scoped CSS Module names.

## Root cause found while reducing

This is introduced by `f8481ac04d898968429a607951c47cc857f70b7e` (`fix(css): preserve distinct media filenames for CSS url() assets (#1725)`). I verified the exact boundary with a source build of `packages/vinext`:

```text
f8481ac0^ (4ffd5bf2):
server = client = css = _hero_ois0b_1 / _icon_ois0b_6

f8481ac0:
server = _hero_ois0b_1 / _icon_ois0b_6
client = css = _hero_50rj6_1 / _icon_50rj6_6
```

Instrumenting `css.modules.generateScopedName` shows that the same `hero.module.scss` reaches PostCSS Modules with different CSS text depending on environment:

```diff
.indicatorIconSelection {
-  mask-image: url("../../assets/home/hero/selection.svg?vinext_css_url_asset=selection.svg");
+  mask-image: url("../../assets/home/hero/selection.svg");
}
```

`postcss-modules`' default scoped name hashes the CSS text, so adding the private `?vinext_css_url_asset=...` marker in only one environment changes every class hash in the module, not only the rule containing `url(...)`.

## Version checks

- `vinext@0.0.52`: good in the original app; SSR/client CSS Module hashes match.
- `vinext@0.1.0` through `vinext@0.1.4`: bad in the original app; SSR/server uses one hash family and client JS/CSS another.
- The commit above is contained by all `vinext@0.1.x` tags.

## Notes

The issue only needs a CSS Module with a relative CSS `url(...)` asset reference to trigger. A custom `css.modules.generateScopedName` based only on filename + local name works around it, but the default config should not emit divergent class maps across SSR/client environments.

james-elicx:
## Summary

- apply the temporary CSS URL provenance marker consistently across production build environments
- keep CSS Module hash input identical between RSC/SSR and client builds
- retain client-only restoration of distinct media filenames
- add regression coverage comparing server JS, client JS, and emitted client CSS class names

## Root cause

The CSS URL asset repair added `?vinext_css_url_asset=...` only during the client CSS transform. Vite's default CSS Modules scoped name hashes the full CSS text, so App Router server builds and client builds generated different class maps for modules containing relative `url(...)` references.

The marker is now added in every production build environment. The client restore phase remains responsible for emitting deduplicated sibling assets and removing the marker from final output.

## Validation

- `vp test run tests/css-url-assets.test.ts tests/ssr-css-assets.test.ts`
- `vp test run tests/build-optimization.test.ts -t "multi-env build scopes treeshake"`
- `vp check packages/vinext/src/index.ts packages/vinext/src/build/css-url-assets.ts tests/css-url-assets.test.ts`

Fixes #2111

### DIFF
--- a/packages/vinext/src/build/css-url-assets.ts
+++ b/packages/vinext/src/build/css-url-assets.ts
@@ -13,12 +13,14 @@ import type { Rolldown } from "vite";
  * vitejs/vite#8632). The one native escape hatch is `this.emitFile({ fileName })`:
  * assets emitted with an explicit `fileName` are *never* deduped. So we:
  *
- *   1. mark — during the client CSS transform, tag each relative asset `url()`
- *      with a private `?vinext_css_url_asset=<source-basename>` query. This is
- *      the only durable carrier of per-reference provenance: once Rolldown
- *      dedupes, the emitted bundle cannot tell which `url()` came from which
- *      source, and the bundle metadata (`originalFileNames`) records the set of
- *      sources but not the mapping per reference.
+ *   1. mark — during every production CSS transform, tag each relative asset
+ *      `url()` with a private `?vinext_css_url_asset=<source-basename>` query.
+ *      Using identical transformed CSS in server and client environments is
+ *      required because the default CSS Modules scoped name hashes the full CSS
+ *      text. In the client build, the marker is also the only durable carrier
+ *      of per-reference provenance: once Rolldown dedupes, the emitted bundle
+ *      cannot tell which `url()` came from which source, and bundle metadata
+ *      records the source set but not the mapping per reference.
  *
  *   2. restore — at `generateBundle`, read each marked reference's source
  *      basename back out. When it differs from the deduped output's basename,
@@ -149,7 +151,7 @@ function rewriteCssUrls(code: string, replace: (rawUrl: string) => string | null
 /**
  * Append the private provenance marker to each relative asset `url()` in a CSS
  * source. Idempotent and side-effect free (only adds a query param), so it is
- * safe to run on every client-environment stylesheet, vendored CSS included.
+ * safe to run on every build-environment stylesheet, vendored CSS included.
  */
 export function markCssUrlAssetReferences(code: string, id: string): string | null {
   if (!isCssRequest(id) || !code.includes("url(")) return null;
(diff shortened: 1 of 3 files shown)


james-elicx:
Addressed the server-emitted CSS marker leak in `c45fc4c8`:

- run the restore pass in every production build environment, not only `client`
- strip the private marker from server CSS and emit distinct sibling assets there too
- force non-inlined assets in the App Router regression build and assert server CSS has no marker and preserves separate `dark` / `dark2` URLs

Validation:
- `vp test run tests/css-url-assets.test.ts tests/ssr-css-assets.test.ts`
- `vp test run tests/build-optimization.test.ts -t "multi-env build scopes treeshake"`
- `vp check packages/vinext/src/index.ts packages/vinext/src/build/css-url-assets.ts tests/css-url-assets.test.ts`

PR #2120, opened 2026-06-17
perf(ssr): isolate open redirect detection

TheAlexLichter:
The App Router SSR imported two small helpers from large runtime modules. In turn Vite has to transform the unrelated request, cache, response, and stream code on the first request.

Moving them into dependency-free modules reduces the initial graph and improves cold start times.

(Yes, these are somewhat micro-optimizations but on the plugin level these are worth it IMO. Here we get ~40ms)

### DIFF
--- a/packages/vinext/src/server/app-page-stream.ts
+++ b/packages/vinext/src/server/app-page-stream.ts
@@ -268,7 +268,6 @@ export async function renderAppPageHtmlStream(
  * response body). This is the correct place to clear per-request context,
  * because the RSC/SSR pipeline is lazy — components execute while the stream
  * is being consumed, not when the stream handle is first obtained.
- */
 export async function renderAppPageHtmlResponse(
   options: RenderAppPageHtmlResponseOptions,
 ): Promise<Response> {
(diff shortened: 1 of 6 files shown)


packages/vinext/src/server/app-page-stream.ts
```diff
@@ -266,54 +269,6 @@ export async function renderAppPageHtmlStream(
  * because the RSC/SSR pipeline is lazy — components execute while the stream
  * is being consumed, not when the stream handle is first obtained.
  */
```
james-elicx:
```suggestion

```

packages/vinext/src/server/request-pipeline.ts
```diff
@@ -78,25 +81,6 @@ export function guardProtocolRelativeUrl(rawPathname: string): Response | null {
  * encoded delimiter) so malformed suffixes can still reach the normal
  * "400 Bad Request" decode path instead of being masked as "404".
  */
```
james-elicx:
```suggestion
```

packages/vinext/src/server/app-page-stream.ts
```diff
@@ -265,55 +268,6 @@ export async function renderAppPageHtmlStream(
  * response body). This is the correct place to clear per-request context,
  * because the RSC/SSR pipeline is lazy — components execute while the stream
  * is being consumed, not when the stream handle is first obtained.
```
james-elicx:
```suggestion
```

james-elicx: what on earth is github doing. completely messing up the suggestions

PR #2138, opened 2026-06-18
refactor(shims): split NextRequest primitives from next server

NathanDrake2406:
## Summary

This repurposes the old route-handler cold-start PR into a no-behavior-change shim refactor.

The public `next/server` shim still exports the same request primitives, but `NextURL`, `RequestCookies`, request sealing helpers, URL validation, and `NextFetchEvent` now live in `server-request-primitives.ts`. `server.ts` keeps the public `NextRequest`, `NextResponse`, cookie response, middleware, user-agent, and `after()` surfaces.

The previous route-handler optimization has been removed. Route handlers continue to construct and receive real `NextRequest` instances.

## What changed

| Area | Before | After |
| --- | --- | --- |
| Request primitives | Implemented inline in `shims/server.ts` | Moved into `shims/server-request-primitives.ts` |
| Public `next/server` API | Exported directly from one large file | Re-exports the same primitives from the split module |
| Route-handler runtime | Was changed in the old perf version | Restored to the existing `NextRequest` construction path |
| Runtime behavior | No intended API change | No intended API change |

## Validation

- `vp check packages/vinext/src/shims/server.ts packages/vinext/src/shims/server-request-primitives.ts tests/shims.test.ts tests/app-route-handler-runtime.test.ts`
- `vp test run tests/app-route-handler-runtime.test.ts tests/shims.test.ts`
- `vp run vinext#build`
- pre-commit checks


NathanDrake2406:
Closing this for now. The split is tidy, but it does not currently change behavior or create a meaningful runtime boundary because consumers still go through `vinext/shims/server`. Since `server.ts` is large but still cohesive as the `next/server` public shim, this feels like churn without enough payoff.\n\nIf we want this boundary later, it should come with the follow-on change that actually benefits from it, such as moving internal runtime consumers to a non-public primitive module or reducing a measured import/cold-start cost.

PR #2149, opened 2026-06-18
chore: upgrade vite-plus to 0.2.1

fengmk2:
v0.2.1 consumes upstream Vitest directly: the @voidzero-dev/vite-plus-test wrapper is removed and vitest now comes in transitively through vite-plus.

- catalog: vite-plus 0.1.24 -> 0.2.1, vite -> @voidzero-dev/vite-plus-core@0.2.1
- catalog: vitest -> upstream 4.1.9 (was the vite-plus-test alias), @vitest/coverage-istanbul 4.1.6 -> 4.1.9
- drop the vitest peerDependencyRules accommodations for the wrapper/old vitest
- decode Vitest's flatted blob format directly in scripts/lib/vitest-blob-timings.mjs (the removed wrapper bundled the parser this script previously reached into)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -94,8 +94,8 @@ catalogs:
       specifier: ^0.5.27
       version: 0.5.27
     '@vitest/coverage-istanbul':
-      specifier: 4.1.6
-      version: 4.1.6
+      specifier: 4.1.9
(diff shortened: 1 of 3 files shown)


james-elicx (approved):
sweet, thanks

PR #2150, opened 2026-06-18
perf(app-router): defer prerender endpoint & file metadata runtimes

TheAlexLichter:
Another PR that defers code until it is needed. This time, prerender endpoint and file-based metadata runtimes.

## Why?

Typical App Router page requests do not use these paths.

Loading them lazily reduces development cold-start and first-page compilation work without changing prerender or metadata behavior.

Cold start time got reduced by ~22ms on my machine


NathanDrake2406:
I was inspired by your PR and did some deferring too 😆 
Is this the secret sauce for optimising?

TheAlexLichter:
@NathanDrake2406 It is definitely part of the mix to reduce cold start times, yes. Most low hanging fruits are in by now (and some tradeoffs for prod perf are not worth making)

PR #2152, opened 2026-06-18
fix(fonts): preload only requested google font subsets

### LINKED ISSUE #1897: `next/font/google` ignores `subsets` config and inlines massive `@font-face` CSS blocks
### Description
When using `next/font/google` in a Vinext project, there are two intertwined issues that cause significant performance regressions compared to standard Next.js (Vercel) deployments:

1. **`subsets` configuration is completely ignored:** Even when explicitly configured to only load the `"latin"` subset, Vinext preloads every available unicode subset (Cyrillic, Greek, Vietnamese, etc.) for that font.
2. **`@font-face` rules are inlined instead of externalized:** Vinext drops all generated `@font-face` declarations into a massive inline `<style data-vinext-fonts>` block directly inside the HTML `<body>`, rather than generating a cacheable external `.css` stylesheet.

Because of issue 1, issue 2 is severely amplified. In my project, using just two fonts (`Inter` and `IBM_Plex_Sans`) adds **13 unnecessary `.woff2` preloads** and **~12.7 KB of uncompressed inline CSS** to every single HTML response.

### Steps to Reproduce
1. Configure a font in `layout.tsx` explicitly restricting it to the latin subset:
```tsx
import { Inter } from "next/font/google";

const inter = Inter({
  subsets: ["latin"],
  variable: "--font-satoshi",
  display: "swap",
});
```
2. Build and deploy to Cloudflare.
3. Inspect the returned HTML `<head>` and `<body>`.

### Expected Behavior (Standard Next.js behavior)
- Only the `latin` subset's `<link rel="preload">` and `@font-face` rules should be generated.
- Generated `@font-face` declarations should be extracted into a standard external stylesheet chunk (e.g., `/_next/static/css/...`) so they can be cached by the browser and CDN, keeping the initial HTML document small.

### Actual Behavior (Vinext behavior)
- The browser is forced to open 13+ unnecessary network connections on the first visit to download fonts it will never use.
- The browser cannot cache the 12.7 KB CSS block. Returning visitors must re-download that payload on every page transition/visit because it is baked into the HTML.

TheAlexLichter:
This PR fixes https://github.com/cloudflare/vinext/issues/1897. Now `vinext` respects subsets and `preload: false` when preloading Google Fonts.

- Ports Next.js’s CSS subset parser into `build/google-fonts/`.
- Computes `preload` URLs during the build.
- Passes an explicit `preload` list to the runtime instead of preloading every URL found in the CSS.
- Keeps existing inline font-face CSS behavior unchanged.
- Deduplicates shared font files.
- Adds focused parser, build, and production coverage.

This intentionally excludes #1899’s external stylesheet refactor so that CSS delivery and possible performance tradeoffs can be evaluated separately (needs a bit of benchmarking).


TheAlexLichter:
+57% seems weird. investigating

couldnt repro locally.

NathanDrake2406:
> +57% seems weird. investigating
> 
> couldnt repro locally.

noisy benchmark. I have an idea, you know how the App Router is famously over complex and over-engineered and that most apps don't use all the machinery? What about building something that detects the size of the app, then load accordingly?

TheAlexLichter:
> > +57% seems weird. investigating
> > couldnt repro locally.
> 
> noisy benchmark.

Workin on fixing this so baseline runs on the same runner too.

> I have an idea, you know how the App Router is famously over complex and over-engineered and that most apps don't use all the machinery? What about building something that detects the size of the app, then load accordingly?

That isn't only size dependent though, no?

NathanDrake2406:
> > > +57% seems weird. investigating
> > > couldnt repro locally.
> > 
> > 
> > noisy benchmark.
> 
> Workin on fixing this so baseline runs on the same runner too.
> 
> > I have an idea, you know how the App Router is famously over complex and over-engineered and that most apps don't use all the machinery? What about building something that detects the size of the app, then load accordingly?
> 
> That isn't only size dependent though, no?

I meant like simple app -> simple handler, complex app -> complex handler

james-elicx:
Something seems fishy. Two re-runs and both with large increases for vinext, with only small increases for nextjs. Merge in latest main and now a decrease for both.

james-elicx (approved):

james-elicx:
Ah the big jump was due to it being based on an older commit and the benchmark running with the head sha rather than github's merge sha for the workflow run

TheAlexLichter:
Ohh, makes sense!

PR #2156, opened 2026-06-18
feat(use-cache): support callable cached functions with rsc plugin api

james-elicx:
## Summary

POC for implementing callable `"use cache"` server references entirely in Vinext user land instead of adding the `serverFunctionDirectives` option and orchestration plugin originally proposed in vitejs/vite-plugin-react#1246.

This now targets the stable `@vitejs/plugin-rsc@0.5.34` release and follows the maintainer's final callable-cache plugin shape:

https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-rsc/examples/use-cache-persistent/callable-cache-plugin.ts

This PR includes the still-open #1871 callable-cache foundation and references it below.

## Plugin composition

Vinext inserts `vinext:server-function-directives` immediately before `rsc:use-server` and uses plugin-rsc's public transform primitives:

When `rsc()` is configured manually, `vinext({ rsc: false })` also installs the Vinext directive plugin. Vinext must appear before `rsc()` in the Vite plugin array; reversed order now fails fast instead of silently dropping cache transforms.

- `transformWrapExport()` for module-level `"use cache"`
- `transformHoistInlineDirective()` for inline `"use cache"`
- `transformDirectiveProxyExport()` for SSR/client proxies
- `RscPluginManager.serverReferences.resolve()` for canonical dev/build identities
- `replaceClaim()` and `deleteClaim()` for Vinext-owned references

The generated RSC transform imports `registerServerReference` from `@vitejs/plugin-rsc/react/rsc/server`.

With plugin-rsc 0.5.34, compatible Vinext and `rsc:use-server` claims can coexist for mixed directive modules. Vinext no longer deletes or takes ownership of plugin-rsc's claims. It preserves leading file directives when injecting runtime imports so the built-in transform can run afterwards.

That supports both mixed cases:

- file-level `"use server"` with an inline `"use cache"` function
- file-level `"use cache"` with an inline `"use server"` function

## Differences from the upstream example

The orchestration now matches the upstream example, while Vinext keeps framework-specific cache behavior:

- parses and validates `"use cache: <kind>"` variants
- passes Vinext cache identity and variant data to `registerCachedFunction()`
- records `argumentCount` and `acceptsSecondArgument` for metadata/viewport behavior
- marks cached App Page default exports for page-specific runtime validation
- leaves layout/template default exports uncached because their `children` slot is not cache-serializable yet
- encrypts closure captures through Vinext's cache runtime envelope
- rejects standalone inline cache functions in client/SSR graphs with a Next-compatible diagnostic
- resolves plugin-rsc from the application project root so linked/workspace installs share the application's Vite instance

Vinext's cache runtime, persistence, invalidation, cache kinds, and request API restrictions remain outside plugin-rsc.

## Cache replay

Cached Flight replay uses plugin-rsc's supported API:

```ts
createFromReadableStream(stream, {}, { preserveServerReferences: true })
```

This preserves opaque server references while replaying cached RSC without importing their implementations into the replaying RSC runtime.

## Upstream requirements

This POC now depends on the public features shipped in `@vitejs/plugin-rsc@0.5.34`:

- low-level directive transform primitives
- `getPluginApi()` access to `RscPluginManager`
- independently owned server-reference claims and compatible claim aggregation
- mixed `"use cache"` / `"use server"` transform composition
- `preserveServerReferences` during cache replay
- split RSC runtime entrypoints

It does **not** depend on:

- a `serverFunctionDirectives` plugin option
- plugin-rsc-owned custom directive orchestration
- private transform markers
- direct writes to `serverReferenceMetaMap`
- local copies of plugin-rsc reference hashing or Vite URL normalization

## Validation

- `vp check`
- frozen lockfile install with `@vitejs/plugin-rsc@0.5.34`
- 36/36 server-function directive transform tests
- 13/13 development App Router cache/HMR E2Es
- 8/8 RSC plugin registration integration tests
- 4/4 production callable-cache E2Es
- production direct Client Component calls for file-level cache modules with inline server overrides
- production nested cached-function action round trips and cached Flight replay

Refs #1871


hi-ogawa:
I've added a support of user land server reference registration via https://github.com/vitejs/vite-plugin-react/pull/1310, which is expected to simplify manual server reference management from use cache transform plugin here.

packages/vinext/src/plugins/use-cache-callable.ts
```diff
@@ -0,0 +1,374 @@
+import { createRequire } from "node:module";
```
james-elicx, line 1: /bigbonk review for issues

james-elicx:
Coordinator mapping for deploy-suite run 31290819291: this merged PR appears to own the two callable-cache failures `test/e2e/app-dir/use-cache-with-server-function-props/use-cache-with-server-function-props.test.ts` (`should be able to use nested cache functions as props`) and `test/e2e/app-dir/cache-components/cache-components.server-action.test.ts` (`should prerender pages with inline server actions`). I am skipping a duplicate implementation because this PR is by `james-elicx`; current-main targeted Next.js E2E re-verification is being tracked separately.

james-elicx:
Correction after checking the post-merge baseline exactly: run 31350589613 tested main commit `99623542d` (which contains merge commit `373e8264a` from this PR). In that report, `use-cache-with-server-function-props > should be able to use nested cache functions as props` **passes**, so this PR does own/fix that row. However `cache-components > should prerender pages with inline server actions` still fails on the same post-merge baseline, so that second row is **not** fixed by #2156 and remains in the active backlog. Please use this correction rather than my preceding broader mapping comment.

PR #2159, opened 2026-06-18
chore(perf): pair pull request benchmark baselines

TheAlexLichter:
With this PR, PR benchmarks now measure the base and head revisions together on the same runner, alternating execution order across rounds.

This produces more reliable comparisons than using a historical baseline, where differences in runner hardware or load. Alternating the order also reduces warmup and run-order bias.

To reduce the additional overhead, Next.js benchmarks are skipped when a trusted fingerprint confirms their inputs are unchanged. 

The dashboard and PR comment now preserve and explain whether each baseline is paired, historical, or unavailable. Publishing also validates fingerprints and skipped implementations before accepting results.


james-elicx:
Codex:
- Kept fork benchmark runs isolated from the trusted baseline and result files.
- Made retries and overlapping runs safer so older results cannot replace newer ones.
- Added safer handling for deployment timing, manual runs, and generated benchmark files.
- Added coverage for the new safeguards and fixed a flaky local test setup.
- Verified the focused tests and the full check command pass.

james-elicx:
Codex:
- Fixed the performance job so isolated builds can finish safely.
- Kept the trusted benchmark files available throughout the run.
- Added checks that real traces still include vinext, Vite, and Rolldown data for the dashboard filters.
- Verified the full performance run and its uploaded trace artifacts.

PR #2166, opened 2026-06-18
refactor(shims): normalize the shim alias base to forward slashes on Windows

shulaoda:
## What

Normalize the plugin-local `shimsDir` (`path.resolve(__dirname, "shims")`) to forward slashes, and switch the `nextShimMap` alias entries from `path.join(shimsDir, …)` to `path.posix.join(shimsDir, …)`, so every shim path derived from `shimsDir` is a forward-slash POSIX id on Windows.

## Why

On Windows `path.resolve` returns backslashes, and `path.join` re-introduces backslashes even when given a forward-slash base. That left the `next/*` → shim `resolve.alias` replacements and the dep-optimizer alias plugin's `resolveId` results as backslash / mixed-separator ids, inconsistent with the module-level `_shimsDir` (already forward-slash) and the canonicalized, forward-slash module ids Rolldown/Vite store. Because `resolveShimModulePath` already joins with `path.posix.join` internally, feeding it a backslash base produced *mixed* separators (e.g. `E:\…\shims/error.react-server.ts`). Normalizing once at the source and using `path.posix.join` downstream keeps all shim ids POSIX, matching the rest of the pipeline. On POSIX it is a no-op: `normalizePathSeparators` returns its input and `path.posix.join` equals `path.join`.

## How

- `shimsDir` now wraps `path.resolve(__dirname, "shims")` in `normalizePathSeparators`.
- All `path.join(shimsDir, …)` entries in `nextShimMap` became `path.posix.join(shimsDir, …)` (single- and multi-line forms).
- The `resolveShimModulePath(shimsDir, …)` call sites need no change — that helper already joins with `path.posix.join`, so they inherit the forward-slash base automatically.
- The unrelated `path.resolve(__dirname, …)` alias entries (e.g. `vinext/cache`, `vinext/instrumentation`) are left as-is — they resolve a different base and are out of scope for this change.

## Testing

- `vp test run --project unit` (full suite, 264 files) — baseline vs. with-change failing sets diffed both ways; every difference re-run in isolation at both states. No test goes from fail→pass or pass→fail because of this change: the diffs are all known parallel-flaky files (asset-prefix, pages-server-exports, run-prerender-concurrency, fixture-dev-server, oxlint) that pass in isolation, plus the pre-existing routing / route-sorting cluster that fails identically with and without the change.
- `vp test run --project unit tests/middleware-server-only.test.ts` — 4 passed both with and without the change (the `serverOnlyShimPath` consumer is unaffected; Windows fs loads the mixed-separator id fine).
- `vp test run --project unit tests/shims.test.ts` — failure count unchanged at 14 (all pre-existing fixture issues); the "every top-level next/* alias has a corresponding .js variant" test still passes.
- `vp check packages/vinext/src/index.ts` — format, lint, and typecheck clean.

Classified `refactor` because the change is behavior-preserving: Vite/Rolldown normalize separators internally, so no unit test distinguishes the old backslash form from the new forward-slash one and none flips state. It removes a latent Windows separator inconsistency in the shim alias / dep-optimizer ids.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -862,8 +862,11 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
   // `__VINEXT_CLASS` stub with a real dispatch table.
   let rscClassificationManifest: RouteClassificationManifest | null = null;
 
-  // Resolve shim paths - works both from source (.ts) and built (.js)
-  const shimsDir = path.resolve(__dirname, "shims");
+  // Resolve shim paths - works both from source (.ts) and built (.js).
+  // Normalize to forward slashes so every downstream `path.posix.join` keeps
+  // the shim ids POSIX on Windows (matching `_shimsDir` and the canonicalized
+  // module-graph ids); raw `path.resolve` yields backslashes there.
+  const shimsDir = normalizePathSeparators(path.resolve(__dirname, "shims"));
 
   // Shared with the Layer 2 renderChunk hook below. Rolldown stores module
   // IDs as canonicalized filesystem paths (fs.realpathSync.native) with forward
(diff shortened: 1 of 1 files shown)


packages/vinext/src/index.ts
```diff
@@ -1573,96 +1576,100 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
...
+            "vinext/fetch-cache": path.posix.join(shimsDir, "fetch-cache"),
+            "vinext/cache-runtime": path.posix.join(shimsDir, "cache-runtime"),
+            "vinext/navigation-state": path.posix.join(shimsDir, "navigation-state"),
+            "vinext/unified-request-context": path.posix.join(shimsDir, "unified-request-context"),
+            "vinext/pages-router-runtime": path.posix.join(shimsDir, "pages-router-runtime"),
+            "vinext/router-state": path.posix.join(shimsDir, "router-state"),
+            "vinext/head-state": path.posix.join(shimsDir, "head-state"),
+            "vinext/i18n-state": path.posix.join(shimsDir, "i18n-state"),
+            "vinext/i18n-context": path.posix.join(shimsDir, "i18n-context"),
             "vinext/cache": path.resolve(__dirname, "cache"),
```
james-elicx, line 1673: do we need to do these ones below as well?

shulaoda: I'll review the remaining ones in other PRs. For now, these changes don't affect any test cases and haven't made any tests pass on Windows, so their impact is limited at the moment. I'll also have AI take a closer look and investigate them further.

shulaoda: The main focus of this PR is the changes related to `shimsDir`.

PR #2173, opened 2026-06-19
feat(benchmarks): track server bundle sizes

TheAlexLichter:
add 2 more bench cases:

* RSC entry size
* General server bundle


TheAlexLichter:
(CI fails bc trusted input is changed)

james-elicx (approved):

PR #2176, opened 2026-06-19
perf(app-router): trim AppElements Flight metadata

NathanDrake2406:
## What changed

This reworks the original `safeJsonStringify` micro-optimization into a payload reduction patch.

Instead of changing the escaping path, this reduces framework-owned App Router metadata that gets serialized through Flight and embedded into inline RSC scripts:

- omit empty `__layoutFlags`
- omit default `__artifactCompatibility`
- stop sending partial `__renderObservation` through the outgoing AppElements Flight payload
- keep complete render observations server-side for ISR/cache finalization
- remove the dispatch/render peek wiring that only existed to build the partial Flight observation

The reader already treats missing layout flags as `{}` and missing artifact compatibility as the default envelope, so the compatibility path remains explicit.

## Profiling

Used Node's CPU profiler after building the package. The local repo's existing vinext production SSR benchmark path is currently skipped, and `codspeed`/`samply`/`autocannon` are not installed here, so this profiles representative embedded AppElements payload serialization rather than HTTP p50.

800,000 iterations comparing the previous payload shape (`__layoutFlags: {}`, default `__artifactCompatibility`, partial `__renderObservation`) with the new omitted-default shape:

| metric | before | after |
| --- | ---: | ---: |
| raw RSC chunk bytes | 1,069 | 157 |
| inline script bytes / payload | 1,481 | 452 |
| total inline script bytes | 1,184,800,000 | 361,600,000 |
| avg serialization wall time | 1.203us | 0.319us |
| `safeJsonStringify` CPU samples | 578 | 130 |
| estimated `safeJsonStringify` self time | 864.6ms | 191.8ms |

Interpretation: this is mainly a critical-payload byte reduction. The CPU reduction follows from less input to serialize/escape, but this PR does not optimize `safeJsonStringify` itself.

## Validation

- `vp check packages/vinext/src/server/app-elements-wire.ts packages/vinext/src/server/app-page-render.ts packages/vinext/src/server/app-page-dispatch.ts tests/app-elements.test.ts tests/app-page-render.test.ts`
- `vp test run tests/app-elements.test.ts tests/app-page-render.test.ts`
- `vp run vinext#build`
- pre-commit hook: staged checks, full check, staged unit/integration tests, knip


packages/vinext/src/server/app-elements-wire.ts
```diff
@@ -666,18 +672,21 @@ export function buildOutgoingAppPayload(input: {
...
+    !isDefaultArtifactCompatibilityEnvelope(input.artifactCompatibility)
+  ) {
+    payload[APP_ARTIFACT_COMPATIBILITY_KEY] = input.artifactCompatibility;
+  }
   if (input.cacheEntryReuseProof) {
     payload[APP_CACHE_ENTRY_REUSE_PROOF_KEY] = input.cacheEntryReuseProof;
   }
-  if (input.renderObservation) {
-    payload[APP_RENDER_OBSERVATION_KEY] = input.renderObservation;
-  }
```
james-elicx, line 695: so we're no longer setting this in the payload but we're keeping the key defined? can we remove it if we're not using it?

from the looks of it, this is eliminating dead code that you added but never finished? was there a reason we never started consuming it?

NathanDrake2406: Removed in 4023282e.

This key came from the render-observation/cache-proof work in #1198. The idea then was to attach render observations to the AppElements Flight payload so later cache-proof code could read what the render observed.

We never added that reader because the real cache-proof consumers needed a different boundary:

- cache artifacts need complete observations after the stream drains, so they use `CachedAppPageValue.renderObservation`
- static layout skip needs layout-scoped proof, so it builds a scoped observation and emits `__cacheEntryReuseProof`
- the client/planner gates cached payload commits on `__cacheEntryReuseProof`, not on page-level observation metadata

So `__renderObservation` became leftover scaffolding: page-level, partial, pre-stream metadata with no consumer. This PR already stopped writing it; I removed the remaining key/type/barrel export and dropped the stale test.

PR #2189, opened 2026-06-19
perf(client): preserve App Router route chunk boundaries

TheAlexLichter:
This PR preserves App Router route chunk boundaries instead of eagerly bundling route-owned `vinext` shims.

This should reduce initial JavaScript by 6–10 KB (gzipped) with no measurable performance regression, at the cost of 1 to 2 additional requests in case the shims are necessary.  These occur only when a route actually needs one of the separated shims, such as `next/link` or `next/image`.

### DIFF
--- a/packages/vinext/src/build/client-build-config.ts
+++ b/packages/vinext/src/build/client-build-config.ts
@@ -16,6 +16,17 @@ type RscFrameworkManualChunksMeta = {
   getModuleInfo(id: string): RscFrameworkModuleInfo | null;
 };
 
+const ROUTE_OWNED_CLIENT_SHIMS = new Set([
+  "dynamic-preload-chunks",
+  "form",
+  "image",
+  "layout-segment-context",
+  "link",
+  "offline",
+  "router",
+  "script",
+]);
+
 // Next.js emits CSS under `static/css/` and CSS url() dependencies (images,
 // fonts, …) under `static/media/`, both with an 8-char content hash. Mirror
 // that layout so migrated apps keep stable, Next-shaped asset URLs.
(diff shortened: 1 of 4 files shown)


packages/vinext/src/build/client-build-config.ts
```diff
@@ -16,6 +16,17 @@ type RscFrameworkManualChunksMeta = {
   getModuleInfo(id: string): RscFrameworkModuleInfo | null;
 };
 
+const ROUTE_OWNED_CLIENT_SHIMS = new Set([
```
james-elicx, line 19: i believe there's a couple of others that could potentially be in here as well - i think `compat-router` might be one for example?

packages/vinext/src/build/client-build-config.ts
```diff
@@ -122,10 +131,11 @@ export function createClientManualChunks(shimsDir: string) {
...
     }
 
-    // vinext shims — small runtime, shared across all pages.
-    // Use the absolute shims directory path to avoid matching user files
-    // that happen to have "/shims/" in their path.
     if (id.startsWith(shimsDir)) {
+      const relativeId = id.slice(shimsDir.length).split("?", 1)[0] ?? "";
+      const extensionIndex = relativeId.lastIndexOf(".");
+      const shimName = extensionIndex === -1 ? relativeId : relativeId.slice(0, extensionIndex);
+      if (preserveRouteBoundaries && ROUTE_OWNED_CLIENT_SHIMS.has(shimName)) return undefined;
```
james-elicx, line 138: all of this could be gated behind preserveRouteBoundaries :)

james-elicx (approved):

packages/vinext/src/build/client-build-config.ts
TheAlexLichter: Good catch. Added `compat-router`, plus the other App Router user-imported wrappers currently exercised by the fixture: `dynamic`, `legacy-image`, and `web-vitals`. Added focused manual-chunk assertions for each in b2ac7958e.

packages/vinext/src/build/client-build-config.ts
TheAlexLichter: Done in b2ac7958e. Shim-name parsing now only runs when `preserveRouteBoundaries` is enabled.

packages/vinext/src/build/client-build-config.ts
TheAlexLichter: Correction: #2189 was already merged, so I moved this follow-up onto current `main`. The fix is now in #2205 at bccf210b6; the old PR branch has been restored to its merged head.

packages/vinext/src/build/client-build-config.ts
TheAlexLichter: Correction: this follow-up is now in #2205 at bccf210b6, branched from current `main`. The old PR branch has been restored to its merged head.

PR #2190, opened 2026-06-19
feat(benchmarks): track eager client entry size

TheAlexLichter:
Another scenario to track is how much JS is loaded eagerly. We only capture the full gzip size which is somewhat misleading given that we add more JS that is loaded lazily.

See e.g. #2189


TheAlexLichter:
(fails bc touching benchmarks from a fork)

PR #2191, opened 2026-06-19
perf(dev): externalize SSR React runtime

TheAlexLichter:
This PR externalizes React, ReactDOM, and the Flight client runtime from Vite’s SSR transform pipeline during local Node development.

This avoids unnecessary parsing and prebundling, improving App Router dev cold start.

Behavior is unchanged.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -2193,6 +2193,8 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
           : config.ssr?.external === true
             ? true
             : nextServerExternal;
+        const externalizeSsrReactInDev =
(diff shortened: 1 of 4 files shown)


packages/vinext/src/index.ts
```diff
@@ -2193,6 +2193,8 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
           : config.ssr?.external === true
             ? true
             : nextServerExternal;
+        const externalizeSsrReactInDev =
+          env.command === "serve" && !hasCloudflarePlugin && !hasNitroPlugin;
```
james-elicx, line 2197: for what it's worth, i dont like that we have these conditions and i want to try and remove some of them over the next month...

james-elicx (approved):

PR #2209, opened 2026-06-20
refactor(dev): defer command and request-path imports

NathanDrake2406:
## Overview

| Field | Details |
| --- | --- |
| Goal | Keep `vinext` CLI/plugin startup from evaluating command-, build-, and request-specific modules before they are needed. |
| Core change | Move heavyweight imports behind cached dynamic imports at command dispatch, Vite hook, or request-branch boundaries. |
| Boundary | This PR is the lazy import-graph cleanup only. The App Router optimizer-entry cold-start optimization is split to #2210. |
| Primary files | `packages/vinext/src/cli.ts`, `packages/vinext/src/index.ts`, `packages/vinext/src/plugins/fonts.ts`, `packages/vinext/src/server/app-ssr-entry.ts`, `packages/vinext/src/entries/pages-server-entry.ts`, `packages/vinext/src/server/react-renderer-env.ts` |

## Why

The CLI and plugin entrypoints sit on the unavoidable startup path. They should not evaluate deploy, build, Pages SSR, API, image, font metadata, typegen, static-export, or build-only CSS asset repair implementations before the selected command, hook, or request path actually needs them.

This cleanup makes those ownership boundaries explicit. It also removes one incidental correctness dependency: SSR/static-export renderer loading now matches the React runtime flavor already loaded in the process instead of relying on an eager dev-server import side effect.

## What changed

| Surface | Before | After |
| --- | --- | --- |
| CLI dispatch | Importing the CLI eagerly pulled in the plugin and most command implementations. | Command modules load only after the command is known; build-only imports are warmed concurrently once `build` is selected. |
| Plugin startup | Dev startup loaded many build-only or request-specific helpers. | Pages entry generation, typegen, deploy checks, image optimization, API handling, SSR dev server creation, static export, build manifests, inline CSS, precompression, and CSS URL asset repair are imported at the hook/request branch that owns them. |
| Font plugin | Registering `next/font/google` support loaded Google font metadata immediately. | Google font validation, axes, and fallback metrics load only when CSS injection needs them. |
| React renderer | Static `react-dom/server.edge` imports masked runtime flavor coupling. | `react-renderer-env` loads the renderer under the already-loaded React dev/prod flavor. |

## Performance

The previous CodSpeed result for this split PR showed this import-graph cleanup is small on the official benchmark by itself: dev cold start moved from `2.39 s` to `2.37 s` (`-0.7%`), below the `1.5%` reporting threshold.

This branch now also defers the build-only CSS URL asset helper from plugin startup. I expect that to be a small incremental improvement, not a headline change. The larger App Router cold-start optimization, focused `optimizeDeps.entries`, is in #2210.

## Validation

- `vp check packages/vinext/src/index.ts`
- `vp check packages/vinext/src/cli.ts packages/vinext/src/index.ts packages/vinext/src/plugins/fonts.ts packages/vinext/src/entries/pages-server-entry.ts packages/vinext/src/server/app-ssr-entry.ts packages/vinext/src/server/react-renderer-env.ts`
- `vp test run tests/cli-args.test.ts tests/entry-templates.test.ts`
- `vp test run tests/css-url-assets.test.ts tests/route-classification-injector.test.ts tests/build-time-classification-integration.test.ts`
- `vp run vinext#build`
- Earlier before the split:
  - `vp test run tests/static-export.test.ts`

## Risk / compatibility

- Public API: no intended API changes.
- Runtime: dynamic imports are cached and remain behind the same observable command, hook, or request branches.
- Build behavior: CSS URL repair still runs only in `apply: "build"` hooks; the implementation module now loads on first matching build hook instead of during dev/plugin startup.
- Dev/prod parity: the renderer helper is the main guard; it prevents prod SSR/static export from depending on an unrelated eager import.
- Review shape: this PR is intentionally separated from #2210 so each optimization can be measured independently.

### DIFF
--- a/packages/vinext/src/cli.ts
+++ b/packages/vinext/src/cli.ts
@@ -655,9 +655,8 @@ async function buildApp() {
     }
   } finally {
     if (pagesClientAssetsBuildSession) {
-      const { clearPagesClientAssetsBuildMetadata } = await import(
-        "./build/pages-client-assets-module.js"
-      );
+      const { clearPagesClientAssetsBuildMetadata } =
+        await import("./build/pages-client-assets-module.js");
       clearPagesClientAssetsBuildMetadata(pagesClientAssetsBuildSession);
       if (
         process.env.__VINEXT_PAGES_CLIENT_ASSETS_BUILD_SESSION === pagesClientAssetsBuildSession
@@ -681,9 +680,8 @@ async function buildApp() {
   }
 
   let prerenderResult;
-  const { formatVinextPrerenderLabel, resolveVinextPrerenderDecision } = await import(
-    "./config/prerender.js"
-  );
+  const { formatVinextPrerenderLabel, resolveVinextPrerenderDecision } =
+    await import("./config/prerender.js");
   const prerenderDecision = resolveVinextPrerenderDecision({
     prerenderAllFlag: parsed.prerenderAll,
     vinextPrerenderConfig: buildConfigMetadata.prerenderConfig,
@@ -809,9 +807,7 @@ async function lint() {
 }
 
 async function deployCommand() {
-  const { deploy: runDeploy, parseDeployArgs } = await import(
-    "@vinext/cloudflare/internal/deploy"
-  );
+  const { deploy: runDeploy, parseDeployArgs } = await import("@vinext/cloudflare/internal/deploy");
   const parsed = parseDeployArgs(rawArgs);
   if (parsed.help) {
     printDeployDeprecationWarning();
@@ -892,9 +888,8 @@ async function initCommand() {
   const port = parsed.port ?? 3001;
   const skipCheck = rawArgs.includes("--skip-check");
   const force = rawArgs.includes("--force");
-  const { INIT_PLATFORMS, resolveInitPlatform, resolveInitPrerender } = await import(
-    "./init-platform.js"
-  );
+  const { INIT_PLATFORMS, resolveInitPlatform, resolveInitPrerender } =
+    await import("./init-platform.js");
   const platform = await resolveInitPlatform(rawArgs);
   const platformOptions = await INIT_PLATFORMS[platform].options(rawArgs);
   const prerender = await resolveInitPrerender(rawArgs);
--- a/packages/vinext/src/cli.ts
+++ b/packages/vinext/src/cli.ts
@@ -14,8 +14,6 @@
  * needed for most Next.js apps.
  */
 
-import vinext from "./index.js";
-import { runPrerender } from "./build/run-prerender.js";
 import path from "node:path";
 import fs from "node:fs";
 import { pathToFileURL } from "node:url";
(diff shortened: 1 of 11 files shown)


james-elicx (commented):
please can you do separate optimisations in separate prs so we can see the impact of each one

NathanDrake2406:
Split implemented. This PR is now only the lazy import graph optimization and has been rebased onto current main. The App Router optimizeDeps.entries scope reduction moved to #2210. I do not have permission to push a stack base branch to cloudflare/vinext, so #2210 is draft against main for now; its body includes the fork compare link that shows the isolated two-file diff against this branch.

NathanDrake2406:
CI update: the remaining failing `Performance scenarios` check is hitting a base perf-harness dependency-resolution bug, not the lazy import graph change in this PR.

The fork PR workflow runs the trusted harness from `$RUNNER_TEMP/vinext-perf-harness` while measuring this branch via `VINEXT_PERF_TARGET_ROOT`. The current `main` harness has a top-level `import { parseAst } from "vite"` in `benchmarks/perf/bundle-size.mjs`, so Node tries to resolve `vite` from the temp harness copy and fails before the bundle-size benchmark runs.

Opened the small harness fix here: https://github.com/cloudflare/vinext/pull/2211

Once that lands on `main`, this PR should be rerunnable against a base harness that resolves Vite from the benchmark target checkout instead of the copied harness directory.

NathanDrake2406:
Addressed the review feedback in 60e0a6f.

**1. Renderer loader / global `NODE_ENV`.** Valid concern — the generic `importWithReactNodeEnv(env, load)` was the wrong shape. Replaced it with a dedicated `importReactDomServerEdge(reactNodeEnv)` that owns the concern and documents the constraints. Two of the suggested mechanisms turned out not to be viable, with evidence:

- *Resolve the dev/prod renderer explicitly via separate entrypoints* — React doesn't expose them. `react-dom/server.edge` is a CJS wrapper that branches on `process.env.NODE_ENV` at eval time, and the per-flavor cjs builds are absent from the package `exports` map: `import('react-dom/cjs/react-dom-server.edge.development.js')` throws `ERR_PACKAGE_PATH_NOT_EXPORTED`. The `NODE_ENV` gate is the only public lever.
- *Make it non-interleavable via a synchronous `createRequire` load* — I implemented this first, and it broke 34 App Router prod-server tests with `dispatcher.getOwner is not a function`. Root cause: `createRequire` escapes the bundle's module graph and pulls a *separate* react-dom that binds a different React instance than the app's bundled React. The renderer load **must** go through the bundler-visible dynamic `import()` to preserve single-React-instance identity, which is inherently async.

So the swap stays, but it's now tightly scoped and documented: it fires only on a genuine flavor/`NODE_ENV` mismatch (no-op in the steady state — dev, and bundled Workers where `NODE_ENV` is build-time fixed), is restored in `finally`, and callers memoize the loader so it happens at most once per process for a single module evaluation. Added `tests/react-renderer-env.test.ts` covering the env discipline (no mutation on the fast path; restore of both a defined and an unset `NODE_ENV` after a mismatch).

**2. Duplicated constants.** Valid — agreed it was the wrong trade. Extracted the cheap facts to their canonical homes: `server/image-optimization-paths.ts` (endpoint constants + `isImageOptimizationPath`) and `server/content-types.ts` (the full MIME table). The heavy modules re-export them so existing importers are unaffected; `index.ts` and `fonts.ts` now import the tiny modules. Lazy-import win kept, drift gone.

**3. Static-export wrappers.** Valid — switched to the real single-options signatures (`StaticExportOptions` / `AppStaticExportOptions` → `StaticExportResult`), preserving arity and dropping the `Parameters<>`/`ReturnType<>` plumbing, while keeping the deferred import.

No runtime behavior change. Targeted suites green locally: App Router prod-server (59), Pages Router (318), deploy (288), static-export + fonts (189), entry-templates, and the new renderer-env test.

PR #2210, opened 2026-06-20
perf(dev): scope App Router optimizer startup entries

NathanDrake2406:
## Overview

| Field | Details |
| --- | --- |
| Goal | Stop the root App Router dev cold start from paying whole-app Vite optimizer discovery. |
| Core change | Seed `optimizeDeps.entries` with app startup conventions and instrumentation instead of `app/**/*.{tsx,ts,jsx,js}`. |
| Boundary | This PR is now independent from #2209 and is based directly on `main`. |
| Primary files | `packages/vinext/src/index.ts`, `tests/app-router-dev-server.test.ts` |

## Why

The official cold-start benchmark waits for `/`, but the old optimizer entry list crawled the entire App Router tree across RSC, SSR, and client environments. That makes first response time scale with total route count, even though route modules are represented as lazy imports in the generated RSC manifest.

This PR keeps startup dependency discovery proportional to files that can participate in the initial app shell, while non-startup route modules are discovered when those routes are first requested.

## What changed

| Surface | Before | After |
| --- | --- | --- |
| App Router `optimizeDeps.entries` | RSC, SSR, and client optimizers crawled `app/**/*.{tsx,ts,jsx,js}`. | Optimizers crawl app startup convention files and instrumentation only. |
| Non-startup client routes | Whole-app crawling pre-discovered their deps during root startup. | They are discovered when the route is requested; a regression test verifies `/interactive` renders after `/` without duplicate React. |
| Framework identity | React/RSC identity stability partly benefited from broad source crawling. | Existing explicit framework include/exclude rules remain responsible for stable React/runtime identity across environments. |

## Performance evidence

The previous CI run for this PR failed later in the bundle-size phase because the trusted base harness could not resolve `vite` from its temp copy. That harness issue has since been fixed on `main` in #2212, and this branch has been rebuilt independently on top of that fix.

Before the failure, the cold-start rounds had already completed and showed about `2437 ms -> 2289 ms`, roughly `-148 ms` / `-6.1%` for `vinext` dev cold start. The new independent CI run should provide the clean PR summary.

## Validation

- `vp check packages/vinext/src/index.ts tests/app-router-dev-server.test.ts`
- `vp test run tests/app-router-dev-server.test.ts -t "optimizeDeps|non-startup client route|framework dependencies|static RSC renderer"`

## Risk / compatibility

- Public API: no intended API changes.
- Dev optimizer behavior: route-specific deps are discovered on first route request instead of during root startup.
- Regression coverage: the new non-startup client route test protects against duplicate React after deferred route discovery.
- Existing-app risk: apps with root convention files still seed startup dependencies. Route-only dependencies shift to route-first-request cost.


NathanDrake2406:
## Triage of review feedback (pushed in `448e63816`)

I took the structural recommendation directly: the bespoke filesystem walk is gone, and `collectAppRouterStartupOptimizeEntries` is now a projection of `buildAppRouteGraph` — the convention modules that rendering the root URL (`/`) loads, plus the filename-resolved root boundaries (`global-error` / `global-not-found`). One source of truth, so the startup set can no longer miss real startup modules or include non-startup ones.

### 1. Missing `@children` startup entries — valid, fixed
Confirmed against the collector directly: for `app/layout.tsx` + `app/@children/page.tsx`, the old walk returned only `["app/layout.tsx"]`, while the route graph registers `app/@children/page.tsx` as the `/` page. The projection now includes it (the graph already treats `@children` as transparent). Regression added.

### 3. Over-includes unrelated top-level route groups — valid, fixed
Confirmed: for `app/page.tsx` + `app/(admin)/layout.tsx` + `app/(admin)/dashboard/page.tsx`, the old walk emitted `app/(admin)/layout.tsx`; the graph's `/` route does not, because `(admin)/layout.tsx` only wraps `/dashboard`. The projection only ever emits files attached to a route whose pattern is `/`, so unrelated `(admin)`/`(marketing)`/`(auth)` shells are excluded by construction. Regression added.

### 2. Slot route-group layouts excluded — no separate change needed
The premise that these are loaded ("the manifest emits loaders for them, the route module loader loads them, the renderer wraps the slot subtree with them") does not hold against the current code:

- `discoverParallelSlots` sets a slot's `layoutPath` to `findFile(slotDir, "layout")` — the **immediate** slot-root layout (`app/@drawer/layout.tsx`). The route-group-nested `app/@drawer/(group)/layout.tsx` is never captured. `findSlotRootPage` descends route groups for the slot's *page*, but there is no matching layout discovery.
- `createRouteManifest` materializes a slot-layout loader only when `slot.layoutPath` is set (`app-route-graph.ts` ~L521). With no `layoutPath`, no loader is emitted, so the renderer cannot wrap with it.
- There is no `configLayoutPaths` symbol in the codebase (`grep -rn configLayoutPath packages/vinext/src` → no matches).

So excluding `app/@drawer/(group)/layout.tsx` from startup entries matches what the renderer actually loads. Crucially, with the projection this is no longer a hand-maintained blind spot: the startup set is now *exactly* what the graph loads for `/`. If vinext ever starts rendering slot route-group layouts, the optimizer picks them up automatically — the two cannot diverge again. (If Next.js renders that nested layout and vinext should too, that's a separate rendering-parity bug, out of scope for this perf change; the projection means the optimizer would follow the graph once it's fixed.)

I added a path-level regression that asserts the slot root page (found through a route group) is in the set while its route-group-nested layout is out — strictly stronger than the previous dev-server assertion, since it checks the exact file rather than a transitively-discovered dep.

### Validation
`vp check` + `vp test run tests/app-router-optimizer-startup.test.ts` green (3 new direct-collector regressions + the existing dev-server end-to-end test, which still passes unchanged — the projection produces identical wiring for the real fixture). Rebased onto latest `main`; no conflicts.

PR #2225, opened 2026-06-21
perf(link): omit Pages runtime in app-only builds

TheAlexLichter:
Pure App Router builds now compile Pages-only navigation, prefetching, interpolation, and hybrid route ownership logic out of `next/link`. Pages Router and hybrid applications retain the existing behavior. If the App navigation runtime is unavailable, App-only links use document navigation so the destination still renders without relying on a missing client runtime.

### DIFF
--- a/packages/vinext/src/shims/link.tsx
+++ b/packages/vinext/src/shims/link.tsx
@@ -226,6 +226,15 @@ function resolvePagesLinkNavigationHref(href: string, locale: string | false | u
   );
 }
 
+function applyPagesNavigationFallback(href: string, replace: boolean): void {
+  if (replace) {
+    window.history.replaceState({}, "", href);
+  } else {
+    window.history.pushState({}, "", href);
+  }
+  window.dispatchEvent(new PopStateEvent("popstate"));
+}
+
 /**
  * Collapse repeated forward-slashes (and convert backslashes to forward-slashes)
  * in the path portion of a URL, preserving any query string.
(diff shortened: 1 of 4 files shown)


TheAlexLichter:
Note: I will clean this up a. bit more after 2222-2225 are merged

packages/vinext/src/shims/link.tsx
```diff
@@ -1152,6 +1160,10 @@ const Link = forwardRef<HTMLAnchorElement, LinkProps>(function Link(
           window.dispatchEvent(new PopStateEvent("popstate"));
         },
       });
+    } else if (replace) {
+      window.location.replace(absoluteFullHref);
+    } else {
+      window.location.assign(absoluteFullHref);
     }
```
james-elicx: the `fallback` callback would have previously ran instead of this new else if and else - can we keep them consistent?

james-elicx: hmm actually i think the fallback probably would have almost never run because it was importing the router.

james-elicx: hard nav sounds safer

TheAlexLichter: Good catch. I restored the existing history push/replace + `popstate` fallback

james-elicx: sorry i started replying again at the same time as you were pushing a change 😅. i think what you were doing originally was probably the safer option if no runtime is available rather than the replacestate/pushstate/popstate

TheAlexLichter: No worries. changed it back 👍

PR #2226, opened 2026-06-21
perf(app-router): skip speculative page probes for HTML renders

southpolesteve:
## Summary

This builds on the request-scoped fetch dedupe work from #1134. That change prevents identical render fetches from hitting the network twice, but HTML document requests can still pay duplicate userland work because `renderAppPageLifecycle()` invokes the App Router page probe before the real RSC render.

For data-heavy App Router pages, the probe can execute the page/server-component code path once, then the real RSC render executes it again. Hydrogen exposed this as extra Storefront GraphQL client work even when fetch dedupe is active.

This PR keeps layout probing/classification intact, but skips the speculative page probe for HTML document renders. Page-level `redirect()` / `notFound()` recovery now uses the existing RSC error tracker path after the real render starts, before HTML bytes flush. RSC requests keep the existing eager page-probe behavior.

For fresh query-bearing HTML renders, dispatch now asks the real page element to observe `searchParams` access and waits for the lazy HTML stream before finalizing the cache policy. That preserves cache safety when the page actually reads `searchParams`, without treating every `?utm_source=...` request as dynamic when the page never uses the query.

## What changed

- Added a `probePageBeforeRender` switch to `probeAppPageBeforeRender()` and disabled page probing for HTML document renders.
- Extended captured special-error recovery so skipped-probe HTML renders can still produce 307/404 responses before flush.
- Added an explicit page-element-builder flag for observing real page `searchParams` access during query-bearing HTML cache classification.
- Added a render-lifecycle option to buffer query-bearing HTML until lazy dynamic usage is known, then recreate the response stream.
- Updated dispatch/render/probe tests so duplicate execution, query-invariant caching, and actual `searchParams` usage are covered separately.

## Why

`#1134` fixed duplicate network fetches, but it does not remove duplicated page/userland execution. This PR targets that remaining layer: avoid running arbitrary page code once in the probe phase and again in the real render for a single HTML request.

Alex also pointed out that simply marking all query-bearing HTML dynamic would regress ISR for common tracking URLs. The revised approach avoids that: a fresh `/products?utm_source=google` render can populate ISR if the page does not read `searchParams`, while `/products?filter=alpha` remains dynamic when the page actually reads the query.

## Quick benchmark

Temporary local benchmark: 40 HTML render lifecycles, each with an 8ms synthetic `probePage()` cost.

- `origin/main`: `probeCalls=40`, `durationMs=423`
- this branch: `probeCalls=0`, `durationMs=19`

The benchmark models framework probe overhead directly; it does not depend on Hydrogen or Shopify.

## Validation

- `./node_modules/.bin/vp check`
  - format, lint, and type checks passed
- `./node_modules/.bin/vp test run tests/app-page-dispatch.test.ts tests/app-page-render.test.ts tests/app-page-probe.test.ts tests/app-page-element-builder.test.ts tests/entry-templates.test.ts`
  - 5 files passed, 176 tests passed
- `./node_modules/.bin/vp test run tests/app-router-production-server.test.ts -t "page ISR + searchParams"`
  - 2 tests passed, including the previous CI failure path

### DIFF
--- a/packages/vinext/src/server/app-page-dispatch.ts
+++ b/packages/vinext/src/server/app-page-dispatch.ts
@@ -230,6 +230,7 @@ export type DispatchAppPageOptions<TRoute extends AppPageDispatchRoute> = {
     opts: AppPageDispatchInterceptOptions | undefined,
     searchParams: URLSearchParams,
     layoutParamAccess?: AppLayoutParamAccessTracker,
+    options?: { observePageSearchParamsAccess?: boolean },
   ) => Promise<AppPageElement>;
   clientReuseManifest?: ClientReuseManifestParseResult;
   cleanPathname: string;
@@ -529,7 +530,13 @@ async function dispatchAppPageInner<TRoute extends AppPageDispatchRoute>(
   const isDynamicError = dynamicConfig === "error";
   const isForceDynamic = dynamicConfig === "force-dynamic";
   const isDraftMode = isDraftModeRequest(options.request, options.draftModeSecret);
+  const hasRequestSearchParams = !isForceStatic && hasSearchParams(options.searchParams);
   const layoutParamAccess = createAppLayoutParamAccessTracker();
+  const hasActiveLoadingBoundary = shouldSuppressLoadingBoundaries(
+    options.renderMode ?? APP_RSC_RENDER_MODE_NAVIGATION,
+  )
+    ? false
+    : Boolean(route.loading?.default);
 
   setCurrentFetchSoftTags(buildAppPageTags(options.cleanPathname, [], route.routeSegments));
   setCurrentFetchCacheMode(options.fetchCache ?? null);
@@ -591,7 +598,7 @@ async function dispatchAppPageInner<TRoute extends AppPageDispatchRoute>(
     const cachedPageResponse = await readAppPageCacheResponse({
       cleanPathname: options.cleanPathname,
       clearRequestContext: options.clearRequestContext,
-      hasRequestSearchParams: !isForceStatic && hasSearchParams(options.searchParams),
+      hasRequestSearchParams,
       isEdgeRuntime: options.isEdgeRuntime,
       isRscRequest: options.isRscRequest,
       isrDebug: options.isrDebug,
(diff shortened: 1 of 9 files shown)


southpolesteve:
@james-elicx @NathanDrake2406 I got the new Hydrogen preview working on workers with vinext and noticed performance was worse than expected. This PR should fix

https://hydrogen-vinext-demo.southpolesteve.workers.dev/

southpolesteve:
@bigbonk review please

packages/vinext/src/server/app-page-dispatch.ts
```diff
@@ -1001,6 +1002,14 @@ async function dispatchAppPageInner<TRoute extends AppPageDispatchRoute>(
   if (pageBuildResult.response) {
     return pageBuildResult.response;
   }
+
+  // HTML RSC work executes lazily while the stream is consumed, after response
+  // headers are chosen. Query-bearing fresh renders without a cache proof must
+  // be conservative now; proven query-invariant cached HTML returned above.
+  if (!options.isRscRequest && hasRequestSearchParams) {
+    markDynamicUsage();
+  }
```
TheAlexLichter: This now runs for every HTML request with query params, even if the page never reads the params.

In turn, `no-store` will be set and all fresh query-bearing HTML requests wont populate the ISR store.

Example:

```tsx
// app/products/page.tsx
export const revalidate = 60;

export default function ProductsPage() {
  return <h1>Products</h1>;
}
```

- `/products` populates ISR normally.
- `/products?utm_source=google` calls markDynamicUsage(), returns no-store, and does not populate ISR.
- If tracking URLs are the first or primary traffic source, every request renders again despite the page never using searchParams.

TheAlexLichter (changes requested):
see `markDynamicUsage()` comment

packages/vinext/src/server/app-page-dispatch.ts
southpolesteve: Fixed now. I removed the broad query-param dynamic mark and replaced it with actual render observation: query-bearing HTML now observes real page `searchParams` access while the lazy HTML stream is consumed, so pages that read `searchParams` still skip ISR, but inert tracking-query requests can populate ISR/cache proof instead. Added coverage for both sides in `tests/app-page-dispatch.test.ts`, plus the existing production `page ISR + searchParams` path is green locally.

southpolesteve:
Addressed the latest review feedback.

The query-bearing HTML path no longer drains the full SSR stream before returning the response. Instead it streams immediately with conservative no-store headers, omits the optimistic cache-state header while dynamic usage is still pending, and lets the cache finalizer decide after stream consumption whether the render was query-invariant enough to write ISR. That removes the request-context cleanup hazard from the buffering path and preserves streaming behavior.

I also rebased onto latest main and reran the focused App Router/cache tests plus vp check.

PR #2241, opened 2026-06-21
fix(router): refresh shallow rewrite query state

james-elicx:
## Summary
- distinguish shallow navigations from deep navigations using the persisted history state
- rebuild shallow-navigation query state from the visible URL while preserving matched dynamic route params
- preserve middleware-injected rewrite query state after deep navigation

## Validation
- 4 focused router unit tests, including explicit shallow and deep rewrite branches
- 1 focused production Playwright test
- 12 existing dev shallow-routing tests
- scoped `vp check`, `git diff --check`, independent review: clean

## Next.js parity
Fixes the remaining shallow-linking assertion from `middleware-general/test/index.test.ts` in deploy-suite run 27858251903. The production regression test extends that upstream scenario with middleware-injected query state to cover both deep and shallow navigation. No raw Next.js E2E was run locally.

### DIFF
--- a/packages/vinext/src/shims/internal/hybrid-client-route-owner.ts
+++ b/packages/vinext/src/shims/internal/hybrid-client-route-owner.ts
@@ -33,6 +33,12 @@ import type {
 
 export type HybridClientOwner = "app" | "document" | "pages";
 
+export type PagesClientRouteResolution = {
+  href: string;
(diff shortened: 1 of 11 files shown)


james-elicx:
Temporarily closing and reopening to retrigger PR workflows for head 6bdba8e after GitHub only emitted CodeQL on synchronize.

PR #2284, opened 2026-06-23
fix(app-router): exclude dev overlay from production client bundle

### LINKED ISSUE #1724: [nextjs-tracker] Avoid bundling HMR refresh reducer in production App Router client
## Next.js change

Next.js moved the `hmrRefreshReducer` import behind a `process.env.NODE_ENV === 'development'` compile-time branch in the App Router client router-reducer. Previously the eager top-level import caused the HMR-only reducer to be included in production browser chunks. A new production browser-chunks test rejects `hmr-refresh-reducer` sources to prevent regressions.

**Commits:** [057b84f](https://github.com/vercel/next.js/commit/057b84fe27b7bdd686d98874d0682a95728ffcd9)
**Next.js files changed:** `packages/next/src/client/components/router-reducer/router-reducer.ts`, `test/production/app-dir/browser-chunks/browser-chunks.test.ts`

## Why this matters to vinext

vinext's App Router client entry (`packages/vinext/src/entries/app-browser-entry.ts`) and any reducer/HMR plumbing under `packages/vinext/src/shims/` or `packages/vinext/src/server/` can have the same problem: dev-only code (HMR refresh action handler, dev overlay glue) eagerly imported from a module that is also reachable in production builds, ending up in production browser chunks.

Worth auditing what we import unconditionally from our client-side router/reducer code and whether we have any `process.env.NODE_ENV === 'development'` guards around HMR-only code paths. If we use Next.js's router-reducer code directly or have our own equivalent, the same gating pattern (or a Vite `import.meta.env.DEV` check + dynamic `require`/import) should apply.

## Priority

**Minor**

- Bundle size / dev-prod parity issue. Not a behavioral break, but increases production payload and ships dev-only code paths to end users.

## Suggested action

Investigate: Grep our App Router client entry for HMR/dev-only imports (e.g., `hmr-refresh-reducer`, dev overlay modules, error overlay reducers) and confirm they are tree-shaken in production builds. If not, gate them behind `import.meta.env.DEV` (or `process.env.NODE_ENV === 'development'`) so the production bundle does not include them. Add a regression test similar to Next.js's `browser-chunks.test.ts` that asserts dev-only modules are absent from production client chunks.

ponharu:
## Summary

- Move App Router dev error overlay imports behind dev-only dynamic imports
- Keep RSC HMR overlay cleanup inside the dev/HMR code path
- Add a production build regression test that rejects dev overlay/HMR-only strings in client chunks

## Verification

- pnpm test tests/app-router-production-build.test.ts
- PLAYWRIGHT_PROJECT=app-router pnpm exec playwright test --project=app-router tests/e2e/app-router/dev-error-overlay.spec.ts
- pnpm run check

Close #1724

### DIFF
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -2164,8 +2164,7 @@ function bootstrapHydration(
       // empty; if it throws again, devOnCaughtError/devOnUncaughtError
       // re-populates it. Without this, an old "DropZone is not defined"
       // error would linger after the developer fixed the bug.
-      const devErrorOverlay = await import("./dev-error-overlay.js");
-      devErrorOverlay.dismissOverlay();
+      devErrorOverlay?.dismissOverlay();
       // Interception context on HMR re-renders is intentionally deferred:
       // preserving intercepted modal state across HMR reloads is out of scope
       // for the previousNextUrl mechanism.
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -145,13 +145,6 @@ import {
 } from "../client/app-nav-failure-handler.js";
 import { createClientReuseManifestHeaderFromVisibleAppState } from "./app-browser-client-reuse-manifest.js";
 import {
-  devOnCaughtError,
-  dismissOverlay,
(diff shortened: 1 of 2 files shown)


james-elicx:
/bigbonk review for issues and edge cases

PR #2285, opened 2026-06-23
feat(benchmarks): add rolling median trendlines

TheAlexLichter:
This PR adds dashed 7-run rolling median trendlines to benchmark charts while preserving existing data points, links, and tooltips.
Missing runs remain visual gaps and do not consume the rolling window.

The chart controls allow each framework to be shown or hidden independently. Individual runs can also be hidden for a trend-only view; their hover targets, tooltips, and links remain available, and the hovered point is revealed temporarily. This choice is persisted as `runs=hidden` in the URL. The control is omitted for scenarios with fewer than seven actual observations, where no rolling median can be shown.

## Before / After

These screenshots use the current 100-run **Dev server cold start** history from [vinext.dev/benchmarks](https://vinext.dev/benchmarks), including both Next.js and vinext measurements. The after screenshot shows the optional trend-only view.

| Before | After: individual runs hidden |
| --- | --- |
| Live Dev server cold start chart before rolling median trendlines | Live Dev server cold start chart showing only selectable 7-run rolling median trendlines |


james-elicx:
Preview environment: https://pr-2285-vinext-web.vinext.workers.dev/benchmarks

PR #2305, opened 2026-06-23
refactor(app-router): descend scanForInterceptingPages with path.posix.join

shulaoda:
## What

In `scanForInterceptingPages`, hoist the child directory join and make it `path.posix.join`, and document that `currentDir` / `routeDir` / `appDir` must be forward-slash:

- `const interceptDir = path.posix.join(currentDir, entry.name)` is computed once and reused by both the intercept-marker branch and the recursive scan.
- The two previous `path.join(currentDir, entry.name)` calls are removed.

## Why

`currentDir` is forward-slash by the route-graph contract, and the scan descends it recursively, so the join must keep it forward-slash. `path.join` is `path.win32.join` on Windows and would rebuild the child path with backslashes, which the intercept page lookups (`findFile` / `path.posix.join`) would then turn into mixed separators. `path.posix.join` keeps the descent canonical. Both branches were joining `currentDir` + `entry.name` independently, so computing it once removes the duplication.

## How

- Hoisted `interceptDir` out of the marker branch, switched it to `path.posix.join`, and reused it in the recursive `scanForInterceptingPages` call.
- Added a doc note: `currentDir` / `routeDir` / `appDir` must be forward-slash.

## Testing

- `vp check packages/vinext/src/routing/app-route-graph.ts` — format, lint, and typecheck clean.
- No-op on POSIX (`path.posix.join` === `path.join`). On Windows the recursive descent stays forward-slash given a forward-slash `currentDir`; the join value is otherwise unchanged. Behavior-preserving, hence `refactor`.

### DIFF
--- a/packages/vinext/src/routing/app-route-graph.ts
+++ b/packages/vinext/src/routing/app-route-graph.ts
@@ -2375,6 +2375,10 @@ export function findOwnerRouteForDir(
 /**
  * Recursively scan a directory tree for page.tsx files that are inside
  * intercepting route directories.
+ *
+ * `currentDir`, `routeDir`, and `appDir` must be forward-slash. `currentDir`
+ * descends with `path.posix.join` and all three reach the `path.posix.join` /
+ * `path.posix.relative` calls that build the intercept page paths and patterns.
  */
 function scanForInterceptingPages(
   currentDir: string,
@@ -2394,12 +2398,12 @@ function scanForInterceptingPages(
 
     // Check if this directory name starts with an interception convention
     const interceptMatch = matchInterceptConvention(entry.name);
+    const interceptDir = path.posix.join(currentDir, entry.name);
 
     if (interceptMatch) {
       // This directory is the start of an intercepting route
       // e.g. "(.)photos" means intercept same-level "photos" route
       const restOfName = entry.name.slice(interceptMatch.prefix.length);
-      const interceptDir = path.join(currentDir, entry.name);
 
       // Find page files within this intercepting directory tree.
       // `currentDir` is the *parent* of the marker dir — used by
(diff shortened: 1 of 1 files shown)


TheAlexLichter:
the windows hero 🙏🏻

Should we consider https://github.com/unjs/pathe/ for paths to make this easier?

shulaoda:
> the windows hero 🙏🏻
> 
> Should we consider https://github.com/unjs/pathe/ for paths to make this easier?

Yeah! I'd actually considered `pathe` before, and I'd noticed that void pulls it in downstream too.

My earlier hesitation was performance. `pathe` reimplements the path logic in JS for both Linux and Windows. But I just went and checked node's own `path` implementation, and it turns out that's JS as well, so `pathe` might actually be just as efficient (or even a bit more), though I'm not fully sure.

The other reason I held off is that I wasn't yet confident about the blast radius of each path. I was also a bit worried that reaching for `pathe` too early might paper over some potential underlying issues instead of surfacing them. Since it normalizes separators automatically, a mismatch that actually points to a real bug could just get smoothed away silently. That's why the recent PRs have been working through each failing Windows test case one by one, doing a bottom-up trace of how the Windows path production leads to the problem across the whole chain. My original plan was to do the full switch to `pathe` at the very end, once all the Windows tests are passing, and then see how it looks holistically. But you're right: we can adopt it earlier.

One concrete example: `pathe` rewrites `\` to `/` on every platform, but on POSIX a backslash is a valid filename character, so a file named `foo\bar` would get split into two segments. `path.posix` leaves it intact. 
Refs: https://github.com/unjs/pathe/issues/236

^cc @james-elicx

james-elicx:
If it gets bundled instead of passing on another dep to users installing the package, sure

TheAlexLichter:
> If it gets bundled instead of passing on another dep to users installing the package, sure

We should check if [`pathe`](https://npmx.dev/package/pathe) is part of other typical user libs/deps though. Inlining might not be worth it if it is 👀

shulaoda:
I'll take a look and see whether this issue https://github.com/unjs/pathe/issues/236 can be resolved, and then give `pathe` a try later.

PR #2312, opened 2026-06-25
fix(config): encode destination query params

james-elicx:
## Summary

- preserve encoded route separators when decoded vinext path captures are interpolated into rewrite and redirect queries
- retain catch-all `/` separators and existing pathname/fragment interpolation behavior
- cover query values, unusual query-key placeholders, repeated placeholders, external destinations, and fragment boundaries

## Why this is needed

Vinext evaluates config routes against an already-decoded `cleanPathname`. Next.js retains encoded source captures through destination interpolation, so an input such as `foo%26admin=true` remains encoded in the emitted query URL. Without compensating at substitution time, vinext turns that path data into new query delimiters.

## Verification

- real Next.js 16.2.7 webpack dev server for the rewrite and redirect wire behavior
- `vp check packages/vinext/src/config/config-matchers.ts tests/shims.test.ts`
- `vp test run tests/shims.test.ts` (1,148 tests)

### DIFF
--- a/packages/vinext/src/config/config-matchers.ts
+++ b/packages/vinext/src/config/config-matchers.ts
@@ -1153,9 +1153,10 @@ function substituteDestinationParams(destination: string, params: Record<string,
 
   return destination.replace(
     paramRe,
-    (_token, key: string, _modifier: string | undefined, offset: number) => {
+    (_token, key: string, modifier: string | undefined, offset: number) => {
       if (!isDestinationQueryOffset(destination, offset)) return params[key];
-      return encodeURIComponent(params[key]);
+      const encoded = encodeURIComponent(params[key]);
+      return modifier === "*" || modifier === "+" ? encoded.replace(/%2F/gi, "/") : encoded;
     },
   );
 }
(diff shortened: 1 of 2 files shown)


james-elicx:
Addressed the actionable catch-all parity finding in `73a9cf4e`.

I also verified the acceptance boundary through a real Next.js 16.2.7 webpack dev server rather than calling `prepareDestination()` in isolation:

- `GET /search/foo%26admin=true` with `/api/search?q=:term&fixed=1` produced `{"query":{"q":"foo%26admin=true","fixed":"1","term":"foo%26admin=true"}}`
- `/docs/intro/guide` with `?path=:path*` produced `{"query":{"path":"intro/guide"}}`
- the catch-all redirect emitted `location: /login?next=/foo%26next%3Dhttps%3A%2F%2Fevil.example&safe=1`

The implementation now preserves `/` for `:param*` / `:param+` query substitutions while encoding separators for ordinary params. The full `tests/shims.test.ts` suite passes (1,148 tests).

/bigbonk review for issues

james-elicx:
Clarified the concrete vinext-specific boundary in `fffc164a`:

- Next.js retains encoded source captures through destination interpolation.
- Vinext calls the shared config matcher with an already-decoded `cleanPathname`.
- This layer therefore re-encodes query substitutions so encoded path data does not become query syntax.
- Catch-all `/` separators remain unescaped, matching the real Next.js server behavior.

The code comments, test comments, and PR description no longer claim that `compileNonPath()` itself performs this encoding. Targeted checks remain green.

/bigbonk review for issues

PR #2348, opened 2026-06-25
fix(build): apply configured webpack loaders

james-elicx:
## Summary
- retain user webpack loader rules discovered during next.config probing
- run matching loader chains before vinext static-image handling
- cover the Next.js webpack-loader-set-environment-variable App Router fixture in a focused production build test

## Validation
- `vp check packages/vinext/src/plugins/webpack-loader-compat.ts packages/vinext/src/config/next-config.ts packages/vinext/src/index.ts tests/webpack-loader-compat.test.ts`
- `vp test run tests/webpack-loader-compat.test.ts`
- `vp test run tests/next-config.test.ts -t "webpack loader callbacks|webpack config"`
- targeted Next.js v16.2.6 suite attempted with concurrency 1, but the machine hit `ENOSPC` before deployment and did not reach assertions; exact-head CI will classify end-to-end

Fixes the baseline failure in `test/e2e/app-dir/webpack-loader-set-environment-variable/webpack-loader-set-environment-variable.test.ts` from run 28143992598.


james-elicx:
Targeted independent review for the baseline `webpack-loader-set-environment-variable` failure:

The PR **does cover the exact run-28143992598 failure** and the fixture is faithful to the Next.js v16.2.6 setup. Please keep this draft for one remaining compatibility blocker before graduation:

- `resolveId()` strips the resolved query at `packages/vinext/src/plugins/webpack-loader-compat.ts:240` and returns a path-only virtual ID at line 244, while `matchesRule()` only evaluates path conditions. Therefore a rule such as `{ test: /\.svg$/ }` force-routes `icon.svg?url` / `?raw` through the webpack loader instead of preserving Vite's query semantics. Rules with `resourceQuery` are also currently applied to every matching path because that condition is ignored. Next.js/webpack uses `resourceQuery` extensively, so this can silently change unrelated asset imports as soon as one custom loader rule is configured.

Suggested focused remediation: carry the full resource query into matching/context and honor `resourceQuery`; if full query compatibility is intentionally out of scope, conservatively leave query-qualified imports to Vite unless a supported `resourceQuery` condition explicitly matches. Add a plugin-level regression covering bare `.svg`, `.svg?url`, and a `resourceQuery: /react/` rule.

I also confirmed the async callback race reported earlier is fixed at current head `688beaa3`. I did not duplicate the PR or modify its branch. Local focused execution was attempted, but the isolated worktree could not resolve linked workspace peer dependencies; CI is the authoritative run for the current head.

james-elicx:
Exact-head CI blocker at `569c492f` is in-scope, not ambient:

`Vitest (unit 1/3)` has 14 failures in `tests/image-imports.test.ts` with:

```
TypeError: Cannot read properties of undefined (reading 'webpackLoaderRules')
  packages/vinext/src/index.ts:4811
```

The image transform unit harness invokes the plugin before `nextConfig` is populated, while the new custom-loader bypass dereferences `nextConfig.webpackLoaderRules` directly. Please use the same defensive fallback as the compat plugin registration (for example `nextConfig?.webpackLoaderRules ?? { client: [], server: [] }`) and add/retain focused image-import coverage proving ordinary static images still transform when config has not resolved and custom-loader matches bypass only when rules exist.

Failed job: https://github.com/cloudflare/vinext/actions/runs/28185426521/job/83485969127

james-elicx:
Coverage/classification for Next.js Deploy Suite run 28478866791, report job 84413308650:

This PR covers the failure in `test/e2e/app-dir/webpack-loader-set-environment-variable/webpack-loader-set-environment-variable.test.ts` (`webpack-loader-set-environment-variable > loader that sets an environment variable should work`). The run failed because the rendered page returned the error shell text instead of `hello world`, consistent with the configured SVG webpack loader not being applied before the env-dependent build path. PR #2348 preserves configured webpack loader rules, applies matching loader chains, and adds focused coverage for this exact fixture.

Classification: covered by this build-plugin compatibility PR; not a separate functional App/Pages Router backlog item.

PR #2358, opened 2026-06-25
fix(app-router): preserve rewritten route identity

james-elicx:
## Summary

- keep browser-visible App Router pathnames separate from the route identity selected after middleware rewrites
- encode RSC route/page keys from the rewritten matched route instead of the original visible URL
- add a production regression for a prefetched route A whose navigation request rewrites to route B without changing the address bar

## Next.js parity

Targets `test/e2e/app-dir/concurrent-navigations/mismatching-prefetch.test.ts` from Next.js v16.2.6.

The baseline deploy suite stops earlier in that test while waiting for Next.js's Suspense-derived partial prefetch request. Vinext's equivalent explicit loading-boundary fixture reproduces the latent next-stage failure: the response had params/segment data for B but `__route` and page element keys for A, causing the client RSC error boundary. This PR fixes that independently reproduced route-identity mismatch. The exact upstream suite still requires the separate Suspense shell-classification parity gap before it can reach this assertion.

## Validation

- `vp check` on all changed files
- `vp test run tests/app-page-render-identity.test.ts tests/app-page-element-builder.test.ts` — 43 passed
- production app-basic build and one-worker Playwright regression — 1 passed
- `vp run vinext#build`
- targeted Next.js deploy suite deployed successfully but retained the baseline first-stage timeout waiting for the Suspense-derived prefetch request

Excludes cacheComponents, PPR, resume, fallback-shell, and external endpoint failures.


james-elicx:
Updated the regression to an isolated production-only browser fixture. This removes shared app-basic middleware changes, avoids dev mode where Link prefetch is intentionally disabled, and resolves the unrelated middleware invocation flake. Local exact browser spec passes at `f1b7ce6c`.

PR #2375, opened 2026-06-26
ci: run deploy suite against Next 16.3 preview

NathanDrake2406:
Updates the daily Next.js deploy suite default from `v16.2.6` to `v16.3.0-preview.5`, keeping checkout, concurrency, and report metadata in sync.

This makes the suite track the newest preview now that the current open PR queue already covers much of the `v16.2.6` fallout.

@james-elicx they have a new instant navigation feature on 16.3 which I wanna explore, would be great if we bump this test suite to use as an oracle


james-elicx:
Not planning to bump this week, sorry

PR #2406, opened 2026-06-29
fix(ci): resolve setup-vp version from catalog

NathanDrake2406:
Deploy Examples failed before deploy because parallel jobs let setup-vp resolve `vite-plus@latest`, and one npm latest lookup returned 1015. Resolve the Vite+ version from `pnpm-workspace.yaml` in the shared setup wrapper so normal catalog bumps stay the source of truth.

Validation: YAML parsed, resolver script tested for catalog and override paths, `vp fmt --check .github/actions/setup/action.yml`, pre-commit `vp check --fix` and full check.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -33,32 +33,7 @@ runs:
         if [ -n "$INPUT_VERSION" ]; then
           version="$INPUT_VERSION"
         else
-          version="$(node <<'NODE'
-        const fs = require("node:fs");
-
-        const workspace = fs.readFileSync("pnpm-workspace.yaml", "utf8");
-        let inCatalog = false;
-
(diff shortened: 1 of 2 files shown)


NathanDrake2406:
Leaving the perf workflow on the direct pinned `voidzero-dev/setup-vp` action intentionally. We tried routing perf through the repo-local setup wrapper earlier and it made the benchmark path not runnable because perf bootstraps the head checkout, base checkout, and trusted benchmark manifests separately. Using the PR-local wrapper there would make the benchmark runner depend on code from the checkout being measured instead of the trusted workflow bootstrap.\n\nThis PR still fixes the shared setup action used by normal CI, but I am not changing `.github/workflows/perf.yml` as part of this review.

PR #2445, opened 2026-07-01
fix(app-router): separate header-only full-route rsc requests

james-elicx:
## Summary
- Treat `RSC: 1` as a full-route RSC request marker for App Router HTML paths
- Redirect header-only HTML-path RSC requests onto a `_rsc` URL before serving Flight, preserving cache separation
- Keep legacy `.rsc` transport URLs accepted without adding `_rsc`

## Validation
- `vp test run tests/app-rsc-request-normalization.test.ts tests/app-rsc-cache-busting.test.ts tests/app-rsc-handler.test.ts`
- `vp check packages/vinext/src/server/app-rsc-cache-busting.ts packages/vinext/src/server/app-rsc-request-normalization.ts tests/app-rsc-cache-busting.test.ts tests/app-rsc-handler.test.ts tests/app-rsc-request-normalization.test.ts`
- `REPO="$(pwd)" NEXTJS_DIR="/Users/jamesanderson/Developer/vinext/.nextjs-ref" NEXT_TEST_CONCURRENCY=1 ./scripts/run-targeted-nextjs-e2e.sh test/e2e/app-dir/ppr-root-param-rsc-fallback/ppr-root-param-rsc-fallback.test.ts`

### DIFF
--- a/packages/vinext/src/server/app-rsc-cache-busting.ts
+++ b/packages/vinext/src/server/app-rsc-cache-busting.ts
@@ -357,7 +357,7 @@ export async function resolveInvalidRscCacheBustingRequest(
   const actualHash = url.searchParams.get(VINEXT_RSC_CACHE_BUSTING_SEARCH_PARAM);
   const expectedHash = await computeRscCacheBustingSearchParam(options.request.headers);
 
-  if (actualHash === null && expectedHash === "") {
+  if (actualHash === null && expectedHash === "" && url.pathname.endsWith(".rsc")) {
     return null;
   }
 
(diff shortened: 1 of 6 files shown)


james-elicx:
I think this PR already covers run 28478866791 / job 84413308650 for `test/e2e/app-dir/ppr-root-param-rsc-fallback/ppr-root-param-rsc-fallback.test.ts`, including the two full-route RSC cases for non-pregenerated and pregenerated root params. The fix treats header-only `RSC: 1` HTML-path requests as full-route RSC while keeping Flight cache-separated via `_rsc`, which matches the failure mode from those assertions.

PR #2451, opened 2026-07-01
fix(pages): align middleware data prefetch caching

james-elicx:
## Summary

- align Pages Router middleware data prefetch handling with Next.js by returning the prefetch skip protocol for matched non-SSG data routes
- preserve `x-middleware-cache` through middleware response shaping so SSG middleware prefetches can opt out of the client cache
- route Pages data prefetch cache keys through the active/default locale path and avoid persisting non-SSG middleware-prefetch data

## Validation

- Targeted local validation passed for Pages data prefetch, pages-data-route, pages-request-pipeline, request-pipeline, middleware-runtime, pages-data-route, link-navigation, shims, and scoped `vp check`.
- Targeted upstream Next.js middleware-rewrites assertions passed for opt-out prefetch caching, dynamic no-cache, non-SSG no-prefetch, and recursive rewrites.
- Caveat: the full upstream `test/e2e/middleware-rewrites/test/index.test.ts` still has an unrelated broader timeout in `should handle static rewrite from next.config.js correctly`, and a later direct Next harness attempt for one assertion failed before assertions due to the local `.nextjs-ref` SWC setup mismatch.

### DIFF
--- a/packages/vinext/src/entries/pages-server-entry.ts
+++ b/packages/vinext/src/entries/pages-server-entry.ts
@@ -7,6 +7,7 @@
  *
  * Extracted from index.ts.
  */
+import { readFile } from "node:fs/promises";
 import { resolveEntryPath } from "./runtime-entry-module.js";
 import { normalizePathSeparators } from "../utils/path.js";
 import { pagesRouter, apiRouter, type Route } from "../routing/pages-router.js";
@@ -14,6 +15,7 @@ import { createValidFileMatcher } from "../routing/file-matcher.js";
 import { type ResolvedNextConfig } from "../config/next-config.js";
 import { isProxyFile } from "../server/middleware.js";
 import { findFileWithExts } from "./pages-entry-helpers.js";
+import { hasExportedName } from "../build/report.js";
 
 const _requestContextShimPath = resolveEntryPath("../shims/request-context.js", import.meta.url);
 const _middlewareRuntimePath = resolveEntryPath("../server/middleware-runtime.js", import.meta.url);
@@ -25,6 +27,13 @@ const _pagesApiRoutePath = resolveEntryPath("../server/pages-api-route.js", impo
 const _serverGlobalsPath = resolveEntryPath("../server/server-globals.js", import.meta.url);
 const _queryUtilsPath = resolveEntryPath("../utils/query.js", import.meta.url);
 const _pagesPageHandlerPath = resolveEntryPath("../server/pages-page-handler.js", import.meta.url);
+
+async function getPagesDataKind(filePath: string): Promise<"static" | "server" | "none"> {
+  const source = await readFile(filePath, "utf8");
+  if (hasExportedName(source, "getStaticProps")) return "static";
+  if (hasExportedName(source, "getServerSideProps")) return "server";
+  return "none";
+}
 
 /**
  * Generate the virtual SSR server entry module.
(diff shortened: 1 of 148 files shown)


packages/vinext/src/entries/pages-server-entry.ts
```diff
@@ -26,6 +28,13 @@ const _serverGlobalsPath = resolveEntryPath("../server/server-globals.js", impor
 const _queryUtilsPath = resolveEntryPath("../utils/query.js", import.meta.url);
 const _pagesPageHandlerPath = resolveEntryPath("../server/pages-page-handler.js", import.meta.url);
 
+async function getPagesDataKind(filePath: string): Promise<"static" | "server" | "none"> {
+  const source = await readFile(filePath, "utf8");
+  if (hasExportedName(source, "getStaticProps")) return "static";
```
NathanDrake2406, line 33: This feels like the point where `dataKind` needs a single owner rather than another local source parser. The client entry already classifies Pages routes for `__VINEXT_PAGES_SSG_PATTERNS__` / `__VINEXT_PAGES_SSP_PATTERNS__`, this adds a second build-time classifier, and dev adds a third request-time classifier in `index.ts`. That leaves one protocol decision split across three file-read/AST paths. Can we pull this into a shared `collectPagesDataKinds(pageRoutes)` (or enrich the route metadata once) and have client entry, server entry, and dev consume the same map? That would delete the duplicated parsing and let the pipeline require a real `dataKind` instead of accepting an optional one.

packages/vinext/src/server/pages-request-pipeline.ts
```diff
@@ -50,6 +51,10 @@ export type PagesRenderOptions = {
 
 export type FilesystemRoutePhase = "direct" | "beforeFiles" | "afterFiles" | "fallback";
 
+type PageRouteMatch = {
+  route: { isDynamic: boolean; pattern?: string; dataKind?: "static" | "server" | "none" };
+};
+
```
NathanDrake2406, line 57: `dataKind` shouldn't be optional at this boundary while the skip branch treats every non-`"static"` value as eligible for the prefetch-skip response. If classification is absent/unknown, we now take the unsafe path and bypass rendering; pre-PR behaviour would have rendered. Please either make `dataKind` required for every `matchPageRoute` result, or make the skip check explicit (`dataKind === "server" || dataKind === "none"`) so `undefined` preserves the old render path.

NathanDrake2406 (changes requested):
Requesting changes on the shape of the new `dataKind` contract. The protocol behaviour looks reasonable, but the implementation makes the route classification both scattered and non-conservative at the pipeline boundary. This is not worth carrying as another optional string through the request flow.

james-elicx:
Worker lane coverage note for deploy-suite run 28478866791 / job 84413308650: this PR covers the middleware-rewrites prefetch failures in this assignment (`should not prefetch non-SSG routes` and `should opt out of prefetch caching for dynamic routes`). I am skipping a duplicate implementation for those rows and keeping the separate trailing-slash branch focused on the uncovered data-path normalization rows.

james-elicx:
Backlog mapping for GitHub Actions run 28478866791 / job 84413308650: this PR appears to own the Pages Router middleware data-prefetch caching failures, alongside the broader middleware rewrite/navigation-state coverage in #2454.

Per instruction I am skipping a duplicate implementation because this PR is by `james-elicx`. Remaining work looks like resolving review feedback and rerunning the exact Pages middleware/data-prefetch targeted suites.

PR #2453, opened 2026-07-01
fix(app-router): recover mismatched prefetch rewrites

james-elicx:
## Summary
- recover App Router navigations when a prefetched URL is rewritten to a different route
- preserve optimistic loading shells by probing nested server-component Suspense fallbacks
- only demote query-string prefetch cache reuse when middleware matcher conditions confidently require skipping prefetch requests

## Validation
- `vp check packages/vinext/src/shims/link.tsx tests/link-navigation.test.ts packages/vinext/src/server/app-optimistic-routing.ts packages/vinext/src/server/app-page-route-wiring.tsx`
- `vp test run tests/link-navigation.test.ts tests/app-page-probe.test.ts tests/app-page-dispatch.test.ts tests/app-optimistic-routing.test.ts tests/app-page-route-wiring.test.ts`
- `vp exec playwright test -c tests/e2e/app-router/nextjs-compat/playwright.nextjs-compat.config.ts tests/e2e/app-router/nextjs-compat/mismatching-prefetch.browser.spec.ts --project=chrome-browser-specific`

Independent review loop completed with no findings after follow-up fixes.

### DIFF
--- a/packages/vinext/src/entries/app-browser-entry.ts
+++ b/packages/vinext/src/entries/app-browser-entry.ts
@@ -6,6 +6,7 @@ import type {
 import type { AppRoute } from "../routing/app-router.js";
 import type { RouteManifest } from "../routing/app-route-graph.js";
 import type { NextRewrite } from "../config/next-config.js";
+import type { StaticMiddlewareMatcher } from "../build/report.js";
 
 /**
  * Generate the virtual browser entry module.
@@ -23,6 +24,7 @@ export function generateBrowserEntry(
     beforeFiles: [],
     fallback: [],
   },
+  middlewareMatcher?: StaticMiddlewareMatcher,
 ): string {
   const entryPath = resolveRuntimeEntryModule("app-browser-entry");
   const navigationRuntimePath = resolveClientRuntimeModule("navigation-runtime");
@@ -39,6 +41,7 @@ window.__VINEXT_LINK_PREFETCH_ROUTES__ = ${JSON.stringify(prefetchRoutes)};
 // the same — whichever entry runs first emits both globals).
 window.__VINEXT_PAGES_LINK_PREFETCH_ROUTES__ = ${JSON.stringify(pagesPrefetchRoutes)};
 window.__VINEXT_CLIENT_REWRITES__ = ${JSON.stringify(rewrites)};
+window.__VINEXT_MIDDLEWARE_MATCHER__ = ${JSON.stringify(middlewareMatcher)};
 registerNavigationRuntimeBootstrap({
     routeManifest: ${buildRouteManifestExpression(routeManifest)}
 });
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -365,7 +365,10 @@ import {
   resolveAppPageChildSegments as __resolveAppPageChildSegments,
 } from ${JSON.stringify(appPageRouteWiringPath)};
 import { buildPageElements as __buildPageElements } from ${JSON.stringify(appPageElementBuilderPath)};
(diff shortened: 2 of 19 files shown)


NathanDrake2406 (changes requested):
Requesting changes on the client-side middleware matcher duplication.

The new App Router prefetch path adds another partial middleware matcher implementation in `packages/vinext/src/shims/link.tsx`: `ClientMiddlewareMatcherObject`, `isClientMiddlewareMatcherObject`, `stripLocaleForMiddlewareMatcher`, `clientMiddlewareSourceMatches`, and a separate `has`/`missing` condition walker. We already have the canonical middleware matcher path in `packages/vinext/src/server/middleware-matcher.ts` backed by `packages/vinext/src/config/config-matchers.ts`, and there is also a client-side Pages copy in `packages/vinext/src/shims/internal/pages-data-target.ts`.

I do not think `link.tsx` should import the server module directly, but this PR should extract the client-safe matcher subset into a shared helper and reuse it from both the App Router prefetch demotion and the Pages data target. The helper can keep the two callers' different policies explicit, e.g. fail closed for App prefetch demotion and fail open for Pages middleware data probing, while sharing matcher-object validation, locale stripping, and source matching. That preserves this PR's behavior without leaving three drift-prone matcher implementations.

james-elicx:
I think this PR covers the functional failure from GH Actions run 28478866791 / job 84413308650 for `test/e2e/app-dir/concurrent-navigations/mismatching-prefetch.test.ts`, especially `mismatching prefetch recovers when a navigation rewrites to a different route than the one that was prefetched`.

Why: the PR specifically handles App Router navigations where a prefetched URL rewrites to a different route, and its validation lists the focused mismatching-prefetch E2E plus the related client navigation/prefetch tests. I did not see this as a cacheComponents/use-cache-only failure.

james-elicx:
Backlog mapping for GitHub Actions run 28478866791 / job 84413308650: this PR appears to own the functional App Router failure in `test/e2e/app-dir/concurrent-navigations/mismatching-prefetch.test.ts` around mismatched prefetch rewrite recovery.

Per instruction I am skipping a duplicate implementation because this PR is by `james-elicx`. Remaining work looks like addressing review feedback and rerunning the exact targeted Next.js E2E proof for that path.

james-elicx:
Run 29871622126 / job 88775791401 still fails `test/e2e/app-dir/concurrent-navigations/mismatching-prefetch.test.ts` at `recovers when a navigation rewrites to a different route than the one that was prefetched`. I think this PR owns that functional loading-shell/rewrite recovery failure.

I reverified the exact targeted Next.js v16.2.6 suite on the current-main campaign baseline (`3d30a3ca3`): it still fails because the click response does not contain `Loading a...`. I am adding this existing PR to the revival backlog rather than opening a duplicate; the refreshed branch will need the same exact targeted proof.

james-elicx:
Correction to my prior run mapping: `apps/web/app/compatibility/suite-support.ts` explicitly classifies `test/e2e/app-dir/concurrent-navigations/mismatching-prefetch.test.ts` as deferred Partial Prerendering. The current-main failure is genuine, but it is intentionally outside this campaign’s fix-now backlog. I am removing #2453 from the active run-29871622126 work queue; no revival action should be inferred from my earlier comment.

james-elicx:
Correction after live freshness review: this PR must **not** be treated as the current fix for Actions run 31290819291.

Its head 18b485522 is from 2026-07-01, is currently DIRTY/conflicting with main, and has not passed the targeted Next.js E2E against the current repository state. The current-main failure remains reproducible. The older implementation also uses structural first-Suspense fallback extraction and broad payload replacement, which do not preserve the actual React partial shell in multi-boundary/static-sibling cases.

I am therefore withdrawing the prior "skip duplicate" mapping. The run-312908 backlog item remains active until a current-main implementation has exact targeted E2E proof, clean independent review, Bonk, and CI.

PR #2455, opened 2026-07-01
fix(app-router): reuse metadata rewrite prefetches

james-elicx:
## Summary
- resolve App Router rewrite aliases when looking up prefetch/navigation cache entries
- consume alternate cached RSC URLs while normalizing alias response URLs back to the visible route
- dedupe the visible rewrite alias probe so metadata-prefetch navigations stay request-free

## Validation
- `vp test run tests/prefetch-cache.test.ts tests/hybrid-client-route-owner.test.ts tests/link-navigation.test.ts`
- `vp check --fix packages/vinext/src/shims/link.tsx tests/e2e/app-router/nextjs-compat/segment-cache-metadata.browser.spec.ts`
- `vp run vinext#build`
- `PLAYWRIGHT_PROJECT=app-router-chrome-browser-specific pnpm run test:e2e -- tests/e2e/app-router/nextjs-compat/segment-cache-metadata.browser.spec.ts`
- targeted Next.js E2E `test/e2e/app-dir/segment-cache/metadata/segment-cache-metadata.test.ts` passed

Independent review loop completed with no findings.


NathanDrake2406 (changes requested):
Requesting changes on structure.

The alias reuse direction looks right, but `resolveHybridClientRewriteHref` duplicates the rewrite traversal already owned by `resolveHybridClientRouteOwner`: beforeFiles, the dynamic-route gate, afterFiles, fallback, and document bailout now exist in two separate flows.

That makes the owner decision and the cache-alias href drift-prone. These are not separate concepts; they are two facts produced by the same route-resolution walk. Can we extract a single resolver that walks rewrites once and returns both the owner and the rewritten href, then have `resolveHybridClientRouteOwner` and the cache-alias path consume that shared result?

Something like:

```ts
type HybridClientRouteResolution =
  | { kind: "document" }
  | {
      kind: "route";
      owner: HybridClientOwner | null;
      href: string;
      rewritten: boolean;
    };
```

That would remove the duplicated state machine and make future rewrite-order changes much safer.

I also think `prefetchUrl` is carrying too much of this policy directly now: rewrite alias resolution, alias cache candidates, shell/full fetch orchestration, and the alias HEAD probe are all embedded in the same already-busy function. After the shared route-resolution helper exists, I’d prefer the cache-candidate construction to move behind a small helper as well, so `link.tsx` is not manually threading `additionalRscUrls` as an ad-hoc side channel.

james-elicx:
I think this PR covers the segment-cache metadata failures from run 28478866791 / report job 84413308650. The report has exactly the two non-cacheComponents failures in `test/e2e/app-dir/segment-cache/metadata/segment-cache-metadata.test.ts`: dynamic content + dynamic metadata with a full prefetch, and runtime-prefetchable content + dynamic metadata with a runtime prefetch. This PR adds the matching dynamic-head/runtime-prefetchable-head metadata-prefetch regression coverage and its validation notes say the exact targeted upstream metadata E2E passed.

james-elicx:
Backlog mapping for GitHub Actions run 28478866791 / job 84413308650: this PR appears to own the App Router metadata rewrite/prefetch failures in `test/e2e/app-dir/segment-cache/metadata/segment-cache-metadata.test.ts`.

Per instruction I am skipping a duplicate implementation because this PR is by `james-elicx`. Remaining work looks like addressing review feedback and rerunning the exact metadata targeted Next.js E2E proof.

PR #2458, opened 2026-07-01
fix(app-router): support route tree prefetch inlining

james-elicx:
## Summary
- add App Router route-tree prefetch handling so inlined segment-cache prefetch payloads include the expected deployment/build metadata
- size route-tree inlining against gzip-compressed bytes
- add focused route-tree unit coverage plus a local compat fixture/spec for prefetch inlining

## Validation
- `vp test run tests/app-route-tree-prefetch.test.ts`
- targeted upstream `test/e2e/app-dir/segment-cache/prefetch-inlining/prefetch-inlining.test.ts` passed 7/7 before PR
- independent review: NO FINDINGS

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -706,6 +706,7 @@ export default createAppRscHandler({
   basePath: __basePath,
   buildId: process.env.__VINEXT_BUILD_ID ?? null,
   ensureRouteLoaded: __ensureRouteLoaded,
+  renderToReadableStream,
   clearRequestContext() {
     __clearRequestContext();
   },
--- a/packages/vinext/src/server/app-route-tree-prefetch.ts
+++ b/packages/vinext/src/server/app-route-tree-prefetch.ts
@@ -1,0 +1,374 @@
+import {
+  NEXT_ROUTER_SEGMENT_PREFETCH_HEADER,
+  NEXTJS_DEPLOYMENT_ID_HEADER,
+  RSC_HEADER,
+  NEXT_ROUTER_PREFETCH_HEADER,
+} from "./headers.js";
+import {
+  VINEXT_RSC_CONTENT_TYPE,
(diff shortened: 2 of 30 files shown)


NathanDrake2406:
Request changes.

The route-tree prefetch implementation is sitting in the wrong layer. handleAppRscRequest now answers /_tree by calling createRouteTreePrefetchResponse in the live request path, and that builder walks the route tree and re-renders layout/page modules with synthetic props just to estimate gzip size.

That turns a metadata request into arbitrary user component execution. It also makes prefetch hints a request-time rendering heuristic rather than stable route/build metadata. If sizing fails, the error is swallowed and the route silently falls back to an unknown size, which makes the output dependent on accidental renderability rather than the canonical app route model.

Can we move this into the build/prerender/route metadata layer instead? The request path should be boring: match the route, load stable route-tree/hint metadata, serialize RootTreePrefetch. If we cannot compute faithful inline sizes yet, the safer behaviour is conservative/outlined hints rather than rendering user components during the prefetch request.

There is also a model-drift issue. This PR creates a new hand-rolled route-tree shape beside the existing route graph/app-page wiring. Dynamic segment params always emit siblings: null, despite staticSiblings already being computed for client cache reuse, and the slot model ignores canonical slot metadata such as config layouts/config layout positions. That is likely to diverge from real rendering as App Router features stack up.

I’d rather see the route tree derived from the canonical graph/elements data, with hint generation owned by build/prerender metadata, not by a request-time component render pass.

james-elicx:
Backlog coordination: I was assigned GitHub Actions run 28478866791 job 84413308650 for `test/e2e/app-dir/segment-cache/prefetch-inlining/prefetch-inlining.test.ts`. This PR already appears to cover that lane: the failure is the missing App Router route-tree `/_tree` prefetch path used by Segment Cache prefetch inlining, where the client expects an inlined prefetch payload with deployment/build metadata and size decisions based on compressed route-tree payloads.

I am not opening a duplicate patch for this lane.

PR #2459, opened 2026-07-01
fix(app-router): preserve dynamic segment-cache navigation

james-elicx:
## Summary
- preserve client route alias resolution for segment-cache navigations with dynamic segments
- add coverage for longest wildcard aliases and exact package aliases in the app browser route manifest

## Validation
- vp test run tests/entry-templates.test.ts tests/link.test.ts tests/navigation-planner-early-intent.test.ts
- vp test run tests/app-browser-client-reuse-manifest.test.ts tests/app-optimistic-routing.test.ts tests/link-navigation.test.ts
- REPO="/Users/jamesanderson/.codex/worktrees/segment-cache-basic-28478866791/vinext" NEXTJS_DIR="/Users/jamesanderson/Developer/vinext/.nextjs-ref" NEXT_TEST_CONCURRENCY=1 ./scripts/run-targeted-nextjs-e2e.sh test/e2e/app-dir/segment-cache/basic/segment-cache-basic.test.ts

### DIFF
--- a/packages/vinext/src/client/vinext-next-data.ts
+++ b/packages/vinext/src/client/vinext-next-data.ts
@@ -13,6 +13,7 @@ export type VinextLinkPrefetchRoute = {
   documentOnly?: boolean;
   isDynamic: boolean;
   patternParts: string[];
+  requiresDynamicNavigationRequest?: boolean;
 };
 
 /**
@@ -30,6 +31,7 @@ export type VinextPagesLinkPrefetchRoute = {
   documentOnly?: boolean;
   isDynamic: boolean;
   patternParts: string[];
+  requiresDynamicNavigationRequest?: boolean;
 };
 
 export type VinextNextData = {
--- a/packages/vinext/src/config/tsconfig-paths.ts
+++ b/packages/vinext/src/config/tsconfig-paths.ts
@@ -36,7 +36,7 @@ type TsconfigPathResolution = {
   pathAliases: TsconfigPathAlias[];
 };
 
-export type TsconfigPathAlias = {
+type TsconfigPathAlias = {
   find: string;
   kind: "exact" | "prefix";
   replacement: string;
--- a/packages/vinext/src/config/tsconfig-paths.ts
+++ b/packages/vinext/src/config/tsconfig-paths.ts
@@ -33,6 +33,18 @@ const TSCONFIG_FILES = ["tsconfig.json", "jsconfig.json"];
 type TsconfigPathResolution = {
   aliases: Record<string, string>;
   baseUrl: string | null;
+  pathAliases: TsconfigPathAlias[];
+};
+
+export type TsconfigPathAlias = {
+  find: string;
+  kind: "exact" | "prefix";
+  replacement: string;
+};
+
+type MaterializedAliases = {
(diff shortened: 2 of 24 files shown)


NathanDrake2406:
Request changes.

The behaviour this PR is aiming for makes sense, but I don’t think this implementation is merge-ready yet.

The blocking issue is the dynamic-shell stream capture. `readAppPageInitialBinaryStream` decides that the shell is complete by racing each `reader.read()` against a hardcoded 25ms timeout. That makes the new `prefetch-dynamic-shell` path timing-dependent: under load or on a cold Workers isolate, we can truncate the shell before it has fully emitted; on a fast synchronous render, we risk capturing more than intended. It also leaks the timeout whenever `reader.read()` wins the race.

The awkward part is that this path already has the better primitive available. `dispatchAppPage` creates and installs a `PprFallbackShellState` for `APP_RSC_RENDER_MODE_PREFETCH_DYNAMIC_SHELL`, and the dynamic request APIs now suspend through that state. So the shell boundary should be driven by the existing fallback-shell abort lifecycle, not by a wall-clock idle guess.

There is also a structural concern: `entries/app-browser-entry.ts` is now doing filesystem/source/import-graph analysis in the same module that generates the virtual browser entry. That file now owns tsconfig alias resolution, package resolution, regex import parsing, recursive source traversal, and dynamic API detection. This is too much responsibility for an entry-template generator. Please move this into a dedicated route dynamic-request analysis module, with a narrow API consumed by `generateBrowserEntry` / route manifest projection.

A smaller but related issue: `dynamicRequestApiPattern` over-matches any local/member call named `headers`, `cookies`, `connection`, etc., and it runs before the import-origin-aware check. That means the precise scanner is bypassed for false-positive cases. If this is intentionally conservative, it should be documented; otherwise the broad pre-check should go.

The navigation-planner change also has a minor inconsistency: the comment says parsed search params are compared to avoid treating equivalent encodings as different, but the new same-URL flight branch also checks raw `current.search === next.search`. That raw check seems to contradict the surrounding invariant.

I’d like the stream-boundary issue fixed before merge. The extraction of the dynamic-request analyser should also happen before this settles, otherwise we’re normalising a fairly large chunk of handrolled source-analysis logic in the wrong layer.

james-elicx:
Deploy-suite run 28478866791 / report job 84413308650 triage: this PR appears to cover the non-excluded App Router navigation/prefetch behavior rows from `test/e2e/app-dir/segment-cache/basic/segment-cache-basic.test.ts`.

Exact rows:
- `segment cache (basic tests) > navigate before any data has loaded into the prefetch cache`
- `segment cache (basic tests) > navigate with prefetched data`
- `segment cache (basic tests) > prefetch interception route`
- `segment cache (basic tests) > prefetch interception route with params`
- `segment cache (basic tests) > skips dynamic request if prefetched data is fully static`
- `segment cache (basic tests) > skips static layouts during partially static navigation`
- `segment cache (basic tests) > refreshes page segments when navigating to the exact same URL as the current location`
- `segment cache (basic tests) > can handle circular references in client component props`

Since this PR is already James-owned, I am skipping duplicate implementation work for these rows.

PR #2463, opened 2026-07-01
fix(app-router): prefetch root-param route trees

james-elicx:
## Summary
- emit root-param route-tree metadata to the App Router Link prefetch manifest
- request segment route-tree prefetches for root-param links without encoding placeholder segments
- keep viewport visibility gated through IntersectionObserver while scheduling the route-tree task immediately with low fetch priority

## Validation
- vp test run tests/link.test.ts tests/link-navigation.test.ts -t "root-param|Link App Router prefetch mode|route-tree segment prefetch"
- vp check on changed root-param files
- targeted Next.js e2e: test/e2e/app-dir/segment-cache/vary-params/root-params-segment-prefetch.test.ts
- independent final re-review: NO FINDINGS

### DIFF
--- a/packages/vinext/src/client/vinext-next-data.ts
+++ b/packages/vinext/src/client/vinext-next-data.ts
@@ -13,6 +13,7 @@ export type VinextLinkPrefetchRoute = {
   documentOnly?: boolean;
   isDynamic: boolean;
   patternParts: string[];
+  shouldPrefetchRouteTree?: boolean;
 };
 
 /**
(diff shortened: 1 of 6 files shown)


james-elicx:
This appears to cover the current run 28478866791 / job 84413308650 failure in `test/e2e/app-dir/segment-cache/vary-params/root-params-segment-prefetch.test.ts`:

- `segment cache - root params segment prefetch > does not encode root param placeholders in segment-prefetch responses`

Since this PR is already owned by james-elicx, I am treating it as the owner for that failure class and skipping duplicate implementation work. The remaining coordinator risk is the current draft/dirty/red state rather than missing ownership.

PR #2464, opened 2026-07-01
fix(app-router): align dynamic-on-hover prefetches

james-elicx:
## Summary

- keep dynamic-on-hover upgrades behind an existing loading-shell prefetch
- omit static layouts from hover-time dynamic segment payloads
- add focused Link and App Router render coverage for shell-first dynamic hover prefetching

## Next.js parity

Targets the App Router Segment Cache behavior covered by:

- `test/e2e/app-dir/segment-cache/dynamic-on-hover/dynamic-on-hover.test.ts`

## Validation

- `vp test run tests/link-navigation.test.ts -t "loading-shell prefetch|shell prefetch fails|dynamic-on-hover"`
- `vp test run tests/app-page-render.test.ts tests/app-page-route-wiring.test.ts -t "dynamic-on-hover|loading-shell prefetches without|omits static layouts"`
- `vp test run tests/link-navigation.test.ts tests/app-page-render.test.ts tests/app-page-route-wiring.test.ts`
- `vp check` on touched files
- `REPO="$(pwd)" NEXTJS_DIR="/Users/jamesanderson/Developer/vinext/.nextjs-ref" NEXT_TEST_CONCURRENCY=1 ./scripts/run-targeted-nextjs-e2e.sh test/e2e/app-dir/segment-cache/dynamic-on-hover/dynamic-on-hover.test.ts` (1/1 passed)

## Review

Independent review found one issue before PR: hover dynamic-after-shell could proceed after a failed/stale shell prefetch and cache a payload missing static layouts. This branch includes the follow-up guard and regression test.

### DIFF
--- a/packages/vinext/src/server/app-page-render.ts
+++ b/packages/vinext/src/server/app-page-render.ts
@@ -37,7 +37,10 @@ import {
   renderAppPageHtmlStreamWithRecovery,
   type AppPageSsrHandler,
 } from "./app-page-stream.js";
-import type { AppRscRenderMode } from "./app-rsc-render-mode.js";
+import {
+  APP_RSC_RENDER_MODE_PREFETCH_DYNAMIC_AFTER_SHELL,
+  type AppRscRenderMode,
+} from "./app-rsc-render-mode.js";
 import {
   createArtifactCompatibilityEnvelope,
   createArtifactCompatibilityGraphVersion,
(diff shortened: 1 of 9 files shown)


packages/vinext/src/shims/link.tsx
```diff
@@ -488,9 +489,11 @@ function prefetchUrl(
         const autoPrefetch =
           mode === "auto"
             ? resolveAutoAppRoutePrefetch(prefetchHref)
-            : mode === "full-after-shell"
-              ? { cacheForNavigation: true, prefetchShellFirst: true, shouldPrefetch: true }
-              : resolveFullAppRoutePrefetch();
+            : mode === "auto-shell"
```
NathanDrake2406, line 494: This adds `auto-shell` as an unconditional prefetch plan, which bypasses the route-manifest eligibility check that normal `auto` gets through `resolveAutoAppRoutePrefetch()`. Since visible `unstable_dynamicOnHover` links are switched to this mode, any same-origin App-runtime link can now issue a shell RSC prefetch even when it has no `__VINEXT_LINK_PREFETCH_ROUTES__` match. This is exactly the kind of mode-string branching that makes `link.tsx` harder to reason about. Can we push this back into a single typed prefetch-plan helper that always does route eligibility first, then chooses shell-only vs navigation-cache behavior for matched routes?

packages/vinext/src/server/app-page-render.ts
```diff
@@ -509,6 +512,36 @@ function createRenderLifecycleSkipDisposition(input: {
   return plan.skipDisposition;
 }
 
+function createDynamicPrefetchAfterShellSkipDisposition(input: {
```
NathanDrake2406, line 515: This reuses `ClientReuseManifestSkipDisposition` and returns `SKIP_STATIC_LAYOUT_VERIFIED` directly from current `layoutFlags`, but this path is not the same invariant as the existing manifest/cache-proof verifier in `skip-cache-proof.ts`; it depends on the client having a fresh shell payload, not on server-side cache proof. Reusing the verified disposition muddies the boundary and makes later readers think the normal proof pipeline ran. Can we add an explicit shell-delta skip model/code, or move this behind a helper in the skip/cache-proof layer with a name that states the prerequisite?

NathanDrake2406 (changes requested):
Requesting changes on two structural issues: the Link prefetch mode now bypasses the existing route eligibility boundary, and the server skip disposition reuses a cache-proof contract for a different shell-delta invariant.

james-elicx:
Deploy-suite run 28478866791 / report job 84413308650 triage: this PR appears to cover the App Router hover-prefetch behavior row from `test/e2e/app-dir/segment-cache/dynamic-on-hover/dynamic-on-hover.test.ts`.

Exact row:
- `dynamic on hover > prefetches the dynamic data for a Link on hover`

Since this PR is already James-owned, I am skipping duplicate implementation work for this row.

PR #2440, opened 2026-06-30
fix(app-router): start optimistic prefetches immediately

james-elicx:
## Summary

- start App Router viewport prefetches immediately instead of waiting for browser idle
- keep low-priority Pages fallback document prefetches on requestIdleCallback
- add a focused regression for visible App Router links dispatching before idle callbacks

## Next.js parity

Fixes the optimistic-routing failures from deploy-suite run 28478866791:

- `test/e2e/app-dir/optimistic-routing/optimistic-routing.test.ts`

## Validation

- `vp check packages/vinext/src/shims/link.tsx tests/link-navigation.test.ts`
- `vp test run tests/link-navigation.test.ts tests/app-optimistic-routing.test.ts --maxWorkers=1` — 67 passed
- `NEXT_TEST_CONCURRENCY=1 REPO="$(pwd)" NEXTJS_DIR="/Users/jamesanderson/Developer/vinext/.nextjs-ref" ./scripts/run-targeted-nextjs-e2e.sh test/e2e/app-dir/optimistic-routing/optimistic-routing.test.ts` — 9 passed

## Notes

The broader app-prefetch protocol cluster from the same run appears covered by #2318, so this PR stays focused on the optimistic-routing dispatch timing issue.

PR #2467, opened 2026-07-01
fix(app-router): avoid runtime route-tree prefetch sizing

james-elicx:
## Summary
- avoid rendering user components during route-tree prefetch size checks
- normalize object-form `experimental.prefetchInlining` thresholds, including `Infinity`, and pass resolved values into the App RSC handler
- add route-tree prefetch sizing/config/codegen coverage plus a focused browser fixture

Stacked on #2440 (`codex/fix-optimistic-routing-prefetch`) to avoid bundling the prerequisite optimistic-prefetch diff.

## Validation
- `vp check` on touched source/test files
- `vp test run tests/app-route-tree-prefetch.test.ts tests/next-config.test.ts`
- `vp test run tests/app-rsc-handler.test.ts -t "route-tree"`
- `vp test run tests/link-navigation.test.ts -t "prefetchInlining|deduped"`
- `vp test run tests/app-router-next-config-codegen.test.ts tests/app-route-tree-prefetch.test.ts tests/next-config.test.ts`
- `vp run vinext#build`
- `REPO="$(pwd)" NEXTJS_DIR="/Users/jamesanderson/Developer/vinext/.nextjs-ref" NEXT_TEST_CONCURRENCY=1 ./scripts/run-targeted-nextjs-e2e.sh test/e2e/app-dir/segment-cache/max-prefetch-inlining/max-prefetch-inlining.test.ts`


NathanDrake2406 (changes requested):
## Review: route-tree prefetch inlining

Requesting changes on one contract issue, plus one structural cleanup I think should happen before this pattern spreads.

### 1. Blocking: route-tree dynamic segment siblings use the wrong data model

`packages/vinext/src/server/app-route-tree-prefetch.ts:211` passes `route.staticSiblings` into every dynamic route-tree segment. That field is route-wide and flattened across all dynamic URL levels; the existing route graph docs say so explicitly. But `TreePrefetchParam.siblings` is segment-local, matching Next's dynamic segment tuple.

For a route like `/[locale]/products/[id]`, any static sibling discovered at one dynamic level can be attached to the other dynamic segment in the route-tree response. That gives the client incorrect segment-cache metadata and makes the new helper depend on a field that was only shaped for the older full RSC payload.

Can we either emit segment-local static siblings into the route manifest and thread that into the route-tree builder, or set route-tree `siblings` to `null` until that data exists? Reusing the flattened field here is the wrong ownership boundary.

### 2. Structural: route-tree prefetches are being modeled as optimistic loading shells

`packages/vinext/src/shims/link.tsx:543` reuses the loading-shell gate path for route-tree prefetches and stores the `/_tree` response with `optimisticRouteShell: true`. That leaks a new transport concept into unrelated optimistic-template learning: `app-browser-entry.ts` scans settled prefetch entries and tries to decode them as App element payloads.

It mostly avoids hard failure because the learning loop awaits `Promise.allSettled`, but the shape is misleading. A route-tree response is not a loading-shell payload and should not be tagged as one. I think this wants an explicit cache kind, e.g. `prefetchKind: "route-tree" | "loading-shell" | "navigation"`, or a dedicated route-tree pending gate outside the navigation prefetch cache.

CI is green, and the earlier renderer-plumbing/type/parity issues are resolved. This review is about keeping the route-tree contract and cache model clean before more segment-cache behavior builds on top of it.

james-elicx:
Deploy-suite run 28478866791 / report job 84413308650 triage: this PR appears to cover the App Router max prefetch inlining rows from `test/e2e/app-dir/segment-cache/max-prefetch-inlining/max-prefetch-inlining.test.ts`.

Exact rows:
- `max prefetch inlining > bundles all segment data into a single request per route`
- `max prefetch inlining > deduplicates inlined prefetch requests for the same route`

Since this PR is already James-owned, I am skipping duplicate implementation work for these rows.

PR #2469, opened 2026-07-01
fix(app-router): publish committed no-prefetch navigations

james-elicx:
## Summary
- keep committed App Router navigation payloads eligible for visited-response publication even when the user immediately navigates back
- publish complete no-prefetch navigation snapshots into the visited-response cache instead of only seeding the Link prefetch cache
- add a Next.js-ported browser regression for the immediate-back cache publication path

## Validation
- `vp test run tests/app-visited-response-cache.test.ts tests/navigation-planner-prefetch-reuse.test.ts`
- `pnpm exec playwright test tests/e2e/app-router/nextjs-compat/client-cache.spec.ts --project=app-router-client-cache --workers=1 --retries=0`
- `vp check`
- `REPO="$(pwd)" NEXTJS_DIR="/Users/jamesanderson/Developer/vinext/.nextjs-ref" NEXT_TEST_CONCURRENCY=1 ./scripts/run-targeted-nextjs-e2e.sh test/e2e/app-dir/segment-cache/cached-navigations/cached-navigations.test.ts`

## Upstream target notes
The exact upstream target now passes `serves a fully static page without any requests on the second navigation`, which previously failed with an unexpected `/_rsc` request. Remaining assertions in that file still fail under the suite's `cacheComponents: true` / `use cache` segment staging behavior and are intentionally out of scope for this PR.

### DIFF
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -2169,7 +2169,7 @@ function bootstrapHydration(
               PREFETCH_CACHE_TTL,
               mountedSlotsHeader,
             );
-          } else if (committedState !== null && isCompleteAppPayloadMetadata(metadata)) {
+          } else if (committedState !== null) {
             const state = committedState as AppRouterState;
             const committedElements = {
               ...state.elements,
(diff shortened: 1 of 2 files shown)


james-elicx:
Deploy-suite run 28478866791 / report job 84413308650 triage: this PR appears to cover the non-excluded App Router cached-navigation behavior rows from `test/e2e/app-dir/segment-cache/cached-navigations/cached-navigations.test.ts`.

Exact rows:
- `cached navigations > serves cached static segments instantly on the second navigation`
- `cached navigations > serves a fully static page without any requests on the second navigation`
- `cached navigations > caches static segments when navigating to a known route without a prefetch`
- `cached navigations > includes static params in the cached static stage`
- `cached navigations > defers fallback params to the runtime stage`
- `cached navigations > caches runtime-prefetchable content from a navigation for instant second visit`
- `cached navigations > caches runtime-prefetchable content from the initial HTML for subsequent navigations`
- `cached navigations > caches a fully static page from the initial HTML for subsequent navigations`
- `cached navigations > caches a partially static page from the initial HTML for subsequent navigations`

Since this PR is already James-owned, I am skipping duplicate implementation work for these rows. I am not counting any cacheComponents/use-cache rows here.

PR #2478, opened 2026-07-01
fix(ci): harden perf publisher tree validation

james-elicx:
## Summary

- filter remote GitHub tree responses in the perf publisher validator before fingerprinting skipped Next.js benchmark inputs
- share the fingerprint input rules with the `gh api --jq` projection so only relevant tree entries cross into Node
- keep a larger child-process buffer as a secondary guard for trusted GitHub API calls
- add regression coverage that records the validator's tree requests, verifies rich/non-blob/irrelevant entries are filtered out, and checks the generated jq predicate against real `jq` when available

## Validation

- `pnpm run fmt:check -- benchmarks/perf/nextjs-input-fingerprint.mts benchmarks/perf/validate-results.mjs tests/performance-benchmarks.test.ts knip.ts`
- `vp test run tests/performance-benchmarks.test.ts`

### DIFF
--- a/benchmarks/perf/nextjs-input-fingerprint.mts
+++ b/benchmarks/perf/nextjs-input-fingerprint.mts
@@ -7,11 +7,38 @@ import { createHash } from "node:crypto";
 import { resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 
-const excludedPerfFiles = new Set([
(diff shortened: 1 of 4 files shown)


james-elicx:
Split the create-next-app/App Router dev resiliency changes out into draft PR #2479. This PR is now back to the perf publisher tree-validation hardening only.

PR #2481, opened 2026-07-01
feat(cloudflare): warm prerendered paths before deploy

james-elicx:
## Summary
- add an opt-in `--warm-cdn-cache` deploy path that uploads a Worker version, warms prerendered paths through the version preview URL, then deploys that exact version
- expose a narrow internal prerender-manifest helper for selecting concrete warmup paths
- guard warmup against fallback shells, 404 documents, missing preview URLs, and stale prerender manifests

## Validation
- `vp check packages/cloudflare/src/deploy.ts packages/cloudflare/src/cdn-warm.ts packages/cloudflare/src/version-deploy.ts packages/cloudflare/src/cli.ts packages/cloudflare/src/deploy-help.ts packages/vinext/src/server/prerender-manifest.ts tests/deploy.test.ts tests/deploy-prerender-config.test.ts tests/cloudflare-cdn-warm.test.ts tests/cloudflare-version-deploy.test.ts tests/pregenerated-concrete-paths.test.ts`
- `vp test run tests/deploy.test.ts tests/deploy-prerender-config.test.ts tests/cloudflare-cdn-warm.test.ts tests/cloudflare-version-deploy.test.ts tests/pregenerated-concrete-paths.test.ts`


james-elicx:
Addressed the actionable part of the Bonk finding: `vinext-cloudflare deploy --skip-build --warm-cdn-cache` now still emits `dist/server/vinext-prerender-paths.json` from the existing build output before reading warm paths.

Intentional contract for review: normal `vite build` / `vinext build` should not run path discovery or emit the warmup path manifest unless it is an explicit prerender build. Pre-warm deploys are the trigger for warmup path discovery, including `--skip-build`.

/bigbonk review for issues

james-elicx:
Addressed the remaining nit from the last Bonk pass: Cloudflare deploy now emits the supplemental path manifest for warm deploys and for non-skip explicit prerender builds only. That means route-root metadata is always loaded when the manifest is emitted, while `--skip-build --prerender-all` keeps the existing contract of not loading Vite config just for a supplemental manifest.

/bigbonk review for issues

james-elicx:
Added the optional contract-lock test from the last Bonk pass: `--skip-build --prerender-all` without `--warm-cdn-cache` now explicitly asserts that no supplemental warmup path manifest is written.

/bigbonk review for issues

james-elicx:
Fixed the workers-cache deploy issue from the attached real run: when fallback warmup is used and `wrangler versions deploy` does not print a production URL, we now parse the URL from `wrangler triggers deploy` and warm that target instead of skipping with “no production URL could be inferred”. Added a regression matching that mixed 100%/0% deployment state.

/bigbonk review for issues

james-elicx:
Updated the deploy logs so the version flow distinguishes staging at 0%, fallback promotion, and final warmed promotion. The real workers-cache run should now read as staging -> triggers -> warm -> promote instead of repeating the same “Deploying warmed” label.

/bigbonk review for issues

PR #2484, opened 2026-07-02
refactor: migrate the CLI cluster path handling to pathslash

shulaoda:
## What

First batch of the `node:path` → [`pathslash`](https://github.com/shulaoda/pathslash) migration. Introduces the dependency and migrates the self-contained CLI cluster:

- `packages/vinext/src/typegen.ts`
- `packages/vinext/src/check.ts`
- `packages/vinext/src/cli.ts`
- `packages/vinext/src/init.ts`
- `packages/vinext/src/init-cloudflare.ts`
- `packages/vinext/src/utils/project.ts`

## Why

The forward-slash invariant currently lives in JSDoc prose ("`root` must be forward-slash — callers normalize it at the CLI entry…") and has to be restated and hand-maintained at every hop, which produced a five-week, 50+-commit whack-a-mole campaign. `pathslash` turns that convention into an API guarantee: it delegates every operation to the real `node:path` (drive letters, per-drive cwd, case-insensitive `relative` all keep native semantics) and only converts Windows output separators to `/`. Once a module imports it, every `join`/`resolve`/`relative`/`dirname` result is canonical forward-slash by construction, so the per-function contracts, the `path.posix.*` rewrites, and the `normalizePathSeparators(...)` wrappers all become unnecessary in migrated code.

Notably, `pathslash.relative` bakes in the rule this repo learned the hard way in #2308: relativize natively, then normalize the output (`E:/...` paths are not POSIX-absolute, so `path.posix.relative` was a dead end).

## How

- `import path from "node:path"` → `import path from "pathslash"` in the six files.
- `path.posix.join/dirname` → `path.join/dirname`; `normalizePathSeparators(path.resolve/relative(...))` → bare `path.resolve/relative(...)`.
- Raw `process.cwd()` reads that never pass through a path API are normalized once at the entry with `toSlash` (build command root, `check` root); `runPrerender`/`printBuildReport` now reuse the already-canonical `root` instead of re-reading and re-normalizing `process.cwd()`.
- Obsolete "must be forward-slash" JSDoc contracts removed outright — the guarantee now lives in the `pathslash` import, not in per-function prose. Contracts on unmigrated files stay until their batch migrates them.
- `init-cloudflare.ts` still emits `import path from "node:path"` literals into _generated user configs_ — those are intentionally untouched.
- `utils/path.ts` (`normalizePathSeparators`) stays: unmigrated files and tests still consume it; it is deleted in the final batch.
- Workspace: `pathslash@^0.1.0` added via catalog; `minimumReleaseAgeExclude` added since it is our own package and the release-age quarantine only delays dogfooding fresh releases.
- Tests keep building inputs with native `node:path` and pin the new output contract by wrapping expectations in `normalizePathSeparators` (5 expectations in `tests/deploy.test.ts` / `tests/typegen.test.ts` — `findInNodeModules` and the typegen `outPath` now return forward-slash on Windows). Keeping tests on `node:path` is deliberate: on a Windows machine the inputs stay backslash, so a source regression that stops normalizing still fails.

## Testing

- `vp check` green (format + lint + types, 1063 files).
- `tests/{check,typegen,init,deploy}.test.ts`: 522/522 pass on Windows.
- Full unit suite compared against a clean `HEAD` baseline worktree on the same Windows machine: identical failure sets (61 pre-existing Windows long-tail failures in unmigrated areas — routing/app-route-graph/scss/build — with a ±4 flaky churn between runs that also flips within the same tree). Zero new failures attributable to this change.
- Adversarial multi-agent review over the diff (call-site equivalence, downstream consumers, test masking, Windows edge cases): no semantic issues; two minor **pre-existing** coverage gaps noted for follow-up — CI runs vitest on ubuntu only, and `cli.ts` root normalization has no direct test (extending the windows-latest smoke job with `vinext check` would close both cheaply).

### DIFF
--- a/packages/vinext/package.json
+++ b/packages/vinext/package.json
@@ -141,6 +141,7 @@
     "image-size": "catalog:",
     "ipaddr.js": "catalog:",
     "magic-string": "catalog:",
+    "pathslash": "catalog:",
     "vite-plugin-commonjs": "catalog:",
     "web-vitals": "catalog:"
   },
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -4,6 +4,8 @@ preferFrozenLockfile: true
 saveExact: true
 blockExoticSubdeps: true
 minimumReleaseAge: 1440
+minimumReleaseAgeExclude:
+  - pathslash
 
 packages:
   - apps/*
@@ -73,6 +75,7 @@ catalog:
   next-view-transitions: ^0.3.5
   nitro: npm:nitro-nightly@latest
   nuqs: ^2.8.8
+  pathslash: ^0.1.0
   playwright: ^1.60.0
   postcss: 8.5.10
   react: ^19.2.7
--- a/packages/vinext/src/check.ts
+++ b/packages/vinext/src/check.ts
@@ -6,10 +6,9 @@
  */
 
 import { detectPackageManager, findDir } from "./utils/project.js";
-import { normalizePathSeparators } from "./utils/path.js";
 import { parseAst, type ESTree } from "vite";
 import fs from "node:fs";
-import path from "node:path";
+import path from "pathslash";
 
 // ── Support status definitions ─────────────────────────────────────────────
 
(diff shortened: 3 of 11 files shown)


packages/vinext/package.json
```diff
@@ -141,6 +141,7 @@
     "image-size": "catalog:",
     "ipaddr.js": "catalog:",
     "magic-string": "catalog:",
+    "pathslash": "catalog:",
```
james-elicx, line 144: can this be moved to devdeps and bundled by any chance? trying to avoid adding more deps that end-users have to download

pnpm-workspace.yaml
```diff
@@ -4,6 +4,8 @@ preferFrozenLockfile: true
 saveExact: true
 blockExoticSubdeps: true
 minimumReleaseAge: 1440
+minimumReleaseAgeExclude:
+  - pathslash
```
james-elicx, line 8:
```suggestion
```

shulaoda:
See #2502

PR #2486, opened 2026-07-02
feat(build): require Vite 8

james-elicx:
## Summary
- raise vinext's Vite peer requirement to Vite 8 and remove the optional vite-tsconfig-paths fallback
- collapse Vite-version branches to Vite 8/Rolldown config shapes
- update tests and docs to use rolldownOptions and assert Vite 7 is rejected

## Validation
- vp install
- vp test run tests/build-optimization.test.ts tests/tsconfig-paths-vite8.test.ts tests/optimize-deps-jsx-in-js.test.ts
- vp test run tests/pages-router.test.ts -t "MODULE_LEVEL_DIRECTIVE|preserves user-supplied"
- vp check $(git diff --name-only | tr '\n' ' ')

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ Use `--env <name>` to target `wrangler.jsonc` `env.<name>`. `--preview` is short
 The init command also auto-detects and fixes common migration issues:
 
 - Adds `"type": "module"` to package.json if missing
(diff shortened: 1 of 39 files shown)


TheAlexLichter:
(updated the title to include a `!` to signal the breaking change)

PR #2487, opened 2026-07-02
fix(app-router): preserve semicolons in redirect digests

NathanDrake2406:
## Overview

| Field | Details |
| --- | --- |
| Goal | Preserve raw redirect URLs when parsing `NEXT_REDIRECT` digests. |
| Core change | Parse semicolon-containing redirect targets through a tiny dependency-free redirect digest utility. |
| Boundary | Only `303`, `307`, and `308` in the final digest segment are treated as redirect status. Other semicolon-delimited content stays in the URL. |
| Primary files | `packages/vinext/src/utils/redirect-digest.ts`, `packages/vinext/src/server/next-error-digest.ts`, `packages/vinext/src/shims/navigation-errors.ts` |
| Expected impact | Redirects whose target URL contains semicolons keep the original destination instead of truncating or misclassifying part of the URL. |

## Why

**Intentional compatibility change:** empty redirect destinations now round-trip as redirects and produce the same self-redirect behavior as Next.js. Previously vinext treated those digests as malformed and surfaced a regular error.

Redirect digest parsing has to separate control metadata from the redirect target without assuming semicolons only appear as digest separators. Vinext formats redirect digests with raw URLs, so the URL portion can contain semicolons. This PR centralizes that parsing in a small neutral utility so the server and navigation shim share behavior without making hot server code import the public shim graph.

| --- | --- | --- |
| Redirect URL parsing | Raw redirect URLs are data, even when they contain semicolons. | Joins the URL portion until a valid final status segment is found. |
| Status parsing | A redirect status must be one of the supported redirect statuses. | Recognizes only final `303`, `307`, or `308`; omitted status still defaults to `307`. |
| Startup graph | Shared parsing should not cross from server code into public navigation shims. | Both server and shim import `utils/redirect-digest.ts`; server does not import `vinext/shims/navigation-errors`. |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| URL contains semicolons | The parser could treat semicolon-delimited URL content as digest structure. | Semicolon content remains part of the decoded URL unless the final segment is a valid redirect status. |
| Status omitted | Defaulted to `307`. | Still defaults to `307`. |
| Malformed status-like suffix | `parseInt()` could accept prefixes like `307garbage`. | The suffix stays URL content unless it exactly matches `303`, `307`, or `308`. |
| Empty destination | Fell through as a malformed redirect error. | Produces a self-redirect, matching Next.js empty-destination behavior. |
| Parser ownership | Server and shim parsing could drift, or server could import the shim. | Both use the same dependency-free utility. |

<details>

| File | Review focus |
| --- | --- |
| `packages/vinext/src/utils/redirect-digest.ts` | Check the shared parser boundary for raw URLs, final status parsing, and malformed status-like suffixes. |
| `packages/vinext/src/server/next-error-digest.ts` | Confirm server-side digest parsing delegates to the utility without importing the navigation shim. |
| `packages/vinext/src/shims/navigation-errors.ts` | Confirm navigation error decoding delegates to the same utility and keeps push / replace behavior. |
| `tests/next-error-digest.test.ts` | Review coverage for semicolon URLs, omitted status, strict status parsing, and unsupported digests. |
| `tests/error-boundary.test.ts` | Confirm redirect boundary behavior still receives the decoded target and navigation type. |

</details>

<details>
<summary>Validation</summary>

- `vp test run tests/next-error-digest.test.ts tests/error-boundary.test.ts`
- `vp check packages/vinext/src/utils/redirect-digest.ts packages/vinext/src/server/next-error-digest.ts packages/vinext/src/shims/navigation-errors.ts tests/error-boundary.test.ts tests/next-error-digest.test.ts`
- Pre-commit staged checks, full check, and knip passed during the final amend.

</details>

<details>

| Surface | Notes |
| --- | --- |
| Public API | No public API change. |
| Redirect behavior | Preserves default status `307` and push / replace decoding; empty destinations now round-trip as redirects, matching Next.js. |
| Compatibility | Tightens malformed final status parsing to exact supported redirect statuses. Valid digests remain supported. |
| Runtime impact | Parser-only change. Shared helper is dependency-free and does not import the navigation shim into server startup. |

</details>

<details>
<summary>Non-goals</summary>

- Does not change how redirect digests are formatted.
- Does not change redirect response shaping, headers, or browser navigation behavior, except that empty destinations now produce the Next.js-compatible self-redirect instead of falling through as malformed.
- Does not address RSC page probing or cache-state behavior.

</details>


NathanDrake2406 (commented):
Small follow-up nit: the server comment still describes the redirect shape as `NEXT_REDIRECT;<type>;<encodedUrl>;<status>`, but the shared parser now intentionally accepts both vinext’s encoded three-part form and Next’s raw five-part form. Might be worth updating that wording so future readers don’t “fix” this back to a single encoding model.

james-elicx:
Addressed the latest Big Bonk notes: added explicit parser invariant comments, documented the prefix-gate/validator distinction, and made the intentional empty-destination behavior change prominent in the PR overview (in addition to the risk and non-goals tables).

PR #2495, opened 2026-07-02
fix(pages): isolate on-demand revalidation requests

james-elicx:
## Summary

- route Pages Router on-demand revalidation through server-owned application origins
- preserve application routing behavior while preventing internal transport metadata from escaping its intended request
- reject nested revalidation and keep Node and Worker adapters aligned

## Validation

- focused revalidation and request-routing Vitest coverage (1,961 passed)
- Node production Playwright (12 passed)
- dev revalidation E2E (3 passed)
- pure Pages Worker E2E (3 passed)
- hybrid Worker E2E (1 passed)
- changed-file format, lint, type, and diff checks


james-elicx:
The current PR description understates the review surface. Here is a functional breakdown of what the PR now contains:

| Area | What changed |
|---|---|
| Revalidation transport | Pins Node loopbacks to a server-owned origin; uses in-process dispatch on Workers; authenticates internal requests; prevents secret/header leakage through redirects, middleware, or hostile paths; rejects nested/self revalidation. |
| Production ISR | Implements `unstable_onlyGenerated`, concurrent revalidation coalescing, `revalidateReason`, `revalidate: false`, `expireTime`, stale/expired handling, and correct replacement after failed regeneration. |
| Cached representations | Reworks how HTML, data envelopes, redirects, `notFound`, cache headers, deployment IDs, and custom cache-handler values are stored and served. |
| Pages rendering pipeline | Substantially rewrites and centralizes logic across `dev-server.ts`, `pages-page-data.ts`, and `pages-page-handler.ts` so development and production use compatible semantics. |
| Custom `_app` parity | Preserves optional, primitive, array, and custom `pageProps`; fixes hydration/navigation merging and custom error envelopes. This is separate from the original revalidation issue. |
| Development parity | Makes GSP run on every dev request, avoids persisting/coalescing dev ISR responses, recomputes redirects/`notFound`, and matches Next.js cache headers. |
| Shared cache behavior | The final commit changes explicit `no-store` propagation in the general cache runtime. This is also broader than the PR title. |

---

The simplest mental model is: this PR started by securing `res.revalidate()`, then expanded because exercising that path exposed gaps in how Pages Router ISR results were generated, stored, and replayed.

```mermaid
flowchart LR
    A["API route calls res.revalidate(path)"] --> B["Validate and pin target"]
    B --> C{"Runtime"}
    C -->|"Node"| D["HEAD request to server-owned origin"]
    C -->|"Worker"| E["In-process request dispatch"]
    D --> F["Authenticate internal request"]
    E --> F
    F --> G["Skip user middleware"]
    G --> H["Run getStaticProps with on-demand reason"]
    H --> I["Store content, redirect, or notFound"]
    I --> J["Return REVALIDATED"]
```

### 1. Revalidation can no longer escape the application origin

Core change.

Previously, a path such as a protocol-relative or backslash-shaped URL could be interpreted as having a different host. The internal request carries the revalidation secret, so that could send the credential somewhere unintended.

Now [pages-revalidate.ts](/Users/jamesanderson/.codex/worktrees/3a33/vinext/packages/vinext/src/server/pages-revalidate.ts:40):

- Requires a path beginning with `/`.
- Builds the destination by appending it to a trusted origin.
- Verifies the resulting origin is still identical.
- Rejects non-HTTP(S) origins.
- Uses the server’s actual bound origin instead of the inbound `Host` header.

Reason: `Host` and the path are request-controlled; neither should choose the destination of a credential-bearing loopback.

Observable effect: `res.revalidate("//other-host/path")`, backslash variants, and spoofed `Host` headers remain on the vinext server or are rejected.

### 2. Workers revalidate in-process instead of using global `fetch`

Core change.

Node still makes an HTTP loopback to its own server. Workers now re-enter the vinext request handler directly using [worker-revalidation-context.ts](/Users/jamesanderson/.codex/worktrees/3a33/vinext/packages/vinext/src/server/worker-revalidation-context.ts:34).

Reason: a same-origin global `fetch()` from a Worker can bypass the Worker and reach the zone origin. Besides being unreliable, that could expose the revalidation credential outside the isolate.

Observable effect: Worker `res.revalidate()` no longer depends on DNS, public routing, or the zone origin. It executes through the same vinext pipeline inside the Worker.

### 3. Internal revalidation requests are authenticated consistently

Core change.

The sender attaches `x-prerender-revalidate`; the receiver treats the request as internal only when its value exactly matches the revalidation secret.

The secret is:

- Generated at build time and shared by every production isolate.
- Generated once per process in development, even though Vite has separate module graphs.
- Compared using a constant-work equality function.
- Never trusted based merely on header presence.

Reason: a forged header must not bypass middleware or trigger privileged cache replacement. Multiple Worker isolates also need to agree on the same credential.

Observable effect: adding a fake `x-prerender-revalidate` header to a public request does nothing special.

### 4. Internal metadata is kept away from user code and external destinations

Core change.

Authenticated revalidation skips middleware in [pages-request-pipeline.ts](/Users/jamesanderson/.codex/worktrees/3a33/vinext/packages/vinext/src/server/pages-request-pipeline.ts:269). Internal protocol headers are stripped from ordinary inbound requests, user code, and external rewrite/proxy requests.

Reason:

- Next.js skips middleware during on-demand revalidation.
- Middleware could otherwise observe the credential or rewrite the request to an external URL.
- Internal transport state must not become application state.

Config redirects are not followed automatically. A redirect returned by `getStaticProps` counts as successful because it carries `REVALIDATED`; an ordinary config redirect does not.

### 5. Production header forwarding is explicit

Core change.

Production only forwards request headers named in:

```js
experimental: {
  allowedRevalidateHeaderKeys: ["authorization"]
}
```

Development automatically forwards cookies, matching Next.js’s local behavior.

Reason: authenticated pages sometimes need selected request context, but blindly forwarding cookies or authorization headers would expand the credential boundary.

Observable effect: production credentials remain excluded unless the application opts in.

### 6. Domain-locale identity survives the trusted Node loopback

Core change.

Pinning the HTTP destination to `localhost` or the bound server origin loses the original logical hostname used for domain-based i18n.

The PR adds a private logical-host header that:

- Is only sent by the Node loopback.
- Is accepted only on an authenticated revalidation request.
- Must exactly match a configured i18n domain.
- Is removed before user code sees it.
- Is not used by the Worker in-process path.

Reason: `/about` on `example.fr` and `/about` on `example.com` may be different cache entries even though both loop back through the same local server.

### 7. Nested and self-targeting revalidation is bounded

Core change.

A route already executing as an internal revalidation cannot call `res.revalidate()` again. Self-targeting loops are also rejected.

Reason: without this boundary, an API or page could recursively revalidate itself until the server or Worker exhausts resources.

Concurrent unrelated requests retain independent request context, so one internal request does not contaminate another.

### 8. `res.revalidate()` now performs synchronous cache replacement

Direct parity work.

An authenticated on-demand request:

- Bypasses fresh and stale cache-hit shortcuts.
- Runs `getStaticProps` immediately.
- Passes `revalidateReason: "on-demand"`.
- Stores the result before `res.revalidate()` resolves.
- Returns `x-nextjs-cache: REVALIDATED`.

Reason: callers expect that after `await res.revalidate(path)`, the cache contains the newly generated representation.

Previously, some paths could serve an existing entry, schedule background regeneration, or report success without synchronously replacing it.

### 9. `unstable_onlyGenerated` matches Next.js

Direct parity work.

For:

```ts
await res.revalidate("/posts/new", {
  unstable_onlyGenerated: true,
});
```

vinext now:

- Regenerates the route if an entry already exists.
- Treats an unseen route as a successful no-op.
- Does not generate a new `fallback: "blocking"` route.
- Returns a `404 + REVALIDATED` internal result for the no-op.

Without `unstable_onlyGenerated`, an invalid/unlisted `fallback: false` route is a genuine failed revalidation.

Reason: this option is intended to refresh existing output without accidentally expanding the generated route set.

### 10. Concurrent production revalidations are coalesced

Direct parity work.

Multiple simultaneous `res.revalidate("/same-path")` calls share one render through [isr-cache.ts](/Users/jamesanderson/.codex/worktrees/3a33/vinext/packages/vinext/src/server/isr-cache.ts:259).

This batch is separate from ordinary stale-background regeneration.

Reason:

- Avoid executing expensive `getStaticProps` several times for the same cache key.
- Avoid competing writes where an older result might overwrite a newer result.
- Preserve ordinary traffic’s ability to read the existing representation.

The entry is removed from the in-flight map in `finally`, so a failed attempt does not permanently block future revalidation.

### 11. ISR cache lifetimes now distinguish fresh, stale, expired, and indefinite

Direct parity work.

The PR formalizes four states:

- Fresh: serve immediately.
- Stale: serve the old value and regenerate in the background.
- Expired: do not serve it, but retain it as input for blocking regeneration.
- `revalidate: false`: cache indefinitely.

A one-year `s-maxage` is only an HTTP formatting convention for `revalidate: false`; it is not stored as a real regeneration deadline.

Reason: the previous representation blurred HTTP cache headers and the actual ISR lifetime. That could regenerate nominally permanent pages, serve hard-expired entries, or apply `expireTime` where it should not.

### 12. Content, redirects, and `notFound` are stored as distinct canonical results

Direct parity work and the largest source change, centered in [pages-page-data.ts](/Users/jamesanderson/.codex/worktrees/3a33/vinext/packages/vinext/src/server/pages-page-data.ts:1248).

The cache now distinguishes:

- `PAGES`: rendered HTML plus the complete data-props envelope, status, and safe headers.
- `REDIRECT`: redirect markers, status, destination, and `basePath` behavior.
- `null`: canonical `notFound`.

Reason: these are not interchangeable HTML responses. A client data request, hard navigation, and stale regeneration need different output forms from the same cached logical result.

This fixes transitions such as:

- Content → `notFound`
- `notFound` → content
- Content → redirect
- Redirect → content
- Redirect/data regeneration followed by an HTML request

### 13. HTML and `/_next/data` responses remain consistent

Direct parity work.

For a cached page:

- HTML requests receive the cached HTML representation.
- Data requests receive JSON from the cached full props envelope.
- `notFound` data responses return `{"notFound":true}`.
- Redirect data responses return `__N_REDIRECT` markers rather than following the redirect and returning HTML.
- Data responses include `x-nextjs-deployment-id` when configured.

Reason: client navigation expects JSON. Returning HTML or an incomplete `pageProps` object causes hydration errors, hard reloads, or stale deployment behavior.

### 14. Redirect validation and replay now match Pages Router rules

Direct parity work.

Redirect results now validate:

- `permanent` versus `statusCode`.
- Allowed status codes.
- String destinations.
- `basePath` type and behavior.
- Local path normalization.
- Dangerous executable schemes.

Cached redirects preserve `permanent` and `basePath` metadata and add the `Refresh` header for 308 responses.

Reason: revalidation must reproduce the same redirect that a fresh request would have produced. Invalid redirect data must fail without replacing a valid cached entry.

### 15. Custom 404 and terminal results are handled specially

Follow-on parity work.

When `getStaticProps` returns `notFound`:

- The source route stores a canonical not-found marker.
- The custom `/404` page is rendered per request.
- Request-specific 404 props, cookies, or CSP nonces are not replayed from shared cache.
- The source route’s lifetime controls the outgoing 404, not the `/404` page’s own cache identity.
- `/404` itself is forbidden from returning `notFound`, matching Next.js.

Reason: caching the fully rendered custom 404 under the source route can leak request-derived data and mix two different cache lifetimes.

### 16. Custom `_app.getInitialProps` retains its complete props envelope

Separate follow-on parity work.

Next.js treats the object returned by `_app.getInitialProps` as the full props passed to the custom App. It is not necessarily `{ pageProps: object }`.

The PR now preserves:

- An intentionally missing `pageProps` key.
- App-level properties beside `pageProps`.
- Primitive, array, or unusual enumerable `pageProps`.
- The same envelope during initial render, hydration, navigation, error rendering, and stale regeneration.

Page data is merged using `Object.assign`, matching Next.js’s observable behavior.

The client no longer forcibly adds `pageProps` when creating the custom App in [pages-client-entry.ts](/Users/jamesanderson/.codex/worktrees/3a33/vinext/packages/vinext/src/entries/pages-client-entry.ts:238).

Reason: forcing `pageProps` or spreading only plain objects silently changes what custom Apps receive and can cause server/client hydration disagreement.

### 17. Development mode deliberately does not use the Pages ISR response cache

Separate but related parity work.

In development, [dev-server.ts](/Users/jamesanderson/.codex/worktrees/3a33/vinext/packages/vinext/src/server/dev-server.ts:1259) now:

- Runs `getStaticProps` on every HTML and data request.
- Does not persist Pages route responses.
- Does not serve cached HIT/STALE representations.
- Does not coalesce concurrent on-demand revalidations.
- Ignores custom Pages CacheHandler response values.
- Recomputes redirects and `notFound` every time.
- Uses `revalidateReason: "stale"` for ordinary requests and `"on-demand"` for `res.revalidate()`.
- Returns Next.js-style development cache headers.

Reason: Next.js development mode exercises user code every request. The old vinext behavior incorrectly behaved like production ISR and could hide code changes or return stale output during development.

`unstable_onlyGenerated` becomes a successful no-op in development because there is intentionally no generated-response cache.

### 18. Shared cache-handler semantics changed too

Separate cross-cutting change.

[cache-handler.ts](/Users/jamesanderson/.codex/worktrees/3a33/vinext/packages/vinext/src/shims/cache-handler.ts:249) now:

- Supports `revalidate: false` as actual metadata.
- Reports hard-expired entries as `expired` instead of deleting them immediately.
- Ensures an explicit request context `revalidate: 0` still prevents storage.
- Prevents a nested indefinite cache value from incorrectly overriding a finite parent cache lifetime.

Reason: Pages ISR needs to represent permanent entries and blocking expiration correctly, but this shared handler also affects App Router/cache-component behavior. This is the most cross-cutting part of the PR.

### What is not actually part of the PR’s net behavior

The latest merge commit contains two newer `main` commits concerning nested App Router loading streaming and `after()` response-close handling. Those appear in the branch history, but they are now in the base branch too, so they are not part of #2495’s net diff.

Overall, I would describe the PR as:

> Secure Pages `res.revalidate()` transport, then make production and development Pages ISR representations behave consistently enough for that transport to be correct.

The custom `_app` envelope and shared cache-handler changes are genuine additional behavioral work beyond that central goal.

PR #2506, opened 2026-07-03
perf(cloudflare): serve static App Router HTML and RSC from Cloudflare Assets, bypassing the Worker

NathanDrake2406:
## Overview

| Field | Details |
| --- | --- |
| Goal | Let fully static App Router document and plain RSC requests avoid the Worker App Router render lifecycle on Cloudflare. |
| Core change | Publish eligible HTML at visible asset paths, publish proven-static RSC under a reserved static transport path, and route variant RSC through a Worker-only transport fallback. |
| Boundary | Cloudflare Workers builds only, gated by assets config and prerender query-invariance proof. Middleware, config transforms, ISR, fallback shells, basePath, i18n, and trailingSlash stay Worker-owned. |
| Primary files | `packages/vinext/src/build/cloudflare-prerender-assets.ts`, `packages/vinext/src/build/cloudflare-static-assets-config.ts`, `packages/vinext/src/server/app-rsc-transport.ts`, `packages/vinext/src/server/app-rsc-cache-busting.ts`, `packages/vinext/src/server/app-router-entry.ts` |
| Expected impact | Static document requests use Cloudflare Assets before Worker invocation. Plain static RSC requests use a reserved static asset path. Variant RSC keeps existing Worker semantics. |

## Follow-up review fixes

Since the initial description, these correctness fixes were added:

| Area | Issue | Fix |
| --- | --- | --- |
| `_headers` rule limit | HTML routes emit one `_headers` rule each; a large static app could exceed Cloudflare's 100-rule limit and produce an invalid asset bundle. | Publisher plans first, budgets preserved + generated rules, and skips publication entirely (no files written) when it would overflow. |
| Deploy env | `runPrerender()` ran outside `withCloudflareEnv`, so the publisher could read an undefined `CLOUDFLARE_ENV` and disagree with the build's selected environment. | Wrapped `runPrerender()` in `withCloudflareEnv(buildEnv)`. |
| Idempotency | A repeated prerender (e.g. `deploy --skip-build`) saw its own output as a collision, planned nothing, and dropped the generated `_headers` block — leaving assets served without headers. | The generated block is the ownership manifest: owned targets are re-published, foreign targets still skipped. |
| Stale bodies | Owned assets kept their headers but not their contents when the prerender output changed. | `publishAsset` overwrites owned/fresh targets with the current source via temp-file + rename (atomic). |
| Token canonicality | Forgiving base64 accepted many spellings of one route token (padding bits, `=` padding, `+`/`/` alphabet), and invalid UTF-8 collapsed distinct byte sequences into replacement-character routes. | Decoder accepts only the canonical base64url spelling (re-encode must round-trip) and decodes with `TextDecoder("utf-8", { fatal: true })`. |
| URL-shaped tokens | A decoded token containing `?`/`#` smuggled query/fragment syntax when concatenated into the mapped URL. | `resolveRscTransportRequest` assigns `mappedUrl.pathname` instead of string-concatenating, and decoded tokens must pass pathname validation. |
| URL rewrite bypass | The validator still accepted characters the URL implementation rewrites on `url.pathname` assignment: `/foo\bar` → `/foo/bar`, `/\evil.com` → `//evil.com`, newlines stripped, `/foo/../bar` → `/bar` — breaking the bijective route-token invariant. | A transport-local `isUrlSerializedPathname` requires an exact URL serialization round-trip (`new URL(value, base).pathname === value`), rejecting anything the pathname setter would rewrite. The shared interception matched-url validator keeps its loose semantics — interception proofs legitimately carry decoded pathnames. |
| Server action POST collision | Server actions POST to the visible page path (fetch client and no-JS form fallback). Cloudflare's asset worker claims a published asset path for every method — `getIntent` runs before the method check, so `POST /about` against a published HTML asset returns 405 from the asset worker and never invokes the Worker (confirmed in the `workers-shared` asset-worker source). | The rsc build records `hasServerActions` in `vinext-server.json`; the publisher skips visible HTML publication when actions exist (or the flag is absent), keeps the GET-only reserved static RSC transport, and prunes previously published HTML on a repeat prerender. |
| Hybrid build flag loss | Hybrid CLI builds run the Pages SSR bundle as a second builder with a separate `vinext()` closure; its ssr env wrote `vinext-server.json` last and dropped `hasServerActions` (secret-keyed preservation never matched across builders), so the publisher skipped visible HTML for every hybrid CLI build. | The ssr environment preserves an existing flag unconditionally (Pages Router has no server actions of its own). Caught by the local benchmark's sanity gate; regression-tested by driving the real plugin's `writeBundle`. |
| Ancestor/descendant path conflict | `/blog` needs `dist/client/blog` as a file while `/blog/post` needs it as a directory; the plan phase checked targets independently, so publishing both aborted the prerender (ENOTDIR or EISDIR depending on route order). | Planning demotes a route whose HTML target is the directory prefix of another planned HTML target (Worker keeps serving it; flat tokenised RSC transport unaffected), a directory at the target is no longer a foreign-file collision, pruning removes emptied directories so ancestors can publish again, and `publishAsset` degrades to "not published" on directory-shape conflicts instead of aborting. |
| Foreign parent file | A pre-existing user asset file (`dist/client/blog`) blocks the parent directory a descendant target needs (`dist/client/blog/post`); the exact-target foreign check missed it, `mkdirSync` threw ENOTDIR, and the catch block's temp-file `rmSync` threw ENOTDIR itself before the fallback could run — aborting prerender. | `publishAsset` temp-file cleanup is best-effort and EEXIST (the Windows mkdir error for this shape) joins the safe "not published" codes; the route degrades to Worker rendering with transport RSC still published and the user asset untouched. |
| `output: "export"` transport mismatch | The client transport define only checked "Cloudflare plugin + assets config allows fallback". Export builds emit plain `about.rsc` next to the HTML and the publisher skips transport publication, so the browser requested reserved transport assets that were never emitted. | The define additionally requires `nextConfig.output !== "export"`; export clients keep requesting RSC at public route URLs. Test proves the gate flips on the same root. |

Also deduped: shared `utils/jsonc` (JSONC parsing) and `resolveWranglerJsonPath`, previously copied across `init-cloudflare` and the assets-config reader.

## Why

Cloudflare Assets matches by path before the Worker can inspect RSC headers. A public HTML asset at `/about` is therefore incompatible with browser RSC requests that also use `/about?_rsc=...`. Correct static publication needs three distinct resources: the public HTML document, the static RSC artifact, and dynamic or variant RSC requests. This PR splits those transports instead of trying to make one public pathname represent all three.

| --- | --- | --- |
| Document HTML | Fully static, request-invariant HTML should not pay App Router render lifecycle cost. | Copies eligible HTML to visible Cloudflare asset paths such as `/about`. |
| Browser RSC | Browser RSC requests must not be shadowed by visible HTML assets. | Changes Cloudflare client RSC URLs to a reserved static namespace for plain proven-static RSC. |
| Variant RSC | Prefetch, mounted slots, interception, state-tree, refresh, and reuse-manifest variants are not the same payload. | Sends variant RSC to a reserved Worker namespace that maps back to the visible route before dispatch. |
| Safety gates | Request-observable behavior must stay Worker-owned. | Skips static publication for middleware/proxy, config transforms, basePath, i18n, trailingSlash, ISR, fallback, `/404`, and `/500`.Visible HTML additionally requires the build to contain no server actions. |
| Cache proof | Query-bearing requests are only safe when prerender proved search params were not observed. | Requires `queryInvariant.html === true` for HTML and `queryInvariant.rsc === true` for static RSC. |
| Cloudflare config | Missing transport assets must fall through to the Worker. | Enables the browser transport only for Cloudflare builds whose wrangler assets config is absent or uses `not_found_handling: "none"`.Disabled entirely for `output: "export"` builds, which never emit transport assets. |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| Static document `/about` | Worker entered App Router dispatch/render/cache lifecycle. | Cloudflare Assets can serve `dist/client/about` directly. |
| Plain RSC navigation for `/about` | Browser requested `/about?_rsc=...`, which collides with a visible HTML asset. | Browser requests `/_next/static/__vinext/prerendered-rsc/about.rsc?_rsc`. |
| Variant RSC request | Same public route URL with variant headers. | Browser requests `/__vinext/rsc/about.rsc?_rsc=...`; Worker maps it back to `/about` before dispatch. |
| Stale transport `_rsc` hash | Could redirect back to the public route URL. | Canonical redirect remains on the transport path. |
| Existing asset collision | Existing target stayed untouched. | Route publication is skipped when the visible HTML target or reserved RSC target already exists. |

## Benchmark

Local paired Workers benchmark (script kept out of the PR diff; two concurrent `wrangler dev` servers kept alive, samples alternate AB/BA per round, sanity-gated: head must serve `/about` with `x-vinext-cache: STATIC`, base must not):

- Fixture: `tests/fixtures/cf-app-basic`, copied to a temp dir with `middleware.ts` removed so middleware cannot observe requests.
- Build: `node packages/vinext/dist/cli.js build --prerender-all` (the real CLI path).
- `head`: generated build with published visible HTML and reserved static RSC assets.
- `base`: byte-identical copy with the published route assets stripped (using the generated `_headers` block as the manifest), so requests fall through to the Worker.
- Method: 8 paired AB/BA rounds, 250 warmup requests per sample excluded, 1600 measured requests per sample, concurrency 8, Node `fetch()` latency including body consumption.
- Latest run (branch head, 2026-07-04): `benchmarks/results/static-artifact-worker-bypass/2026-07-04T08-43-32Z/summary.md` in the local ignored benchmark output.

| scenario | base mean ms | head mean ms | diff ms | diff % |
| --- | ---: | ---: | ---: | ---: |
| document-about | 8.1366 | 4.6565 | -3.4801 | -42.77% |
| rsc-about-static-transport | 8.4376 | 4.2110 | -4.2266 | -50.09% |
| document-blog | 8.3508 | 4.5483 | -3.8025 | -45.53% |
| rsc-blog-static-transport | 9.0687 | 4.2572 | -4.8116 | -53.06% |
| route-api-control | 4.3568 | 4.3278 | -0.0290 | -0.67% |

The control route is effectively neutral. The useful signal is that static document and static RSC requests leave the App Router lifecycle rather than shaving a small allocation inside it. (An earlier ad-hoc run on the pre-review head showed the same shape: -51%/-36%/-50%/-32%, control +0.15%.)

## Maintainer review path

<details>
<summary>Suggested order</summary>

1. `packages/vinext/src/server/app-rsc-transport.ts`
   - Review transport path encoding, plain RSC eligibility, and reverse mapping.
2. `packages/vinext/src/server/app-rsc-cache-busting.ts`
   - Review the Cloudflare-only switch from public route `_rsc` URLs to split transport URLs.
3. `packages/vinext/src/server/app-router-entry.ts`
   - Review Worker transport mapping and stale `_rsc` redirect handling before normal App Router dispatch.
4. `packages/vinext/src/build/cloudflare-prerender-assets.ts`
   - Review publication gates, visible HTML publication, reserved static RSC paths, and `_headers` output.
5. `packages/vinext/src/index.ts`
   - Review the client define gate for Cloudflare transport enablement.
6. `tests/app-rsc-cache-busting.test.ts` and `tests/cloudflare-prerender-assets.test.ts`
   - Review the transport URL, redirect, publication, and collision regressions.

</details>

## Validation

<details>
<summary>Commands and probes</summary>

- `vp fmt packages/vinext/src/build/cloudflare-prerender-assets.ts packages/vinext/src/build/run-prerender.ts packages/vinext/src/index.ts packages/vinext/src/server/app-router-entry.ts packages/vinext/src/server/app-rsc-cache-busting.ts packages/vinext/src/server/app-rsc-transport.ts tests/app-rsc-cache-busting.test.ts tests/cloudflare-prerender-assets.test.ts`
- `vp check packages/vinext/src/build/cloudflare-prerender-assets.ts packages/vinext/src/build/run-prerender.ts packages/vinext/src/index.ts packages/vinext/src/server/app-router-entry.ts packages/vinext/src/server/app-rsc-cache-busting.ts packages/vinext/src/server/app-rsc-transport.ts tests/app-rsc-cache-busting.test.ts tests/cloudflare-prerender-assets.test.ts`
- `vp test run tests/app-rsc-cache-busting.test.ts tests/cloudflare-prerender-assets.test.ts`
- `vp test run tests/cloudflare-rsc-transport-config.test.ts tests/deploy-prerender-config.test.ts`
- `vp test run tests/prerender.test.ts -t "Cloudflare Workers hybrid build"`
- Server-action gating tests: publisher keeps transport RSC but skips visible HTML when `hasServerActions` is set, repeat prerender prunes previously published HTML, and the hybrid Cloudflare build records the flag in `vinext-server.json`.
- Follow-up regression tests: `_headers` rule-limit skip, idempotent re-run preserves the generated block, owned-asset overwrite when prerender output changes, and deploy sets `CLOUDFLARE_ENV` during prerender.
- `vp test run tests/app-rsc-transport.test.ts` — token round-trip/uniqueness plus rejection of non-canonical base64url, invalid UTF-8, unrooted pathnames, query/fragment syntax, and URL-rewritable tokens (backslash, protocol-relative, newline, dot segments).
- `vp run vinext#build`
- Fresh no-middleware `tests/fixtures/cf-app-basic` build with `node packages/vinext/dist/cli.js build --prerender-all`
- Local `wrangler dev` probes verified:
  - `/about` returns static HTML with `x-vinext-cache: STATIC`.
  - `/_next/static/__vinext/prerendered-rsc/about.rsc?_rsc` returns `text/x-component` with static cache headers.
  - `/__vinext/rsc/about.rsc?_rsc=stale` redirects to `/__vinext/rsc/about.rsc?_rsc=<hash>`, then returns `text/x-component` through the Worker path.
- Pre-commit hook passed staged checks, full check, staged unit/integration tests, and Knip.
- Local paired benchmark described above.

</details>

## Risk / compatibility

<details>
<summary>Compatibility notes</summary>

- Public API: no API changes.
- Runtime scope: Cloudflare App Router entry maps reserved Worker RSC transport requests back to visible routes before dispatch.
- Client scope: Cloudflare builds use split RSC URLs only when the wrangler asset config can fall through to Worker on misses.
- Build output: eligible HTML is published at visible paths; eligible RSC is published under `/_next/static/__vinext/prerendered-rsc`.
- Existing apps with middleware/proxy, config headers/redirects/rewrites, basePath, i18n, trailingSlash, ISR, fallback shell, `/404`, or `/500` stay Worker-owned.
- The main review risk is whether another RSC variant header should force Worker transport. Current gating covers prefetch, segment prefetch, state tree, next URL, reuse manifest, interception context, mounted slots, and non-navigation render modes.

</details>

## Non-goals

<details>
<summary>Out of scope</summary>

- No route-handler fast path.
- No server action optimization.
- No lazy cookies, headers, or request-store changes.
- No ISR asset publication.
- No middleware or config transform bypass.
- No cache lifecycle refactor.

</details>


NathanDrake2406:
@codex review

james-elicx:
If a file is SSG'd, it should end up in the client directory regardless of if it's cloudflare or not. If it's not SSG'd, it should go down the normal ISR route through stubbing the cache.

NathanDrake2406:
> If a file is SSG'd, it should end up in the client directory regardless of if it's cloudflare or not. If it's not SSG'd, it should go down the normal ISR route through stubbing the cache.

agree, will follow up for node

NathanDrake2406:
Follow-up TODOs from James' SSG/publication point, so this PR does not try to solve the general model in the Cloudflare-specific path:

- [ ] Keep this PR scoped to the Cloudflare Workers problem: Cloudflare Assets can claim the public pathname before the Worker can inspect RSC/action semantics, so this PR needs the split HTML/static-RSC/Worker-RSC transport.
- [ ] Follow up with a shared static App Router publication planner that decides which prerendered artifacts are safe to expose publicly, independent of platform.
- [ ] Move the common eligibility rules into that shared planner: rendered App route, non-ISR, non-fallback, not `/404`/`/500`, and separate `queryInvariant.html` / `queryInvariant.rsc` proof.
- [ ] Add a Node/static-file executor for the shared plan so truly SSG'd App artifacts can land in the client/static output where safe, while ISR/dynamic routes continue through the normal cache/runtime path.
- [ ] Keep platform-specific mechanics in executors rather than duplicating the decision model: Cloudflare owns Wrangler/assets config, `_headers`, ownership/pruning, server-action visible-HTML blocking, and the reserved RSC transport; Node should only materialize the safe static artifacts it can serve correctly.

PR #2513, opened 2026-07-03
fix(app-router): preserve client state on server-action revalidatePath under loading.tsx

### LINKED ISSUE #2505: App Router: server-action revalidatePath() remounts client components on segments with loading.tsx (loses useState)
## Problem

When a Server Action calls `revalidatePath()`, vinext refreshes the **current route by
rebuilding its subtree**, which unmounts and remounts the route's Client Components and
destroys their `useState`. Next.js performs a soft refresh that reconciles the existing
tree and preserves client state.

**The trigger is a `loading.tsx` on the current segment** (i.e. a Suspense boundary).
With `loading.tsx` present the client components remount; remove it and the exact same
flow reconciles correctly and state is preserved. It also fires **regardless of which
path is revalidated** — even `revalidatePath('/somewhere-else')` remounts the current
route.

This breaks any "mutate but stay on the page" flow (open a dialog after a mutation, keep
in-progress form/cart state, optimistic UI, etc.).

## Reproduction

Minimal App Router example — **three** files. `loading.tsx` is what makes it fire:

```tsx
// app/demo/loading.tsx  — the trigger: a Suspense boundary on this segment
export default function Loading() {
  return <p>loading…</p>
}
```

```tsx
// app/demo/page.tsx
'use client'
import { useState } from 'react'
import { touch } from './actions'

  const [n, setN] = useState(0)
  return (
    <>
      <button onClick={() => setN((v) => v + 1)}>count: {n}</button>
      <button onClick={() => touch()}>run action</button>
    </>
  )
}
```

```ts
// app/demo/actions.ts
'use server'
import { revalidatePath } from 'next/cache'
export async function touch() {
  // Remounts even when revalidating an UNRELATED path:
  revalidatePath('/somewhere-else')
}
```

Steps:
1. Open `/demo`, click **count** a few times → `count: 3`.
2. Click **run action** (the server action returns after `revalidatePath`).

**Expected (Next.js):** `count` stays `3` (client component reconciled).
**Actual (vinext):** `count` resets to `0` — the client component remounted.

### A/B that isolates the trigger

Delete `app/demo/loading.tsx` and repeat — `count` is preserved. Re-add it — it resets
again. Verified on this machine:

| Runner                                     | `loading.tsx` | count after action |
| ------------------------------------------ | ------------- | ------------------ |
| `vinext dev`                               | present       | 0 ❌ (remount)     |
| `vinext dev`                               | absent        | 3 ✅               |
| `vinext build` + `wrangler dev` (prod build) | present     | 0 ❌ (remount)     |

Rows 1 vs 2 pinpoint `loading.tsx` as the trigger; row 3 shows the production build has
the same defect (it's client-runtime navigation logic, not a dev-only artifact).

## Likely area

The browser server-action result handler treats an action's revalidation as a same-URL
navigation and rebuilds the route subtree rather than reconciling. In
`dist/server/app-browser-server-action-client.js`, on a non-`none` revalidation it calls
`clearClientNavigationCaches()` then `commitSameUrlNavigatePayload(...)`. Suspect
`commitSameUrlNavigatePayload` re-enters the segment through its `loading.tsx` Suspense
boundary (re-suspend → fresh mount) instead of reconciling the already-committed tree —
which is exactly why segments **without** a `loading.tsx` have no boundary to re-suspend
and reconcile correctly. Separately, revalidating a *foreign* path shouldn't trigger a
current-route refresh at all.

## Impact / workaround

- Breaks mutate-and-stay flows on any route that has a `loading.tsx`. Affects **dev and
  production builds** (both confirmed above).
- Workarounds (app side), neither a real fix: remove the segment's `loading.tsx`, or don't
  `revalidatePath` from actions where the user stays on the page (only safe when the
  affected routes are dynamic and re-fetch anyway).

## Environment

- vinext 0.2.0 and 0.2.1
- next 16.2.7, react / react-dom 19.2.7
- @vitejs/plugin-rsc 0.5.27, vite 8.1.3
- Cloudflare Workers runtime (workerd via @cloudflare/vite-plugin), dev and build

ruralad:
Fixes #2505

## Problem

A Server Action calling `revalidatePath()` (and `router.refresh()` flows delivered through the same-URL action path) remounted the current route's Client Components — destroying their `useState` — whenever the segment defined a `loading.tsx`. In-progress form input, open dialogs, and optimistic UI were lost on every "mutate but stay on the page" flow. Next.js soft-refreshes instead: the segment reconciles in place and client state survives.

Segments **without** a `loading.tsx` were unaffected, which made the trigger easy to isolate (see the A/B in #2505).

## Root cause

The "preserve-ui" re-render modes (`action-rerender-preserve-ui`, `refresh-preserve-ui`) omitted the loading `<Suspense>` boundary from the payload (`shouldSuppressLoadingBoundaries` in `app-page-route-wiring.tsx`) — but the already-committed navigation tree always contains that boundary. Dropping it on the re-render changed the element type at the page position, so React unmounted the boundary's subtree and mounted a fresh one. The content was already resolved, so no fallback ever flashed — the remount was silent, which is why the suppression looked harmless.

Notably it was *not* a re-suspension problem: the fallback never rendered, and wrapping the commit in a transition alone did not fix it. The bug was purely structural.

## Fix

1. **`app-page-route-wiring.tsx`** — keep the loading Suspense boundary in the tree for every render mode (route-level and parallel-slot-level). The preserve-ui re-render is now structurally identical to the committed tree, so React reconciles in place and client state survives.
2. **`app-browser-navigation-controller.ts`** — commit same-URL server-action revalidations inside `startTransition`, matching how ordinary navigations commit (`dispatchApprovedVisibleCommit` in "transition" mode). This is defense-in-depth, not the fix itself: with the boundary now always present, a future payload that genuinely suspends would drop to the fallback under an urgent update, while a transition keeps the committed tree visible until content resolves — Next.js's soft-refresh behavior. The HMR and history-traversal callers of `dispatchSynchronousVisibleCommit` are deliberately untouched.

The original purpose of the suppression (no loading flash on refresh) is preserved — the existing `action-refresh-no-rerender` tests still pass, and the new test also asserts the fallback never mounts.

## Test

New fixture `tests/fixtures/app-basic/app/nextjs-compat/action-revalidate-remount/` (a client `useState` counter under a route with `loading.tsx`, plus a `revalidatePath` server action) and a regression test in `actions-revalidate.spec.ts` asserting, after the action:

- the server **did** re-render (render counter changed) — revalidation still works
- the client counter still shows its pre-action value — no remount
- the `loading.tsx` fallback never mounted (armed-sentinel console check, same pattern as `action-refresh-no-rerender`)

## Verification

- `pnpm run check` — clean (format, lint, types)
- `npx playwright test app-router/nextjs-compat/actions-revalidate.spec.ts --project=app-router` — 7/7 passing, including the two pre-existing "does not mount route loading" guards most at risk from this change
- Verified end-to-end in a browser against the minimal repro from #2505: with the fix, count survives the action (`3 → 3`) with no remount and no fallback flash, in dev; the same repro reproduced the bug on both `vinext dev` and production builds before the fix

### DIFF
--- a/packages/vinext/src/server/app-browser-navigation-controller.ts
+++ b/packages/vinext/src/server/app-browser-navigation-controller.ts
@@ -906,10 +906,29 @@ export function createAppBrowserNavigationController(
       }
 
       if (latestApproval.approvedCommit) {
-        dispatchSynchronousVisibleCommit(latestApproval.approvedCommit);
+        const approvedRevalidationCommit = latestApproval.approvedCommit;
+        // Commit the revalidation inside a transition, matching how ordinary
+        // client navigations commit (`dispatchApprovedVisibleCommit` in
+        // "transition" mode). This is defense-in-depth, not the fix for the
+        // loading.tsx remount bug — that fix is structural, in
+        // app-page-route-wiring.tsx (the loading Suspense boundary now stays in
+        // the tree for preserve-ui re-renders, so React reconciles in place).
+        // The transition matters because this code runs after the action's
+        // `await fetch`, outside React's event batch, so a plain setter would
+        // be an urgent update: if a future payload delivered through this path
+        // genuinely suspends inside the (now always-present) loading.tsx
+        // boundary, an urgent update would drop the subtree to its fallback,
+        // while a transition keeps the committed tree visible until the
+        // content resolves — Next.js's soft-refresh behavior. The HMR and
+        // history-traversal callers of dispatchSynchronousVisibleCommit are
+        // deliberately left non-transition (see the comment above
+        // dispatchApprovedVisibleCommit).
+        startTransition(() => {
+          dispatchSynchronousVisibleCommit(approvedRevalidationCommit);
+        });
         syncHistoryStatePreviousNextUrl(
-          latestApproval.approvedCommit.previousNextUrl,
-          latestApproval.approvedCommit.action.bfcacheIds,
+          approvedRevalidationCommit.previousNextUrl,
+          approvedRevalidationCommit.action.bfcacheIds,
         );
       } else {
         notifyDiscardedServerActionRevalidation(lifecycleOptions);
--- a/packages/vinext/src/server/app-page-route-wiring.tsx
+++ b/packages/vinext/src/server/app-page-route-wiring.tsx
@@ -47,7 +47,6 @@ import { probeReactServerSubtree } from "./app-page-probe.js";
 import {
   APP_RSC_RENDER_MODE_NAVIGATION,
   APP_RSC_RENDER_MODE_PREFETCH_LOADING_SHELL,
-  shouldSuppressLoadingBoundaries,
   type AppRscRenderMode,
 } from "./app-rsc-render-mode.js";
 import {
@@ -903,7 +902,7 @@ export function buildAppPageElements<
     }
 
     const slotLoadingComponent = getDefaultExport(slot.loading);
-    if (slotLoadingComponent && !shouldSuppressLoadingBoundaries(renderMode)) {
+    if (slotLoadingComponent) {
       const SlotLoadingComponent = slotLoadingComponent;
       slotElement = (
         <Suspense key={slotResetKey} fallback={<SlotLoadingComponent />}>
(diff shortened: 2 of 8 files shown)


james-elicx:
Thanks for taking a crack at this. Taken a slightly different approach in #2517

PR #2519, opened 2026-07-04
perf(build): filter virtual module hooks

camc314:
Move several virtual-module resolveId/load hooks to Vite's object hook form with native id filters.

This lets Vite/Rolldown skip invoking vinext JavaScript handlers for unrelated module ids during dev and build module graph walks, reducing repeated string comparisons on the hot plugin hook path. The filtered hooks now only run for vinext virtual modules, instrumentation client injection, and the React canary shim ids they can actually handle.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -3930,7 +3930,8 @@ export const loadServerActionClient = ${
       },
 
       load: {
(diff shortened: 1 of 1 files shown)


camc314:
image

nicee i didn't expect that much of an improvement!

PR #2538, opened 2026-07-05
Fix route param percent-literal decoding

### LINKED ISSUE #1963: Routing: dynamic route params are decoded twice, corrupting percent-literal values (App + Pages Router)
### Problem
Every route param (single dynamic [id] and catch-all [...slug]) in both App Router (RSC matchRoute + matchAppRoute) and Pages Router (matchRoute) is decoded twice. Path-delimiter sequences (%2F/%23/%3F/%5C) survive because encodePathDelimiters re-encodes them, which is why the only existing test (app-rsc-route-matching.test.ts:78, /files/a%252Fb -> a%2Fb) passes. But every other double-encoded value is wrong vs Next.js single-decode: /blog/a%2520b yields id="a b" instead of "a%20b"; %2541 yields "A" instead of "%41"; caf%25C3%25A9 yields "café" instead of "caf%C3%A9". Any app whose param legitimately contains a literal percent sequence (encoded once on the client) receives a corrupted, over-decoded value in useParams(), getServerSideProps params, and route handlers.

### Evidence
- `packages/vinext/src/routing/utils.ts:119` — normalizePathnameForRouteMatch() fully decodes every URL segment via decodeRouteSegment -> decodeURIComponent (re-encoding only the path delimiters /,#,?,\ and their %2F/%23/%3F/%5C forms).
- `packages/vinext/src/routing/route-matching.ts:52` — matchRouteWithTrie() runs normalizePathnameForRouteMatch (first decode) then splits and calls trieMatch on the already-decoded parts.
- `packages/vinext/src/routing/route-trie.ts:145` — trieMatch() calls decodeMatchedParams(result.params), applying decodeURIComponent a SECOND time to captured param values.
- `packages/vinext/src/routing/route-pattern.ts:141` — matchRoutePattern() likewise calls decodeMatchedParams on params whose source urlParts callers (e.g. navigation-planner splitMatchedUrlIntoRouteParts) already decoded.
- `packages/vinext/src/server/app-rsc-route-matching.ts:110` — App RSC matchRoute does trieMatch(routeTrie, appRscPathnameParts(url)) where appRscPathnameParts->splitPathnameForRouteMatch decodes first, then trieMatch decodes again.

Next.js decodes route params exactly once. The request pathname is parsed via parseUrl (new URL().pathname / parseRelativeUrl), which preserves percent-encoding, and in next-server.ts handleCatchallRenderRequest it is passed to matchers.match still encoded (only removeTrailingSlash applied). getRouteMatcher then runs re.exec against that ENCODED pathname and applies decodeURIComponent a single time to each captured group (route-matcher.ts:27 for single [id], :38 per-segment for catch-all [...slug]). I simulated /blog/a%2520b, /blog/%2541, and /blog/caf%25C3%25A9 through this exact logic: Next.js yields a%20b, %41, caf%C3%A9 — the correct single-decode results the bug report expects — whereas vinext yields a b, A, café because it pre-decodes the whole pathname in normalizePathnameForRouteMatch and then decodes the captured param a second time in decodeMatchedParams. vinext's decodeMatchedParams step alone mirrors Next.js correctly; the defect is the extra normalizePathnameForRouteMatch/splitPathnameForRouteMatch pre-decode layered on top, which has no counterpart in the Next.js code path. This is therefore a vinext-only regression; Next.js handles the scenario correctly.

Citations: .nextjs-ref/packages/next/src/shared/lib/router/utils/route-matcher.ts:21-45 (re.exec(pathname) on encoded pathname; decode() applies decodeURIComponent exactly once per group — line 27 for single, line 38 for repeat/catch-all); .nextjs-ref/packages/next/src/server/next-server.ts:1074,1106,1111 (pathname taken from parsedUrl, only removeTrailingSlash applied, then passed still-encoded to matchers.match — no pre-decode); .nextjs-ref/packages/next/src/server/route-matchers/route-matcher.ts:28 + app-page-route-matcher.ts:1-8 (both Pages and App routers use the same single-decode getRouteMatcher);

### Suggested fix
The upstream normalize step already produces the canonical single-decoded value (with delimiters re-escaped); the params should not be decoded a second time. Remove the redundant decodeMatchedParams call from trieMatch (route-trie.ts:145) and matchRoutePattern (route-pattern.ts:141) — or, equivalently, stop fully decoding in normalizePathnameForRouteMatch and let decodeMatchedParams be the single decode. Add coverage for non-delimiter double-encoded values (%2520, %2541) since the current test only exercises the delimiter case that is accidentally protected.

### Test plan
Extend tests/app-rsc-route-matching.test.ts with `%2520`/`%2541` cases for `[id]` and `[...slug]` in both routers (current test only covers the accidentally-protected `%252F` delimiter case).

### Related / notes
Distinct from #1555 (unicode searchParams/useParams identity) and #1466 (Pages useParams empty). Fixing this also resolves the slot/route param decode asymmetry TODO at `app-optimistic-routing.ts:245`. Constraint: fix by removing the second decode (`decodeMatchedParams` in `trieMatch`/`matchRoutePattern`); do NOT change `normalizePathnameForRouteMatch` (the trailing-slash fix depends on its current contract).

---

WilliamK112:
﻿## Summary
- avoid a second full `decodeURIComponent` pass for params captured after `normalizePathnameForRouteMatch`
- keep protected path delimiters (`%2F`, `%23`, `%3F`, `%5C`) decoding correctly while preserving literal percent sequences such as `%20` and `%41`
- cover Pages Router, App Router, App RSC route matching, and optimistic route manifest matching

Closes #1963

## Test plan
- `corepack pnpm test tests/routing.test.ts tests/app-rsc-route-matching.test.ts tests/app-optimistic-routing.test.ts -t "percent-literal"` (failed before the fix with `a%2520b` becoming `a b`, passes after)
- `corepack pnpm test tests/app-rsc-route-matching.test.ts tests/app-optimistic-routing.test.ts tests/route-trie.test.ts tests/route-pattern.test.ts`
- `corepack pnpm test tests/routing.test.ts -t "percent-literal|encoded slashes|decoded URL-encoded routes|keeps encoded slashes"`
- `corepack pnpm run check`
- `git diff --check HEAD^..HEAD`

Local note: I also ran the broader related command including full `tests/routing.test.ts`; the decode-related tests passed, while the remaining failures on this Windows checkout were existing route-discovery/path separator or invalid Windows filename cases unrelated to this change.

### DIFF
--- a/packages/vinext/src/routing/route-matching.ts
+++ b/packages/vinext/src/routing/route-matching.ts
@@ -54,5 +54,5 @@ export function matchRouteWithTrie<R extends { patternParts: string[] }>(
   // Split URL once, look up via trie
   const urlParts = normalizedUrl.split("/").filter(Boolean);
   const trie = getOrBuildTrie(cache, routes);
-  return trieMatch(trie, urlParts);
+  return trieMatch(trie, urlParts, { decodeParams: "path-delimiters" });
 }
--- a/packages/vinext/src/routing/route-pattern.ts
+++ b/packages/vinext/src/routing/route-pattern.ts
@@ -1,6 +1,10 @@
-import { decodeMatchedParams } from "./utils";
+import { decodeMatchedParams, decodeMatchedPathDelimiterParams } from "./utils";
 
 export type RoutePatternParams = Record<string, string | string[]>;
+
+export type RoutePatternMatchOptions = {
+  decodeParams?: boolean | "path-delimiters";
+};
 
 function routePatternPart(segment: string): string {
   if (segment.startsWith("[[...") && segment.endsWith("]]")) {
@@ -91,6 +95,7 @@ export function fillRoutePatternSegments(
 export function matchRoutePattern(
   urlParts: readonly string[],
   patternParts: readonly string[],
+  options: RoutePatternMatchOptions = {},
 ): RoutePatternParams | null {
   const params: RoutePatternParams = Object.create(null);
 
@@ -138,7 +143,11 @@ export function matchRoutePattern(
   }
 
   if (!matchFrom(0, 0)) return null;
-  decodeMatchedParams(params);
+  if (options.decodeParams === "path-delimiters") {
+    decodeMatchedPathDelimiterParams(params);
+  } else if (options.decodeParams !== false) {
+    decodeMatchedParams(params);
+  }
   return params;
 }
 
--- a/packages/vinext/src/routing/route-trie.ts
+++ b/packages/vinext/src/routing/route-trie.ts
@@ -1,4 +1,4 @@
-import { buildParams, decodeMatchedParams } from "./utils";
+import { buildParams, decodeMatchedParams, decodeMatchedPathDelimiterParams } from "./utils";
 
 /**
  * Trie (prefix tree) for O(depth) route matching.
@@ -20,6 +20,10 @@ export type TrieNode<R> = {
   catchAllChild: { paramName: string; route: R } | null;
   optionalCatchAllChild: { paramName: string; route: R } | null;
   route: R | null;
+};
+
+export type TrieMatchOptions = {
+  decodeParams?: boolean | "path-delimiters";
 };
 
 function createNode<R>(): TrieNode<R> {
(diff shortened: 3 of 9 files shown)


WilliamK112:
Closing this as a duplicate of #2059, which already covers #1963 and predates this PR. I missed it during preflight because my initial gh search prs query did not return #2059; gh pr list --search 1963 does. Sorry for the noise.

PR #2543, opened 2026-07-06
fix(build): honor inline next config for static export

### LINKED ISSUE #2542: output: "export" requires next.config.ts; vinext({ nextConfig }) does not trigger static export
## Summary

`vinext` documents `vinext({ nextConfig })` as a way to configure projects from `vite.config.*` without a separate `next.config.*` file. However, `output: "export"` only appears to trigger static export when it is defined in `next.config.ts`.

When `output: "export"` is provided through `vinext({ nextConfig })`, `vinext build` completes but does not enter static export mode and does not emit prerendered HTML pages.

## Environment

```text
vinext: 1.0.0-beta.0
vite: 8.1.3
@cloudflare/vite-plugin: 1.43.0
wrangler: 4.107.0
runtime/package manager: bun 1.3.14
```

## Reproduction

Use a `vite.config.ts` like this, with no `next.config.*` file:

```ts
import vinext from 'vinext';
import { defineConfig } from 'vite';

export default defineConfig({
  plugins: [
    vinext({
      nextConfig: {
        output: 'export',
      },
    }),
  ],
});
```

Then run:

```sh
vinext build
```

## Actual behavior

The build completes as a normal vinext build, but it does not enter static export mode.

In our app this meant:

- No `Pre-rendering all routes (output: 'export')...` message.
- No route HTML emitted into `dist/client`.
- A downstream packaging script found zero `.html` files from vinext export.

## Expected behavior

If `nextConfig.output === "export"` is supplied through `vinext({ nextConfig })`, `vinext build` should behave the same as when `output: "export"` is supplied through `next.config.ts`.

Expected build output should include:

```text
Pre-rendering all routes (output: 'export')...
```

## Workaround

Adding a root `next.config.ts` works:

```ts
const nextConfig = {
  output: 'export',
} as const;

export default nextConfig;
```

With that file present, `vinext build` correctly enters static export mode:

```text
Pre-rendering all routes (output: 'export')...
Prerendered 43 routes (0 skipped).
```

## Suspected cause

From the installed `vinext/dist/cli.js`, `vinext build` appears to decide the output mode before Vite plugin config metadata is used:

```js
const resolvedNextConfig = await resolveNextConfig(
  await loadNextConfig(root, PHASE_PRODUCTION_BUILD),
  root
);

const outputMode = resolvedNextConfig.output;
```

Later, prerender/export mode is selected from that `resolvedNextConfig.output`:

```js
const prerenderDecision = resolveVinextPrerenderDecision({
  prerenderAllFlag: parsed.prerenderAll,
  vinextPrerenderConfig: buildConfigMetadata.prerenderConfig,
  nextOutput: resolvedNextConfig.output
});
```

This means `output: "export"` is only visible if it came from `loadNextConfig()`, i.e. from `next.config.*`.

By contrast, `vinext({ nextConfig })` appears to be loaded as Vite plugin configuration. That configuration affects the plugin, but it does not appear to be fed back into the CLI-level `resolvedNextConfig` used to decide `outputMode`.

## Why this matters

For vinext-first projects, requiring `next.config.ts` just to enable static export is surprising because:

1. The documented `vinext({ nextConfig })` API says projects can configure vinext from `vite.config.*` without a separate `next.config.*`.
2. `output: "export"` is specifically documented as supported.
3. The build silently proceeds without static export rather than warning that inline `nextConfig.output` is ignored by the CLI export decision.
4. It creates two possible config locations, but only one of them works for this specific option.

## Possible fixes

Any of these would resolve the mismatch:

1. Make `vinext build` load inline `nextConfig` from the Vite plugin before deciding `outputMode`.
2. Include `nextConfig` in the existing build Vite config metadata path, similar to `prerenderConfig`.
3. If inline `nextConfig.output` cannot be supported, document that `output: "export"` must be defined in `next.config.*`.
4. Emit a warning/error if `vinext({ nextConfig: { output: "export" } })` is present but the CLI will not use it to select static export mode.

## Additional related observation

`vinext({ prerender: true })` is a vinext-native option and does trigger prerendering, but it is not the same documented static export mode as `output: "export"`. The issue here is specifically about the documented `output: "export"` behavior not working through the documented inline `nextConfig` path.

james-elicx:
## Summary

- expose `vinext({ nextConfig })` through Vite plugin metadata for CLI/deploy config discovery
- resolve inline object/function config before output-mode decisions and pass the resolved config into prerender execution
- keep Cloudflare deploy config resolution inside the selected `CLOUDFLARE_ENV`
- support promised/nested Vite plugin composition when discovering inline config

## Root cause

`vinext build` and `vinext-cloudflare deploy` decided whether to use static export from `next.config.*` before reading the Vite plugin's inline `nextConfig`. Even after selecting prerendering, the prerender runner reloaded disk config, so inline-only `output: "export"` did not receive export output/error semantics.

## Validation

- `vp test run tests/deploy-prerender-config.test.ts tests/prerender-config.test.ts` (23 tests)
- `vp check packages/cloudflare/src/deploy.ts packages/vinext/src/cli.ts packages/vinext/src/config/next-config.ts packages/vinext/src/index.ts tests/deploy-prerender-config.test.ts tests/prerender-config.test.ts`
- `vp run vinext#build`
- manual real CLI reproduction with inline `nextConfig.output = "export"`; confirmed export-mode log and HTML at `dist/client/index.html` and `dist/client/old-school.html`
- two independent reviews: no findings

Closes #2542

### DIFF
--- a/packages/vinext/src/utils/plugin-options.ts
+++ b/packages/vinext/src/utils/plugin-options.ts
@@ -1,0 +1,27 @@
+export async function flattenPluginOptions(value: unknown): Promise<unknown[]> {
+  if (value instanceof Promise) {
+    return flattenPluginOptions(await value);
+  }
+  if (Array.isArray(value)) {
+    return (await Promise.all(value.map((item) => flattenPluginOptions(item)))).flat();
+  }
+  return value ? [value] : [];
+}
+
+export async function selectHybridPagesUserPlugins(
+  value: unknown,
+): Promise<import("vite").Plugin[]> {
+  const flattened = await flattenPluginOptions(value);
+  return flattened.filter(
+    (plugin): plugin is import("vite").Plugin =>
+      !!plugin &&
+      typeof plugin === "object" &&
+      "name" in plugin &&
+      typeof plugin.name === "string" &&
+      !plugin.name.startsWith("vinext:") &&
+      !plugin.name.startsWith("vite:react") &&
+      !plugin.name.startsWith("rsc:") &&
+      plugin.name !== "vite-rsc-load-module-dev-proxy" &&
+      !plugin.name.startsWith("vite-plugin-cloudflare"),
+  );
+}
--- a/packages/vinext/src/utils/plugin-options.ts
+++ b/packages/vinext/src/utils/plugin-options.ts
@@ -7,21 +7,3 @@ export async function flattenPluginOptions(value: unknown): Promise<unknown[]> {
   }
   return value ? [value] : [];
 }
-
-export async function selectHybridPagesUserPlugins(
-  value: unknown,
-): Promise<import("vite").Plugin[]> {
-  const flattened = await flattenPluginOptions(value);
-  return flattened.filter(
-    (plugin): plugin is import("vite").Plugin =>
-      !!plugin &&
-      typeof plugin === "object" &&
-      "name" in plugin &&
-      typeof plugin.name === "string" &&
-      !plugin.name.startsWith("vinext:") &&
-      !plugin.name.startsWith("vite:react") &&
-      !plugin.name.startsWith("rsc:") &&
-      plugin.name !== "vite-rsc-load-module-dev-proxy" &&
-      !plugin.name.startsWith("vite-plugin-cloudflare"),
-  );
-}
--- a/packages/cloudflare/src/deploy.ts
+++ b/packages/cloudflare/src/deploy.ts
@@ -276,7 +276,9 @@ async function loadDeployViteConfigMetadata(root: string): Promise<DeployViteCon
   );
   const plugins = loaded?.config.plugins;
   return {
-    cacheConfig: await findVinextCacheConfigInPlugins(plugins),
+    cacheConfig: viteConfigHasCacheAdapter(root)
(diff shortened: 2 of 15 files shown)


packages/vinext/src/utils/plugin-options.ts
```diff
@@ -0,0 +1,27 @@
...
+    (plugin): plugin is import("vite").Plugin =>
+      !!plugin &&
+      typeof plugin === "object" &&
+      "name" in plugin &&
+      typeof plugin.name === "string" &&
+      !plugin.name.startsWith("vinext:") &&
+      !plugin.name.startsWith("vite:react") &&
+      !plugin.name.startsWith("rsc:") &&
+      plugin.name !== "vite-rsc-load-module-dev-proxy" &&
+      !plugin.name.startsWith("vite-plugin-cloudflare"),
```
jon-hotaisle: This looks like a ton of slop and very brittle.

jon-hotaisle: Also, I can see it got moved from cli.ts... without any of the nice comments that were in there before.

Image

james-elicx: i've not looked at the diff yet - this may change.

PR #2544, opened 2026-07-06
fix(pages): preserve fast refresh state

james-elicx:
## Summary

- preserve Pages Router Fast Refresh state instead of forcing full document reloads for ordinary script edits
- consume Pages SSR dead-end updates after downstream HMR plugins run, while retaining full reloads for route additions and deletions
- preload CSS introduced by client transforms, including virtual CSS modules, so dev HTML has blocking first-paint styles
- compile MDX before the development React transform and extend the default dev-only React include so MDX routes also receive Fast Refresh without changing production parsing
- add unit and browser regression coverage for state preservation, syntax-error recovery, transformed CSS, and MDX route updates

## Motivation

DigitecGalaxus/next-yak#569 added vinext to its bundler compatibility suite and exposed two dev-only gaps:

- seven HMR cases failed because vinext reloaded the document for every Pages Router script edit, clearing browser state
- pseudo-element styles raced the first computed-style read because next-yak's transformed virtual CSS was only injected during hydration

The HMR behavior now matches Next.js Pages Router Fast Refresh. Dev stylesheet discovery now follows Vite's transformed client module graph and emits blocking stylesheet links for generated CSS, matching production's first-paint behavior.

The automatic React integration now uses command-scoped plugin instances: production keeps the existing configured include behavior, while development adds `.mdx` only when the user did not explicitly configure `react.include`. The MDX delegate is split into an early transform proxy and a later config proxy so compiled JSX reaches React Fast Refresh while the delegate is still initialized during `vinext:config`.

## Validation

- `vp check packages/vinext/src/index.ts packages/vinext/src/server/dev-server.ts tests/pages-router.test.ts tests/fixtures/pages-basic/pages/hmr-state.tsx tests/e2e/pages-router/hmr.spec.ts`
- targeted Pages Router transformed-CSS tests — pass
- Pages Router HMR Playwright regression — pass
- MDX Pages Router Fast Refresh regression — pass
- failed CI Vitest shard (`integration 4/10`) — 398/398 pass locally
- next-yak `vinext-pages` dev matrix — 27/27 pass, including all seven HMR cases and `pseudo-elements`

Related: DigitecGalaxus/next-yak#569

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -1,6 +1,7 @@
 import type {
   Alias,
   CSSModulesOptions,
+  HotUpdateOptions,
   Logger,
   Plugin,
   PluginOption,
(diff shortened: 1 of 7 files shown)


Mad-Kat:
I think

> pseudo-element styles raced the first computed-style read because next-yak's transformed virtual CSS was only injected during hydration

is wrong, but the solution to follow vite's CSS resolving behaviour sounds correct to me. Thanks for looking into it

PR #2563, opened 2026-07-08
fix(actions): enforce server action route ownership

james-elicx:
## Summary

- enforce route ownership before dispatching production Server Actions
- collect export-level action provenance from plugin-RSC's existing lexer analysis pass instead of performing a second vinext AST crawl
- preserve ownership through client component boundaries, re-export barrels, dynamic imports, inline actions, layouts, and same-named non-action exports
- use native per-environment `typeof window` defines for real builds
- run vinext's scope-aware `typeof window` fold only during plugin-RSC's write-less lexer analysis builds, so dead imports are removed before the synthetic scan graph is generated

## Why

Production action identifiers are deterministic and can be computed offline. Dispatch must therefore verify that the requested action is reachable from the current route before decoding or executing it.

The initial ownership implementation also performed a separate source traversal. This version reuses plugin-RSC's existing lexer scan metadata and keeps the extra `typeof window` transform limited to the analysis phase where native Oxc define folding has not run yet.

## Validation

- `pnpm exec vitest run tests/type-of-window.test.ts tests/action-owner-manifest.test.ts tests/action-runtime-security.test.ts`
  - 33 tests passed
- Same-worktree hyperfine A/B on `tests/fixtures/app-action-ownership`, 15 runs in each command order:
  - scan-only fold: 1.265s and 1.297s mean
  - no scan-only fold: 1.334s and 1.407s mean
  - no measured regression; the folded scan trended 5-8% faster by avoiding dead synthetic imports

Benchmark output:

- `/tmp/vinext-actions-scan-fold-ab.json`
- `/tmp/vinext-actions-scan-fold-ab-reversed.json`

## Upstream dependency

This currently expects the plugin-RSC scan observer API used by the patched workspace dependency. The plugin-RSC implementation itself remains on its existing lexer/synthetic-import scan path.


james-elicx:
Closing because this draft accidentally included unrelated server-action ownership work. I’ll replace it with a clean typeof-window-only PR.

PR #2564, opened 2026-07-08
perf(build): use native typeof window folding

james-elicx:
## Summary

- configure native per-environment `typeof window` defines so real client/server builds use Oxc/Rolldown folding
- retain vinext's scope-aware fold only for plugin-RSC's write-less analysis builds, where the lexer replaces source before native define folding runs
- remove the broad custom transform and its mutable cache-directory hook filter

## Performance

The custom AST fold no longer runs for normal production modules. It is limited to plugin-RSC's two analysis builds and is invoked only for modules matching the native `typeof window` hook filter.

Same-worktree hyperfine A/B on the action fixture showed no regression; the analysis fold trended 5-8% faster because dead imports were omitted from the synthetic scan graph.

## Validation

- `pnpm exec vitest run tests/type-of-window.test.ts`
- 10 tests passed

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -236,7 +236,6 @@ import { createHash, randomBytes, randomUUID } from "node:crypto";
 import commonjs from "vite-plugin-commonjs";
 import { createIgnoreDynamicRequestsPlugin } from "./plugins/ignore-dynamic-requests.js";
 import { stripJsExtension, stripViteModuleQuery } from "./utils/path.js";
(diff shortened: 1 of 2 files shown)


james-elicx:
/bonk review the latest PR head for issues

james-elicx:
/bigbonk review for issues, scoped to the latest commit 09b04d4e369975733d65153615d705b0bec3a6f9 only (the apply: build addition and cacheDir regression test). Post a concise blocker/no-blocker verdict.

Issue #2007, opened 2026-06-13
Async generateMetadata tags render into <div hidden> in <body> instead of <head>


jerome281:
# Async `generateMetadata` tags render into a `<div hidden>` in `<body>` instead of `<head>`

## Summary

On routes that resolve metadata via **async `generateMetadata`**, the produced tags — `<title>`, `<meta name="description">`, `<link rel="canonical">`, and all `og:*` / `twitter:*` tags — are emitted inside a `<div hidden>` in the `<body>`, **not** in `<head>`. Routes that use a **static** `export const metadata` render the same tags correctly into `<head>`.

This breaks `rel=canonical` (Google ignores canonical outside `<head>`) and Open Graph / Twitter Card previews (non-JS scrapers — Facebook, Slack, LinkedIn, WhatsApp, X — only parse `<head>`), and fails Lighthouse's `meta-description` SEO audit.

## Expected behavior

Match Next.js: `generateMetadata` is awaited before the `<head>` shell is streamed, so the resolved tags land in `<head>` — identical to static `metadata`.

## Actual behavior

The `<head>` shell is flushed before the async page metadata resolves. The late-resolved tags are then rendered inline in the document body inside a `<div hidden>` and never hoisted into `<head>`.

## Reproduction

1. App Router project on vinext, Cloudflare Workers target.
2. Root layout exports static `metadata` (works) — e.g. a `description`.
3. A nested dynamic route exports async `generateMetadata`:

```ts
// app/[locale]/deals/page.tsx
export async function generateMetadata({ searchParams }): Promise<Metadata> {
  const description = 'Browse the latest UK running shoe deals…';
  const canonical = canonicalFor('/uk/deals', page);
  return {
    title: 'Running Shoe Deals',
    description,
    alternates: { canonical },
    openGraph: { title: 'Running Shoe Deals', description, url: canonical },
    twitter: { card: 'summary_large_image', title: 'Running Shoe Deals', description },
  };
}
```

4. `pnpm build`, then serve the built Worker with `npx wrangler dev --local`.
5. `curl http://localhost:8787/uk/deals` and inspect where the tags land.

### Observed output (abridged)

`<head>` contains only the synchronously-available content (charset, viewport, stylesheet, font preloads, JSON-LD from the explicit `<head>` JSX in the root layout). The page metadata is in the body:

```html
<body>
  …
  <div hidden>
    <title>Running Shoe Deals | …</title>
[...]

PR #2572, opened 2026-07-08
fix(app-router): preserve streamed metadata placement parity

LubomirGeorgiev:
## Summary

Restore Next.js-compatible metadata placement across request-time streaming and static artifact rendering.

- Streaming-capable runtime requests keep async `generateMetadata` output in `<body>`.
- HTML-limited bots receive blocking metadata in `<head>`.
- Full build prerenders and ISR/cache regeneration artifacts resolve metadata into `<head>`.
- PPR fallback-shell behavior remains unchanged until metadata-level staticness is tracked.

Refs #2007.

## Root cause

The original change rendered streamed metadata as real React elements so React 19 hoisted it into `<head>` on every request. That diverged from Next.js: for normal request-time streaming, delayed metadata intentionally remains in `<body>`. Reverting that change restored the primary metadata-streaming suite, but exposed a second distinction: complete static artifacts must place generated metadata in `<head>`.

Vinext uses a manually serialized metadata outlet, so it cannot infer this distinction from React flush timing. This change threads an explicit placement decision from the App Router dispatcher into element construction while keeping UA-driven bot and metadata-error semantics separate.

## Verification

- Next.js `metadata-streaming.test.ts`: 19/19 passed
- Next.js customized-rule suite: 2/2 passed
- Next.js static-generation suite: 7/7 passed
- Next.js parallel-routes suite: 4/4 passed
- Focused vinext unit/integration tests: 212/212 passed
- Vinext metadata Playwright suite: 12/12 passed
- `vp check` passes on all changed files
- Two fresh independent sub-agent reviews found no remaining issues

The upstream PPR metadata-streaming suite is currently `describe.skip`; this change preserves existing fallback-shell placement rather than guessing at unsupported metadata staticness classification.


james-elicx:
There's a regression in the streamed metadata nextjs e2e. Will look at fixing.

LubomirGeorgiev:
@james-elicx Thanks

PR #2604, opened 2026-07-13
perf(build): split react-dom/server into its own client chunk

blitss:
## Problem

`createClientManualChunks` keys the always-loaded `framework` chunk on the bare package name (`react-dom`), so the server renderer entrypoints — `react-dom/server.browser`, `react-dom/static.*`, and their sizeable cjs implementation (Fizz: `renderToString` / `renderToReadableStream` / `renderToStaticMarkup` + the prerender APIs) — ride along on **every** page, even though only client code that explicitly renders to a string (e.g. an embedded Sanity Studio) ever imports them.

## Change

Route the server/static renderer entrypoints (`server.*`, `static.*`, `cjs/react-dom-server*`) to a dedicated `react-dom-server` chunk. The rest of `react-dom` (client renderer, shared internals, the `server-rendering-stub`) stays in `framework` as before.

Result: the `framework` chunk is ~35KB brotli lighter on every page; the server renderer chunk loads lazily, only on routes that import it.

## Testing

Added unit tests in `tests/build-optimization.test.ts` covering:
- `server.browser` / `server.edge` / `static.edge` / `cjs/react-dom-server*` (incl. legacy) → `react-dom-server`
- `react-dom/index`, `cjs/react-dom-client.*`, `cjs/react-dom.*` shared internals, and `server-rendering-stub` → `framework`

`pnpm test tests/build-optimization.test.ts` — 148 passed. `pnpm run check` — clean.


blitss:
Addressed the Windows path finding: `sub` in the `react-dom` branch of `createClientManualChunks` was sliced from the raw id with a `"react-dom/"` literal, so on Windows (`C:\app\node_modules\react-dom\server.browser.js`) `lastIndexOf` returned -1 and every react-dom entrypoint silently fell into `framework`, disabling the split.

Fixed in 0ae641c7 by slash-normalizing the id with the file's existing `toSlash` helper (same convention `getPackageName` and the shims-dir match already use) before computing `sub`.

Added a test next to the existing react-dom-server chunk tests asserting backslash ids classify correctly (`server.browser.js` / `cjs/react-dom-server.*` → `react-dom-server`; `client.js` / `server-rendering-stub.js` → `framework`). It's gated `it.runIf(process.platform === "win32")` matching the existing Windows test convention, because `toSlash` is deliberately a no-op on POSIX (backslash is a legal filename character there). Pre-fix the test fails on win32: with `lastIndexOf("react-dom/") === -1`, `sub` becomes `id.slice(9)` which never matches the server-entry prefixes, yielding `framework`.

blitss:
CI note: the Vitest (unit 1/3) failure was `link-navigation.test.ts > re-prefetches visible links after the prefetch cache is invalidated` — unrelated to this PR (chunk classification never intersects Link prefetch scheduling; the test mocks IntersectionObserver/fetch). It passes 5/5 locally on this exact head, and the file has prior flake history (#1218). Pushed an empty commit to retrigger.

blitss:
@james-elicx any chance u can re-trigger review; all findings should be addressed here and on the #2605, #2607

PR #2605, opened 2026-07-13
fix(fonts): keep immutable font assets query-free

blitss:
## Problem

Font preload URLs carried `?dpl=` deployment-skew tokens while vinext's self-hosted font `@font-face` URLs were bare. Because preload and CSS URLs must be byte-identical, browsers could discard the preload and download the same immutable font again.

Fonts emitted by `next/font/google` and `next/font/local` are content-hashed immutable assets, so they do not need deployment-ID cache busting. This matches Next.js's immutable-asset deployment path: when `supportsImmutableAssets` applies, the client asset token is empty and static immutable URLs remain query-free.

## Change

- Keep App Router font preload URLs bare so they match `@font-face` and HTTP `Link` URLs.
- Apply the same behavior to Pages Router HTML and HTTP `Link` preloads.
- Cover App and Pages production output with both Google and local fonts, asserting query-free CSS/HTML/Link URL identity.
- Verify the emitted Google assets, and App Router local asset, are served with immutable caching.

Parity reference: Next.js `test/production/deployment-id-handling/deployment-id-handling.test.ts` and its `NEXT_DEPLOYMENT_ID_IMMUTABLE` / `supportsImmutableAssets` fixture configuration.

## Testing

- `vp test run tests/app-router-font-google-prod.test.ts tests/pages-router-font-google-prod.test.ts tests/font-google.test.ts tests/font-local-transform.test.ts` (163 passed)
- targeted `vp check` for all changed source and test files
- independent cumulative parity/correctness review: CLEAN


james-elicx:
/bigbonk review for issues and nextjs parity

blitss:
Addressed the blocking Pages Router parity concern in 12d9f4e9 — both sites fixed plus a regression test:

- `pages-page-response.ts` (`buildPagesFontHeadHtml`): the HTML `<link rel="preload">` href now emits `preload.href` bare instead of wrapping it in `appendAssetDeploymentIdQuery(...)`, with the same explanatory comment as the App Router fix. The stylesheet `<link>` hrefs on the line above still use the query (those are matched by URL-with-query requests), so the import stays.
- `pages-page-handler.ts` (HTTP `Link:` header): now built from bare `p.href`, matching App Router's `buildAppPageFontLinkHeader`. That was the file's only `appendAssetDeploymentIdQuery` callsite, so the now-unused import is removed.
- New test `tests/pages-router-font-google-prod.test.ts` with a Pages fixture (`tests/fixtures/font-google-pages`, `next/font/google` in `_app.tsx`, hermetic mocked Google Fonts CDN): pins `deploymentId` and asserts both the rendered preload hrefs and the `Link:` header URLs contain no `dpl` and are byte-identical to the `@font-face` `src: url(...)` values. Verified the test fails on the pre-fix code with both assertions catching `?dpl=dpl-pages-font-prod` leaking into the URLs.

`vp test run` on the touched test files (83 tests) passes; `fmt`/`check` clean.

james-elicx:
sorry, my agent went a little rogue with deciding what was correct there.

PR #2607, opened 2026-07-13
fix(font-google): share SSR collection state via globalThis

blitss:
## Problem

In Vite's multi-environment dev mode the `font-google-base` shim can be loaded more than once in the same worker (resolved through different IDs across the rsc/ssr environments), so each module copy gets its own freshly-initialized closure variables. The `fontLoader` call site and the SSR `getSSRFontStyles()` reader can land on **different copies** — the reader returns an empty array even though the loader's pushes succeeded — and the collected `<style data-vinext-fonts>` block silently never makes it into the HTML head. Preload links and dedup sets drop the same way.

Production builds are unaffected because the build plugin inlines `_selfHostedCSS` into each font-loader call, so that path doesn't depend on cross-copy state visibility.

## Change

Back every piece of mutable module state (injection-tracking Sets, SSR collection arrays, preload dedup) with a `Symbol.for` slot on `globalThis` — the same pattern `navigation.ts` already uses for its layout segment context. All module copies collapse onto one shared store, so writes from one copy are visible to reads from another.

## Testing

`pnpm test tests/shims.test.ts tests/app-router-font-google-prod.test.ts` — 1205 passed. `pnpm run check` — clean. (`tests/font-google.test.ts` network-dependent build test fails identically on pristine main in my sandbox — pre-existing, unrelated.)


blitss:
Addressed the review in 1c0ecca0:

1. **font-local.ts had the identical multi-copy bug — fixed.** Every per-copy mutable binding in the local-font shim is now backed by a `Symbol.for` slot on `globalThis` (namespaced `vinext.fontLocal.*` to stay disjoint from the google shim's `vinext.font.*` slots): `injectedFonts`, `injectedClassRules`, `injectedVariableRules`, `injectedRootVariables`, `ssrFontStyles`, `ssrFontPreloads`, `ssrFontPreloadHrefs`, and `classCounter`. The counter is *reassigned* rather than mutated in place, so a bare `??=` alias would still diverge per copy — it's boxed in a shared `{ value: number }` object instead (commented in-source).

2. **Regression test added** (`tests/shims.test.ts`, "shares SSR collection state and class counter across module copies"): calls `localFont()` through one module copy, then uses `vi.resetModules()` + dynamic re-import to create a genuinely fresh second copy, and asserts (a) the second copy's `getSSRFontStyles()`/`getSSRFontPreloads()` see the first copy's pushes and (b) the class counter continues across copies (the second copy does not restart at `__font_local_0`). Verified it fails on the pre-fix shim (SSR-styles assertion trips) and passes post-fix. Since `Symbol.for` slots deliberately survive `vi.resetModules()`, the test clears them in a `finally` block so the rest of the suite is unaffected. Full `tests/shims.test.ts` passes (1202 tests).

3. **`[k: symbol]: unknown` index signature tightened.** Both font-google-base.ts and font-local.ts now use the named-key `typeof globalThis & { [_KEY]?: T }` augmentation style from navigation.ts, which also removed the per-site `as` casts.

`pnpm run fmt` and `pnpm run check` are clean.

PR #2609, opened 2026-07-13
fix(router): replace stale optimistic layouts across dynamic params

### LINKED ISSUE #2597: App Router: soft navigation between different dynamic params reuses the committed shared-layout subtree without fetching the new RSC payload
### Summary

On soft navigation between two URLs that match the **same shared dynamic layout** but with **different params** (`/projects/A` → back → `/projects/B`), the client router reuses the committed subtree from `/projects/A` **without fetching the RSC payload for `/projects/B`**. The URL and `<title>` update to B, but everything rendered from the layout's server props (data fetched with A's param) remains A's. A layout-level `key={param}` does not help — the reused subtree is never re-rendered with fresh server output, so no new key is ever applied.

### Repro

App-router app with a shared dynamic layout that fetches per-param data:

```
app/
  dashboard/page.tsx          // links to /projects/A and /projects/B
  projects/[pid]/layout.tsx   // async layout: fetches project by pid, renders client shell with it as props
  projects/[pid]/page.tsx
```

Steps (client-side navigation throughout, e.g. `<Link>` clicks + browser back):

1. Load `/dashboard`.
2. Click a link to `/projects/A` → renders A correctly.
3. Browser **back** to `/dashboard`.
4. Click a link to `/projects/B`.

**Expected:** `/projects/B` renders B's layout data (an `?_rsc` fetch for `/projects/B`, or a fresh render of the layout with `pid=B`).

**Observed:** URL bar and `<title>` show B (metadata appears to refetch), but the rendered content is A's. The network panel shows **no `GET /projects/B?_rsc` request at all** for step 4 — the only activity is A's own subresources (in our app, an iframe still pointing at A's preview host). Reproduces both in `vinext dev` and consistently across pairs of params; order doesn't matter (B → back → A shows B under A's URL).

### What we ruled out

- **Server side is fine:** when the `?_rsc` request *is* issued (first visit to a pid, or hard load), the correct payload is returned.
- **`key={pid}` on the layout's client child doesn't fix it** — the committed subtree is reused as-is, so the fresh keyed render never applies.
- **`router.refresh()` from a client guard "works" but races navigation:** a popstate while the refresh's RSC stream is in flight hits `abortSupersededNavigation` → uncaught `BodyStreamBuffer was aborted` (stack below), and the stale subtree survives. Our current containment is a `window.location.reload()` when a client component sees `useParams().pid` disagree with its server props, which can't be superseded — but it's a full page load on every cross-param navigation into the route.

```
BodyStreamBuffer was aborted
  abortSupersededNavigation  vinext/dist/server/app-browser-entry.js:836:36
  navigateRsc                vinext/dist/server/app-browser-entry.js:843:4
  <anonymous>                vinext/dist/server/app-browser-popstate.js:29:49
  <anonymous>                vinext/dist/server/app-browser-entry.js:1181:3
```

(The `useParams()` context inside the reused subtree **does** update to B — only the server-rendered output is stale. That's what makes the client-side detection possible.)

### Suspected area

Behaviorally this looks like the committed client cache / BFCache-style payload reuse treating the route pattern (`/projects/[pid]`) as the reuse key rather than the concrete URL, when the navigation goes dashboard → back → different param. Possibly related to the committed-payload reuse work (#2251) and history-entry restoration (#1366) — the back-traversal restores dashboard from cache, and the subsequent forward navigation to `/projects/B` appears to reuse `/projects/A`'s committed entry instead of issuing a new RSC request.

### Environment

- vinext `1.0.0-beta.1` (also reproduced on `0.2.0`)
- vite `8.0.16`, react `19.2.6`, node `25.2.1`, macOS (darwin 25.5.0)
- `vinext dev` targeting Cloudflare Workers (workerd via miniflare)

Happy to provide more traces or test a patch — this bites any app where per-param data flows through a shared dynamic layout (in ours, the layout binds a per-project WebSocket + iframe, so the wrong project's editor renders under the new URL).

NathanDrake2406:
## Overview

| | |
| --- | --- |
| Goal | Prevent stale server-rendered dynamic-layout output from surviving an optimistic-to-authoritative navigation handoff |
| Core change | Prepare every payload in one navigation against the immutable router state captured when that navigation starts |
| Commit authority | Continue checking the live router state before any payload becomes visible |
| Expected behavior | Cross-param layouts receive fresh output and identity; shared roots and same-param sibling layouts retain client state |
| Scope | Browser navigation preparation, controller input typing, and focused BFCache identity coverage |

Closes #2597.

## Why

A single navigation can render a detached optimistic shell before its authoritative RSC payload arrives. Both payloads must derive layout reuse from the same pre-navigation state. Previously, commit preparation sampled whichever router state was visible at that moment. Once the detached shell committed, the authoritative response could treat that shell as its reuse base and preserve stale dynamic-layout output from another param.

| --- | --- | --- |
| Navigation preparation | Payloads from one navigation share one historical identity base | `navigateRsc()` captures the initiation state once, before redirects, prefetch consumption, or detached commits |
| Commit approval | Only current work may mutate visible state | Approval still reads the live router state and rejects superseded navigation IDs or unrelated visible commits |
| Layout identity | Concrete dynamic-param changes require fresh identity | Existing identity comparison replaces cross-param or newly introduced optimistic layouts while preserving proven shared layouts |
| Ownership | Decision inputs should cross boundaries explicitly | The private payload wrapper now uses a typed options object and passes the initiation state into controller preparation |

This follows the navigation architecture in #726: deterministic preparation receives explicit inputs, while the visible commit gate owns lifecycle authority. The trace work in #2598 helped isolate this boundary, but this fix does not depend on diagnostics.

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| Dashboard to optimistic B to authoritative B | Authoritative preparation could reuse stale A output installed by the detached shell | Authoritative preparation compares against the dashboard initiation state and installs B |
| Same-param sibling navigation | Shared dynamic layout state could be disturbed by a broad replacement fix | Existing layout object and BFCache identity remain stable |
| Dynamic-layout-owned slot | A slot introduced by the detached owner could survive the authoritative payload | The authoritative owner and slot receive fresh identities together |
| Redirect or visited-response handoff | A later payload could prepare from state committed by an earlier hop | Redirected, cached, detached, and live payloads share the initiation state |
| Cancellation and supersession | Live state guarded visible commits | Unchanged |

The browser regression holds the B RSC response long enough to inspect the detached shell, then releases it and verifies that server props, pathname, params, client state, and BFCache identity all agree on B.

<details>
<summary>Validation</summary>

- `vp check` on all touched source, test, and fixture files
- `vp test run tests/app-browser-entry.test.ts` (227 tests)
- `PLAYWRIGHT_PROJECT=app-router vp exec playwright test tests/e2e/app-router/nextjs-compat/use-router-bfcache-id.spec.ts --workers=1` (13 tests)
- Server-action invalidation race repeated three times after replacing response-timing with a BFCache-epoch commit assertion
- `vp run vinext#build`
- `git diff --check`
- Repository pre-commit staged checks, full check, staged unit and integration tests, and knip

</details>

<details>
<summary>Risk and compatibility</summary>

- No public API, configuration, wire-format, or build-output change.
- The runtime change only affects the state used to prepare navigation identities.
- The live commit gate still owns cancellation, supersession, root-boundary checks, and visible commit approval.
- The fix reuses the existing layout and slot identity rules instead of adding a parallel cache lookup or fallback policy.
- The new loading boundary and server-derived markers exist only in the app-router test fixture.

</details>

<details>
<summary>Non-goals</summary>

- Add segment-cache entries or a new segment identity model.
- Change `NavigationPlanner` decisions or scan caches during commit preparation.
- Introduce a global `reuseCurrentBfcacheIds` fallback.
- Couple correctness to the opt-in diagnostics from #2598.

</details>

## References

| --- | --- |
| [Issue #2597](https://github.com/cloudflare/vinext/issues/2597) | Reproduction and expected cross-param behavior |
| [Navigation architecture #726](https://github.com/cloudflare/vinext/issues/726) | Defines the separation between deterministic decisions and visible commit authority |
| [Navigation diagnostics #2598](https://github.com/cloudflare/vinext/pull/2598) | Trace evidence that localized the stale state to the optimistic-to-authoritative handoff |
| [Committed payload reuse #2251](https://github.com/cloudflare/vinext/pull/2251) | Prior reuse path implicated by the reported sequence |
| [Next.js optimistic navigation preparation](https://github.com/vercel/next.js/blob/31541356c21bc389150b958379acb26aa7e97b22/packages/next/src/client/components/router-reducer/ppr-navigations.ts#L178-L206) | Upstream models the optimistic tree as a provisional result reconciled by the navigation response |
| [Next.js segment identity rules](https://github.com/vercel/next.js/blob/31541356c21bc389150b958379acb26aa7e97b22/packages/next/src/client/components/router-reducer/ppr-navigations.ts#L1352-L1363) | Distinguishes fresh dynamic-param identity from search-only reuse |
| [Next.js BFCache identity tests](https://github.com/vercel/next.js/blob/31541356c21bc389150b958379acb26aa7e97b22/test/e2e/app-dir/use-router-bfcache-id/use-router-bfcache-id.test.ts#L59-L88) | Covers sibling layout preservation and cross-group reset semantics |


NathanDrake2406 (commented):
`navigationInitiationState` is a correctness input for lifecycle-spanning navigations, but it remains optional and silently falls back to `getBrowserRouterState()`. That leaves future callers able to recreate the same temporal bug while still compiling.

Please strengthen the API so initiation-anchored payload preparation requires this state. At minimum, make it required in the private `renderNavigationPayload` wrapper used by `navigateRsc()`. If the lower-level controller still needs current-state preparation for isolated operations, expose that as an explicit separate path rather than an implicit fallback.

NathanDrake2406:
Addressed in `63453313d`. `navigationInitiationState` is now required at both the browser-entry wrapper and controller boundary, and payload preparation no longer falls back to `getBrowserRouterState()`. Direct controller tests that intentionally prepare from current state now do so through the explicitly named test-only `renderCurrentStateNavigationPayload()` helper.

NathanDrake2406:
@codex review

quantizor:
👏 thanks for all these improvements, I've been loving vinext

NathanDrake2406:
> 👏 thanks for all these improvements, I've been loving vinext

Thanks bro! But it's all this gigachad -> @james-elicx

PR #2621, opened 2026-07-14
fix(app-router): defer browser runtime controller creation

### LINKED ISSUE #2620: App Router prod build: circular chunk init crashes hydration on every route (`RestorableClientStateController is not a constructor`) — beta.0 & beta.1
## Summary

Production **client** bundles crash during hydration on **every route** with `TypeError: <Class> is not a constructor`. The failing class is vinext's own `RestorableClientStateController`, instantiated by `AppBrowserHistoryController` at the top level of the browser entry. `vinext dev` is fine; only `vinext build` output is broken. Reproduces on **`1.0.0-beta.0`** and **`1.0.0-beta.1`**. Last working version for us is `0.0.46`.

Because the crash happens at module-evaluation time in the client bootstrap, server-rendered HTML still appears but the page never hydrates: static pages *look* alive, while any client component (ours: an `/internal` route) freezes on its loading state, and client-side navigation is dead everywhere.

## Environment

- **vinext:** `1.0.0-beta.0` and `1.0.0-beta.1` (both reproduce)
- **Bundler:** vinext-bundled `vite` = `npm:@voidzero-dev/vite-plus-core` (`0.2.1` in beta.0, `0.2.2` in beta.1) — i.e. Rolldown
- **Peers:** `vite@8.0.10`, `@vitejs/plugin-rsc@0.5.26`, `@vitejs/plugin-react@6`, `react`/`react-dom@19.2.6`
- **Target:** Cloudflare Workers (`@cloudflare/vite-plugin`), App Router
- **Node:** 24/25, macOS arm64

## Error

Minified (beta.1):

```
TypeError: I is not a constructor
    at <client-entry chunk>
```

With `build: { minify: false }` the names resolve (beta.0 dist):

```
TypeError: RestorableClientStateController is not a constructor
    at AppBrowserHistoryController (index-*.js)   // this.#restorableClientState = new RestorableClientStateController({...})
    at <top-level of client entry> (index-*.js)   // const historyController = new AppBrowserHistoryController({...})
```

## Root cause — circular chunk dependency

`dist/server/app-browser-entry.js` runs a **top-level side effect**:

```js
const historyController = new AppBrowserHistoryController({
  initialHistoryState: window.history.state,
  maxHistoryStateSnapshots: 50,
  ...
});
```

and `AppBrowserHistoryController`'s constructor does `this.#restorableClientState = new RestorableClientStateController({ initialHistoryState, maxHistoryStateSnapshots })`.

In the production build these two land in **separate chunks with a direct cycle**:

- the **entry/`index`** chunk contains `AppBrowserHistoryController` + the top-level `new AppBrowserHistoryController(...)`, and imports `RestorableClientStateController` from the **`navigation`** chunk;
- the **`navigation`** chunk defines `RestorableClientStateController` **and imports back** from the `index` chunk.

Chunk execution enters `navigation` first; `navigation` imports `index`, which runs its top-level `new AppBrowserHistoryController(...)` **before** `navigation` has reached the `RestorableClientStateController` class definition. So the imported binding is still `undefined` → "is not a constructor".

`0.0.46` (plain Vite/**Rollup**, no bundled vite-plus) handled this cycle correctly. `dev` works because native ESM evaluates the dependency before the importer's body. Only the Rolldown production chunking reorders it unsafely.

## What I ruled out

- **Not minification** — reproduces identically with `build.rolldownOptions`/`build.minify:false`.
- **Not treeshake side-effects** — patching the client build's `getClientTreeshakeConfig()` from `{ moduleSideEffects: "no-external" }` to `{ moduleSideEffects: true }` had no effect.
- **Not app-fixable** — vinext owns `output.codeSplitting`, so app-level `build.rolldownOptions.output.manualChunks` / `advancedChunks` are ignored ("`advancedChunks`/`manualChunks` option is ignored because the `codeSplitting` option is specified").

## Reproduce

1. App Router app on `vinext@1.0.0-beta.1`, Cloudflare target, Vite 8.
2. `vinext build` then `vinext start` (or deploy).
3. Load any route in a browser → console throws `TypeError: … is not a constructor`; page never hydrates. `vinext dev` is unaffected.
4. Add `build: { minify: false }` to see the real names (`RestorableClientStateController` / `AppBrowserHistoryController`).

I haven't reduced this to a from-scratch minimal repo yet (it depends on the app's module graph producing the `index`↔`navigation` chunk split), but the failing code is entirely within vinext's own client runtime. Happy to provide a minimal reproduction repo if that helps triage.

## Suggested direction

Either keep the interdependent runtime modules (`app-browser-entry` / `app-browser-history-controller` / `navigation` `RestorableClientStateController`) in one chunk, or avoid the top-level `new AppBrowserHistoryController(...)` side effect during client-entry evaluation (lazy/first-use init), so a cross-chunk cycle can't observe the class before it's defined.

NathanDrake2406:
## Overview

| | |
|---|---|
| Goal | Stop circular production chunks from crashing App Router hydration |
| Change | Lazily create the history controller and MPA navigation scheduler |
| Boundary | Controller lifetime; routing and history behavior stay the same |
| Files | `app-browser-entry.ts` supplies factories; `app-browser-runtime-controllers.ts` caches instances |
| Impact | Module evaluation no longer dereferences either imported controller class |

Closes #2620.

## Why

Rolldown can evaluate `index` before `navigation` finishes defining its exports. The browser entry previously constructed both controllers at module scope, read an uninitialized binding, and stopped hydration. In the reporter's first retest, the lazy history controller let evaluation reach the eager `AppBrowserMpaNavigationScheduler`, which threw the same constructor error.

|---|---|---|
| Module evaluation | Cross-chunk class bindings must initialize before use | Stores constructor calls in runtime factories |
| Lifecycle | Each document owns one history controller and one MPA scheduler | Creates each instance on first use, then reuses it |
| Ownership | Cycle-sensitive construction needs a module outside the runtime dependency graph | Caches both instances in a dependency-free module |
| Next.js reference | Next.js creates mutable App Router state inside `hydrate()` | Moves vinext's controller setup from module load to runtime |

## What changed

| Scenario | Before | After |
|---|---|---|
| Module evaluation | Constructed both imported controller classes | Captures their constructors in factory closures |
| History or MPA work | Used module-scoped instances | Gets the matching lazy instance |
| Regression test | Covered the history controller alone | Verifies independent, deferred, once-only creation for both controllers |

<details>
<summary>Validation</summary>

- `vp test run tests/app-browser-runtime-controllers.test.ts tests/app-browser-history-controller.test.ts tests/app-browser-mpa-navigation.test.ts`: 23 tests passed.
- `vp run vinext#build`: passed.
- Cloudflare Workers hydration Playwright test: passed.
- `vp check`: 2,444 files formatted; no lint, warning, or type errors across 1,090 files.
- Pre-commit staged checks, unit/integration tests, full check, and knip: passed.
- Built entry inspection: both constructors remain inside lazy factories.
- External package test: installed the preview in a fresh Cloudflare app. Production build/start, hydration, `next/link`, `router.push`, and `router.replace` passed with no browser or console errors. Rolldown extracted a navigation chunk but not the reverse import; the reporter's exact A/B retest remains pending.

</details>

<details>

- No public API or configuration changes.
- Each controller still has document scope and one stable identity.
- History, MPA navigation, RSC, and Workers behavior remain unchanged after construction.
- Chunk assignment remains unchanged.

</details>

<details>
<summary>Non-goals</summary>

- Reconfigure Rolldown chunk assignment.
- Remove the source-module relationships that produced the cycle.
- Change history or MPA navigation semantics.

</details>

## References

|---|---|
| [Issue #2620](https://github.com/cloudflare/vinext/issues/2620) | Reports the production hydration crash |
| [Reporter follow-up](https://github.com/cloudflare/vinext/pull/2621#issuecomment-4974615354) | Identifies the remaining eager scheduler after the first fix |
| [Emitted artifact bundle](https://gist.github.com/reynard93/7c39a3c859de2ea69f8cb7421630b8ed) | Contains the circular `index` and `navigation` chunks |
| [Next.js hydration source](https://github.com/vercel/next.js/blob/303f7ffd4a0db19948a71eba73cd85f366625a65/packages/next/src/client/app-index.tsx#L346-L386) | Creates mutable App Router state inside `hydrate()` after reading the initial RSC payload |
| [Next.js production hydration test](https://github.com/vercel/next.js/blob/303f7ffd4a0db19948a71eba73cd85f366625a65/test/e2e/app-dir/app/index.test.ts#L1042-L1058) | Covers interactive client-component hydration |
| [Rolldown code-splitting groups](https://rolldown.rs/reference/TypeAlias.CodeSplittingGroup) | Documents the chunking mechanism |


reynard93:
Tested this branch against the app from #2620 via the preview build (`npm i https://pkg.pr.new/vinext@2621`), production build + `vinext start`, loading routes in a real browser.

**Result: the lazy `historyController` change works, but hydration still crashes on every route** — the same circular-chunk pattern has a *second* module-eval-time instantiation this PR doesn't cover.

## What I saw

```
TypeError: AppBrowserMpaNavigationScheduler is not a constructor
    at <module evaluation of entry chunk>  (index-D9OUIRZ7.js:5633)
```

(minified it surfaces as `he is not a constructor`; a `build.minify:false` build gives the real name.)

`getHistoryController()` is now deferred ✔, but `/internal` (a client component) still freezes on "Loading…" and hydration never completes, because the entry chunk still throws while evaluating.

## Cause — a second eager `new` in the same `index ↔ navigation` cycle

In the preview build's `dist/server/app-browser-entry.js`:

- line **74**: `const getHistoryController = createLazyAppBrowserHistoryController(() => new AppBrowserHistoryController({…}))`  ← fixed by this PR ✔
- line **130**: `const mpaNavigationScheduler = new AppBrowserMpaNavigationScheduler();`  ← **still eager** ✗

`AppBrowserMpaNavigationScheduler` is defined in the **navigation** chunk, imported by the **entry/`index`** chunk, which the navigation chunk imports back — the exact `index ↔ navigation` cycle from #2620. The entry chunk runs the top-level `new` before the navigation chunk finishes defining the class → binding is `undefined` → "is not a constructor".

It's only used inside callbacks (`app-browser-entry.js:474` `mpaNavigationScheduler.navigate(…)`, `:1228` `mpaNavigationScheduler.reset()`), so it's lazy-safe with the identical treatment you applied to `historyController`.

I enumerated the entry chunk: `mpaNavigationScheduler` is the **only** remaining top-level `new X()` whose `X` crosses the cycle (the other two module-scope `new`s are `TextEncoder` / `Promise` built-ins — safe). So deferring this one *should* complete the fix for this app, but it'd be worth auditing `app-browser-entry` for any other module-eval-time instantiation of a class reachable across the entry↔navigation boundary. A categorical fix (avoid the top-level side effects entirely, or keep these runtime modules out of a cross-importing chunk split) would close the whole class rather than one `new` at a time.

## Environment

- vinext `1.0.0-beta.1` (also repros on `beta.0`); bundled `vite` = `@voidzero-dev/vite-plus-core@0.2.2`; `vite@8.0.10`, `@vitejs/plugin-rsc@0.5.26`, `@vitejs/plugin-react@6`, `react`/`react-dom@19.2.6`, `@cloudflare/vite-plugin@1.34`; App Router; Cloudflare Workers target; Node 24.
- `vite.config.ts` plugins: `[vinext(), cloudflare({ viteEnvironment: { name: "rsc", childEnvironments: ["ssr"] } })]`.

## Artifacts (the failing **beta.1** prod build you asked for)

Vite manifest + the entry(`index`) and `navigation` chunks that import each other, unminified (`build.minify:false`): https://gist.github.com/reynard93/7c39a3c859de2ea69f8cb7421630b8ed

- `index-Cq6-7B5c.js`: L11 imports `RestorableClientStateController` **and** `AppBrowserMpaNavigationScheduler` from `./navigation-V2InicPx.js`; L5570 `new AppBrowserHistoryController(…)`; L5622 `new AppBrowserMpaNavigationScheduler()`.
- `navigation-V2InicPx.js`: L129 `var RestorableClientStateController = class`; L296 `var AppBrowserMpaNavigationScheduler = class`; L11 imports back from `./index-Cq6-7B5c.js`.

Happy to test a follow-up commit that also defers `mpaNavigationScheduler`, or to put together a from-scratch minimal repro if that's more useful.

NathanDrake2406:
Could you please retest the new commit?

NathanDrake2406:
I also tried to reproduce this independently in a fresh external Cloudflare App Router app using the reported package boundary and several shared next/link and next/navigation client entries. The beta.1 build naturally extracts the navigation runtime and contains both eager constructors, while the PR preview keeps both constructors inside lazy factories. The preview hydrates and completes Link, router.push, and router.replace navigation in a real browser with no page or console errors.

My reconstructed beta graph does not import the entry back from the navigation chunk, though, so it does not produce the original failure and is not an honest failing-before/passing-after reproduction. Could you share the from-scratch minimal repo you offered, or a sanitized copy of the affected app graph? I would like to run beta.1 and commit 70c88de against the exact same app before moving the PR out of draft.

NathanDrake2406:
@codex review

reynard93:
Tested `70c88de` (installed the PR head via `npm i https://pkg.pr.new/vinext@6fe9d36`) against the same app, production build, real browser.

Both deferrals are present in the installed dist and work — `getHistoryController()` / `getMpaNavigationScheduler()` via `createAppBrowserRuntimeControllerAccessors` ✔ — but **hydration still crashes on every route with a third manifestation:**

```
TypeError: ReadonlyURLSearchParams is not a constructor
    at getClientNavigationState  (navigation chunk)   ← new ReadonlyURLSearchParams(window.location.search)
    at setClientParams           (navigation chunk)
    at applyClientParams         (entry chunk)
    at applyRuntimeRscBootstrap  (entry chunk)
    at readInitialRscStream      (entry chunk)
    at main                      (entry chunk)
    at <module top-level>        (entry chunk)        ← dist/server/app-browser-entry.js:1233 `main();`
```

Source pointers: `app-browser-entry.js:1233` calls `main()` at module scope; its synchronous prefix runs `readInitialRscStream → applyRuntimeRscBootstrap → applyClientParams → setClientParams` (from `shims/navigation.js`), which hits `getClientNavigationState()`'s `cachedReadonlySearchParams: new ReadonlyURLSearchParams(...)` (`shims/readonly-url-search-params.js`).

## Why deferring individual `new`s can't converge

With the `entry ↔ navigation` chunk cycle, the entry chunk's body evaluates **before** the navigation chunk's body. At that moment:

- hoisted **function declarations** from the navigation chunk (`setClientParams`, `getClientNavigationState`, …) are fully callable, but
- every `var X = class { … }` in the navigation chunk is still `undefined` (Rolldown emits classes as `var`-assigned, so no TDZ error — just `undefined`).

So *any* code the entry runs at module-eval time — direct `new`s **or any call chain that reaches a navigation-chunk class** — can observe an undefined class. That's three distinct sites across two commits now (`RestorableClientStateController`, `AppBrowserMpaNavigationScheduler`, `ReadonlyURLSearchParams`), and nothing bounds the set: it depends on which modules Rolldown groups into the far side of the cycle for a given app graph.

Two categorical options:

1. **Defer the bootstrap itself out of module evaluation** — e.g. `queueMicrotask(main)` (or equivalent) in `app-browser-entry`. Chunk-graph evaluation completes synchronously before microtasks run, so every `var`-class is assigned by the time `main()` executes, regardless of how chunks got grouped. This closes the whole bug class in one place.
2. **Make the client chunk split acyclic** (or keep the app-browser runtime + navigation shim in one chunk), so no chunk body can run before its class definitions.

## Repro status

A from-scratch minimal app so far does **not** trigger the extraction, matching your reconstruction: I tried client pages, `next/dynamic` sections, a dynamic route, layout/nav/footer client components shared by client pages *and* referenced from a server page, `useRouter`/`useSearchParams`/`useParams`/`usePathname` — the router runtime always colocates into the entry chunk (single dominator), so no back-edge and no crash. The extraction that creates the cycle seems to need a larger/shaped module graph.

Next I'll delta-minimize a sanitized copy of the affected app down to the smallest graph that still produces the `navigation` chunk with the back-edge, and share that repo. Meanwhile the exact failing graph (unminified `index` + `navigation` chunks + Vite manifest, beta.1) is in the gist from my previous comment, and I'm happy to run any candidate commit against the real app on request — turnaround is a few minutes.

reynard93:
Here's the from-scratch minimal repro: **https://github.com/reynard93/vinext-2620-repro** — delta-minimized from the affected app, exact-pinned, `npm install && npm run build && npm run check` prints the cycle; `npm start` + a browser shows the hydration crash.

Minimizing it also surfaced the missing variable that explains why your reconstruction has no back-edge:

## The cycle is a `rolldown@1.0.0-rc.17` chunking behavior, fixed in `rc.18`

Bisected on the same app graph (everything else held constant):

| vite | bundled rolldown | result |
| --- | --- | --- |
| **8.0.10** | **1.0.0-rc.17** | **`index ↔ navigation` cycle → crash** |
| 8.0.11 | 1.0.0-rc.18 | no cycle — the `navigation` extraction doesn't even happen |
| 8.0.13 / 8.0.14 / 8.1.0 / 8.1.5 | ≥ rc.18 | no cycle |

Not the lever (all still cycle on vite 8.0.10): `@vitejs/plugin-rsc` 0.5.26/0.5.27/0.5.28, `@cloudflare/vite-plugin` 1.34.0/1.45.1.

The affected app's lockfile pinned `vite@8.0.10` from an earlier install; your monorepo and any fresh `npm install` resolve `vite@>=8.0.11`, so the reconstructed graph can't produce the cyclic split. Graph shape still matters (the repro needs two `"use client"` pages sharing client components + a `next/dynamic` section + the Cloudflare plugin — a flat app colocates the runtime into the entry and never splits), but the version is the gate.

## Confirmation on the real app

Published `vinext@1.0.0-beta.1` + `vite@8.1.5` (rolldown 1.1.5), minified production build: hydration recovers on every route (the previously-frozen client-component route is interactive), navigation + history traversal work, zero page errors, full unit-test suite green. So the acute production fix for anyone hitting this is simply moving `vite` past 8.0.10.

## Where that leaves this PR, IMO

- The lazy accessors in `70c88de` are good **defense-in-depth**: they remove real module-eval-time side effects, so a future bundler regression or an unlucky chunk grouping can't take the whole client down the same way. The third manifestation I reported above (`ReadonlyURLSearchParams` via the top-level `main()` → `applyRuntimeRscBootstrap` chain) shows per-`new` deferral alone can't fully close the class though — if you want categorical immunity, deferring the bootstrap call itself out of module evaluation (microtask) covers every variant at once.
- Since the trigger is the bundler version, it may also be worth vinext enforcing/nudging `vite >= 8.0.11` (peer range floor, or a build-time warning when the resolved rolldown is `<= 1.0.0-rc.17`), so locked-older apps fail loudly at build time instead of silently shipping a dead client bundle.

Happy to re-run any follow-up commit against both the repro and the real app.

NathanDrake2406:
@james-elicx please let me know if you want to keep this PR as defense in depth (it's also a slight refactor)

PR #2633, opened 2026-07-16
fix: honor custom TypeScript config path

### LINKED ISSUE #1449: Build failure: custom `typescript.tsconfigPath` from `next.config` not honored

## Problem

When `next.config.js` specifies a non-default `typescript.tsconfigPath`, the vinext production build fails entirely with `Custom deploy script failed: undefined (1)`. The build pipeline appears to assume the default `tsconfig.json` and does not forward the user-specified path to the TypeScript / Vite compiler. Affects App Router, Pages Router, and middleware variants of the test fixture.

```
Error: Custom deploy script failed:  undefined (1)
    at NextDeployInstance.deployUsingCustomScript
```

## Estimated Impact

~6 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/tsconfig-path/index.test.ts` (3 failures)
- `test/e2e/typescript-custom-tsconfig/test/index.test.ts` (3 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test fixture with `tsconfig.app.json` and `typescript: { tsconfigPath: './tsconfig.app.json' }` in `next.config.js`. Confirm the build currently fails.

2. **Study how Next.js loads the typescript config.** Search `.nextjs-ref/packages/next/src/build/` for `tsconfigPath` to understand the resolution order and what the path is used for (paths, baseUrl, JSX runtime).

3. **Forward the custom path to Vite/Rolldown.** Read `typescript.tsconfigPath` from the resolved next config and pass it as the `tsconfig` option to Vite's TypeScript handling (or load the file and merge its `compilerOptions.paths`/`baseUrl` into the Vite resolve options).

4. **Check dev/prod parity.** Apply the same resolution in both `server/dev-server.ts` and the production build.

akim136:
## What changed

- preserve `typescript.tsconfigPath` while resolving `next.config`
- materialize path aliases from the explicitly selected TypeScript config
- disable Vite's default `tsconfig.json` auto-discovery when a custom config is selected

## Why

vinext always loaded aliases from the default `tsconfig.json`, even when Next.js configuration selected another file. As a result, production and development resolution could use the wrong `paths` mappings and fail builds that work in Next.js.

The custom path is resolved relative to the project root, matching Next.js configuration semantics. Existing default discovery and explicit Vite `resolve.tsconfigPaths` overrides remain unchanged.

Closes #1449.

## Validation

- `pnpm test tests/tsconfig-paths-vite8.test.ts tests/next-config.test.ts` — 225 tests passed
- `pnpm exec vp check packages/vinext/src/index.ts packages/vinext/src/config/next-config.ts tests/tsconfig-paths-vite8.test.ts` — passed with no warnings or errors
- commit hook `vp check --fix` — passed

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -871,20 +871,26 @@ function sortTsconfigAliasesBySpecificity(aliases: Record<string, string>): Reco
   return Object.fromEntries(Object.entries(aliases).sort((a, b) => b[0].length - a[0].length));
 }
 
-function resolveTsconfigAliases(projectRoot: string): Record<string, string> {
-  if (_tsconfigAliasCache.has(projectRoot)) {
-    return _tsconfigAliasCache.get(projectRoot)!;
+function resolveTsconfigAliases(
+  projectRoot: string,
+  configuredPath?: string,
+): Record<string, string> {
+  const configPath = configuredPath ? path.resolve(projectRoot, configuredPath) : undefined;
+  const cacheKey = configPath ?? projectRoot;
+  if (_tsconfigAliasCache.has(cacheKey)) {
+    return _tsconfigAliasCache.get(cacheKey)!;
   }
 
   let aliases: Record<string, string> = {};
-  for (const name of TSCONFIG_FILES) {
-    const candidate = path.join(projectRoot, name);
+  for (const candidate of configPath
+    ? [configPath]
+    : TSCONFIG_FILES.map((name) => path.join(projectRoot, name))) {
     if (!fs.existsSync(candidate)) continue;
     aliases = sortTsconfigAliasesBySpecificity(loadTsconfigPathAliases(candidate, projectRoot));
     break;
   }
 
-  _tsconfigAliasCache.set(projectRoot, aliases);
+  _tsconfigAliasCache.set(cacheKey, aliases);
   return aliases;
 }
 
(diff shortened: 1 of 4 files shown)


packages/vinext/src/index.ts
```diff
@@ -2013,6 +2007,28 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
...
+        const configuredTsconfigPath = isRecord(nextConfig.typescript)
+          ? typeof nextConfig.typescript.tsconfigPath === "string"
+            ? nextConfig.typescript.tsconfigPath
+            : undefined
+          : undefined;
+        tsconfigPathAliases = resolveTsconfigAliases(root, configuredTsconfigPath);
+        // Native tsconfig discovery always starts from tsconfig.json. For an
+        // explicitly selected file, use the aliases materialized above so the
+        // default config cannot override the user's next.config selection.
+        if (configuredTsconfigPath) shouldEnableNativeTsconfigPaths = false;
```
james-elicx: Ideally we would be able to eventually get rid of our custom tsconfig plugin etc. in favour of the native option in the future. In looks like this would prevent that though? Is there a way we could support the native handling with the custom config?

akim136: I definitely agree on this long-term. The issue is vite's native option currently only accepts a boolean and discovers tsconfig.json so it can't receive Next’s explicit typescript.tsconfigPath. The change keeps native handling for the standard case and uses the custom resolver only for an explicitly selected config. An explicit user Vite setting remains untouched. I've updated the naming and comment and added coverage for that boundary so the fallback can be removed once Vite supports an explicit config path.

james-elicx: i was doing some research and it looks like we dont have an option for non rolldown prod builds at the moment, so dev server seems like it would be a bit of a gap, so can merge this for now. thanks!

PR #2641, opened 2026-07-17
fix(app-router): stream nested loading boundaries

### LINKED ISSUE #1528: App Router: `loading.js` fallback not streamed in initial HTML

## Problem

Pages that have a slow `layout.js`/`page.js` should stream the `loading.js` fallback content in the initial HTML response. vinext serves an empty body for these tests, indicating the Suspense boundary does not flush the loading fallback before the slow content resolves.

```
Expected: "Loading..." Received: ""
```

```
Expected: "Loading layout..." Received: ""
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/app-dir/app/index.test.ts` (2 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a slow async server component with an adjacent `loading.tsx`. Assert the initial response body contains the loading fallback text.

2. **Flush the suspense boundary fallback eagerly.** During streaming SSR, emit each Suspense fallback as soon as the boundary suspends, then patch in the resolved content as it completes. Match Next.js's streaming output shape.

---

Part of #1328.

akim136:
## Summary

- discover `loading.tsx` conventions at every App Router segment, including segments without a `layout.tsx`
- carry each loading module and its loader-tree position through the generated RSC manifest and lazy module loader
- compose ancestor `Suspense` boundaries around child layouts while preserving the existing leaf loading and prefetch behavior
- keep inherited loading boundaries correctly positioned for synthetic parallel-slot routes

## Root cause

App route metadata only retained the leaf route's `loading.tsx`. That works when the page itself suspends, but an ancestor loading convention was absent from the flat route element tree when a nested layout suspended. The response therefore waited for the layout instead of streaming the ancestor fallback.

This change models loading conventions the same way the route graph already models other segment-owned boundaries: as module paths paired with loader-tree positions. Ancestor boundaries are inserted during bottom-up segment composition; the leaf boundary stays on the existing route-level path to avoid changing page error/access ordering, scroll behavior, or link-prefetch shells.

The regression fixtures and assertions follow the nested-layout cases covered by Next.js's App Router E2E suite.

## Validation

- `pnpm test tests/routing.test.ts tests/app-route-module-loader.test.ts tests/entry-templates.test.ts tests/app-page-route-wiring.test.ts` — 226 passed
- `PLAYWRIGHT_PROJECT=app-router pnpm exec playwright test tests/e2e/app-router/loading.spec.ts` — 6 passed
- `pnpm run check` — format, lint, types, Next.js type sync, and public shim checks passed
- `pnpm run build` — passed

Risk class: behavior-changing App Router rendering fix; no public API, dependency, or data changes.

Fixes #1528

### DIFF
--- a/packages/vinext/src/entries/app-browser-entry.ts
+++ b/packages/vinext/src/entries/app-browser-entry.ts
@@ -49,12 +49,12 @@ import ${JSON.stringify(entryPath)};`;
  * emitting the same manifest for hybrid builds — see issue #1526 and
  * `pages-client-entry.ts`.
  */
-export function isLinkPrefetchRoute(route: AppRoute): boolean {
+function isLinkPrefetchRoute(route: AppRoute): boolean {
   if (route.pagePath !== null) return true;
   return route.routePath === null && route.layouts.length > 0;
 }
 
-export function toDocumentOnlyAppRoute(route: AppRoute): VinextLinkPrefetchRoute {
+function toDocumentOnlyAppRoute(route: AppRoute): VinextLinkPrefetchRoute {
   return {
     canPrefetchLoadingShell: false,
     documentOnly: true,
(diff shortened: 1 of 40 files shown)


akim136:
@james-elicx got a fix up that tracks loading boundaries per segment and streams the parent fallback while a nested layout resolves. checks pass locally. Would you mind taking a look?

PR #2650, opened 2026-07-19
test(build): cover extensionless optimized imports

Boyeep:
## What changed

- add a real App Router integration regression for an optimized package barrel that re-exports `./button` while only `button.js` exists
- exercise the existing extensionless absolute rewrite in RSC and SSR development transforms
- verify both the development request and a `createBuilder().buildApp()` production handler render successfully
- keep `packages/vinext/src/plugins/optimize-imports.ts` byte-identical to `main`

## Why

The original implementation change was unnecessary. The existing optimizer intentionally emits an absolute extensionless import, and Vite already applies its environment-aware resolver to that generated import. A real dev and production reproduction confirmed this behavior before the source changes were removed.

This PR now adds coverage only. It introduces no resolver probing, filesystem reads, or other work in the transform hot path.

## Validation

- `vp test run tests/optimize-imports.test.ts tests/optimize-imports-integration.test.ts tests/page-extensions-resolve.test.ts` — 51 passing
- `vp check tests/optimize-imports-integration.test.ts vite.config.ts packages/vinext/src/plugins/optimize-imports.ts tests/optimize-imports.test.ts`
- `vp run vinext#build`
- independent correctness and performance re-review: clean


james-elicx:
Vite should already support this I thought.

Boyeep:
well, you’re right. I dropped the fix and just kept a test so we catch it if this ever breaks.

PR #2663, opened 2026-07-21
fix: client-safe next/constants shim and memoized hot-path dynamic imports

### LINKED ISSUE #2661: next/constants shim throws ReferenceError: process is not defined in client bundles
### Summary

Importing anything from `next/constants` in client code crashes production browser bundles with `ReferenceError: process is not defined`.

The shim (`packages/vinext/src/shims/constants.ts`) has a top level bare reference to `process`:

```js
export const CONFIG_FILES = [
  "next.config.js",
  "next.config.mjs",
  "next.config.ts",
  // process.features can be undefined on Edge runtime
  ...(process?.features?.typescript ? ["next.config.mts"] : []),
];
```

Optional chaining does not guard an undeclared identifier, so evaluating the module in a browser throws. Real Next.js has the same expression in its constants module but survives because webpack injects a `process` polyfill into client bundles. Vite does not polyfill `process`.

Since `next/constants` is aliased for every environment (public-shim-map applies to the shared `resolve.alias`), the shim is a legitimate client import. `@sentry/nextjs` for example imports `PHASE_PRODUCTION_BUILD` in code that ends up in the client graph. And because the shim lands in the shared `vinext-*` client manual chunk, the failed evaluation takes down that whole runtime chunk, not just the importer. In our case the visible symptom was Sentry's client SDK never initializing in production.

### Reproduction

Add a `"use client"` page that imports `PHASE_PRODUCTION_BUILD` from `next/constants` to the `app-basic` fixture, build for production, open in a real browser: the page logs `ReferenceError: process is not defined` (we saw 3 page errors from the shared chunk).

### Suggested fix

```js
...(typeof process !== "undefined" && process.features?.typescript ? ["next.config.mts"] : []),
```

Behavior is unchanged in any runtime that has `process`, and the module becomes safe to evaluate in the browser. Verified locally: 0 page errors after the change, same rendered output.

Happy to send a PR. Thanks for vinext, it has been great to run Next apps on Vite!

### LINKED ISSUE #2662: Prod server re-runs dynamic import() on every request, expensive with ESM loader hooks
### Summary

The production server re-runs several dynamic `import()` calls on every request. Node caches the modules, but each `import()` still pays ESM resolution, and when ESM loader hooks are registered (`@sentry/node` or OpenTelemetry via `module.register()`) every resolution becomes a synchronous round trip to the loader hooks thread (`makeSyncRequest` in `node:internal/modules/esm/hooks`). Profiling a production app with Sentry instrumentation, this path alone accounted for about 17% of process CPU.

### Measurements

Counted with a `diagnostics_channel.tracingChannel("module.import")` subscriber inside the prod server of the `app-basic` fixture, 30 iterations over 4 routes (120 requests):

| module | imports |
|---|---|
| `config/config-matchers.js` (app-rsc-handler) | 720 (6 per request) |
| `config-headers.js` | 150 |
| `metadata-route-response` thunk (runs for every request, not only metadata routes) | 120 |
| `app-page-cache.js` (app-page-dispatch) | 60 |
| `file-based-metadata` thunk | 60 |
| `loadModule("ssr", "index")` | 60 |
| `app-route-handler-dispatch` thunk | 30 |
| `headers.js` (via `connection()` in the server shim) | 30 |
| total | about 10 per request |

### Suggested fix

Cache the import promises. Two cases seem worth distinguishing:

1. Internal framework modules (`headers.js`, `config-matchers`, `config-headers`, `app-page-cache`, the generated entry thunks, `react-dom/static.edge`): memoize unconditionally at module scope. The targets are framework code, never user code, so this is safe in dev too (when the importer is invalidated, its module scope cache resets with it).
2. `import.meta.viteRsc.loadModule("ssr" | "rsc", "index")`: the target bundles user code, so memoize only when `process.env.NODE_ENV === "production"` to keep the per call runner import that dev HMR relies on. `NODE_ENV` matches the dev detection these files already use.

With this change the same measurement drops from about 10 imports per request to 0, and the prod, dev/HMR and pages router test suites stay green.

Happy to send a PR with the change. Thanks for the great project!

jlucaso1:
Closes #2661
Closes #2662

Two related production fixes, one commit each. Happy to split into separate PRs if you prefer.

### 1. `fix(shims)`: guard the `process` reference in `next/constants`

The shim is a valid client import (it is aliased for every environment, and `@sentry/nextjs` for example reaches it from client code), but `CONFIG_FILES` referenced `process` at module top level. Browsers have no `process` and optional chaining does not guard an undeclared identifier, so evaluating the module threw `ReferenceError: process is not defined` and took down the shared `vinext-*` client chunk. Real Next.js has the same expression but is saved by webpack's `process` polyfill, which Vite does not inject. Fixed with a `typeof process` check, identical behavior on any runtime that has `process`.

Verified with a `"use client"` page importing `PHASE_PRODUCTION_BUILD` in the `app-basic` fixture, production build, real Chrome: 3 page errors before, 0 after, same rendered output.

### 2. `perf(server)`: memoize hot-path dynamic imports in production

Several `import()` calls run on every request. The modules are cached by Node, but each call re-runs ESM resolution, and with loader hooks registered (Sentry, OpenTelemetry via `module.register()`) every resolution is a synchronous round trip to the hooks thread. On a production app with Sentry instrumentation this path alone was about 17% of process CPU.

Counted with `diagnostics_channel.tracingChannel("module.import")` inside the prod server of `app-basic`, 120 requests across 4 routes:

| module | before | after |
|---|---|---|
| `config-matchers` | 720 | 0 |
| `config-headers` | 150 | 0 |
| `metadata-route-response` thunk | 120 | 0 |
| `app-page-cache` | 60 | 0 |
| `file-based-metadata` thunk | 60 | 0 |
| `loadModule("ssr", "index")` | 60 | 0 |
| `app-route-handler-dispatch` thunk | 30 | 0 |
| `headers.js` via `connection()` | 30 | 0 |
| total | 1230 (~10 per request) | 0 |

Internal framework modules are memoized unconditionally (safe in dev: when the importer is invalidated its module scope cache resets with it). The `import.meta.viteRsc.loadModule("ssr" | "rsc", "index")` sites bundle user code, so those are memoized only when `process.env.NODE_ENV === "production"`, keeping the per call runner import that dev HMR relies on. `headers.js` stays a dynamic import on purpose (the existing `use cache` constraint), only the promise is cached.

### Tests

`entry-templates` (codegen assert updated accordingly), `shims`, `app-router-production-server`, `app-router-production-build`, `app-flight-framing-production`, `app-router-dev-server`, `dev-route-discovery`, `vite-hmr-websocket`, `app-page-dispatch`, `app-page-render`, `metadata-route-build-data` and `pages-router` suites pass, and `tsc` is clean.

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -270,7 +270,7 @@ export function generateRscEntry(
   const loadPrerenderPagesRoutesCode = hasPagesDir
     ? `
 async function __loadPrerenderPagesRoutes() {
-  const __gspSsrEntry = await import.meta.viteRsc.loadModule("ssr", "index");
+  const __gspSsrEntry = await __loadSsrModule();
   return __gspSsrEntry.pageRoutes;
 }
 `
@@ -330,20 +330,31 @@ ${
 } from ${JSON.stringify(appRouteHandlerResponsePath)};`
     : ""
 }
-const __loadAppRouteHandlerDispatch = () => import(${JSON.stringify(appRouteHandlerDispatchPath)});
+const __memoizeLoad = (load) => {
+  let promise;
+  return () => (promise ??= load());
+};
+let __ssrModulePromise;
+function __loadSsrModule() {
+  if (process.env.NODE_ENV !== "production") {
+    return import.meta.viteRsc.loadModule("ssr", "index");
+  }
+  return (__ssrModulePromise ??= import.meta.viteRsc.loadModule("ssr", "index"));
+}
+const __loadAppRouteHandlerDispatch = __memoizeLoad(() => import(${JSON.stringify(appRouteHandlerDispatchPath)}));
 ${
   hasServerActions
-    ? `const __loadAppServerActionExecution = () => import(${JSON.stringify(appServerActionExecutionPath)});`
+    ? `const __loadAppServerActionExecution = __memoizeLoad(() => import(${JSON.stringify(appServerActionExecutionPath)}));`
     : ""
 }
 ${
   (metadataRoutes?.length ?? 0) > 0
-    ? `const __loadMetadataRouteResponse = () => import(${JSON.stringify(metadataRouteResponsePath)});`
+    ? `const __loadMetadataRouteResponse = __memoizeLoad(() => import(${JSON.stringify(metadataRouteResponsePath)}));`
     : ""
 }
 ${
   (metadataRoutes?.length ?? 0) > 0
-    ? `const __loadFileBasedMetadata = () => import(${JSON.stringify(fileBasedMetadataPath)});
+    ? `const __loadFileBasedMetadata = __memoizeLoad(() => import(${JSON.stringify(fileBasedMetadataPath)}));
 async function __applyFileBasedMetadata(...args) {
   const { applyFileBasedMetadata } = await __loadFileBasedMetadata();
   return applyFileBasedMetadata(...args);
@@ -592,7 +603,7 @@ const __fallbackRenderer = __createAppFallbackRenderer({
   globalNotFoundEnabled: ${config?.globalNotFound === true},
   metadataRoutes,
   ssrLoader() {
-    return import.meta.viteRsc.loadModule("ssr", "index");
+    return __loadSsrModule();
   },
   fontProviders: {
     buildFontLinkHeader: __buildAppPageFontLinkHeader,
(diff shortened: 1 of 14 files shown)


james-elicx:
Looks like there's a regression in bundle size.

jlucaso1:
> Looks like there's a regression in bundle size.

Nice catch, I've submitted this commit to mitigate that: https://github.com/cloudflare/vinext/pull/2663/changes/c796f39890c22151b6da33a740c437d780699044

Also rebased the PR. Thank you

PR #2668, opened 2026-07-21
fix(pages): normalize decoded edge responses

james-elicx:
## Summary

- strip `content-length`, `content-encoding`, and `transfer-encoding` after Node decodes a Pages Edge API response
- apply the boundary in both Vite dev and the Node production server, matching the Next.js edge sandbox
- leave Cloudflare Worker responses untouched so encoded bodies and their headers continue streaming together

## Failure

Actions run `29871622126`, job `88775791401`:

- `test/e2e/edge-compiler-can-import-blob-assets/index.test.ts` — `allows to fetch a remote URL`
- `test/e2e/edge-compiler-can-import-blob-assets/index.test.ts` — `allows to fetch a remote URL with a path and basename`

Both returned raw gzip bytes instead of `Example Domain`.

The three imported text/image/node_modules asset failures are intentionally out of scope. PR #1833 owns that build/compiler lane; its diff does not touch runtime response decoding.

## Root cause

Node fetch decodes compressed response bodies but retains the origin encoding metadata. Vinext forwarded those stale headers through an Edge API response; the production response buffer then recompressed the already-decoded body while retaining the old encoding metadata. Clients therefore received an invalid encoding chain and exposed raw gzip bytes.

Next.js removes the three wire-level body headers in `packages/next/src/server/web/sandbox/sandbox.ts`. Vinext now applies that same rule only for Node-hosted Edge API execution. The Worker path keeps the original response object: Workers require an encoded body and `Content-Encoding` to remain paired, and direct streaming avoids buffering.

## Validation

Pinned Next.js oracle: `v16.2.6` at `ee6e79b1792a4d401ddf2480f40a83549fe8e722`.

```bash
REPO="/private/tmp/vinext-worker-fetch-content-encoding" \
NEXTJS_DIR="/Users/jamesanderson/Developer/vinext/.nextjs-ref" \
./scripts/run-targeted-nextjs-e2e.sh test/e2e/edge-compiler-can-import-blob-assets/index.test.ts
```

Result:

- 2/2 scoped remote-fetch assertions pass
- 3/3 out-of-scope imported-asset compiler assertions remain red

Focused vinext validation:

- `vp test run tests/pages-api-route.test.ts tests/api-handler.test.ts tests/deploy.test.ts tests/entry-templates.test.ts` — 451 passed
- targeted `vp check` for all 9 changed files — clean
- `vp run vinext#build` — passed as part of the exact targeted wrapper

Cloudflare runtime references:

- https://developers.cloudflare.com/workers/runtime-apis/response/#the-encodebody-option
- https://developers.cloudflare.com/workers/best-practices/workers-best-practices/#stream-request-and-response-bodies

Review and Bonk have not been requested yet.

### DIFF
--- a/packages/vinext/src/entries/pages-server-entry.ts
+++ b/packages/vinext/src/entries/pages-server-entry.ts
@@ -467,11 +467,12 @@ export async function renderPage(request, url, manifest, ctx, middlewareHeaders,
 
 
 
-export async function handleApiRoute(request, url, ctx, trustedRevalidateOrigin) {
+export async function handleApiRoute(request, url, ctx, trustedRevalidateOrigin, edgeRuntime = "worker") {
   __registerConfiguredCacheAdapters();
   const match = matchRoute(url, apiRoutes);
   return __handlePagesApiRoute({
     ctx,
+    edgeRuntime,
     match,
     nextConfig: vinextConfig,
     request,
--- a/packages/vinext/src/server/api-handler.ts
+++ b/packages/vinext/src/server/api-handler.ts
@@ -20,7 +20,7 @@ import {
   urlQueryToSearchParams,
 } from "../utils/query.js";
 import { PagesBodyParseError, getMediaType, isJsonMediaType } from "./pages-media-type.js";
-import { isEdgeApiRuntime } from "./edge-api-runtime.js";
+import { finalizeEdgeApiResponse, isEdgeApiRuntime } from "./edge-api-runtime.js";
 import {
   DEFAULT_PAGES_API_BODY_SIZE_LIMIT,
   resolveBodyParserConfig,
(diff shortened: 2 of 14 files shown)


james-elicx:
Addressed the independent review findings in 5b5d5d0d18bb5af8df41805c86a0dedd163aa913. The request execution context now carries the actual host runtime: vinext start stamps node, the Worker wrapper stamps worker, and app-pages-bridge forwards it through the PagesEntry handleApiRoute contract. Added focused hybrid coverage proving Node strips content-encoding, content-length, and transfer-encoding after decoding while Workers preserve the encoded body/header pair. Updated tests/after-deploy.test.ts for the explicit worker argument. Proof: the six-file review suite passed 482/482; focused app-pages-bridge plus after-deploy passed 31/31 after final assertions; scoped vp check is clean across 14 files; vp run vinext#build passed; the pinned Next.js v16.2.6 targeted edge-compiler-can-import-blob-assets suite still passes both remote-fetch assertions, with only the same three explicitly excluded imported-asset compiler assertions failing. PR remains draft; Bonk was not invoked.

PR #2677, opened 2026-07-23
perf(app): use `@vitejs/plugin-rsc/rsc/*` entries

hi-ogawa:
This PR updates the `@vitejs/plugin-rsc` runtime entries consumed by vinext.

Vinext previously imported internal-ish `/core/*`, `/react/*`, and `/vendor/*` entries. These are now replaced and consolidated to public [`/rsc/*` runtime API family](https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-rsc/README.md#rsc-runtime-react-server-dom-api).

Specifically:
- Use `/rsc/static`, released in `@vitejs/plugin-rsc@0.5.29`, instead of deep import of `/vendor/` static API.
- Use the slimmer `/rsc/server` entry for the main RSC runtime because it only consumes server-side APIs.
- Lazily load `/rsc/server` and `/rsc/client` separately for the `"use cache"` runtime.

The server/client split entries are a part of perf works by @TheAlexLichter (Thank you!). The background context is found in https://github.com/cloudflare/vinext/pull/2080 and https://github.com/vitejs/vite-plugin-react/pull/1252.

The public static API was added in https://github.com/vitejs/vite-plugin-react/pull/1285 with minimal example to exercise the PPR concept.

This is not by no means against Vinext using non-public API forever, but I think, from time to time, it's good to catch up with what API Vinext uses and make `@vitejs/plugin-rsc` align the runtime API to prove some features are framework agnostic concept. If there's a future occasion to reach for core/vendor API, please feel free to do so 😃 

---

For perf number, the agent ran `benchmarks/perf/run-scenarios.mjs` script locally and claimed a following (I don't yet understand what the script does):

Ran six alternating base/head rounds against `main` using the generated 33-route benchmark app with framework caches cleared before each round.

| Metric | `main` | PR | Change |
|---|---:|---:|---:|
| Dev cold start, mean | 1280.7 ms | 1190.5 ms | **-90.2 ms (-7.0%)** |
| Dev cold start, median | 1283.7 ms | 1181.4 ms | **-102.3 ms (-8.0%)** |
| Server bundle, gzip | 181,712 B | 181,444 B | -268 B |

The PR was faster in all six cold-start rounds. Production build time was noisy and showed no meaningful change. These are local macOS results; the PR performance workflow will provide the authoritative Linux CI comparison.

### DIFF
--- a/packages/vinext/package.json
+++ b/packages/vinext/package.json
@@ -194,7 +194,7 @@
   "peerDependencies": {
     "@mdx-js/rollup": "^3.0.0",
     "@vitejs/plugin-react": "^5.1.4 || ^6.0.0",
-    "@vitejs/plugin-rsc": "^0.5.26",
+    "@vitejs/plugin-rsc": "^0.5.29",
     "react": "^19.2.6",
     "react-dom": "^19.2.6",
     "react-server-dom-webpack": "^19.2.6",
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -289,17 +289,12 @@ import {
   createTemporaryReferenceSet,`
       : ""
   }
-} from ${JSON.stringify(
-    hasServerActions ? "@vitejs/plugin-rsc/rsc" : "@vitejs/plugin-rsc/react/rsc",
-  )};
-import { createClientManifest as _createClientManifest } from "@vitejs/plugin-rsc/core/rsc";
-import { prerender as _prerender } from "@vitejs/plugin-rsc/vendor/react-server-dom/static.edge";
+} from "@vitejs/plugin-rsc/rsc/server";
+import { prerender as _prerender } from "@vitejs/plugin-rsc/rsc/static";
 import { createRscPrerenderer, createRscRenderer } from ${JSON.stringify(rscStreamHintsPath)};
 
 const renderToReadableStream = createRscRenderer(_renderToReadableStream);
-const prerenderToReadableStream = createRscPrerenderer(async (model, options) =>
-  _prerender(model, _createClientManifest(), options),
-);
+const prerenderToReadableStream = createRscPrerenderer(_prerender);
 import { createElement } from "react";
 import { getNavigationContext as _getNavigationContext } from "next/navigation";
 import { configureMemoryCacheHandler as __configureMemoryCacheHandler } from "vinext/shims/cache-handler";
--- a/packages/vinext/src/shims/cache-runtime.ts
+++ b/packages/vinext/src/shims/cache-runtime.ts
@@ -167,7 +167,7 @@ export function getCacheContext(): CacheContext | null {
 // ---------------------------------------------------------------------------
 
 /**
- * RSC serialization APIs from @vitejs/plugin-rsc/react/rsc.
+ * RSC serialization APIs from @vitejs/plugin-rsc/rsc/server and /rsc/client.
  * Lazily loaded because these are only available in the Vite RSC environment
  * (they depend on virtual modules set up by @vitejs/plugin-rsc).
  * In test environments, the import fails and we fall back to JSON.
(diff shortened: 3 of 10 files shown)


packages/vinext/src/entries/app-rsc-entry.ts
```diff
@@ -289,17 +289,12 @@ import {
   createTemporaryReferenceSet,`
       : ""
   }
-} from ${JSON.stringify(
-    hasServerActions ? "@vitejs/plugin-rsc/rsc" : "@vitejs/plugin-rsc/react/rsc",
-  )};
-import { createClientManifest as _createClientManifest } from "@vitejs/plugin-rsc/core/rsc";
-import { prerender as _prerender } from "@vitejs/plugin-rsc/vendor/react-server-dom/static.edge";
+} from "@vitejs/plugin-rsc/rsc/server";
+import { prerender as _prerender } from "@vitejs/plugin-rsc/rsc/static";
```
hi-ogawa, line 293:
```js
hasServerActions ? "@vitejs/plugin-rsc/rsc" : "@vitejs/plugin-rsc/react/rsc"
```

This `hasServerActions` branch is introduced by https://github.com/cloudflare/vinext/pull/2206, but this shouldn't matter since tree shaking server action API is mostly done by the import branch above:

https://github.com/cloudflare/vinext/blob/f91363f3b3aed8a150a79f41a7d95b36cb87bfb3/packages/vinext/src/entries/app-rsc-entry.ts#L281-L291

So this has been consolidated to just `/rsc/server` import unconditionally.

packages/vinext/package.json
```diff
@@ -194,7 +194,7 @@
   "peerDependencies": {
     "@mdx-js/rollup": "^3.0.0",
     "@vitejs/plugin-react": "^5.1.4 || ^6.0.0",
-    "@vitejs/plugin-rsc": "^0.5.26",
+    "@vitejs/plugin-rsc": "^0.5.29",
```
hi-ogawa, line 197: I'm not sure if peer dep bump should be breaking change. From the past commit history, there doesn't seem to be strict policy.

NathanDrake2406:
@james-elicx BOOOOOM! PPR time baby

packages/vinext/package.json
james-elicx: The other times i've switched to new Vite related functionality, i've left the old behaviour behind a version check to avoid breaking changes. Will probably have to see if the same thing can be done for this too. I can have a look at that no worries

packages/vinext/src/shims/cache-runtime.ts
```diff
@@ -232,7 +232,11 @@ let _rscModule: RscModule | null | typeof NOT_LOADED = NOT_LOADED;
 async function getRscModule(): Promise<RscModule | null> {
   if (_rscModule !== NOT_LOADED) return _rscModule;
   try {
-    _rscModule = (await import("@vitejs/plugin-rsc/react/rsc")) as RscModule;
+    const [server, client] = await Promise.all([
+      import("@vitejs/plugin-rsc/rsc/server"),
+      import("@vitejs/plugin-rsc/rsc/client"),
+    ]);
```
hi-ogawa, line 238: It looks like using `/rsc/*` here breaks for pages router setup because no rsc plugin means no virtual module used by `/rsc/*` entry.

I think ideally page router only mode can entirely nullifying `packages/vinext/src/shims/cache-runtime.ts` (and other related app router only modules). Let me check a bit.

hi-ogawa: Attempting to remove "cache-runtime" import from pages router ssr entry https://github.com/cloudflare/vinext/pull/2677/changes/95d98620eeb4aaf9e5e8a2608c4446093c8c0624. The existing test appears to be asserting internal request context, which has no user facing pages router feature, so it's adjusted accordingly.

hi-ogawa:
Sorry, I haven't had a time to finish it through. Let me closed this for now.

PR #2681, opened 2026-07-23
fix(pnpm-workspace): disable global virtual store option

314systems:
## Summary

Disable pnpm's global virtual store for this project.

When `enableGlobalVirtualStore` is enabled, Vite+ fails to resolve its native Rolldown binding and exits with the following error:

```text
Error: Cannot find native binding
Cannot find module 'vite-plus/binding'
Cannot find module './rolldown-binding.darwin-arm64.node'
```

Setting `enableGlobalVirtualStore` to `false` restores the project-local virtual store layout and allows Vite+ to resolve the native binding correctly.

## Changes

* Set `enableGlobalVirtualStore: false` in `pnpm-workspace.yaml`

## Verification

1. Removed the existing `node_modules` directory
2. Reinstalled dependencies with `pnpm install`
3. Ran the command that previously failed
4. Confirmed that Vite+ starts without the native binding error

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -4,6 +4,7 @@ preferFrozenLockfile: true
 saveExact: true
 blockExoticSubdeps: true
 minimumReleaseAge: 1440
+enableGlobalVirtualStore: false
 
 packages:
   - apps/*


james-elicx:
The docs suggest this already defaults to false?
https://pnpm.io/settings#enableglobalvirtualstore

314systems:
In my environment, I use `enableGlobalVirtualStore` set to `true` in the global configuration (`config.yaml`). However, due to compatibility issues in this project, I suggest disabling `enableGlobalVirtualStore` in `pnpm-workspace.yaml`.

james-elicx:
That doesn't feel like a compelling reason to make changes to the config in this repo if the issue is the global config on your machine. Further, I would encourage seeing how you can get your global config to work with vite-plus, as it is quickly becoming the default choice for many projects.

314systems:
https://github.com/voidzero-dev/vite-plus/issues/2054

PR #2686, opened 2026-07-23
fix(rsc): handle Windows paths in client reference dedup

MaxtuneLee:
## Summary

- normalize incoming module IDs in the client reference dedup plugin
- replace the POSIX-only absolute-path check with a platform-aware check
- add coverage for native Windows `node_modules` paths
- related to #1605 

## Test plans

Added a Windows-only regression case.
tests/client-reference-dedup.test.ts all green.


shulaoda (approved):
LGTM

PR #2687, opened 2026-07-23
fix(build): normalize jsx-in-js module ids on Windows

MaxtuneLee:
## Summary

- Normalize Vite-reported `.js` and `.mjs` module IDs after removing their query string.
- Preserve the directive-only JSX transform policy for dependencies whose Windows paths use native backslashes.
- Add a Windows-only regression test for a non-directive module inside `node_modules`.
- related to #1605

## Problem

The `vinext:jsx-in-js` transform uses the module ID to distinguish project source from dependencies. Dependencies are intentionally transformed only when they contain a real `"use client"` or `"use server"` directive.

Vite can report native Windows IDs such as:

```text
C:\project\node_modules\dependency\component.js
```

The existing `cleanId.includes("/node_modules/")` check only recognized forward-slash paths. A native Windows ID therefore bypassed the dependency gate, and vinext transformed an ordinary dependency as JSX even when it had no React directive. This widened the transform beyond its intended scope and could add unnecessary work or parse third-party JavaScript with the wrong language mode.

## Fix

Normalize the bundler-provided ID with `toSlash(stripViteModuleQuery(id))` before directory and module-ID comparisons. This keeps query removal and path normalization at the external input boundary while leaving the existing JSX transform behavior unchanged.

## Test plan

- Added a Windows-only regression test
- `vp test run tests/jsx-in-js-node-modules.test.ts` — all passed.
- `vp test run tests/jsx-in-js.test.ts` — all passed.


tests/jsx-in-js-node-modules.test.ts
```diff
@@ -59,6 +59,25 @@ async function buildApp(root: string) {
...
+
+      const transform = plugin!.transform as {
+        handler(code: string, id: string): Promise<unknown>;
+      };
+      const dependencyId = path.join("C:\\project", "node_modules", "dependency", "component.js");
+
+      await expect(
+        transform.handler("export default function Component() { return <div />; }", dependencyId),
+      ).resolves.toBeUndefined();
+    },
```
shulaoda, line 78: If it's handled correctly, shouldn't it return something like `{ code: ..., map: ... }` instead of `undefined`?

MaxtuneLee: The `vinext:jsx-in-js` plugin intentionally transforms dependency files only when they contain a real React directive. Therefore, it reaches a bare return, which produces undefined.

Before the fix, the native Windows path did not match `"/node_modules/"`, so the plugin incorrectly transformed the file and returned compiled code. After normalization, it shall correctly recognizes the dependency and returns undefined.

shulaoda: Got it. My bad, I mistakenly thought `<div />` was a directive.

shulaoda (approved):
LGTM

shulaoda:
Normally, module IDs resolved through Vite's internal resolve plugin have already been slash-normalized.

Some IDs used internally by Vinext may be resolved and returned by Vinext's own plugins via `resolveId`, and we generally normalize the path separators there as well.

However, there are still cases we can't fully control. For example, a third-party plugin may handle `resolveId` and return an ID with non-normalized path separators. So the normalization here is mainly a fallback to cover those cases.

PR #2691, opened 2026-07-24
fix(css): load Sass partials from paths containing tildes

### LINKED ISSUE #2690: Sass CSS URL asset importer fails when an imported partial path contains `~`
## Description

`createSassCssUrlAssetImporter()` fails to load an imported Sass partial when any segment of the partial's filesystem path contains `~`. Such as `C:\Users\RUNNER~1\...` in [Github Action Runner](https://github.com/cloudflare/vinext/actions/runs/29840986614/job/88669534942?pr=1792) are a common trigger, but the issue is not Windows-specific and can also occur on Linux or macOS when a directory name contains a literal `~`.

The importer reports that it cannot find the stylesheet even though the partial was found, read, prepared, and registered during `rewriteImports()`.

## Error

```text
Can't find stylesheet to import.
  ╷
1 │ @use "vinext-css-url-asset:C%3A%2FUsers%2FRUNNER~1%2FAppData%2FLocal%2FTemp%2Fvinext-sass-namespace-url-8Z2xql%2F_my_colors.scss" as my_colors;
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  - 1:1  root stylesheet
```

## Reproduction conditions

The failure occurs when all of the following conditions are met:

- Any segment of the filesystem path contains `~`.
- The Sass source is rewritten by `createSassCssUrlAssetImporter()`.
- The imported partial contains a CSS `url()` whose source provenance must be preserved.
- Sass calls `load()`, and vinext looks up the prepared stylesheet using its serialized `.href`.

## Root cause

The importer stores prepared stylesheets in `markedStylesheets` using `pathToFileURL(candidate).href`. Node serializes a path containing `~` using `%7E`, so the stored key looks like this:

```text
file:///tmp/path%7Ewith%7Etilde/_card.scss
```

When the canonical URL passes through Sass's internal URI representation, Sass normalizes `%7E` back to the equivalent unreserved character `~` before calling `load()`:

```text
file:///tmp/path~with~tilde/_card.scss
```

These URLs identify the same file, but their serialized `.href` strings are different. As a result, `markedStylesheets.get(canonicalUrl.href)` returns `undefined`, and Sass reports that it cannot find the stylesheet.

## Expected behavior

The importer should load the prepared stylesheet regardless of whether an equivalent file URL serializes the tilde as `%7E` or `~`.

## Actual behavior

The importer compares serialized URL strings, fails to find the prepared stylesheet, and returns `null` from `load()`.

## Notes

An existing namespace test can expose this issue on GitHub Actions because the Windows runner temporary directory may use an 8.3 short path such as `C:\Users\RUNNER~1\...`. That coverage is environment-dependent: the same test passes when `os.tmpdir()` returns a path without `~`.

A focused regression test should construct a tilde-containing path explicitly rather than relying on the runner's temporary directory configuration.

## Suggested direction

Use a normalized filesystem path as the stylesheet map key instead of the serialized URL `.href`. Converting canonical file URLs with `fileURLToPath()` and normalizing the resulting external path with `toSlash()` would make `%7E` and `~` resolve to the same key.

MaxtuneLee:
## Summary

- key Sass importer stylesheets by normalized filesystem paths instead of serialized URL strings
- preserve stylesheet lookup when Sass normalizes equivalent file URLs
- add a focused regression test using a path containing `~`
- related to #1792
- fix #2690 

## Problem

The Sass CSS URL asset importer stored prepared stylesheets using `pathToFileURL(candidate).href`.

Node serializes a filesystem path containing `~` as `%7E`:

```text
file:///C:/Users/RUNNER%7E1/.../_card.scss
```

When the canonical URL passes through Sass's internal URI representation, Sass normalizes `%7E` back to the equivalent unreserved character `~` before calling `load()`:

```text
file:///C:/Users/RUNNER~1/.../_card.scss
```

Although these URLs identify the same file, comparing their serialized `href` strings caused the stylesheet map lookup to fail. Sass then reported:

```text
Can't find stylesheet to import.
```

`RUNNER~1` are a common way to encounter this problem, but the issue is not Windows-specific. A path containing `~` on Linux or macOS goes through the same URL normalization.

## Fix

Convert canonical file URLs back to filesystem paths with `fileURLToPath()` and normalize the resulting external path with `toSlash()` before using it as the stylesheet map key.

This makes the key independent of equivalent URL serialization forms such as `%7E` and `~`, while retaining vinext's canonical forward-slash path format.

## Testing

Added a regression test that:

1. creates a Sass partial under a path containing `~`
2. rewrites an `@use` rule through the CSS URL asset importer
3. compiles the rewritten stylesheet with Sass
4. verifies that the normalized canonical URL still resolves to the prepared
   stylesheet

An existing namespace test happened to expose this failure on GitHub Actions because that runner's temporary directory used the Windows short path `C:\Users\RUNNER~1\...`. However, that coverage depended on the runner's environment. The same test passed locally when `os.tmpdir()` returned a path without `~`, and it would stop covering this regression if the runner's
temporary directory changed.

The focused regression test constructs the tilde-containing path explicitly, making the coverage deterministic. It is kept separate from the existing namespace test because the two tests cover different behavior:

- the existing test verifies that underscores are preserved in Sass default namespaces
- the new test verifies stylesheet lookup across equivalent canonical URL serializations

Keeping these concerns separate also makes future failures easier to diagnose.

Commands run:

- `vp test run tests/sass-options.test.ts`
- `vp check tests/sass-options.test.ts packages/vinext/src/plugins/sass.ts`
- `vp test run tests/sass-options.test.ts -t "loads partials when Sass normalizes Windows short-path URLs"`

All checks pass.

### DIFF
--- a/packages/vinext/src/plugins/sass.ts
+++ b/packages/vinext/src/plugins/sass.ts
@@ -179,7 +179,6 @@ export function createSassCssUrlAssetImporter(): {
 } {
   const markedStylesheets = new Map<string, SassLoadedStylesheet>();
   const importUrlPrefix = "vinext-css-url-asset:";
-  const stylesheetKey = (url: URL): string => toSlash(fileURLToPath(url));
 
   const prepareStylesheet = (
     filename: string,
@@ -235,7 +234,7 @@ export function createSassCssUrlAssetImporter(): {
     },
 
     load(canonicalUrl) {
-      return markedStylesheets.get(stylesheetKey(canonicalUrl)) ?? null;
+      return markedStylesheets.get(toSlash(fileURLToPath(canonicalUrl))) ?? null;
     },
 
     rewriteImports,
--- a/packages/vinext/src/plugins/sass.ts
+++ b/packages/vinext/src/plugins/sass.ts
@@ -217,8 +217,7 @@ export function createSassCssUrlAssetImporter(): {
           if (!fs.statSync(candidate, { throwIfNoEntry: false })?.isFile()) continue;
           const prepared = prepareStylesheet(candidate, entryDirectory);
           if (prepared === null) return match;
-          const canonicalUrl = pathToFileURL(candidate);
-          markedStylesheets.set(stylesheetKey(canonicalUrl), prepared);
+          markedStylesheets.set(candidate, prepared);
           const namespace =
             rule === "use" && !/\bas\s+(?:\*|[-\w]+)/.test(suffix)
               ? ` as ${deriveSassNamespace(importUrl)}`
--- a/packages/vinext/src/plugins/sass.ts
+++ b/packages/vinext/src/plugins/sass.ts
@@ -25,7 +25,7 @@
 import fs from "node:fs";
 import { createRequire } from "node:module";
 import path, { toSlash } from "pathslash";
-import { pathToFileURL } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import { preprocessCSS, type PreprocessCSSResult, type ResolvedConfig } from "vite";
 import { markCssUrlAssetReferences, rebaseCssUrlAssetReferences } from "../build/css-url-assets.js";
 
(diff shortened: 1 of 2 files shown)


shulaoda (approved):
LGTM

PR #2702, opened 2026-07-24
fix(cloudflare): correct TPR's zone-analytics query and export its selection helpers

kHorozhanov:
## What

This fixes the Traffic-aware Pre-Rendering analytics query and exports its reusable traffic-selection helpers.

- Use the current `httpRequestsAdaptiveGroups` schema: `count_DESC` plus `count`.
- Bind zone, time, hostname, and optional route path through GraphQL variables.
- Scope analytics to the deployed hostname and path, honor explicit Wrangler `zone_id` / `zone_name`, and use that zone's account.
- Exclude API, framework, RSC, asset, and error rows before the 10,000-row ranked limit; retain defensive client filtering.
- Skip TPR explicitly for multiple traffic scopes, wildcard-only hosts, or scheme-qualified routes rather than reporting partial coverage.
- Export `queryTraffic`, `filterTrafficPaths`, `selectRoutes`, `resolveZoneId`, and their result types from `@vinext/cloudflare/internal/tpr`.

## Why

The previous query requested `sum_requests_DESC` and `sum { requests }`, which are not fields on `httpRequestsAdaptiveGroups`; Cloudflare rejects it during schema validation. The corrected query uses the documented `count_DESC` / `count` shape.

Traffic must also be scoped before ranking. A zone can contain several hostnames and Workers routes, and applying the 10,000-row limit before removing non-page traffic can omit real page rows.

## Validation

- 340 focused TPR, deploy, and KV tests
- Scoped format, lint, and type checks
- `@vinext/cloudflare` package build
- `knip`
- Independent review loop


kHorozhanov:
Anything blocking here? CI is green, and the two divergences from the earlier rounds are resolved and covered by tests (the JSON `routeTargets.length > 1` guard and the pattern-less `zone_name` fallback).

The three remaining bot notes all read as non-blocking to me:

- **`tpr.ts:337`** — the per-environment path takes `extractRouteTargets(...)[0]?.hostname` without the multiple-route / scheme-qualified guards. Unreachable today: `runTPR` reads `wranglerConfig.customDomain` directly and never merges `env[...]` into the traffic scope. Happy to add a comment marking the intent.
- **`tpr.ts:386`** — whether `clientRequestPath_like` honours `\` as the LIKE escape. I could not settle this from the docs either way; it needs a real route with `_` or `%` in the path. Over-matching would be a soft degradation rather than a wrong result.
- **`tpr.ts:785`** — the sampling note is not quite right: for the adaptive datasets Cloudflare scales the sampled rows back up rather than returning a raw sampled count, so `count` is already an estimate of real volume. What does survive is that these are statistical estimates, so re-running the same query returns slightly different numbers, and a low-traffic zone is estimated less precisely. Worth a comment if you want one.

Glad to push any of those if you would like them in.

kHorozhanov:
Closing this. \`main\` has since restructured \`tpr.ts\` (most of it now lives in \`wrangler-config.ts\`) and already carries the core fix here — the \`count_DESC\`/\`count\` field names, the GraphQL variables, and exported \`filterTrafficPaths\`/\`selectRoutes\`. Rebuilding the branch on top of that would leave a different change than the one under review, so a fresh, narrower PR is the honest way to land what is still missing (host and path scoping of the traffic query, pre-limit row exclusion, explicit \`zone_id\`/\`zone_name\` with that zone's account, skip rules).

PR #2711, opened 2026-07-26
fix(dev): preserve public asset conditional requests

Boyeep:
## What changed

- mirror Vite and sirv public-file identity in development so equivalent strong or weak If-None-Match validators can take the correct 304 path
- keep public-file ETags and verified request-path aliases in an in-memory index rather than doing filesystem work on each request
- preserve Vite filesystem behavior for base paths, encoded paths, case and Unicode normalization, and file or directory symlinks
- refresh the index from Vite watcher events when public files or filesystem capabilities change
- fail closed for ambiguous aliases, redirect collisions, malformed paths, and symlink cycles

## Why

Vite public-file serving delegates to sirv, whose development ETags use file size and mtime. Equivalent request validators and filesystem aliases could miss sirv exact identity and return 200 instead of 304. This normalizes only matching GET and HEAD validators to the exact indexed weak ETag while leaving non-matching requests untouched.

## Scope note

This branch originally implemented static byte ranges. That work landed independently through #2715 while the branch was being prepared. The current diff against main is only the development public-asset conditional-request fix described above.

## Validation

- exact-head CI is green: https://github.com/cloudflare/vinext/actions/runs/30286133162
- performance benchmarks are green: https://github.com/cloudflare/vinext/actions/runs/30286136302
- focused dev public ETag unit and Vite integration coverage passes
- exact-head Big Bonk review at ec3ec0bdf identified the stale title and description, corrected here


Boyeep:
@codex review

Boyeep:
opened an upstream Vite fix for this: https://github.com/vitejs/vite/pull/23104 

Since vinext serves these files through Vite/sirv, fixing it there should also fix the issue for vinext.

PR #2256, opened 2026-06-22
fix(app-router): resolve interception-only RSC targets

james-elicx:
## Summary
- recognize valid interception-only RSC targets before Pages fallbacks and rewrites
- dispatch the interception source route for root-originated dynamic-locale navigation
- keep direct document visits to interception-only targets as 404

## Next.js parity
Targets the interception-from-root dynamic locale failure in the latest deploy-suite report. This is distinct from PR #2231, which fixes precedence between competing interception routes.

## Validation
- 2 focused App handler tests
- generated-entry delegation test
- 3/3 targeted Playwright assertions
- scoped `vp check`
- package build
- independent review: no actionable findings

Full suites are deferred to CI.

PR #2732, opened 2026-07-27
fix(app-router): reject Route Handlers as interception source routes

NathanDrake2406:
> [!WARNING]
> **This is a high severity security fix.** It closes an authorization bypass in the App Router RSC request path that lets an unauthenticated remote client execute a Route Handler which middleware path checks were intended to protect. Please treat this as release-blocking.

## Overview

| | |
| --- | --- |
| Goal | Stop a client-supplied interception context from selecting a Route Handler as the route that gets dispatched |
| Core change | `findIntercept` rejects Route Handler concrete sources and validates the final fallback owner before promotion; page owners and descendants remain unchanged |
| Key boundary | The interception source pathname is an unauthenticated request header, so it may select *which slot owner renders*, never *which arbitrary route executes* |
| Expected impact | Route Handlers become unreachable through interception promotion. Descendant page routes and their dynamic source params are unchanged |

## Why

Route interception is gated on a client header. vinext reads `x-vinext-interception-context`; Next.js reads `Next-URL`. Neither is authenticated, and neither can be, because it is just the browser reporting its current pathname. That makes the header safe to use as a *gate* on an intercept the developer already declared, and unsafe to use as a *selector* for which route the server dispatches.

Next.js keeps to that distinction. `generateInterceptionRoutesRewrites` emits a rewrite whose `destination` is fixed at build time to the intercepting route's app path, gated by a `has` condition matching `Next-URL` against `^<interceptingRoute>(?:/.*)?$`. Source params come from matching the header against the *intercepting route pattern*. Next.js never resolves `Next-URL` to a concrete route and never dispatches whatever it lands on.

vinext's `findIntercept` does resolve it. After the source pathname passes the descendants-allowed pattern gate, the matcher re-resolves it through the route trie and returns whichever concrete route it hits as `sourceRouteIndex` (added in #2042 so dynamic descendant source params survive). Since #2256, an RSC request whose path has no direct App route match promotes that resolved route into the request's `match`, which then flows into the normal dispatch branches.

Those two behaviours compose into a route confusion. Middleware has already run, for the *requested target* path. A request naming a target that has no direct route, plus a header naming a descendant of the intercepting route, promotes that descendant. When the descendant is a Route Handler, dispatch executes it. An application that authorizes route handlers with middleware path checks never sees the protected path, so the guard does not fire and the handler runs.

| --- | --- | --- |
| Interception source resolution | An unauthenticated header may gate a declared intercept, not choose an arbitrary dispatch target | Route Handler matches are skipped; a Route Handler fallback owner is rejected rather than dispatched |
| Route Handler reachability | A `route.ts` is not a renderable interception source, although the route graph may retain slots discovered beside it | Route Handlers are no longer promotable as interception sources |
| Next.js parity | The interception rewrite destination is the intercepting route, fixed at build time | Falling back to the slot owner matches the upstream destination |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| RSC request to an interception-only target, header names a descendant Route Handler | Handler is promoted and executed, having only run middleware for the target path | Promotion resolves to the slot owner; the handler is unreachable |
| Header names a descendant *page* route | Concrete descendant promoted, source params preserved | Unchanged |
| Header names a page slot owner exactly | Slot owner promoted | Unchanged |
| Header names a Route Handler slot owner | Handler could be promoted and executed | Interception is rejected; direct handler requests are unchanged |
| Header names a non-descendant, or is absent | No intercept | Unchanged |
| Lazy Route Handler whose module has not loaded yet | Classified as a page, so promotable | Classified as a handler via `__loadRouteHandler`, consistently before and after first load |

<details>

1. `packages/vinext/src/server/app-rsc-route-matching.ts` — `findIntercept`'s `concreteSourceRoute` resolution. This is the whole behavioural change and the ownership decision: the guard sits in the shared matcher so every caller inherits it rather than each dispatch site guarding separately.
2. `tests/app-rsc-route-matching.test.ts` — regression proof, including the lazy-handler classification case and the preserved page-descendant behaviour.

</details>

<details>
<summary>Validation</summary>

- Added regression coverage for descendant and fallback-owner Route Handlers, including lazy, static, dynamic, catch-all, and optional-catch-all shapes; descendant page routes still resolve concretely with their source params.
- Confirmed the pre-fix behaviour by driving the real `createAppRscRouteMatcher` through `createAppRscHandler`: a request to an interception-only target with a header naming a middleware-protected Route Handler returned that handler's response, while a direct request to the same path was denied by middleware. After the change the handler is never dispatched. A full matcher-to-handler regression also proves a Route Handler slot owner remains directly reachable while forged interception promotion is rejected.
- Ran the interception-adjacent suites: `app-rsc-route-matching`, `app-rsc-handler`, `app-server-action-execution`, `app-page-request`, `app-page-dispatch`. 364 tests pass.
- Ran `pnpm run check` (format, lint, types, Next.js type sync, shim types) clean.
- No e2e fixture places a `route.ts` beneath an intercepting route, so no existing covered behaviour changes.

<details>
<summary>Commands</summary>

```text
pnpm test tests/app-rsc-route-matching.test.ts tests/app-rsc-handler.test.ts \
  tests/app-server-action-execution.test.ts tests/app-page-request.test.ts \
  tests/app-page-dispatch.test.ts
  Test Files  5 passed (5)
       Tests  364 passed (364)

pnpm run check
  pass: All 2681 files are correctly formatted
  pass: Found no warnings, lint errors, or type errors in 1162 files
  Next.js types are in sync with next@16.2.7 (347 files)
  Public shim values match their vendored types (24 modules)
```

</details>
</details>

<details>

Low risk, and the reason is that the removed capability has no legitimate use. Interception renders a page into a parallel slot, so a Route Handler could never have been a working interception source. Any request that previously reached a handler this way was already a route confusion.

- Public API: unchanged. No signature, manifest, or config change.
- Behaviour: narrows one branch of interception source resolution. Page descendants, exact slot-owner sources, dynamic source params, sibling intercepts, and the non-intercept paths are untouched.
- Build output: unchanged. The manifest shape is the same; only request-time resolution differs.
- Deliberate divergence: none introduced. This moves closer to the upstream rewrite semantics.

</details>

<details>
<summary>Non-goals</summary>

- **Descendant page promotion is left as is.** A header naming a middleware-guarded *page* under the intercepting route can still promote it, and slot intercepts render the source route's own page as `children`. Closing that means deciding whether the concrete descendant resolution from #2042 should exist at all, versus always dispatching the slot owner and taking source params from the intercepting-route pattern as Next.js does. That is a behavioural change to a shipped fix with e2e fixtures behind it and belongs in its own PR.
- **Re-running middleware for the claimed source path is deliberately not done.** Next.js does not re-run middleware for `Next-URL`, and doing so would execute user middleware twice per interception navigation, duplicating side effects such as `Set-Cookie`, session rotation, and rate limiting.
- No changeset is included; add one if this should ship in the next release.

</details>

## References

| --- | --- |
| [`generate-interception-routes-rewrites.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/lib/generate-interception-routes-rewrites.ts) | Upstream interception rewrite: fixed destination, `Next-URL` used only as a `has` gate |
| #2256 | Introduced interception-only RSC target promotion, which put the resolved source route into the dispatch path |
| #2042 | Introduced concrete descendant source resolution so dynamic source params survive |

### DIFF
--- a/packages/vinext/src/server/app-rsc-route-matching.ts
+++ b/packages/vinext/src/server/app-rsc-route-matching.ts
@@ -284,17 +284,30 @@ export function createAppRscRouteMatcher<Route extends AppRscRouteForMatching>(
         if (params === null) continue;
         canonicalizeAppPageParams(params);
 
-        const concreteSourceRouteIndex =
-          matchedSourceRoute && entry.sourceMatchPatternParts !== null
-            ? (routeIndexes.get(matchedSourceRoute.route) ?? entry.sourceRouteIndex)
-            : entry.sourceRouteIndex;
+        // Resolving the claimed source pathname to its concrete descendant
+        // route (#2042) keeps dynamic source params intact, but the source
+        // pathname is an unauthenticated client header and the resolved route
+        // becomes the route that renders or dispatches. A Route Handler has no
+        // page, layouts, or parallel slots, so it can never own or sit inside
+        // an interception source tree; promoting one only lets a crafted
+        // interception context execute a `route.ts` that happens to live under
+        // the intercepting route. Fall back to the slot owner instead, which is
+        // the fixed destination Next.js' generated interception rewrite uses.
+        const concreteSourceRoute =
+          matchedSourceRoute &&
+          entry.sourceMatchPatternParts !== null &&
+          !isAppRouteHandlerRoute(matchedSourceRoute.route)
+            ? matchedSourceRoute
+            : null;
+        const concreteSourceRouteIndex = concreteSourceRoute
+          ? (routeIndexes.get(concreteSourceRoute.route) ?? entry.sourceRouteIndex)
+          : entry.sourceRouteIndex;
         const sourceRoute = routes[concreteSourceRouteIndex];
-        const matchedSourceParams =
-          matchedSourceRoute && entry.sourceMatchPatternParts !== null
-            ? matchedSourceRoute.params
-            : sourceRoute
-              ? matchRoutePatternRaw(sourceParts, sourceRoute.patternParts)
-              : null;
+        const matchedSourceParams = concreteSourceRoute
+          ? concreteSourceRoute.params
+          : sourceRoute
+            ? matchRoutePatternRaw(sourceParts, sourceRoute.patternParts)
+            : null;
 
         // Secondary gate (from #1249): when the entry has no
         // `sourceMatchPatternParts` declared (older manifest shapes), reject
--- a/packages/vinext/src/server/app-rsc-route-matching.ts
+++ b/packages/vinext/src/server/app-rsc-route-matching.ts
@@ -306,7 +306,11 @@ export function createAppRscRouteMatcher<Route extends AppRscRouteForMatching>(
         const matchedSourceParams = concreteSourceRoute
           ? concreteSourceRoute.params
           : sourceRoute
-            ? matchRoutePatternRaw(sourceParts, sourceRoute.patternParts)
+            ? matchSlotOwnerSourceParams(
+                sourceParts,
+                sourceRoute.patternParts,
+                entry.sourceMatchPatternParts !== null,
+              )
             : null;
 
         // Secondary gate (from #1249): when the entry has no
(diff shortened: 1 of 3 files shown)


NathanDrake2406:
@codex review

PR #2739, opened 2026-07-28
fix(middleware): stop restoring credentials middleware deleted before external rewrites

NathanDrake2406:
## The decision boundary

`x-middleware-override-headers` is the **complete post-middleware header set, not a diff.**

- `NextResponse.next`/`rewrite` encode *every* key of the `Headers` object middleware passes ([`response.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/spec-extension/response.ts) — `handleMiddlewareField`).
- Next.js then **deletes any request header not in that list** ([`resolve-routes.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/router-utils/resolve-routes.ts)):

  ```js
  // Delete headers.
  for (const key of Object.keys(req.headers)) {
    if (!overriddenHeaders.has(key)) {
      delete req.headers[key]
    }
  }
  ```

There is no tombstone for a deleted header because none is needed. **Absence is the tombstone.**

## The bug

`preserveCredentialHeaders` (#1121) read a short override list as a *partial* override and copied the base request's `cookie`/`authorization` back in. But the documented deletion pattern produces exactly such a list:

```ts
const headers = new Headers(request.headers)
headers.delete('cookie')
return NextResponse.rewrite(externalUrl, { request: { headers } })
```

`cookie` is simply absent from the override list — so the option resurrected it. The option was enabled **only for external rewrites**, so `proxyExternalRequest` then forwarded first-party session cookies and bearer tokens to a cross-origin target, despite the app explicitly stripping them.

Same-origin restoration would have been merely a parity bug; gating it to cross-origin destinations turned it into a credential leak.

## Why removal, not a narrower guard

The "partial override" the option guarded against **cannot occur**. `encodeMiddlewareRequestHeaders` is the only producer of the override list in the codebase, and it always emits the full key set:

```ts
const overrideHeaderNames = [...requestHeaders.keys()];
targetHeaders.set(MIDDLEWARE_OVERRIDE_HEADERS, overrideHeaderNames.join(","));
```

#1121's own "explicit deletion" test passed a list naming `authorization,cookie` with no corresponding `x-middleware-request-*` values — an encoding the public API never emits — so it never exercised the real deletion path. Deleting the option restores Next.js-exact semantics and removes the parameter from three call sites.

## Scenario-level behavior

| Middleware does | Override list | Before | After |
|---|---|---|---|
| `NextResponse.next()` (no `request`) | *not sent* | credentials forwarded | credentials forwarded (unchanged) |
| clones headers, adds one | full set incl. `cookie` | credentials forwarded | credentials forwarded (unchanged) |
| clones headers, **deletes `cookie`** | full set minus `cookie` | 🔴 **`cookie` restored and sent cross-origin** | ✅ `cookie` dropped |

Only the third row changes.

## Validation

Both regression tests were confirmed **failing before the fix, passing after**:

- `tests/app-router-external-rewrite.test.ts` — end to end through the dev server; fixture middleware deletes `cookie`/`authorization` before an external rewrite, mock upstream asserts neither arrives. Pre-fix: `expected 'session=secret123' to be undefined`.
- `tests/shims.test.ts` — the `proxyExternalMiddlewareRewrite` boundary, driven through the real `NextResponse.rewrite` API against a live `node:http` upstream. Pre-fix: `expected 'Bearer secret' to be undefined`.
- `tests/pages-request-pipeline.test.ts` — Pages Router pipeline, both the no-override-list (credentials forwarded) and deleted-from-override-list (dropped) cases.
- `tests/request-pipeline.test.ts` — unit coverage for both override-list states, replacing the two #1121 tests that encoded the buggy expectation.

`pnpm run check` passes (format, lint, types, Next.js type sync, shim types).

## Risk / review path

Behavior change is confined to the one row above — restoring a header the app deleted. Any app relying on the old behavior was relying on a credential leak.

Suggested review order: `middleware-request-headers.ts` (the semantics), then the three call sites (mechanical parameter removal), then the tests.

### DIFF
--- a/packages/vinext/src/config/config-matchers.ts
+++ b/packages/vinext/src/config/config-matchers.ts
@@ -448,13 +448,8 @@ function shouldEvaluateRule(ruleBasePath: false | undefined, state: BasePathMatc
 export function applyMiddlewareRequestHeaders(
   middlewareHeaders: Record<string, string | string[]>,
   request: Request,
-  options: { preserveCredentialHeaders?: boolean } = {},
 ): { request: Request; postMwReqCtx: RequestContext } {
-  const nextHeaders = buildRequestHeadersFromMiddlewareResponse(
-    request.headers,
-    middlewareHeaders,
-    options,
-  );
+  const nextHeaders = buildRequestHeadersFromMiddlewareResponse(request.headers, middlewareHeaders);
 
   for (const key of Object.keys(middlewareHeaders)) {
     if (key.startsWith(MIDDLEWARE_HEADER_PREFIX) && key !== MIDDLEWARE_CACHE_HEADER) {
--- a/packages/vinext/src/server/app-middleware.ts
+++ b/packages/vinext/src/server/app-middleware.ts
@@ -120,9 +120,7 @@ function requestWithMiddlewareRequestHeaders(
   middlewareHeaders: Headers | null,
 ): Request {
   const nextHeaders = middlewareHeaders
-    ? buildRequestHeadersFromMiddlewareResponse(request.headers, middlewareHeaders, {
-        preserveCredentialHeaders: true,
-      })
+    ? buildRequestHeadersFromMiddlewareResponse(request.headers, middlewareHeaders)
     : null;
   if (!nextHeaders) return request;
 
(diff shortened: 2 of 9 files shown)


NathanDrake2406:
@codex review

PR #2026, opened 2026-06-15
fix(actions): return body-limit errors without closing responses

james-elicx:
## Summary

- preserve Node response sockets when oversized request bodies are cancelled
- adapt Node request streams to Web streams with backpressure-aware pause/resume behavior
- keep ordinary 500 Server Action failures as serialized action errors while limiting access fallbacks to 401/403/404
- add deployed overflow and browser error-boundary regressions

## Next.js parity

Fixes the plaintext and multipart failures in:

- `test/e2e/app-dir/actions/app-action-size-limit-invalid.test.ts`

## Validation

- pinned Next.js v16.2.6 targeted deploy suite: 5/5 passed
- browser fixture renders `Something went wrong!` and excludes `NEXT_HTTP_ERROR_FALLBACK;500`
- request adapter/action/middleware and production integration tests passed
- scoped `vp check` and vinext build passed
- two independent review rounds; no actionable findings remain

PR #2741, opened 2026-07-28
fix(server): transfer request bodies into NextRequest instead of teeing

NathanDrake2406:
## Overview

| | |
| --- | --- |
| Goal | Stop retaining whole request bodies in memory when the runtime wraps an incoming `Request` |
| Core change | `NextRequest` transfers the body (`super(input, init)`, matching Next.js) rather than `input.clone()` |
| Key boundary | Only code that genuinely needs two live branches clones, and it owns cancelling the branch it does not consume |
| Expected impact | Streaming route handlers stay O(1) memory; a remote unauthenticated client can no longer force full-body buffering |

## Why

`Request.clone()` is specified to `tee()` the body stream, and `tee` applies no backpressure to the slower branch: whenever one branch pulls a chunk from the source, that chunk is also enqueued into the other branch's queue. If a branch is never read, its queue grows to the full body size. Dropping the reference does not help, because the tee closure holds both branch controllers and stays reachable from the branch that *is* being read.

The `NextRequest` constructor cloned every body-bearing input. Ordinary App Router route handling wraps the incoming request through `createTrackedAppRouteRequest`, so the branch left behind on the caller-owned request was never read and never cancelled. A handler that streams an upload straight to storage, which should hold no more than one chunk at a time, instead retained the entire body for the life of the request. That is remotely reachable, unauthenticated on any public body-bearing route, and repeatable until the process or Worker isolate runs out of memory.

The clone also disabled an existing mitigation. `executeMiddleware` already cancels the middleware body branch in a `finally` (added in #2026). The constructor's clone inserted a second tee between that cancel and the branch actually accumulating chunks, so the cancel released a branch nothing was filling while the real one kept growing one level up. Removing the clone reconnects that cancel to the stream it was written for.

Upstream Next.js does `super(input, init)` in the same constructor, so transferring is also the parity behaviour. The isolation added in #1132 is preserved: the callers that need the source request to stay readable already clone explicitly before wrapping.

| --- | --- | --- |
| `NextRequest` constructor | A wrapper takes ownership of the body it wraps | Transfers the body instead of teeing it |
| `createTrackedAppRouteRequest` basePath re-prefix | Do not branch a stream whose source is discarded | Builds the prefixed request directly from the incoming one |
| `executeMiddleware` | The one place with two genuinely live branches cancels the one it does not consume | Unchanged; its existing cancel becomes effective again |

## What changed

| Scenario | Before | After |
| --- | --- | --- |
| Route handler streams a body-bearing request | Whole body retained on an unread tee branch | Streams through, nothing retained |
| Route handler under a configured `basePath` | Second dead tee added while re-adding the prefix | Body transferred into the prefixed request |
| Middleware matches and never reads the body | Full body buffered despite the `finally` cancel | Branch released, downstream streams normally |
| Middleware reads the body, then `NextResponse.next()` | Downstream still reads the body (#1132) | Unchanged |
| Userland `new NextRequest(sourceRequest)` | Source stayed readable, diverging from Next.js | Source body is transferred, matching Next.js |
| Userland `request.clone()` inside a handler | Standard `Request` tee semantics | Unchanged; branching stays the caller's choice |

<details>

1. `packages/vinext/src/shims/server.ts` for the ownership decision in the constructor.
2. `packages/vinext/src/server/app-route-handler-runtime.ts` for the dead tee on the basePath path.
3. `packages/vinext/src/server/middleware-runtime.ts` (unchanged) for the clone that is genuinely needed and the `finally` that cancels it.
4. `tests/shims.test.ts` and `tests/app-route-handler-runtime.test.ts` for the regression proofs.

</details>

<details>
<summary>Validation</summary>

- Reproduced the leak against the built package with a 128 MiB streamed body, reading and discarding chunk by chunk, holding the source request alive across the measurement the way a real call frame does.
- Confirmed the same retention through `applyAppMiddleware` for middleware that does not touch the body.
- Updated `tests/shims.test.ts` to pin the constructor as transferring rather than teeing, and added coverage in `tests/app-route-handler-runtime.test.ts` for the basePath path. Both fail if a clone is reintroduced.
- The #1132 regression test (middleware reads the body, downstream still reads it) continues to pass unchanged.
- Ran `tests/shims.test.ts`, `tests/app-route-handler-runtime.test.ts`, `tests/app-router-middleware-next-request.test.ts`, `tests/middleware-runtime.test.ts`, `tests/app-server-action-execution.test.ts`, `tests/app-route-handler-execution.test.ts`, `tests/app-post-middleware-context.test.ts`, `tests/routing.test.ts`, `tests/pages-api-route.test.ts`, `tests/api-handler.test.ts`, `tests/app-route-handler-dispatch.test.ts`, `tests/app-route-handler-policy.test.ts`, `tests/app-router-production-server.test.ts`, plus `vp check` on the changed files.

<details>
<summary>Measurements</summary>

Node 26.5.0, 128 chunks of 1 MiB, `--expose-gc`, delta measured around a full streaming read.

```text
before
  baseline (no wrap)           read=128.0MiB  arrayBuffers +0.0MiB
  new NextRequest(source)      read=128.0MiB  arrayBuffers +128.0MiB

after
  baseline (no wrap)           read=128.0MiB  arrayBuffers +0.0MiB
  new NextRequest(source)      read=128.0MiB  arrayBuffers +0.0MiB
```

Through `applyAppMiddleware`, downstream draining the request body:

```text
before
  middleware ignores body   downstream=128.0MiB  arrayBuffers +128.0MiB
  middleware reads body     downstream=128.0MiB  arrayBuffers +0.0MiB

after
  middleware ignores body   downstream=128.0MiB  arrayBuffers +0.0MiB
  middleware reads body     downstream=128.0MiB  arrayBuffers +0.0MiB
```

A second symptom disappears with the fix. Because a tee only releases its source once every branch has cancelled, cancelling the downstream body while the middleware branch was still live returned a promise that never settled. After the fix it settles and the source stream's `cancel()` runs.

</details>
</details>

<details>

- Public API: `new NextRequest(request)` now disturbs the source request's body, which is what Next.js does. Code that wrapped a request and then read the original is relying on a vinext-only divergence and must clone explicitly. No in-tree caller does this.
- Runtime: no new allocation, no new async work. One tee removed from every body-bearing request, and a second from the basePath path.
- Middleware isolation from #1132 is unchanged, and is still covered by its original test.

</details>

<details>
<summary>Non-goals</summary>

- No framework-level request body size limit is introduced here. This PR removes the amplification the framework was adding; capping upload size remains a deployment and userland concern.

</details>

## References

| --- | --- |
| [Next.js `NextRequest`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/spec-extension/request.ts) | Upstream does `super(input, init)`, the behaviour this restores |
| [Streams spec, `ReadableStreamTee`](https://streams.spec.whatwg.org/#abstract-opdef-readablestreamdefaulttee) | Defines the unbounded queueing on the unread branch, and that the source is only cancelled once every branch cancels |
| #1132 | Introduced the clone to isolate middleware body reads; that isolation is preserved |
| #2026 | Added the middleware branch cancel that the clone had been defeating |

### DIFF
--- a/packages/vinext/src/entries/app-rsc-entry.ts
+++ b/packages/vinext/src/entries/app-rsc-entry.ts
@@ -1236,7 +1236,7 @@ export default createAppRscHandler({
   },
   ${
     middlewarePath
-      ? `runMiddleware({ cleanPathname, context, hadBasePath, isDataRequest, request }) {
+      ? `runMiddleware({ cleanPathname, context, hadBasePath, isDataRequest, middlewareRequest, request }) {
     return __applyAppMiddleware({
       basePath: __basePath,
       cleanPathname,
@@ -1246,6 +1246,7 @@ export default createAppRscHandler({
       i18nConfig: __i18nConfig,
       isDataRequest,
       isProxy: ${JSON.stringify(isProxyFile(middlewarePath))},
+      middlewareRequest,
       module: middlewareModule,
       request,
       trailingSlash: __trailingSlash,
--- a/packages/vinext/src/server/app-middleware.ts
+++ b/packages/vinext/src/server/app-middleware.ts
@@ -5,7 +5,11 @@ import { setNavigationContext } from "vinext/shims/navigation";
 import { FLIGHT_HEADERS, VINEXT_MW_CTX_HEADER } from "./headers.js";
 import { buildRequestHeadersFromMiddlewareResponse } from "../utils/middleware-request-headers.js";
 import { mergeMiddlewareResponseHeaders } from "./middleware-response-headers.js";
-import { executeMiddleware, type MiddlewareModule } from "./middleware-runtime.js";
+import {
+  executeMiddleware,
+  type MiddlewareModule,
+  type MiddlewareResult,
+} from "./middleware-runtime.js";
 import { cloneRequestWithHeaders, processMiddlewareHeaders } from "./request-pipeline.js";
 import { internalServerErrorResponse } from "./http-error-responses.js";
 
@@ -28,6 +32,8 @@ export type ApplyAppMiddlewareOptions = {
   isDataRequest?: boolean;
   filePath?: string;
   isProxy: boolean;
+  /** An App Router-created body branch dedicated to middleware. */
+  middlewareRequest?: Request;
   module: MiddlewareModule;
   request: Request;
   /**
@@ -66,7 +72,7 @@ function isForwardedMiddlewareContext(value: unknown): value is ForwardedMiddlew
   return !!value && typeof value === "object";
 }
 
-function requestWithoutFlightHeaders(request: Request): Request {
+function requestWithoutFlightHeaders(request: Request, isolateBody = false): Request {
   let hasFlightHeader = false;
   const headers = new Headers();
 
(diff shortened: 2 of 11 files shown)


NathanDrake2406:
@codex review

Issue #1489, opened 2026-05-22
App Router draft mode (`unstable_cache` bypass + `draftMode().isEnabled`) not implemented


james-elicx:
> *This issue was created by an agent analysing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/26282973805) (vinext `main` vs Next.js `v16.2.6`, 2026-05-22).*

## Problem

Draft mode tests find that `unstable_cache` is not bypassed when draft cookies are set, no new cached value is written in draft mode, and `draftMode().isEnabled` reads return empty strings. The cookie/draft-mode plumbing is not threaded through the request context.

```
draftMode().isEnabled returns empty string; unstable_cache not bypassed in draft mode
```

## Estimated Impact

~3 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/app-dir/app-static/app-static.test.ts` (3 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add an action that calls `draftMode().enable()` and verify the next request sees `draftMode().isEnabled === true` and bypasses `unstable_cache`.

2. **Thread the draft cookie through the request context.** Read the Next.js draft-mode cookie at request start and expose it via the AsyncLocalStorage scope so `draftMode()`, cache helpers, and `unstable_cache` can react.

PR #2744, opened 2026-07-28
fix(cache): bypass shared "use cache" entries in draft mode

NathanDrake2406:
Draft-mode requests can currently poison shared `"use cache"` entries.

## The gap

`registerCachedFunction` builds its shared cache key from the function id, the build seed and the serialized arguments — never from draft-mode state. Since #1489 made `draftMode().isEnabled` readable inside cache scopes (correctly matching Next.js, which allows reads and only forbids `enable()`/`disable()` there), a cached function that branches on it becomes a request-dependent input to a key that does not vary by request:

```ts
async function getPost(slug: string) {
  "use cache";
  const { isEnabled } = await draftMode();
  return isEnabled ? cms.preview(slug) : cms.published(slug); // seeds one shared entry
}
```

An editor request stores the unpublished result; later public requests are served it from cache without re-executing. The reverse also holds — a public request can pin published content into the preview view.

`unstable_cache` received this guard in #2319. The public `"use cache"` path was left without it.

## Change

Skip the shared cache lookup and store when the request is in draft mode, alongside the existing dev-mode bypass, reusing the `isDraftModeEnabled()` helper #2319 added for exactly this purpose.

This mirrors Next.js, which guards the same two operations at request level rather than by widening the key:

- read — `use-cache-wrapper.ts` gates `cacheHandler.get` on `!shouldForceRevalidate(...)`, and `shouldForceRevalidate` returns `true` when `workStore.isDraftMode`
- write — `if (!workStore.isDraftMode)` around the cache-handler save, commented "When draft mode is enabled, we must not save the cache entry."

`"use cache: private"` is deliberately untouched: it is a per-request in-memory map that never crosses a request boundary, so it cannot leak across the draft/public split.

## Validation

`tests/shims.test.ts` — "bypasses the shared cache in draft mode": a draft request renders `DRAFT`, then a non-draft request for the same key must observe `PUBLIC`, then a second draft request must observe `DRAFT` again. Reverting the source change alone fails it at the second assertion with `expected 'DRAFT' to be 'PUBLIC'` — the leak the fix closes.

Also run green: `tests/shims.test.ts`, `tests/cache-proof.test.ts`, `tests/use-cache-transform.test.ts`, `tests/draft-secret-exposure.test.ts` (1310 tests), plus the repo pre-commit full check, staged tests and knip.

## Out of scope

`patchedFetch` in `shims/fetch-cache.ts` has no draft-mode bypass either, where Next.js falls straight through to the native fetch (`patch-fetch.ts`). That is a preview-freshness parity gap rather than the same leak: vinext's fetch cache key already covers URL, method, headers and body, so a draft request and a public request only share an entry when the outgoing request is byte-for-byte identical. Worth fixing separately.

### DIFF
--- a/packages/vinext/src/shims/cache-runtime.ts
+++ b/packages/vinext/src/shims/cache-runtime.ts
@@ -49,7 +49,7 @@ import {
   getRequestContext,
   runWithUnifiedStateMutation,
 } from "./unified-request-context.js";
-import { markDynamicUsage } from "./headers.js";
+import { isDraftModeEnabled, markDynamicUsage } from "./headers.js";
 import { trackPprFallbackShellCacheTask } from "./ppr-fallback-shell.js";
 import { isMarkedAppPagePropsObject } from "./internal/app-page-props-cache-key.js";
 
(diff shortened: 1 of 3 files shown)


NathanDrake2406:
@codex review

NathanDrake2406:
@codex review

PR #2748, opened 2026-07-29
fix(deps): pin vitest and coverage-istanbul to 4.1.10

shulaoda:
Bumps two catalog entries from `4.1.9` to `4.1.10`:

- `vitest`: `4.1.9` → `4.1.10`
- `@vitest/coverage-istanbul`: `4.1.9` → `4.1.10`

vite-plus 0.2.6 bundles vitest 4.1.10 and now asserts the coverage provider matches the runner version, so the `4.1.9` pins broke CI: https://github.com/cloudflare/vinext/actions/runs/30404544308/job/90426980201

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -94,8 +94,8 @@ catalogs:
       specifier: ^0.5.27
       version: 0.5.27
     '@vitest/coverage-istanbul':
-      specifier: 4.1.9
-      version: 4.1.9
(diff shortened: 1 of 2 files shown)


shulaoda:
I think we should use `vite-plus/test` instead of `vitest`, we’ve already done this in some places, but it looks like a few cases still haven’t been updated. I’ll open a follow-up PR after this one is merged to try it. 🤔

PR #2757, opened 2026-07-30
fix(app-router): stop flooring dynamic prefetch stale times

NathanDrake2406:
## Contract

Next keeps its two client stale-time dimensions on deliberately separate rules:

- the cacheLife / `Next-Router-Stale-Time` bound goes through [`getStaleTimeMs`](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/components/segment-cache/cache.ts) — `Math.max(staleTimeSeconds, 30) * 1000`;
- the dynamic bound goes through [`computeDynamicStaleAt`](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/components/segment-cache/bfcache.ts) — raw `seconds * 1000`, **no floor**.

The missing floor on the second is the point: `experimental.staleTimes.dynamic` defaults to `0` so a dynamic payload is never reused across a navigation without a fresh server round-trip.

## Root cause

`resolvePrefetchedRscResponseExpiresAt` floored the *combined* value that `resolveRscResponseStaleTimeSeconds` had already min-combined:

```ts
return timestamp + Math.max(seconds * 1000, minimumTtlMs);
```

`serverStaleTimeSeconds` floors the cacheLife half *before* the min — its own comment says the floor "never raises the config bound". So the outer floor had exactly one live effect: raising the dynamic bound, which is the one value that must never be raised.

Every dynamic render does report that bound. `app-page-render.ts:730` defaults `dynamicStaleTimeSeconds` to `resolveConfiguredDynamicStaleTimeSeconds()` (i.e. `staleTimes.dynamic`, `0`), emitted as `X-Vinext-Dynamic-Stale-Time` when the render is known dynamic up front and in the completion-metadata footer when it turns dynamic mid-stream. Both arrived at the client as `0` and both were floored to 30s.

Note the asymmetry this created: the visited-response path (`resolveCachedRscResponseExpiresAt`) never floored, so it honored the `0` correctly. Only the prefetch path diverged.

## Impact

`<Link>` prefetches are fetched with `credentials: "include"`, and a `cacheForNavigation` prefetch stores the full RSC payload. A navigation could therefore replay authenticated or personalized server-component output for 30s after a logout, session change, role revocation, or permission update, with no server request and so no fresh authorization check. The consumed expiry was then carried into the committed snapshot (`app-browser-entry.ts`) and honored by `app-visited-response-cache.ts`, extending the window past the navigation.

Reported as a medium-severity finding against `ad14e9b`. The report also alleged a second hole — that a dynamic page emitting no stale-time signal falls back to the 300s static TTL. That half does **not** reproduce: the `dynamicStaleTimeSeconds !== undefined` gates at `app-page-render.ts:854` and `:905` are always true because the value is defaulted 170 lines earlier. The server side is correct; the defect was entirely client-side.

## Fix

Floor only the unsignalled fallback, mirroring `STATIC_STALETIME_MS = getStaleTimeMs(config)`. A signalled bound is authoritative.

Also removes the `minimumTtlMs` plumbing. It was a partial workaround that zeroed the floor for routes with a dynamic *pattern* segment (`route.isDynamic`), which is the wrong axis — a statically-patterned route that renders dynamically (`/dashboard` reading `cookies()`) never matched it, and that is precisely the vulnerable shape. With the dimensions separated correctly the option has no remaining effect on any production call site: the policy always yields `fallbackTtl: "static"`, so its only other branch was `Math.max(300_000, 0)`.

## Verification

Three unit regressions, each confirmed to fail with the floor restored and pass without it:

- `expires a zero dynamic stale time immediately instead of flooring it` — `X-Vinext-Dynamic-Stale-Time: 0` under the 300s static fallback must expire at `now` and refuse consumption.
- `keeps a dynamic bound unfloored even when cacheLife would floor higher` — dynamic `5` against cacheLife `60` resolves to 5s, not the 30s floor.
- `allows automatic dynamic full prefetches to expire immediately` — rewritten to drive the real mechanism (the header a dynamic render sends) instead of the removed synthetic option.

The test previously named `uses Next.js's minimum prefetch stale window for server stale time zero` asserted the buggy 30s and cited `getStaleTimeMs`, but its header was the *dynamic* one — it applied the cacheLife rule to the wrong dimension. Renamed and corrected.

Existing coverage that must not move, and did not: cacheLife `1` → 30s, `min(dynamic 300, server 30)` → 30s, pending marker → 30s, `fallbackTtlMs: 5_000` → 30s.

Ran locally: `vp check` on all five changed files, and `vp test run` over `prefetch-cache`, `link`, `link-navigation`, `app-browser-entry`, `app-visited-response-cache`, `slot` — 544 passed. The pre-commit full check and knip also passed.

**Not verified locally:** `tests/e2e/app-router/nextjs-compat/client-cache.spec.ts`. Every test in that spec, including ones unrelated to this change, times out at `waitForAppRouterHydration` in my worktree, so I could not attribute results honestly — leaving it to CI. Reasoning through the fixtures suggests no movement: `client-cache/[id]` exports `unstable_dynamicStaleTime = 30` (30s either way), `client-cache-late-dynamic` exports `60`, and `client-cache/no-loading/[id]` is `force-dynamic` with no override, so it reaches its existing "expires immediately" assertion through the `0` header instead of the removed `minimumTtlMs`. CI should confirm.

## Review path

`packages/vinext/src/shims/navigation.ts` is the fix; the other three source files are the `minimumTtlMs` removal.

## Update after merging current main

Main now includes #2760, which separates runtime-prefetch `cacheLife` freshness from dynamic visited/BFCache freshness. The conflict resolution preserves that Next.js v16.2.6 behavior: a completed `cacheLife` claim controls runtime-prefetch expiry; only when no `cacheLife` claim exists does an automatic prefetch honor the raw dynamic signal. The earlier validation bullet describing dynamic 5 min-combined with cacheLife 60 is superseded and that test was removed during the merge.

Final head `3e2ba48e046db352a90fc1ea32ff1a335b5b8cb0`: 547 targeted tests passed, exact-head CI passed 47/47 jobs, independent review was clean, and BigBonk found no blocking issues.

### DIFF
--- a/benchmarks/vinext/tsconfig.json
+++ b/benchmarks/vinext/tsconfig.json
@@ -11,6 +11,7 @@
     "esModuleInterop": true,
     "resolveJsonModule": true,
     "isolatedModules": true,
+    "types": ["vinext/types"],
     "paths": {
       "@/*": ["./*"]
     }
--- a/examples/app-router-cloudflare/tsconfig.json
+++ b/examples/app-router-cloudflare/tsconfig.json
@@ -7,7 +7,7 @@
     "strict": true,
     "esModuleInterop": true,
     "skipLibCheck": true,
-    "types": ["@cloudflare/workers-types"]
+    "types": ["vinext/types", "@cloudflare/workers-types"]
   },
   "include": ["app", "worker", "*.ts"],
 }
--- a/examples/app-router-nitro/.gitignore
+++ b/examples/app-router-nitro/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 .output
 .vercel
+next-env.d.ts
--- a/examples/app-router-nitro/tsconfig.json
+++ b/examples/app-router-nitro/tsconfig.json
@@ -7,7 +7,7 @@
     "strict": true,
     "esModuleInterop": true,
     "skipLibCheck": true,
-    "types": []
+    "types": ["vinext/types"]
   },
   "include": ["app", "worker"]
 }
--- a/examples/app-router-playground/.gitignore
+++ b/examples/app-router-playground/.gitignore
@@ -35,4 +35,5 @@ yarn-error.log*
 
 # typescript
 *.tsbuildinfo
+next-env.d.ts
 .env*.local
--- a/examples/app-router-playground/tsconfig.json
+++ b/examples/app-router-playground/tsconfig.json
@@ -14,6 +14,7 @@
     "isolatedModules": true,
     "jsx": "react-jsx",
     "incremental": true,
+    "types": ["vinext/types", "node"],
     "baseUrl": ".",
     "paths": { "#/*": ["./*"] },
     "plugins": [{ "name": "next" }]
--- a/examples/benchmarks/tsconfig.json
+++ b/examples/benchmarks/tsconfig.json
@@ -11,7 +11,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "types": ["@cloudflare/workers-types"]
+    "types": ["vinext/types", "@cloudflare/workers-types"]
   },
   "include": ["app", "worker"]
 }
--- a/examples/hackernews/tsconfig.json
+++ b/examples/hackernews/tsconfig.json
@@ -10,9 +10,7 @@
     "allowJs": true,
     "baseUrl": ".",
     "noEmit": true,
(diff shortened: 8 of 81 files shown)


NathanDrake2406:
@codex review

NathanDrake2406:
@codex review

PR #2770, opened 2026-07-31
test(pages): enable complex shallow routing coverage

james-elicx:
## Summary

- wait for the pages-router-complex client-only hydration sentinel before exercising controlled inputs
- enable the shallow routing, router event, and compat-router trial reset tests
- assert the gSSP-derived render timestamp stays stable across the shallow transition
- link the relevant Next.js parity tests and update the documented pass count

All three former fixmes shared one cause on current `main`: Playwright could change the server-rendered `<select>` before React attached its `onChange` handler. The shallow router behavior itself now matches Next.js.

## Validation

- `vp check tests/e2e/pages-router-complex/client-routing.spec.ts`
- negative control: changing the fixture call to `{ shallow: false }` fails on the render timestamp changing
- `PLAYWRIGHT_PROJECT=pages-router-complex pnpm exec playwright test tests/e2e/pages-router-complex/client-routing.spec.ts --retries=0 --reporter=line` (4 passed)
- `PLAYWRIGHT_PROJECT=pages-router-complex pnpm exec playwright test --retries=0 --reporter=line` (62 passed, 11 unrelated skipped)

### DIFF
--- a/examples/pages-router-complex/README.md
+++ b/examples/pages-router-complex/README.md
@@ -98,12 +98,11 @@ PLAYWRIGHT_PROJECT=pages-router-complex pnpm run test:e2e
   degrades under it, so the `@atlas/*` alias is wired into both bundlers
   explicitly (webpack hook + Vite `resolve.alias`); tsconfig `paths` (without
   the removed `baseUrl`) stays authoritative for the type checker.
-- **vinext dev (Cloudflare plugin): 59/73 specs pass; the 14 known gaps are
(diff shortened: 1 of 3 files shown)


james-elicx:
Superseded by #2777. The three client-routing failures were the same Pages dev hydration-order bug: React handlers were not installed before browser load, rather than a separate shallow-router defect. The exact original no-wait cases pass under real Next.js and under #2777, which now carries all six affected no-wait regressions plus the gSSP timestamp assertion from this PR. Keeping this branch for now.

PR #2796, opened 2026-08-03
fix(font): not emitting 400 as default value when font weight is omitted

### LINKED ISSUE #2793: `next/font/local` emits `font-weight: 400` when weight is omitted, clamping variable fonts
## Summary

The `next/font/local` shim emits `font-weight: 400` whenever neither the source nor the top-level options specify a weight.

For variable fonts with a `wght` axis, this clamps the font face to the regular 400 instance. Elements using `font-weight: 600` or `700` still have the expected computed CSS value, but they do not render using the corresponding variable-font instance and look different from Next.js.

This affects both development and production output.

## Reproduction

Use any variable font containing a `wght` axis, for example `NotoSansSC-VariableFont_wght.ttf`:

```tsx

const notoSans = localFont({
  src: [
    {
      path: "./NotoSansSC-VariableFont_wght.ttf",
      style: "normal",
    },
  ],
});

  return (
    <main className={notoSans.className}>
      <p style={{ fontWeight: 400 }}>Regular text</p>
      <p style={{ fontWeight: 600 }}>Semibold text</p>
      <p style={{ fontWeight: 700 }}>Bold text</p>
    </main>
  );
}
```

Run the application with vinext and inspect the generated `data-vinext-fonts` style block.

## Actual behavior

vinext generates:

```css
@font-face {
  font-family: "notoSans";
  src: url(...) format("truetype");
  font-weight: 400;
  font-style: normal;
  font-display: swap;
}
```

The explicit single-value descriptor clamps the variable font's `wght` axis to 400.

The relevant code is at:
https://github.com/cloudflare/vinext/blob/ced0881690a5d5861be9913745a9eb01baa9dd90/packages/vinext/src/shims/font-local.ts#L120

## Expected behavior

When no weight was specified, vinext should not invent a `400` descriptor.

Next.js only emits `font-weight` when `weight` or `defaultWeight` is provided:

- Webpack loader:
  https://github.com/vercel/next.js/blob/canary/packages/font/src/local/loader.ts
- Turbopack implementation:
  https://github.com/vercel/next.js/blob/canary/crates/next-core/src/next_font/local/stylesheet.rs

The generated rule should therefore omit the descriptor:

```css
@font-face {
  font-family: "notoSans";
  src: url(...) format("truetype");
  font-style: normal;
  font-display: swap;
}
```

Alternatively, `font-weight: auto` would preserve the CSS Fonts Level 4 initial behavior, although omitting the descriptor would match Next.js more closely.

CSS Fonts Level 4 specifies that the initial value of the `@font-face` `font-weight` descriptor is `auto`. For variation-axis clamping, `auto` does not clamp the axis:

https://www.w3.org/TR/css-fonts-4/#font-prop-desc

Explicit user-provided values such as `400`, `700`, or `100 900` should continue to be emitted unchanged.

## Verification

I compared vinext and Next.js using the exact same variable TTF file; the source and emitted font files had identical SHA-256 hashes.

In the same Chrome browser:

- A `font-weight: 400` text run had identical dimensions in vinext and Next.js.
- A `font-weight: 700` text run differed:
  - vinext: `136.0729px`
  - Next.js: `139.4271px`

The font loaded successfully in both cases, and the computed style reported `font-weight: 700`. This rules out a missing font or fallback-font issue and isolates the difference to the generated `@font-face` weight descriptor.

## Environment

- vinext: `1.0.0-beta.4`
- Vite: `8.2.0`
- React: `19.2.8`
- Node.js: `22.22.1`
- App Router
- Reproduced in production output and browser rendering

## Notes

Related but not a duplicate: #885. That issue concerns `next/font/google` inventing a weight range when none was specified; this issue concerns `next/font/local` inventing a single weight and clamping variable fonts.

btw, this issue is found during my project's migration to vinext.

NriotHrreion:
Closes #2793

## Overview

The `next/font/local` shim emits `font-weight: 400` when no weight was specified. For variable fonts with a `wght` axis, this descriptor clamps the font face to its regular 400 instance, preventing heavier weights from rendering correctly even when elements use `font-weight: 600` or `700`.

This aligns vinext with Next.js and the CSS Fonts specification by allowing the descriptor to retain its default `auto` behavior when omitted, while preserving explicit single weights, weight ranges, and source-level overrides. The fix applies consistently to both development and production output.

## What changed

The original behavior sets font weight to `400` if `weight` is not provided

https://github.com/cloudflare/vinext/blob/d65ff09985e95bc40d5564457331d0b96ca74c42/packages/vinext/src/shims/font-local.ts#L120
https://github.com/cloudflare/vinext/blob/d65ff09985e95bc40d5564457331d0b96ca74c42/packages/vinext/src/shims/font-local.ts#L132-L138

The new behavior doesn't set font weight if `weight` is not provided, leaving it as `auto`

```diff
rules.push(`@font-face {
  font-family: '${escapeCSSString(family)}';
  src: url('${escapeCSSString(src.path)}') format('${format}');
- font-weight: ${weight};
- font-style: ${style};
+ ${weight === undefined ? "" : `  font-weight: ${weight};\n`}  font-style: ${style};
  font-display: ${display};
}`);
```

## Testing

Corresponding tests are added or updated.

- `pnpm test tests/app-router-font-google-prod.test.ts tests/font-local-transform.test.ts tests/pages-router-font-google-prod.test.ts`

### Fixtures

Related fixtures (`font-google-pages` and `font-google-multiple`) are updated. Noto sans variable font is used in the fixtures because Noto sans makes the visual difference before and after this fix much more noticeable in the fixtures.

**Before:**
image

**After:**
image

## Notes

I ran into this issue when migrating my project to vinext. Applying the fix is quite important to me. Thanks for reviewing!


NriotHrreion:
@james-elicx

Hi! I ran into this issue while migrating my project from nextjs to vinext. The font issue can be quite noticeable after the migration, so I'd really appreciate it if you could take a look and review this PR when you have a chance. Thanks!

PR #2814, opened 2026-08-05
fix(build): throw error if there is `_next` folder inside the public folder

### LINKED ISSUE #2778: Files under `public/_next/static/` are served as build assets, ahead of middleware
Files copied from `public/_next/static/` into the client output are served ahead of middleware, because `/_next/static/*` is classified as a build asset by URL shape alone. Under `next start` those files are not reachable at all, since `/_next/static/x` only ever resolves against `.next/static`.

Verified by reading source on `main` @ `c9a4a843c`, plus one executed unit test (noted below). I did not build and run a production server, so the end-to-end repro is stated but unconfirmed.

## Possible root cause

`packages/vinext/src/server/prod-server.ts:1187`:

```ts
if (pathname.startsWith(nextStaticDir)) return pathname;
```

`resolveAppRouterAssetPath()` decides "this is a build asset" from the request path and nothing else. It never consults the build manifest, so it cannot tell a generated hashed chunk from a file the app dropped into `public/_next/static/`.

Its two callers pass that classification straight to `tryServeStatic()`, before middleware, RSC, and route handling:

- `packages/vinext/src/server/prod-server.ts:1416-1427` (App Router)
- `packages/vinext/src/server/prod-server.ts:1774-1780` (Pages Router)

The namespace collision is created at build time. Vite's default `publicDir` copy writes `public/**` into `dist/client/**`, and `build.assetsDir` is `_next/static` (`packages/vinext/src/utils/asset-prefix.ts:30`), so both kinds of file land in the same directory with nothing distinguishing them. `packages/vinext/src/server/static-file-cache.ts:153-155` then applies `Cache-Control: public, max-age=31536000, immutable` to anything under that directory, so a copied public file is cached as though it were content-hashed.

`6e998ec1` (#2242) changed the miss path, so asset-shaped requests that find no file now fall through to middleware. The hit path was left as-is: an existing file is still served first.

## Reachability

Config required: none. Default `assetPrefix` (empty), default `publicDir`. The only precondition is application-side, a file at `public/_next/static/<name>`.

Request: `GET /_next/static/<name>`, unauthenticated, no special headers or encoding.

Result: 200 with the file contents and `Cache-Control: public, max-age=31536000, immutable`, middleware not invoked. `tryServeStatic` keeps its path containment and `.vite` blocking, so there is no traversal and no read outside what the app itself placed there.

Impact is bounded. Middleware acting as an auth guard for such a file is skipped, but Next.js also skips middleware for `/_next/static/*`, so an app relying on middleware there was never portable. The vinext-specific part is that the file is reachable at all.

## Runtime differences

**Node production server.** As above.

**Cloudflare Workers.** Wider, and likelier to be hit. The ASSETS binding serves any file in `dist/client` before the Worker runs, as documented in `packages/vinext/src/build/assets-ignore.ts:5-20`. The generated wrangler config sets `assets: { directory: "dist/client", not_found_handling: "none", binding: "ASSETS" }` (`packages/vinext/src/init-cloudflare.ts:194-196`, default injected at `:402`) with no `run_worker_first` — grepping `run_worker_first` across `packages/vinext`, `tests`, and `examples` returns nothing. So on a default Workers deployment *every* public-directory file bypasses middleware, not just `_next/static`. `app-router-entry.ts:135-138` and `pages-router-entry.ts:140` both encode that ordering as an assumption. The Node server's post-middleware public-file serving at `prod-server.ts:1968-1994` has no Workers equivalent, so the two runtimes disagree on this ordering.

**Dev server.** Vite's own publicDir middleware runs ahead of vinext's post-middleware handler, so dev probably behaves like Workers here. Unconfirmed — I read the middleware registration sites in `packages/vinext/src/index.ts` but did not run dev.

**Standalone output.** `packages/vinext/src/build/standalone.ts:288-295` copies `public/` separately into `standalone/public`, but runs the same `prod-server.ts`, so behavior is identical rather than additional.

Callers of `resolveAppRouterAssetPath`: `prod-server.ts:1416`, `prod-server.ts:1774`, `tests/asset-prefix.test.ts`. No others.

## Reproduction

Executed: `npx vitest run tests/asset-prefix.test.ts -t "resolveAppRouterAssetPath"`, 6 passed. `tests/asset-prefix.test.ts:220-225` already pins the classification behind this:

```ts
expect(resolveAppRouterAssetPath("/_next/static/foo-abc.js", "", "")).toBe("/_next/static/foo-abc.js");
```

The path string is the entire input, which is the defect.

Not executed, end to end:

1. Take an app fixture whose `middleware.ts` returns `new NextResponse("denied", { status: 403 })` for all paths, or sets a marker header.
2. Add `public/_next/static/private.txt` with known content, and `public/private.txt` as a control.
3. `vinext build && vinext start`.
4. `curl -i http://localhost:3000/private.txt` → 403 from middleware. Control passes.
5. `curl -i http://localhost:3000/_next/static/private.txt` → 200 with the file contents, no middleware marker.

As a test: an integration test in the prod-server suites asserting that step 5 returns the middleware response fails today.

## Possible fix direction

The cheapest option is to let the build own the namespace instead of inferring it at request time: fail or warn during build when `public/` contains anything under the configured `assetsDir`, or skip copying those files. That costs nothing on the request path and matches Next.js, where the two namespaces cannot merge.

Checking each request against the build manifest is the alternative, but it adds a lookup to the hottest path and has to account for every emitted-file category (chunks, `media/`, fonts, precompressed `.br`/`.gz` variants).

Neither addresses the Workers case. That one needs a decision about whether middleware should see public-file requests on Workers at all, which in practice means either `run_worker_first` for non-`_next/static` paths, at the cost of latency and invocations on every asset request, or documenting that public files are unprotected on Workers so app authors stop assuming otherwise.

NriotHrreion:
Closes #2778

## Root cause

In `packages/vinext/src/server/prod-server.ts:1441`, the function `resolveAppRouterAssetPath()` does not consider the case where the provided pathname that starts with `_next/static` is from public folder.
https://github.com/cloudflare/vinext/blob/255be4e2ab20f1f98a6b6fddf7a669a02335e2ed/packages/vinext/src/server/prod-server.ts#L1467-L1470

But in the upstream next.js source, it is not allowed to put a `_next` folder inside the public folder, or it will throw an error during `npm run build`.

## What changed

I created an assert function `assertNoPublicDirAssetConflict()`. This function checks the paths. Once it found there is any `_next` inside public folder, it will throw an error and interrupt the build.

This assert function is executed during `configResolved()` (in `packages/vinext/src/index.ts`):
```diff
async configResolved(config) {
+  const publicDir = config.publicDir === "" ? null : config.publicDir;
+  const assetsDir =
+    config.environments.client.build.assetsDir ??
+    config.build.assetsDir ??
+    resolveAssetsDir(nextConfig.assetPrefix ?? "");
+
+  assertNoPublicDirAssetConflict({
+    root: config.root,
+    publicDir,
+    assetsDir,
+  });

  // ...
```

The workaround is quite simple and is aligned with the behavior of upstream next.js.

## Testing

- `pnpm test tests/public-dir-conflict.test.ts`

## Notes

The error message is as following:
```ts
const PUBLIC_NEXT_CONFLICT_ERROR =
  "You can not have a '_next' folder inside of your public folder. " +
  "This conflicts with the internal '/_next' route. " +
  "https://nextjs.org/docs/messages/public-next-folder-conflict";
```
I'm not sure if the link should be replaced with vinext's one.

Thanks for reviewing!

### DIFF
--- a/packages/vinext/src/build/public-dir-conflict.ts
+++ b/packages/vinext/src/build/public-dir-conflict.ts
@@ -1,0 +1,45 @@
+import { existsSync } from "node:fs";
+import path from "pathslash";
+
+export type PublicDirConflictOptions = {
+  root: string;
+  publicDir: string | null;
+  assetsDir: string;
+};
+
+const PUBLIC_NEXT_CONFLICT_ERROR =
+  "You can not have a '_next' folder inside of your public folder. " +
+  "This conflicts with the internal '/_next' route. " +
+  "https://nextjs.org/docs/messages/public-next-folder-conflict";
+
+export function assertNoPublicDirAssetConflict({
+  root,
+  publicDir,
+  assetsDir,
+}: PublicDirConflictOptions): void {
+  if (!publicDir) return;
+
+  const resolvedPublicDir = path.resolve(root, publicDir);
+  const publicNextDir = path.join(resolvedPublicDir, "_next");
+
+  if (existsSync(publicNextDir)) {
+    throw new Error(PUBLIC_NEXT_CONFLICT_ERROR);
+  }
+
+  if (!assetsDir) return;
+
+  const publicAssetsDir = path.resolve(resolvedPublicDir, assetsDir);
+  const relativeAssetsDir = path.relative(resolvedPublicDir, publicAssetsDir);
+
+  const isInsidePublicDir =
+    relativeAssetsDir !== "" &&
+    !relativeAssetsDir.startsWith("../") &&
+    !path.isAbsolute(relativeAssetsDir);
+
+  if (isInsidePublicDir && existsSync(publicAssetsDir)) {
+    throw new Error(
+      `[vinext] The public directory contains a path reserved for build assets: ` +
+        `${relativeAssetsDir}`,
+    );
+  }
+}
--- a/packages/vinext/src/build/public-dir-conflict.ts
+++ b/packages/vinext/src/build/public-dir-conflict.ts
@@ -7,27 +7,63 @@ export type PublicDirConflictOptions = {
   assetsDir: string;
 };
 
+export type PublicNextRequestConflictOptions = Pick<
+  PublicDirConflictOptions,
+  "root" | "publicDir"
+> & {
+  basePath: string;
+  requestUrl: string;
+};
+
 const PUBLIC_NEXT_CONFLICT_ERROR =
   "You can not have a '_next' folder inside of your public folder. " +
   "This conflicts with the internal '/_next' route. " +
   "https://nextjs.org/docs/messages/public-next-folder-conflict";
+
+function assertNoPublicNextDir(root: string, publicDir: string | null): void {
+  if (!publicDir) return;
+
+  const publicNextDir = path.join(path.resolve(root, publicDir), "_next");
+  if (existsSync(publicNextDir)) {
+    throw new Error(PUBLIC_NEXT_CONFLICT_ERROR);
+  }
+}
+
+export function assertNoPublicNextRequestConflict({
+  root,
+  publicDir,
+  basePath,
+  requestUrl,
+}: PublicNextRequestConflictOptions): void {
+  const internalOrigin = "http://vinext.invalid";
+  // Concatenate origin-form targets so a leading `//` stays in the pathname
+  // instead of being interpreted as a scheme-relative host. Next handles that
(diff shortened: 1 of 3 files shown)


NriotHrreion:
Oops, it seems that my formatter just replaced all import relative paths with import alias...

PR #2817, opened 2026-08-06
fix(lint): transform filename from linter context to the absolute one for `prefer-import-alias` rule

NriotHrreion:
## Overview

`context.filename` provided by the linter is not guaranteed to be absolute, but `prefer-import-alias` just directly uses the `context.filename` to resolve the import paths, which causes unintended lint error in editor.

For example, in `packages/vinext/src/server/api-handler.ts`
```ts
import "./server-globals.js";
// Use alias 'vinext/internal/server/server-globals' instead of relative path './server-globals.js'. oxc(vinext-local(prefer-import-alias))
```

## What changed

I added a function `getPhysicalFilename(context)` to get absolute `filename` from the linter context. This function resolves `context.physicalFilename` (if there is not, fallback to `context.filename`) to absolute with cwd.

And this function is called during `createOnce()`:
```diff
+ const filename = getPhysicalFilename(context);
+ if (!filename) return;
- const aliases = loadAliasesForFile(context.filename);
+ const aliases = loadAliasesForFile(filename);
if (aliases.length === 0) return;
- const importerDir = path.dirname(context.filename);
+ const importerDir = path.dirname(filename);
```

### DIFF
--- a/oxlint-plugins/prefer-import-alias.js
+++ b/oxlint-plugins/prefer-import-alias.js
@@ -167,6 +167,30 @@ function loadAliasesForFile(filename) {
 }
 
 /**
+ * Get an absolute on-disk filename for filesystem and path comparisons.
+ *
+ * `context.filename` is not guaranteed to be absolute (notably when linting
+ * through an editor), while `physicalFilename` is the ESLint API intended for
+ * accessing the file on disk. Keep the fallback for Oxlint/ESLint callers that
+ * do not provide it, resolving relative filenames against the lint context's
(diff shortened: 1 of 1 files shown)


NriotHrreion:
@james-elicx Could you please have a review on this PR?

The oxfmt in my vscode keeps mis-fixing `prefer-import-alias` lint error when I save files. And every time, I have to let codex help me revert the changes.

I would quite appreciate it if this annoying bug could be fixed :)

> Note: I've verified the change of this PR on my PC and it just worked fine.

PR #2828, opened 2026-08-06
feat(client): expose host React to Module Federation remotes

kdy1:
## Summary

- add a browser-only `vinext/client` entry point with `getVinextReact()` so Module Federation remotes can reuse the host React instance
- register the host React instance before App Router and Pages Router client runtimes, including Pages Router development hydration
- document the required singleton sharing configuration, usage, and browser-only limitations

## Why

Module Federation remotes can evaluate with a different React module instance from the vinext host. Components that read hooks from that copy can hit React dispatcher mismatches and fail with invalid hook call errors.

The client bridge stores the first browser React instance under a global symbol and returns it to remotes explicitly. First-registration semantics also keep the host instance stable when duplicate modules are evaluated or HMR reloads the bridge. Server and RSC environments continue to use their condition-specific React instances without global registration.

## User impact

Client-side Module Federation integrations can import `getVinextReact` from `vinext/client` and resolve hooks against the vinext host React instance. Consumers still need compatible singleton `react`, `react/`, `react-dom`, and `react-dom/` sharing configuration. This does not add App Router Module Federation SSR or rewrite React imports in third-party packages.

## Validation

- `pnpm test tests/client-react-bridge.test.ts tests/entry-templates.test.ts tests/deploy.test.ts` — 369 tests passed
- `pnpm run check` — formatting, lint, type checks, Next.js type sync, and public shim checks passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,6 +367,49 @@ export default defineConfig({
 > Adding an explicit `rsc()` call fails the build with `[vinext] Duplicate @vitejs/plugin-rsc detected`.
 > Pass `rsc: false` to `vinext()` only if you want to own that registration.
 
+#### Module Federation (client-side)
+
+For client-side Module Federation, configure React and React DOM as singleton shared modules in both the host and remotes:
+
+```ts
+import { federation } from "@module-federation/vite";
+import { defineConfig } from "vite";
+import vinext from "vinext";
+
+export default defineConfig({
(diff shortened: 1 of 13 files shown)


Nsttt (approved):
This will hep us a lot to leverage https://github.com/module-federation/vinext/

PR #2834, opened 2026-08-07
fix(app-router): replay Back/Forward missed before hydration

### LINKED ISSUE #2822: App Router client navigation: replay Back/Forward traversal missed before hydration (Navigation API detection)
## Next.js Change

**Commit:** [`0fd2e51`](https://github.com/vercel/next.js/commit/0fd2e51ed65726c5b2e2e2650ce9f8ca4e007b3c)
**PR:** [#96252](https://github.com/vercel/next.js/pull/96252) (relands #95682)

## What changed

Fixes a race when the user presses **Back/Forward before the App Router's `popstate` listener is installed** (i.e. before/during hydration). The browser moves to a different history entry than the one the document was activated on, and the resulting `popstate` fires with nobody listening — so the traversal is silently lost and the UI stays on the wrong entry.

The fix uses the **Navigation API** to detect during hydration that the current history entry differs from the activation entry, and if so replays the missed traversal with logic mirroring `onPopState`.

### Mechanism (from the diff, `client/components/app-router.tsx`)

- New `hasMissedTraversal()`: returns true when `window.navigation` exists and `navigation.activation.entry.key !== navigation.currentEntry.key`, **and** `window.history.state?.__NA === true` (only app-router-written entries can be restored; others are left unhandled as before). The activation entry is fixed for the document's lifetime and entry keys are stable across `replaceState`, so a key mismatch before the listener is installed means a traversal went unobserved.
- `onPopState` logic is extracted into a standalone `handlePopState(state)` so it can be invoked both from the real event and from the missed-traversal replay. It still reloads when `!state.__NA` (entry pushed by the Pages Router) and dispatches the traverse action (via `startTransition` + `dispatchTraverseAction`) otherwise.
- Two one-shot guards, `checkedMissedTraversalBeforeHistoryWrite` and `checkedMissedTraversalBeforeReplay`:
  - In `HistoryUpdater`, before the first history write, if `hasMissedTraversal()` it **skips the write** (which would overwrite the traversed-to entry's state) and instead just records `setLastCommittedTree(tree)`.
  - In `Router`'s effect that installs the `popstate` listener, after adding the listener, if `hasMissedTraversal()` it calls `handlePopState(window.history.state)` to replay the missed traversal.

The original reland was reverted (#95853) due to a React hang under Activity applying the state update; that was fixed upstream (react/react#37135) and synced, so the fix is safe again.

## Impact on vinext

vinext reimplements App Router client-side navigation and history handling (its own `next/navigation` + router shims / browser entry). If vinext only handles `popstate` via an event listener installed after hydration, it has the same race: a Back/Forward pressed before hydration completes will be dropped, leaving the app on the wrong route/history entry with no visible navigation.

What to check/do:

1. **Detect missed traversals during hydration** using the Navigation API: compare `navigation.activation.entry.key` to `navigation.currentEntry.key`, gated on the current `history.state` being an app-router-written entry.
2. **Skip the first history write when a traversal was missed**, so the router doesn't clobber the traversed-to entry's state; still commit the rendered tree.
3. **Replay the missed traversal** right after installing the `popstate` listener, using the same handler path as a normal `popstate` (reload on non-app-router entries, otherwise dispatch the traverse/restore action).
4. **Use one-shot guards** so the missed-traversal check runs at most once for the history-write path and once for the replay path.
5. **Port the tests.** Next.js added `test/e2e/app-dir/back-before-hydration/` covering Back before hydration across plain pages, search-param routes, Suspense boundaries, and third-party `pushState`.

### Notes

- The Navigation API is available in Chromium-based browsers; the guard no-ops where `window.navigation` is undefined (Firefox/Safari), matching Next.js's behavior of leaving the traversal unhandled there.

## Related

- #2810 — App Router navigation scroll: respect root `scroll-padding-top`
- #2809 — App Router client navigation: strip hash from stored canonical URL

NathanDrake2406:
Closes #2822. Ports [vercel/next.js#96252](https://github.com/vercel/next.js/pull/96252) (`0fd2e51`).

## Problem

Navigate client-side (`/` → `/about` via `Link`), then reload. Chrome preserves the same-document association across the reload, so the entry pushed before the reload is still a same-document sibling. Pressing Back while the new document has committed but not hydrated is therefore an *instant same-document traversal*: the URL bar moves to `/`, `popstate` fires, and nobody is listening — vinext installs its listener at the end of `bootstrapHydration()`, after the `await`s in `main()` and after the entry script itself has loaded.

Reproduced against `tests/fixtures/app-basic` before the fix (scripts stalled via Playwright routing to make the window deterministic):

```
Expected: "Welcome to App Router"
Received: "About"
```

The URL said `/`, the page rendered `/about`, and every later Back/Forward moved the URL bar without changing the content. Two mechanisms kept it wrong once hydration finished:

1. `writeBootstrapHistoryMetadata()` / `writeHydratedHistoryMetadata()` replaced the traversed-to entry's own metadata (its traversal index, bfcache ids) with the activation entry's.
2. The hydration cache publication seeded the visited-response cache with the reloaded `/about` payload keyed by `window.location.href` — which was already `/`. A traversal to `/` then resolved from that cache and rendered `/about` again.

Point 2 is vinext-specific: Next.js stores the router tree in the history entry and restores from it, while vinext refetches by URL, so the mis-keyed hydration payload had to be withheld too.

## Fix

- `packages/vinext/src/server/app-browser-missed-traversal.ts` (new): `hasMissedInitialTraversal()` compares `navigation.activation.entry.key` with `navigation.currentEntry.key`, gated on the live entry being one the App Router owns.
  Ownership needs more than a traversal index: vinext's patched `history.pushState`/`replaceState` install during module evaluation — before `main()`'s awaits — and `createExternalHistoryStatePreservingMetadata()` copies `__vinext_historyIndex` onto the caller's state, so a raw push inside that window mints a new Navigation API entry key carrying inherited vinext metadata. The patched writers now record an external-write flag on the shared client navigation state (`hasObservedExternalHistoryWrite()`), and any such write disqualifies the check. Without it, a shallow/third-party push was misread as a Back/Forward: the hydration payload was discarded and the push replayed as a traverse (verified: 3 RSC fetches for the pushed URL). vinext's own writes go through `originalPushState`/`originalReplaceState` and never set the flag. The activation entry is fixed for the document's lifetime and entry keys survive `replaceState`, so a mismatch before the listener exists *is* a missed traversal. No-ops where `window.navigation` is undefined (Firefox/Safari), matching Next.js.
- `AppBrowserHistoryController.markMissedInitialTraversal()` makes both initial metadata writes no-op, leaving the traversed-to entry authoritative; the replayed traversal owns every later write.
- `bootstrapHydration()` invalidates the hydration cache publication and skips `initialPrefetchRouterState` when a traversal was missed.
- The popstate listener body is extracted into `handleAppRouterTraversal(state)` (and `createPopstateRestoreHandler` now takes a history state instead of an event) so the replay runs the exact same path as a real `popstate`. The replay re-checks detection after installing the listener, so a third-party history write in between leaves the traversal unhandled instead of replaying onto the wrong entry.

## Verification

- `tests/e2e/app-router/back-before-hydration.spec.ts` (new): fails on `main` with the output above; passes here. Three cases — the replay (including Forward/Back afterwards, and that hydration did not reload the document); a third-party `pushState` landing before the entry script runs; and a third-party `pushState` landing *after* the history patch is installed but before hydration, which is the inherited-metadata case. That third test stalls only the dev-overlay module `main()` awaits, so the entry graph has evaluated (patch installed) while hydration has not started, and asserts no RSC fetch is issued for the pushed URL. It fails with the external-write guard removed.
- `vp test run tests/shims.test.ts tests/app-browser-entry.test.ts tests/app-browser-history-controller.test.ts tests/link-status-registry.test.ts tests/pages-router-i18n-sticky-locale.test.ts` — 1552 passed, including new coverage for the predicate (traversal / same entry / third-party entry / inherited index / no Navigation API), for the suppressed initial writes, and for the shim seam itself: a raw `pushState` preserves the traversal index, sets the external-write flag, and the predicate then declines the entry.
- `playwright --project=app-router navigation.spec.ts navigation-flows.spec.ts navigation-regressions.spec.ts scroll-restoration.spec.ts hydration.spec.ts` — 33 passed.
- `playwright --project=app-router-bfcache` — 7 passed.
- Repo pre-commit gate (full `vp check`, staged unit/integration tests, knip) passed.

## Caveats

- Detection depends on the Navigation API, so Firefox and Safari keep the current behavior (traversal left unhandled), as upstream does.
- Not ported: upstream's Suspense-boundary and `cacheComponents` variants of the scenario, and the search-param variant. The e2e here exercises the shared mechanism on the `app-basic` fixture rather than adding a dedicated fixture app.

### DIFF
--- a/packages/vinext/src/server/app-browser-entry.ts
+++ b/packages/vinext/src/server/app-browser-entry.ts
@@ -138,6 +138,10 @@ import {
   restoreSynchronousPopstateScrollPosition,
 } from "./app-browser-popstate.js";
 import {
+  hasMissedInitialTraversal,
+  readBrowserNavigationEntryKeySource,
+} from "./app-browser-missed-traversal.js";
+import {
   DevRecoveryBoundary,
   GlobalErrorBoundary,
   RedirectBoundary,
@@ -1507,7 +1511,26 @@ function bootstrapHydration(
   devErrorOverlay: DevErrorOverlayModule | null,
   initialRscBootstrap: NavigationRuntimeRscBootstrap | undefined,
 ): void {
+  // Back/Forward pressed before the popstate listener exists (the awaits in
+  // main(), or the entry script still loading) is an instant same-document
+  // traversal that nobody handled: the browser now sits on a different history
+  // entry than the one this document was activated on. The initial payload
+  // describes the activation entry, so it must not seed the URL-keyed
+  // navigation caches and the initial history writes must not overwrite the
+  // traversed-to entry's metadata. The traversal is replayed once the listener
+  // is installed below.
+  const missedInitialTraversal = hasMissedInitialTraversal({
+    historyState: window.history.state,
+    navigation: readBrowserNavigationEntryKeySource(),
+  });
+  if (missedInitialTraversal) {
+    historyController.markMissedInitialTraversal();
+  }
+
   const hydrationCachePublication = createHydrationCachePublication();
+  if (missedInitialTraversal) {
+    hydrationCachePublication.invalidate();
+  }
   const cacheGeneration = clientNavigationCacheGeneration;
   const [reactBranch, cacheBranch] = rscStream.tee();
   const root = decodeAppElementsPromise(createFromReadableStream<AppWireElements>(reactBranch));
@@ -1521,6 +1544,8 @@ function bootstrapHydration(
   void Promise.all([root, initialCacheBuffer])
     .then(async ([elements, buffer]) => {
       if (cacheGeneration !== clientNavigationCacheGeneration) return;
+      // The payload belongs to the activation entry, not the live URL.
+      if (missedInitialTraversal) return;
       const metadata = AppElementsWire.readMetadata(elements);
       initialPrefetchRouterState = {
         pathAndSearch: initialPathAndSearch,
@@ -2446,7 +2471,7 @@ function bootstrapHydration(
     shouldSkipScrollRestore: (navId) => synchronousPopstateScrollRestoreNavigationId === navId,
   });
 
-  window.addEventListener("popstate", (event) => {
+  const handleAppRouterTraversal = (state: unknown): void => {
     // The browser has already applied the history entry by the time popstate
     // fires. App Router state does not include hashes, so matching the
     // committed pathname/search proves this traversal does not need a new RSC
@@ -2456,13 +2481,13 @@ function bootstrapHydration(
     const href = window.location.href;
     if (isSameAppRoutePopstateTarget(href)) {
       notifyAppRouterTransitionStart(href, "traverse");
-      historyController.commitTraversalIndexFromHistoryState(event.state);
-      restorePopstateScrollPosition(event.state);
+      historyController.commitTraversalIndexFromHistoryState(state);
+      restorePopstateScrollPosition(state);
       return;
     }
     const snapshotNavigationId = browserNavigationController.beginNavigation();
     if (
-      restoreHistoryStateSnapshot(event.state, snapshotNavigationId, () => {
+      restoreHistoryStateSnapshot(state, snapshotNavigationId, () => {
         abortSupersededNavigation();
         notifyAppRouterTransitionStart(href, "traverse");
       })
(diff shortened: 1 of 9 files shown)


NathanDrake2406:
@codex review

PR #2835, opened 2026-08-07
fix(app-router): keep usePathname in sync across back/forward around shallow routing

### LINKED ISSUE #1541: App Router: `history.pushState` with new pathname not reflected after back/forward

## Problem

After using `history.pushState` to set a custom pathname (e.g. `/pushstate-new-pathname`), and then navigating back and forward, the pathname is lost — `usePathname()` reports an older value. Distinct from #1348 (which is about middleware rewrites) — this is shallow `pushState` + history traversal.

```
history.pushState(/pushstate-new-pathname) not preserved across back/forward
```

## Estimated Impact

## Affected Test Suites

- `test/e2e/app-dir/shallow-routing/shallow-routing.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Push a shallow state with a new pathname, navigate elsewhere, then back, and assert `usePathname()` returns the originally-pushed pathname.

2. **Persist the shallow pathname in the history entry metadata.** Tie this to the history-entry restoration foundation from #1366.

---

Part of #1328.

sensei-woo:
Fixes #1541.

## Problem

`usePathname()`/`useSearchParams()` freeze on a stale value after Back/Forward when the app uses shallow routing (native `history.pushState`/`replaceState`, per the [Next.js SPA guide](https://nextjs.org/docs/app/guides/single-page-applications#shallow-routing-on-the-client)). Minimal repro and a vinext-vs-`next dev` step table are in [this comment on #1541](https://github.com/cloudflare/vinext/issues/1541#issuecomment-5219915377): the desync needs a pushState performed **after** a traversal, and then it's the second Back that loses sync — the URL moves while the hooks stay put.

## Root cause — two defects

**1. Externally written entries inherited the current entry's traversal index.** The patched `history.pushState`/`replaceState` built the new entry's state with `createExternalHistoryStatePreservingMetadata`, which copied `__vinext_historyIndex` from the entry being pushed from. Two distinct history entries then shared one slot in the index-keyed restorable-snapshot cache (`HistoryStateSnapshotCache`), so a later traversal could resolve a "restore" against the *other* entry's remembered router state — committing a router tree whose `navigationSnapshot` carries the wrong URL, and short-circuiting the full traverse navigation that would have synced from `location`.

**2. The same-route popstate fast path skipped the URL-state resync.** With distinct indices in place, a second defect surfaced: when a traversal target's pathname+search already matches the *committed router tree* (`isSameAppRoutePopstateTarget`), the popstate handler returned after committing the traversal index — but an interleaved shallow pushState had moved the navigation shim's cached URL state elsewhere, and nothing brought it back. The tree matched; the hooks didn't.

## Fix

- `shims/navigation.ts`: the patched `pushState`/`replaceState` allocate a fresh traversal index for the externally written entry through a new `allocateExternalHistoryTraversalIndex` runtime function (skipped for vinext-internal suppressed writes, and absent-runtime callers — Pages Router, pre-hydration — keep the previous copy behavior). `replaceState` allocates too: its URL rewrite makes the snapshot remembered under the old index unrepresentative of the entry it now describes.
- `client/navigation-runtime.ts`: the new optional runtime function, registered by the App Router browser entry and wired to `AppBrowserHistoryController` (`allocateNavigationHistoryTraversalIndex("push")` + `commitHistoryTraversalIndex`).
- `server/app-browser-entry.ts`: the same-route popstate fast path now runs `commitClientNavigationState()` so `usePathname()`/`useSearchParams()` resync from the restored location.
- `server/app-history-state.ts`: `createExternalHistoryStatePreservingMetadata` accepts the allocated index; `undefined` preserves the legacy copy.

## Tests

- **e2e** (`tests/e2e/app-router/advanced.spec.ts`): drives the aliasing sequence — push `/shallow-test/sub`, push `?filter=active`, Back, push again, Back, Forward — asserting `usePathname`/`useSearchParams` match the URL at every settle point. Fails on `main` (frozen pathname at the second Back), passes with the fix. The fixture gains a real `/shallow-test/sub` route so the traverse-fallback RSC navigation resolves instead of 404ing (that page's pathname was already being shallow-pushed by the existing fixture buttons).
- **unit** (`tests/app-browser-history-controller.test.ts`): allocated index is stamped (including onto metadata-less entries), legacy copy preserved without allocation, and the aliasing scenario at the controller level — an externally pushed entry gets its own snapshot slot, so restoring the original entry yields the original state.

`pnpm test`, `PLAYWRIGHT_PROJECT=app-router pnpm run test:e2e -- tests/e2e/app-router/advanced.spec.ts` (38/38), and `pnpm run check` are green.

---

Found while root-causing a glyph-soup animation bug in a production app whose account tabs shallow-route with native pushState; the frozen `usePathname` compounds with cacheComponents' hidden-Activity keep-alive (stale state keeps morphing offscreen). Happy to iterate on review feedback.

https://claude.ai/code/session_01Y4w3KwfeSpsDQWFeLnZMXd

### DIFF
--- a/.changeset/shallow-history-entry-identity.md
+++ b/.changeset/shallow-history-entry-identity.md
@@ -1,5 +1,0 @@
----
-"vinext": patch
----
-
-App Router: `usePathname()`/`useSearchParams()` no longer freeze on a stale URL after Back/Forward around shallow routing. Externally written history entries (app-called `history.pushState`/`replaceState`) now get their own traversal index instead of inheriting the current entry's, so a traversal can no longer restore the aliased entry's router snapshot; and the same-route popstate fast path resyncs the navigation shim's cached URL state from the restored location (#1541).
--- a/.changeset/shallow-history-entry-identity.md
+++ b/.changeset/shallow-history-entry-identity.md
@@ -1,0 +1,5 @@
+---
+"vinext": patch
+---
+
+App Router: `usePathname()`/`useSearchParams()` no longer freeze on a stale URL after Back/Forward around shallow routing. Externally written history entries (app-called `history.pushState`/`replaceState`) now get their own traversal index instead of inheriting the current entry's, so a traversal can no longer restore the aliased entry's router snapshot; and the same-route popstate fast path resyncs the navigation shim's cached URL state from the restored location (#1541).
--- a/packages/vinext/src/client/navigation-runtime.ts
+++ b/packages/vinext/src/client/navigation-runtime.ts
@@ -55,15 +55,21 @@ export type NavigationRuntimeNavigate = (
 
 export type NavigationRuntimeFunctions = {
   /**
-   * Allocates and commits a traversal index for a history entry created or
-   * rewritten by an app-called `history.pushState`/`history.replaceState`
-   * (shallow routing). Registered by the App Router browser entry, wired to
-   * `AppBrowserHistoryController`. The patched history methods in
-   * shims/navigation.ts call this so an externally-written entry gets its own
-   * identity instead of inheriting the current entry's traversal index, which
-   * would alias two entries onto one restorable-snapshot slot (#1541).
+   * Allocates (without committing) a traversal index for a history entry
+   * created or rewritten by an app-called `history.pushState`/
+   * `history.replaceState` (shallow routing). Registered by the App Router
+   * browser entry, wired to `AppBrowserHistoryController`. The patched history
+   * methods in shims/navigation.ts call this so an externally-written entry
+   * gets its own identity instead of inheriting the current entry's traversal
+   * index, which would alias two entries onto one restorable-snapshot slot
+   * (#1541). Allocation is a pure peek; the caller commits through
+   * `commitExternalHistoryTraversalIndex` only after the native history write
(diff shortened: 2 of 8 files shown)


sensei-woo:
Addressed all three Codex findings in b1198f3:

- **Commit after write succeeds (P2):** `allocateExternalHistoryTraversalIndex` is now a pure peek; a separate `commitExternalHistoryTraversalIndex` runs only after the native `pushState`/`replaceState` returns, so a throwing call (cross-origin URL, uncloneable state) leaves the traversal bookkeeping untouched — the peeked index is simply re-allocated by the next write.
- **State-only replaceState keeps its index (P2):** `replaceState` now allocates only when the target URL changes the app-route identity (pathname/search — App Router state does not include hashes, matching `isSameAppRoutePopstateTarget` semantics). `history.replaceState(data, "")` and hash-only replaces keep the entry's index and its remembered snapshot.
- **Manual changeset removed (P1).**

`pnpm test` (29), `PLAYWRIGHT_PROJECT=app-router` advanced.spec (38/38), and `pnpm run check` are green.

PR #2850, opened 2026-08-10
fix(app-router): support runtime instant prefetch shells

james-elicx:
## Summary

- detect runtime `unstable_instant` on pages, layouts, slots, aliases, and local re-export chains
- downgrade explicit/full prefetches to non-reusable partial shells that preserve completed cache and request-data branches while deferring `connection()` content
- keep instant shells out of navigation/CDN caches, learn exact-href optimistic templates, and invalidate them with their source prefetch
- refresh route policy in dev when route or followed config modules change

## Validation

- `vp check`
- independent review: no findings after two fix/review cycles
- 1,528 focused cache/render/analyzer tests plus the external-config HMR regression
- targeted Next.js E2E: `test/e2e/app-dir/prefetch-true-instant/prefetch-true-instant.test.ts` — 2/2 passed

Failure source: Actions run 31290819291.


james-elicx:
Mapping note for the cache-components backlog from Actions run `31290819291`: I think this PR fixes both `test/e2e/app-dir/prefetch-true-instant/prefetch-true-instant.test.ts` failures. The targeted Next.js wrapper passes 2/2.

PR #2855, opened 2026-08-10
fix(cache): support segment prefetch inlining

james-elicx:
## Summary

- measure concrete prerendered Flight segment sizes without rerunning user components
- persist and inject validated prefetch-inlining hint trees for Cache Components builds
- preserve worker affinity and clean up per-render head measurements on every render path
- add local production parity coverage for the upstream prefetch-inlining suite

## Upstream parity target

`test/e2e/app-dir/segment-cache/prefetch-inlining/prefetch-inlining.test.ts`

## Validation

- targeted Next.js wrapper: 7/7 tests, 7/7 snapshots
- local production Playwright parity E2E: 1/1
- focused Vitest suites: 296/296
- targeted format, lint, and types: clean
- independent final review: no findings


james-elicx:
Addressed every actionable Bonk item in 22a52d19b:

- children-slot pages now reuse the existing slot measurement, with a regression assertion that the cached React node is rendered exactly once
- affinity-scoped head data is deleted before route mismatch/route-handler early returns
- verified the head-inlining gate directly against current Next.js packages/next/src/server/app-render/collect-segment-data.tsx: upstream uses !hasChildren at its leaf gate, matching this implementation

Re-ran the exact targeted Next.js wrapper: 7/7 tests and 7/7 snapshots.

james-elicx:
Final follow-up on a64a91547:

- fixed Check by registering the generated-entry runtime with Knip; local vp run knip and full vp run check pass
- removed the feature-only prefetch measurement runtime from generated App Router entries when prefetchInlining is disabled, addressing the default RSC/server bundle regression with codegen coverage
- exact targeted Next.js wrapper remains 7/7 tests and 7/7 snapshots on this head

james-elicx:
Addressed the exact-head Bonk finding in 4862dcbc5:

- a 404/no matching route from build-time hint collection now soft-skips hints without losing the successfully rendered route
- non-404 collector failures remain fatal
- added a regression test covering the 404 path and documented pattern as the explicit precomputed-tree join key

Validation: 326/326 affected tests, full check + Knip, and the exact targeted Next.js wrapper at 7/7 tests + 7/7 snapshots.

james-elicx:
Backlog mapping for Actions run `31290819291`: this PR appears to fix all three reported `test/e2e/app-dir/segment-cache/prefetch-inlining/prefetch-inlining.test.ts` failures (outlined chain, restart groups, and concrete dynamic-route hints). The exact targeted Next.js wrapper passes 7/7 with 7/7 snapshots at head `4862dcbc59f39251fea76d9c88652c9dd630ab04`.

PR #2866, opened 2026-08-10
fix(cache): keep encoded dynamic prefetches learning-only

james-elicx:
## Summary

- keep Cache Components automatic dynamic-route prefetches learning-only so encoded path identity remains authoritative at navigation time
- dedupe pending and fresh learning-only Link prefetches, while evicting and refetching settled entries after expiry
- cover pending, settled-fresh, and expired Link remount behavior

## Original deploy-suite rows

Next.js v16.2.6 suite: test/e2e/app-dir/segment-cache/encoded-slash-params/encoded-slash-params.test.ts

- unencoded param > back navigation does not refetch the route tree
- encoded slash in param > back navigation does not refetch the route tree

## Verification

- exact upstream v16.2.6 baseline: 0/2 passed
- exact upstream v16.2.6 with this patch: 2/2 passed
- focused link/navigation units: 274/274 passed
- scoped format, lint, and type checks passed for all 4 changed files
- git diff --check passed

Draft pending CI, Big Bonk, and an exact-head deploy-suite audit.

### DIFF
--- a/packages/vinext/src/shims/internal/app-route-prefetch-policy.ts
+++ b/packages/vinext/src/shims/internal/app-route-prefetch-policy.ts
@@ -26,6 +26,7 @@ declare global {
 const __basePath: string = process.env.__NEXT_ROUTER_BASEPATH ?? "";
 
 const linkPrefetchRouteTrieCache = createRouteTrieCache<VinextLinkPrefetchRoute>();
+const ENCODED_PATH_DELIMITER_RE = /%(?:2f|5c)/i;
 
 /**
  * How an App Router prefetch for a given href should behave: whether to issue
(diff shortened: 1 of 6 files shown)


james-elicx:
Final exact-head deploy-suite audit for `804e16fdf5cce928b898fc40a1ae50db0c0cc266`:

- Full deploy suite: https://github.com/cloudflare/vinext/actions/runs/31398197230
- Report totals: **2619 passed / 185 failed / 631 skipped** (nightly baseline `31350589613`: 2617 / 187 / 631)
- Both encoded-slash-param target assertions pass.
- All five pass-to-fail rows from the previous #2866 deploy run `31393161755` are recovered (the interception-route row plus all four vary-params rows).
- Assertion-level comparison against baseline surfaced two apparent pass-to-fail rows, both cleared as ambient noise:
  - `reuses dynamic data within the per-page stale time window`: passed in exact same-head targeted execution and passed again on clean tracked main; it has also failed on unrelated run `31382866933`.
  - `should behave consistently on recursive rewrites`: reproduced identically on exact same-head and clean tracked main (`Home Page` instead of `About Page`), and also appears on unrelated run `31391938306`.
- Big Bonk LGTM: https://github.com/cloudflare/vinext/pull/2866#issuecomment-5241672666
- All 66 PR checks are green; GitHub reports the PR mergeable and clean.

Audit verdict: **no regression attributable to #2866**.

PR #2878, opened 2026-08-11
test(compat): classify legacy Edge Runtime suite unsupported

james-elicx:
## Summary

- remove the route-specific runtime-isolation implementation from this PR with additive revert commits
- classify the mixed Next.js dual-runtime suite as unsupported at the compatibility-policy boundary
- keep all raw deploy-suite results visible while excluding this non-portable file from the supported parity rate

## Exact failure mapping

Baseline run [31439707085](https://github.com/cloudflare/vinext/actions/runs/31439707085), job [93624401572](https://github.com/cloudflare/vinext/actions/runs/31439707085/job/93624401572), failed exactly these three assertions from `test/e2e/app-dir/next-after-app-deploy/index.test.ts`:

- `after() in edge runtime > triggers revalidate from a page`
- `after() in edge runtime > triggers revalidate from a server action`
- `after() in edge runtime > triggers revalidate from a route handler`

The file contains eight assertions total. All four `nodejs` cases already pass, and the edge middleware case already passes because it derives its route identity from the request URL. Therefore this file has three legacy Edge Runtime failures and five existing passes.

## Why this is unsupported

The upstream fixture exports `runtime = "edge"` from the edge layout and route, then derives its revalidation identity from `process.env.NEXT_RUNTIME`. Next.js compiles separate Node.js and legacy Edge Runtime route bundles, so that constant differs per route graph.

Vinext deliberately builds one Workers/Node-compat server graph and defines `process.env.NEXT_RUNTIME` as `"nodejs"` for that graph. Cloudflare Workers are the real edge deployment target; vinext does not intend to recreate a second Vercel legacy Edge Runtime module graph solely to expose a route-specific compile-time constant. Workers behavior should be covered by checked-in Cloudflare fixtures instead.

The compatibility manifest is file-level. Because upstream mixes the portable Node.js assertions with legacy Edge Runtime variants in one file, the policy necessarily classifies all eight rows together. Raw pass/fail totals remain visible in the UI.

## Validation

- `vp test run tests/compatibility-support.test.ts` — 5/5 passed
- `vp check apps/web/app/compatibility/suite-support.ts tests/compatibility-support.test.ts` — format, lint, and types passed
- `git diff --check` — passed

The previous product implementation remains in history for auditability and is removed only through additive revert commits; this branch was not force-pushed.


james-elicx:
Independent exact-head review of `d9672c11c88047ae12f680b484b6b10b27432286`: **NO FINDINGS**.

I reviewed the full eight-file diff independently, including static route-runtime inheritance, query-qualified RSC module identity and dependency propagation, client/framework graph sharing, `process.env.NEXT_RUNTIME` AST folding, route manifest code generation, dynamic loaders, generateStaticParams sources, and the final Windows path canonicalization fix.

The design matches Next.js's per-route compile-time runtime model and keeps the edge identity scoped to project-local RSC modules. The existing immutable targeted proof for `test/e2e/app-dir/next-after-app-deploy/index.test.ts` passed 8/8, including edge and node page, action, route-handler, and middleware revalidation.

Fresh exact-head validation:
- focused runtime/compiler/manifest/production tests: **145/145 passed**;
- scoped format, lint, and types: clean;
- `git diff --check`: clean;
- remote and local head SHAs matched.

I found no actionable runtime-parity gap, graph-isolation regression, or missing high-value test.

james-elicx:
The latest Big Bonk server-reference finding does not reproduce. The upstream fixture already exercises the claimed path directly: `test/e2e/app-dir/actions/app/client/edge/page.js` is a `"use client"` module, imports and invokes functions from the sibling `"use server"` actions module, and exports `runtime = "edge"`. On exact head `f34f8e36539719b88288fe9568d1255745ae2b24`, the targeted deploy-mode run of `test/e2e/app-dir/actions/app-action.test.ts` passed 70/70 runnable tests with no retry, including `Edge SSR > should allow cookie and header async storages` against `/client/edge`. The same test also passed once with no retry in the original full deploy suite (run 31470520106). No server-reference change is warranted without a failing reproduction.

james-elicx:
The original route-runtime isolation implementation has been intentionally replaced with a manifest-only policy classification. The additive revert commits preserve the full implementation history without force-pushing. The reason is architectural: these three failures require Next.js dual Node.js/legacy Edge Runtime bundles and a route-specific `process.env.NEXT_RUNTIME`, while vinext deliberately uses one Workers/Node-compat server graph and covers Workers as the actual edge target.

Issue #1503, opened 2026-05-22
Build: transitive externals resolve to wrong package version


james-elicx:
> *This issue was created by an agent analysing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/26282973805) (vinext `main` vs Next.js `v16.2.6`, 2026-05-22).*

## Problem

When `serverExternalPackages` (or equivalent) is configured, transitive dependencies resolve to the wrong version. Test expects `depB: 4.17.21, 4.17.21, 5.0.0` but got `depB: 3.10.1, 4.17.21, 5.0.0` — the first depB instance is wrongly resolved to depA's version.

```
Transitive externals resolution picks wrong package version
```

## Estimated Impact

~1 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/externals-transitive/externals-transitive.test.ts`

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a fixture with depA -> depB@4 and root -> depB@3, mark depA as external, and assert each module sees the correct depB version.

2. **Preserve transitive version resolution.** Externalising a package must not collapse nested copies of a transitive dep to a single version.

PR #2887, opened 2026-08-11
fix(build): preserve transitive external versions

james-elicx:
## Summary

- mirrors Next.js baseResolveCheck behavior for serverExternalPackages by comparing importer-relative and project-root package resolution
- bundles nested package instances when relocation would otherwise collapse them to the root version
- applies the same merged default and explicit external set, including transpile-package exclusions
- limits the resolver to server environments and native-filtered bare package imports

## Failure mapping

Fixes the non-cache externals-transitive failure from Actions run 31439707085, job 93624401572:

- test/e2e/externals-transitive/externals-transitive.test.ts
- before: depB resolved lodash 3.10.1 instead of its nested 4.17.21
- after: exact Next.js v16.2.6 targeted deploy suite passes 1/1

## Validation

- exact Next.js targeted E2E: 1 suite, 1 test passed
- local production-build regression proves both an explicit external and default pg-style external preserve distinct root/nested versions
- adjacent external-package suites: 6/6 passed
- vp check on changed files
- vp run vinext#build
- independent review loop: no findings

Refs #1503.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -1422,7 +1422,7 @@ export default function vinext(options: VinextOptions = {}): PluginOption[] {
   let hasCloudflarePlugin = false;
   let warnedInlineNextConfigOverride = false;
   let hasNitroPlugin = false;
-  let nitroTraceDepsFromServerExternals: string[] = [];
+  let resolvedServerExternalPackages: string[] = [];
   let isServeCommand = false;
   let pagesOptimizeEntries: string[] = [];
   const pagesClientAssetsOutputDirs = new Set<string>();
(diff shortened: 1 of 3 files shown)


james-elicx:
NO FINDINGS — full independent closure review of exact head `d5dfeb963b396b71edac017a631b09e6a6c28903`.

I reviewed the complete diff and the queried-importer follow-up. The build-only, non-client resolver strips Vite query/hash suffixes from absolute importer ids, compares canonical importer-relative and project-root package instances, and returns an absolute id only when the nested copy differs so Rolldown bundles that copy instead of collapsing it to the root external. Same-instance packages remain external. The package filter, import/require resolver split, symlink realpath handling, missing-root fallback, Nitro trace-dependency rename, and dev/preview/build boundaries do not expose a correctness or regression gap.

Validation on this exact SHA:
- Required upstream `test/e2e/externals-transitive/externals-transitive.test.ts` — 1/1 passed, retry 0/0.
- Local `tests/externals-transitive.test.ts` — 1/1 passed, covering a queried importer plus distinct explicit `shared-version` and default `pg` versions in production.
- Targeted `vp check` — clean across all 3 changed files.
- [Final Big Bonk review](https://github.com/cloudflare/vinext/pull/2887#pullrequestreview-4902807485) is attached to this exact SHA and reports no blocking issues.
- All CI is green, the PR is mergeable, the remote head matches, and the worktree remained clean.

PR #2890, opened 2026-08-11
test(compat): cover Worker import conditions

james-elicx:
## Why

The exact Next.js suite `test/e2e/import-conditions/import-conditions.test.ts` remains **2 of 12 passing**. Its other ten assertions assume the Next compiler's per-route node/edge split and private `edge-light` condition. The vinext targeted harness intentionally uses the Node adapter, so its server paths correctly select `node`; emulating Next compiler labels there would be wrong. Cloudflare Workers instead activate the portable `worker` and `workerd` conditions.

This keeps the exact suite in the `needs-vite-equivalent` bucket and records the passing Vite-equivalent coverage against a real Wrangler Worker.

## Coverage

- Uses a local dependency with conditional package `exports`, matching the upstream resolver/externalization path.
- Reverses `worker` and `workerd` key order across two exports so both active conditions are independently verified.
- The equivalent Worker test covers middleware, Pages and App API routes, Pages and App rendering, client hydration, and server actions.
- Verifies browser bundles select `browser`.

## Validation

- Exact Next.js v16.2.6 targeted suite: 2 passed, 10 failed; all server/middleware mismatches resolve to `node`, while client `browser` and RSC `react-server` behavior are correct.
- Equivalent Worker test: `tests/e2e/cloudflare-workers/route-handler-draft-cache.spec.ts` passes against Wrangler (the complete spec has 5 passing tests; one is the import-condition equivalent).
- `vp test run tests/compatibility-support.test.ts`: 5 passed.
- Targeted `vp check`: clean.
- Independent review: no findings after the package-exports parity gap and result-count wording were corrected.


james-elicx:
Independent policy/parity review at exact head `e32186d6bd2dc7fa34e511a5b441c3aacdbbcc86`: **NO FINDINGS**.

Verified independently from a clean detached exact-head checkout:
- Required targeted wrapper reproduced the precise Next-suite baseline for `test/e2e/import-conditions/import-conditions.test.ts`: **2 passed, 10 failed, 0 skipped, retry 0/0**.
- Every failure is the expected Next-compiler-private condition mismatch: vinext’s targeted Node adapter selects `node` for server/middleware paths where Next asserts `browser`/`edge-light`; client `browser` and RSC `react-server` behavior remain correct.
- The real Wrangler Worker equivalent passed **5/5**. Its condition test covers middleware, Pages/App APIs, Pages SSR/hydration, App RSC/hydration, server actions, and browser bundles; reversed export-map order independently proves both active `worker` and `workerd` conditions.
- Manifest status correctly remains `needs-vite-equivalent`, with a precise feature/reason and a focused non-brittle policy test; `tests/compatibility-support.test.ts` passed **5/5**.
- `vp check` and `git diff --check` pass; remote PR head and merge state remain exact/CLEAN.

PR #2912, opened 2026-08-14
fix(headers): expose live view of headers to api

### LINKED ISSUE #2908: next/headers: make headers() a live view of the incoming request (observe writes, hide internal headers on read)
## Upstream change

Next.js restored `headers()` so it reads through to the underlying incoming request instead of returning a detached snapshot.

- Commit: [`3c97df5`](https://github.com/vercel/next.js/commit/3c97df56ead9d1df81b36f891ba5ac0724c4eec0) "Restore the live `headers()` view of the incoming request" (#97166)
- fixes vercel/next.js#97049, closes vercel/next.js#97145

## Background

Between vercel/next.js#94703 and vercel/next.js#95116, `headers()` became detached from the request it describes:

- #94703 started stripping dev-only request-id headers alongside flight headers, but `HeadersAdapter.from` returns the `Headers` instance unchanged (no copy), so the deletes mutated the shared `req.headers`.
- #95116 fixed the mutation by **copying** the headers before stripping — but the copy detached the sealed view from the live request.

Result (since 16.3.0): a Proxy/middleware that writes a header onto `request.headers` and reads it back via `headers()` sees the value from *before* the write. Reading the same header directly off `request.headers` returns the new value, so the two APIs disagree. Whether a write is observed depends on ordering, because the view is built lazily on first access.

## Correct behavior (what to implement/verify)

1. `headers()` must be a **live view** of the incoming request — writes made to the request (e.g. by middleware/proxy) are reflected in `headers()`, regardless of ordering relative to the first `headers()` call.
2. A second `headers()` call must return the same live view (not a stale snapshot).
3. Internal/framework headers (flight headers, dev request-id headers) must stay **on the request** for framework plumbing, but stay **hidden from userland `headers()`** on read — even when written after the first `headers()` call. Hiding is done on *read* (via a sealed view that omits a set of header names), not by a one-time delete.
4. `forEach` on the sealed view must pass the **sealed proxy** as the `parent` argument, not the unsealed target (otherwise the callback gets a mutable handle that defeats the seal). `getSetCookie` must also respect the hidden set.
5. `cookies()` is unaffected — it stays a snapshot because `RequestCookies` parses the `cookie` header at construction time.

## Upstream test to port

`test/e2e/app-dir/proxy-headers-live-view/` — a proxy writes `set-during-proxy` onto `NextRequest.headers` and asserts via a `/probe` route:

```js
{
  valueBeforeMutation: null,
  valueOnRequest: 'set-during-proxy',
  valueOnFirstView: 'set-during-proxy',
  valueOnSecondView: 'set-during-proxy',
  sameView: true,
  internalHeaderOnRequest: 'set-during-proxy',
  internalHeaderOnView: null,
  internalHeaderIsIterated: false,
}
```

Also port the `HeadersAdapter` unit tests (`packages/next/src/server/web/spec-extension/adapters/headers.test.ts`, +130 lines) covering the omit-on-read seal, `forEach` parent, and `getSetCookie`.

## vinext relevance

vinext reimplements `next/headers`. We should verify our `headers()` shim reflects request mutations made by middleware and hides internal headers on read (not via one-time deletes), matching this parity fix. Check dev/prod parity across `entries/app-rsc-entry.ts`, `server/dev-server.ts`, `server/prod-server.ts`, and `cloudflare/worker-entry.ts`.

NriotHrreion:
Closes #2908. Ports https://github.com/vercel/next.js/pull/97166 (https://github.com/vercel/next.js/commit/3c97df56ead9d1df81b36f891ba5ac0724c4eec0)

## Overview

Next.js restored `headers()` as a live, read-only view of the incoming request after a regression made it a detached snapshot. The snapshot caused ordering-dependent behavior: headers written by Proxy or middleware after the first `headers()` call were visible on `request.headers` but not through `headers()`.

This PR aligns vinext with the restored nextjs semantics. Request header mutations remain observable through the same `headers()` view, framework-internal headers remain unavailable to userland, and `cookies()` retains its snapshot behavior.

## What changed

- No longer making a cleaned copy of the original Headers. Instead, return the sealed one.

```diff
function _getReadonlyHeaders(ctx: HeadersContext): Headers {
  if (!ctx.readonlyHeaders) {
-   const cleaned = new Headers(ctx.headers);
-   for (const header of FLIGHT_HEADERS) {
-     cleaned.delete(header);
-   }
-   cleaned.delete(NEXT_REQUEST_ID_HEADER);
-   cleaned.delete(NEXT_HTML_REQUEST_ID_HEADER);
-   ctx.readonlyHeaders = _sealHeaders(cleaned);
+   ctx.readonlyHeaders = _sealHeaders(ctx.headers, ...);
  }

  return ctx.readonlyHeaders;
}
```

- `_sealHeaders()` and `_createHidingHeaderMethods()`: Wrap the original Headers, and proxy each method of it to dynamically hide the internal headers.

- `applyMiddlewareRequestHeaders()`:
  - **Before:** Directly set the `ctx.headers` to the upcoming new headers reference. So the old values still hold the old reference, not updated.
  - **After:** Use the new `replaceHeadersContents()` to loop through the upcoming headers and do updates on the original `ctx.headers`.

- For cookies (`rebuildCookiesFromHeader()` and `mergeMiddlewareSetCookies()`):
  - **Before:** Loop through the upcoming cookies and do updates on the original `ctx.cookies`, merging old cookies and new cookies.
  - **After:** Directly set the `ctx.cookies` to the upcoming new cookies reference. So the old values still retain the old, not-updated reference, which aligns with the upstream nextjs's behavior.

`headers()` returns a live view of the real headers, dynamically proxied to hide the internal headers, while `cookies()` still returns a readonly copy of the real cookies.

## Testing

- `pnpm test tests/shims.test.ts`

### DIFF
--- a/packages/vinext/src/server/app-rsc-handler.ts
+++ b/packages/vinext/src/server/app-rsc-handler.ts
@@ -1111,15 +1111,10 @@ async function handleAppRscRequest<TRoute extends AppRscHandlerRoute>(
       options.clearRequestContext();
       return notFoundResponse();
     }
-    let addedSourceHeader = false;
     for (const [name, value] of sourceHeadersContext.headers) {
       if (!targetRequestHeaders.has(name)) {
         targetHeadersContext.headers.set(name, value);
-        addedSourceHeader = true;
       }
-    }
-    if (addedSourceHeader) {
-      targetHeadersContext.readonlyHeaders = undefined;
     }
     if (sourceMiddlewareResult.rewritten) {
       // Rewrites such as locale insertion are valid only when they resolve to
--- a/packages/vinext/src/shims/headers.ts
+++ b/packages/vinext/src/shims/headers.ts
@@ -157,27 +157,23 @@ function setCookieNameValue(setCookie: string): { name: string; value: string } 
 }
 
 function rebuildCookiesFromHeader(ctx: HeadersContext, cookieHeader: string | null): void {
-  ctx.cookies.clear();
-  if (cookieHeader === null) return;
-
-  const nextCookies = parseEdgeRequestCookieHeader(cookieHeader);
-  for (const [name, value] of nextCookies) {
-    ctx.cookies.set(name, value);
-  }
+  ctx.cookies = cookieHeader === null ? new Map() : parseEdgeRequestCookieHeader(cookieHeader);
 }
 
 function mergeMiddlewareSetCookies(ctx: HeadersContext, rawHeader: string | null): boolean {
   if (rawHeader === null) return false;
 
-  let merged = false;
+  let nextCookies: Map<string, string> | null = null;
   for (const setCookie of splitMiddlewareSetCookieHeader(rawHeader)) {
     const entry = setCookieNameValue(setCookie);
     if (!entry) continue;
-    ctx.cookies.set(entry.name, entry.value);
-    merged = true;
+    nextCookies ??= new Map(ctx.cookies);
+    nextCookies.set(entry.name, entry.value);
   }
 
-  return merged;
+  if (!nextCookies) return false;
+  ctx.cookies = nextCookies;
+  return true;
 }
 
 function _getState(): VinextHeadersShimState {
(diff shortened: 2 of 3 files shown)


NriotHrreion:
It seems the failed e2e test is not directly related to the changes of this PR

PR #2915, opened 2026-08-14
fix(isr): ignore interception context on HTML renders

### LINKED ISSUE #2780: Client-supplied X-Vinext-Interception-Context is folded into the RSC ISR cache key on HTML renders, where interception is disabled
A client-supplied `X-Vinext-Interception-Context` header is folded into the RSC ISR cache key on the **HTML** render path, where interception is disabled. The direct render's captured RSC payload therefore gets written under a key reserved for intercepted renders, and a later intercepted RSC navigation reads it back.

## Relationship to #2732 and #2733

Distinct from both, and not fixed by either.

#2732 (merged, `81db6e314`) stops a Route Handler from being promoted as the concrete interception source route. #2733 (open) runs middleware against the claimed source pathname before rendering from it. Both harden interception-match *resolution*, and both only run for RSC requests — #2733's new block is guarded by `filesystemRouteEligible && isRscRequest && interceptionContextHeader !== null`.

This issue is about a non-RSC HTML request, where `resolveAppPageInterceptState` short-circuits before any of that machinery runs. The two PRs govern what gets rendered; this one is about which cache key the render is stored under.

## Possible root cause

`packages/vinext/src/server/app-rsc-handler.ts:1252` passes the raw `interceptionContextHeader` into `dispatchMatchedPage` with no `isRscRequest` gate, unlike every other consumer of that header. It reaches the renderer via `app-page-dispatch.ts:1145`, and on the HTML branch `app-page-render.ts:1198` hands it to `finalizeAppPageHtmlCacheResponse`, which builds the RSC key at `app-page-cache-finalizer.ts:132-137`:

```ts
const rscKey = options.isrRscKey(
  options.cleanPathname,
  null,
  undefined,                    // renderMode -> defaults to NAVIGATION
  options.interceptionContext,  // client header
);
```

and writes the captured RSC payload under it at `app-page-cache-finalizer.ts:195-205`.

Interception itself never fires on this path: `app-page-request.ts:661-663` returns `{ kind: "none" }` when `!isRscRequest`, so the direct target route renders. `appIsrRscKey` (`isr-cache.ts:446-465`) still turns the header into a `source:<fnv1a64>` key variant, because `normalizeAppPageInterceptionProofPathname` only checks that the value is pathname-shaped, not that an interception occurred. The HTML-triggered background regeneration write at `app-page-cache.ts:455-461` reuses the same forged context.

The header is normalized for every request, RSC or not (`app-rsc-request-normalization.ts:232`), and nothing nulls it for the HTML path.

## What reaches it

Config: a production build (`isProduction` gates both `shouldReadAppPageCache` at `app-page-dispatch.ts:505-522` and the RSC capture in `app-page-render.ts:882`), a shared ISR cache adapter, and a cacheable App Router page that is the target of an intercepting route — `app/photos/[id]/page.tsx` with `revalidate`, plus `app/feed/@modal/(.)photos/[id]/page.tsx`. Draft mode, `force-dynamic`, or dynamic API usage during render all skip the write.

Request: a plain unauthenticated `GET /photos/42` — no `RSC: 1`, no `.rsc` suffix — carrying `X-Vinext-Interception-Context: /feed`.

Result: the direct `/photos/[id]` render is stored at `appIsrCacheKey("/photos/42", "rsc:source:<fnv1a64('/feed')>")`. That is the key a legitimate intercepted RSC navigation reads at `app-page-cache.ts:346-352` — same pathname, null mounted slots, and NAVIGATION render mode on both sides (explicit on the read, defaulted on the write).

Victim: a user clicking `/photos/42` from `/feed` receives the direct page tree instead of the modal tree. Wrong UI, not disclosure — the payload is a page the user could have requested directly anyway.

Persistence: until that entry's `revalidate`/`expire` or a tag revalidation. One request re-poisons it, and the legitimate intercepted render writes the same key, so it is a race an attacker can re-enter cheaply.

The same forgery over RSC does not collide: with a matching source context the RSC path genuinely intercepts and writes the correct tree, and with a non-matching one the resulting key is never read. The HTML path is the vector precisely because interception is off there.

## Reproduction

The request path above comes from reading `main` at `c9a4a843c` — I did not run a server or a full request.

The key collision itself I did execute, against `appIsrRscKey` with the arguments the two call sites pass:

```
appIsrRscKey("/photos/42", null, undefined,     "/feed")  // HTML write, forged header
  -> app:/photos/42:rsc:source:e94cc6a403e844d0
appIsrRscKey("/photos/42", null, "navigation",  "/feed")  // intercepted RSC read
  -> app:/photos/42:rsc:source:e94cc6a403e844d0
appIsrRscKey("/photos/42", null, undefined,     null)     // direct render, no header
  -> app:/photos/42:rsc
```

The forged HTML write and the intercepted RSC read produce the same key, and both differ from where a direct render belongs.

The cheapest failing test fits the existing harness in `tests/app-page-dispatch.test.ts`, where the cases at lines 2380-2450, 3007, and 3102 already stub `isrRscKey` as:

```ts
isrRscKey(pathname, mountedSlotsHeader, _renderMode, interceptionContext) {
  return `rsc:${pathname}:${mountedSlotsHeader ?? "none"}:${interceptionContext ?? "none"}`;
}
```

A new case through `createDispatchOptions` with `isRscRequest: false`, `isProduction: true`, `revalidateSeconds: 60`, `cleanPathname: "/photos/123"`, `interceptionContext: "/feed"`, and an SSR handler that populates `captureOptions.capturedRscDataRef` should assert the RSC write lands on `rsc:/photos/123:none:none`. Against current code it lands on `rsc:/photos/123:none:/feed`.

There is no coverage of this branch today: every existing interception cache test sets `isRscRequest: true`, and `tests/isr-cache.test.ts` has no test for the `source:` key variant at all.

End to end: build the two-route fixture above, serve the production build, `curl -H 'X-Vinext-Interception-Context: /feed' https://host/photos/42`, then issue the intercepted navigation (`RSC: 1` plus the same header) and check for `x-vinext-cache: HIT` with the direct-render payload.

## Possible fix direction

The narrow version gates the header at the dispatch boundary — `isRscRequest ? interceptionContextHeader : null` where `app-rsc-handler.ts:1252` sets `interceptionContext` — which matches the existing `isRscRequest` gates on `renderMode` and `clientReuseManifest` in the same normalization step, and covers all four downstream key builders since they share that one value.

The broader version keys off proven render identity instead of the request header: `AppPageRenderIdentity.interception.sourceMatchedUrl` (`app-page-render-identity.ts`) is non-null only when a slot actually intercepted. That closes the class rather than the instance, at the cost of plumbing the identity into the finalizer.

Either way the change only redirects HTML-captured RSC writes from a source-variant key to the plain `rsc:` key, which is where a direct render belongs. Existing intercepted-cache tests all set `isRscRequest: true` and should be unaffected; entries under the old keys expire on their own. Worth checking first whether any HTML-triggered background regeneration depends on refreshing an intercepted variant (`app-page-cache.ts:455-461`).

JamesbbBriz:
## Summary

- ignore client-supplied interception context when dispatching plain HTML page renders
- preserve interception context for legitimate RSC navigation and Server Action requests
- lock the real cache-write contract for both fresh HTML capture and stale background regeneration: RSC data lands on `rsc:/photos/123:none:none`, never a forged intercepted variant
- add a regression test matching Next.js hard-refresh behavior for interception routes

Closes #2780

## Tests

- `vp test run tests/app-rsc-handler.test.ts tests/app-page-dispatch.test.ts tests/isr-cache.test.ts tests/fetch-cache.test.ts tests/kv-cache-handler.test.ts` — 524 passed
- focused fresh/stale cache-key regressions — 2 passed
- `vp check` on all changed files
- `git diff --check`

The existing Server Action regression also verifies that legitimate actions retain the raw interception source before the HTML-only dispatch gate.


JamesbbBriz:
Rebased onto latest `main` (1 conflict resolved in `server/app-rsc-handler.ts`): kept upstream's newer `interceptionId: interceptionIdHeader` field and applied this PR's RSC-gating only to `interceptionContext`, as originally scoped. Reviewer note: if the newer `interceptionId` mechanism should also be ignored on plain HTML renders, that'd be an explicit follow-up.

Verified locally: `vp test run tests/app-rsc-handler.test.ts tests/app-page-dispatch.test.ts tests/isr-cache.test.ts tests/fetch-cache.test.ts tests/kv-cache-handler.test.ts` — **576 passed** (52 more than at original submission; upstream additions all green).

james-elicx:
/bigbonk review for issues

You have a hard 10-minute workflow time limit. Review the cumulative diff at the current PR head, prioritize correctness, Next.js parity, split response-stage parity, and cache-key safety, then post an explicit LGTM/no-findings verdict before the deadline.

PR #2758, opened 2026-07-30
test(app-router): cover consumed prefetch cache handoff

NathanDrake2406:
## Summary

- add a deterministic App Router browser regression for the consumed-prefetch to committed-cache handoff
- delay the redundant cache tee after a full prefetch is consumed and assert that remounting the destination Link does not issue a duplicate RSC request
- retain the existing upstream Next.js five-minute parallel-route reuse coverage

## Why

The original production fix on this branch was superseded by the broader cache-publication work merged in #2760. After merging current `main`, the conflict was resolved in favor of that newer implementation, leaving this PR as focused regression coverage for the same race.

The test remains useful because it deterministically delays the cache branch that previously created the false-miss window. Current `main` passes by publishing the already-buffered consumed-prefetch snapshot without waiting for that redundant branch to drain.

Upstream parity reference:

- `test/e2e/app-dir/app-client-cache/client-cache.parallel-routes.test.ts`
- https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/app-client-cache/client-cache.parallel-routes.test.ts

Closes #2756.

## Validation

- `vp check tests/e2e/app-router/nextjs-compat/client-cache.spec.ts`
- targeted browser regression: 1/1 passed
- targeted browser regression repeated 10 times in parallel: 10/10 passed

PR #2926, opened 2026-08-15
test(app-router): cover cross-path hash push URL replacement

### LINKED ISSUE #2809: App Router client navigation: strip hash from stored canonical URL to avoid double-hash (`#foo#bar`) on cross-hash same-route nav
## Next.js Change

**Commit:** [`459617a`](https://github.com/vercel/next.js/commit/459617a12598de63c34cec4d5c267207c591d13f)
**PR:** [#93132](https://github.com/vercel/next.js/pull/93132)

## What changed

Fixes a client-navigation bug where the URL hash is appended twice, producing a malformed URL like `/abc#foo#bar`.

Repro (from the PR):

1. From `/`, navigate to `/abc#foo`
2. On `/abc#foo`, click a link to `/abc#bar`
3. The URL becomes `/abc#foo#bar` instead of `/abc#bar`

Fixes vercel/next.js#93126 and vercel/next.js#95551.

### Mechanism (from the diff)

The fix changes what gets stored in the segment cache for a freshly-navigated route. When creating the cache entry for an unknown route, the canonical URL is now stored **without** its hash:

```ts
// packages/next/src/client/components/segment-cache/navigation.ts
createRouteCacheEntry(
  navigationSeed.routeTree,
  metadataVaryPath,
  couldBeIntercepted,
  // Store a hashless canonical URL: the entry is shared across hashes, and
  // a later same-route hash nav appends `url.hash` to it.
  createHrefFromUrl(canonicalUrl, false), // <-- second arg `false` strips the hash
  supportsPerSegmentPrefetching,
  false
)
```

`createHrefFromUrl(url, includeHash = true)` normally includes the hash. Passing `false` stores a hashless canonical href. Because a single route cache entry is **shared across all hash targets** of the same route, a subsequent same-route hash navigation appends the new `url.hash` to the stored canonical href. When the stored href already carried a hash (`#foo`), the new hash (`#bar`) was appended on top of it, yielding `#foo#bar`.

Unlike the earlier fix attempt (vercel/next.js#93855, which sliced away an existing hash), this fix prevents the bad state at the source by never storing the hash in the shared entry.

## Impact on vinext

vinext implements App Router client-side navigation, including same-route hash navigation and canonical-URL tracking. If vinext stores a hash-bearing canonical href in a per-route cache entry and appends `url.hash` on a subsequent same-route hash navigation, it will reproduce the same `#foo#bar` doubling.

What to check/do:

1. **Store the canonical href without the hash** wherever a per-route navigation/cache entry is keyed or shared across hash targets. The hash is a per-navigation property, not a per-route one.
2. **Append the current navigation's hash at navigation time**, not at cache-store time, so `/abc#foo` -> `/abc#bar` replaces rather than concatenates.
3. **Port the regression test.** The Next.js test (`test/e2e/app-dir/navigation/app/hash-cross-path-push/`) navigates `/` -> `/abc#foo` -> `/abc#bar` (via `router.push`) and asserts the final URL is `/abc#bar`, not `/abc#foo#bar`. vinext should add an equivalent E2E case for cross-hash, same-path `router.push` navigation.

### Notes

- This is a general client-navigation correctness bug independent of Cache Components / Partial Prefetching. It applies to any vinext App Router app that uses hash links or `router.push` with a hash.
- The fix is in the navigation store, not a display-time slice, so make sure vinext fixes it at the point the canonical href is stored rather than post-processing the URL bar.

## Related

- (none — no existing hash-navigation tracking issue)

NathanDrake2406:
## Summary

- port the Next.js `hash-cross-path-push` navigation regression as an App Router browser test
- add the matching `nextjs-compat` fixture routes (start page + destination page)
- cover both navigation entry points: `<Link>` and `router.push`

Closes #2809.

## Context

vercel/next.js#93132 fixes a client-navigation bug where the URL hash is
appended twice (`/abc#foo#bar`). Its cause is specific to the Next.js segment
cache: `createRouteCacheEntry` stored one canonical URL per route entry, shared
across every hash target of that route, and a later same-route hash navigation
appended `url.hash` to it. The upstream fix stores a hashless canonical href.

## Triage

The defect does not reproduce in vinext, and the reason is structural rather
than incidental.

- vinext has no per-route canonical URL. `rg 'canonicalUrl' packages/vinext/src`
  matches only the Sass plugin.
- Every navigation resolves its own target against the live location
  (`toBrowserNavigationHref` -> `resolveRelativeHref`, `shims/url-utils.ts`), and
  the committed history write uses that resolved target
  (`AppBrowserHistoryController.commitNavigationHistory`,
  `createCanonicalBrowserHistoryHref`). Nothing concatenates a new hash onto a
  stored href from an earlier navigation.
- Reproduced the upstream scenario against `main` (`<Link>` clicks,
  `router.push`, and a search+hash variant): all landed on the correct single
  hash.

Existing coverage did not reach this case. `hash-popstate-scroll.spec.ts` and
`hash-rsc-requests.browser.spec.ts` only change hashes after a direct document
load of the route; the upstream regression requires the route entry to be
created by a *cross-path client push that already carries a hash*.

Since the behavior is correct, this follows the precedent of #2758: a
tracking issue resolved with focused regression coverage rather than a
production change.

## Verification

- `vp check` on the five new files: pass (format, lint, types)
- pre-commit full check + knip: pass
- `PLAYWRIGHT_PROJECT=app-router npx playwright test hash-cross-path-push.spec.ts`: 2/2 passed
- repeated 5x (`--repeat-each=5 --retries=0`): 10/10 passed
- neighbouring suites `hash-popstate-scroll.spec.ts` + `router-autoscroll.spec.ts`: 38/38 passed

**Mutation check (the tests have real signal).** Temporarily injected the
upstream failure mechanism into `navigateClientSide`'s `sameDocumentScroll`
branch — committing `location.pathname + location.search + location.hash +
intent.hash` instead of the resolved target — then rebuilt `vinext`. Both new
tests failed with exactly the reported malformed URL:

```
Expected: ".../hash-cross-path-push/destination#baz"
Received: ".../hash-cross-path-push/destination#foo#baz"
```

The mutation was reverted and `vinext` rebuilt before committing; the working
tree contains no production changes (`git diff` against `packages/` is empty).

## Scope

Test and fixture only — no runtime behavior changes, so no changeset. The
Pages Router hash lifecycle (`shims/router.ts`, `hashChangeStart`) is a separate
navigation path and is out of scope for this upstream port.

### DIFF
--- a/tests/e2e/app-router/nextjs-compat/hash-cross-path-push.spec.ts
+++ b/tests/e2e/app-router/nextjs-compat/hash-cross-path-push.spec.ts
@@ -1,0 +1,43 @@
+import { expect, test } from "@playwright/test";
+import { waitForAppRouterHydration } from "../../helpers";
+
+const BASE = "http://localhost:4174";
+const START = "/nextjs-compat/hash-cross-path-push";
+const DESTINATION = "/nextjs-compat/hash-cross-path-push/destination";
+
+// Ported from Next.js: test/e2e/app-dir/navigation/navigation.test.ts
+// ("cross-pathname Link then same-pathname hash change")
+// https://github.com/vercel/next.js/pull/93132
+//
+// Next.js regressed here because its segment cache stored one canonical URL per
+// route entry, shared across every hash target, and a later same-route hash
+// navigation appended `url.hash` to it — producing `/destination#foo#bar`.
+// vinext resolves each navigation target against the live location instead
+// (`toBrowserNavigationHref`) and writes that resolved href to history
+// (`commitNavigationHistory`), so the hash is replaced rather than concatenated.
+// These tests lock that in for both navigation entry points.
+test.describe("Next.js compat: cross-path push with hash then same-path hash change", () => {
+  test("<Link> replaces the hash instead of concatenating it", async ({ page }) => {
+    await page.goto(`${BASE}${START}`);
+    await waitForAppRouterHydration(page);
+    await expect(page.locator("h1")).toHaveText("Hash Cross Path Push Start");
+
+    await page.click("#link-to-target-foo");
+    await expect(page).toHaveURL(`${BASE}${DESTINATION}#foo`);
+
+    await page.click("#link-to-target-baz");
+    await expect(page).toHaveURL(`${BASE}${DESTINATION}#baz`);
+  });
+
+  test("router.push replaces the hash instead of concatenating it", async ({ page }) => {
+    await page.goto(`${BASE}${START}`);
+    await waitForAppRouterHydration(page);
+    await expect(page.locator("h1")).toHaveText("Hash Cross Path Push Start");
+
+    await page.click("#push-to-target-foo");
+    await expect(page).toHaveURL(`${BASE}${DESTINATION}#foo`);
+
+    await page.click("#push-to-target-baz");
+    await expect(page).toHaveURL(`${BASE}${DESTINATION}#baz`);
+  });
+});
--- a/tests/fixtures/app-basic/app/nextjs-compat/hash-cross-path-push/destination/page.tsx
+++ b/tests/fixtures/app-basic/app/nextjs-compat/hash-cross-path-push/destination/page.tsx
@@ -1,0 +1,18 @@
+import Link from "next/link";
+import { HashCrossPathPushDestinationActions } from "./push-actions";
+
+export default function HashCrossPathPushDestinationPage() {
+  return (
+    <main>
+      <h1>Hash Cross Path Push Destination</h1>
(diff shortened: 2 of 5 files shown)


NathanDrake2406:
@codex review

NathanDrake2406:
Closing: this turned out to be parity coverage for a defect vinext does not have, and it is not worth a PR on its own.

The triage summary is on #2809. In short, the Next.js bug is specific to its segment cache storing one canonical URL per route entry shared across every hash target; vinext resolves each navigation target against the live location and writes that resolved href to history, so the hash is replaced rather than concatenated. The upstream scenario was reproduced against `main` and behaved correctly.

No production changes were proposed here.

PR #2934, opened 2026-08-15
fix(fonts): normalize self-hosted URLs on Windows

### LINKED ISSUE #2933: next/font: self-hosted font URLs keep the build machine's absolute path on Windows
## Summary

On Windows, `_rewriteCachedFontCssToServedUrls` never matches, so the `@font-face` CSS ships with the font cache's absolute filesystem path instead of the served `/assets/_vinext_fonts/...` URL:

```css
@font-face { src: url(C:/Users/me/project/.vinext/fonts/geist-mono-00e989178794/geist-mono-013b2f2f.woff2) }
```

Browsers refuse the scheme (`Cross origin requests are only supported for protocol schemes: chrome, ..., http, https`), so every self-hosted font fails to load and the page silently falls back to system fonts. The woff2 files are copied to the right place by `writeBundle` — only the URLs pointing at them are wrong.

It reproduces on Windows only. macOS and Linux are unaffected, which makes it easy to ship: CI is green, the site renders, and a screenshot looks plausible.

## Affected versions

| version | behaviour |
|---|---|
| 0.0.50 | absolute path survives into the bundle |
| 0.2.1 | identical logic, same bug |
| 1.0.0-beta.6 | path no longer leaks, but the URL keeps backslashes: `url(/assets/_vinext_fonts\geist-abc\geist-123.woff2)` |

## Root cause

Two halves of the same rewrite disagree about separators.

**The writer** normalises to forward slashes — `fetchAndCacheFont`, `dist/plugins/fonts.js` (line 305 in 0.0.50, 314 in 0.2.1):

```js
css = css.split(fontUrl).join(filePath.replaceAll("\\", "/"));
```

**The rewriter** compares against a `path.join` result — `configResolved` (line 416) and `_rewriteCachedFontCssToServedUrls` (line 90):

```js
cacheDir = path.join(config.root, ".vinext", "fonts");   // C:\Users\me\project\.vinext\fonts
if (!cacheDir || !css.includes(cacheDir)) return css;    // forward vs back → false
```

The CSS holds `C:/Users/...` and `cacheDir` holds `C:\Users\...`, so `includes()` is false and the function returns the CSS untouched. On POSIX `path.sep` is already `/`, the `replaceAll` is a no-op, the strings match, and the rewrite works.

The function's own doc comment describes this exact failure — *"Without this rewrite, all three emit the dev-machine filesystem path ... and any production request fetches `<origin>/home/user/...` → 404"* — so the intent is right; the guard is what fails. Per that comment the blast radius is three consumers: the injected `<style data-vinext-fonts>` block, the `<link rel="preload">` tags, and the HTTP `Link:` response header.

## Reproduction

No Windows machine needed — the exported helper can be driven directly with Windows-shaped input:

```js
// node repro.mjs, from a project with vinext installed
import { _rewriteCachedFontCssToServedUrls as rewrite } from "./node_modules/vinext/dist/plugins/fonts.js";

const cacheDir = "C:\\Users\\me\\project\\.vinext\\fonts";           // what path.join produces
const css = `@font-face{src:url(${cacheDir.replaceAll("\\", "/")}/geist-abc/geist-123.woff2)}`;  // what the writer produces

console.log(rewrite(css, cacheDir, "assets"));
// actual:   @font-face{src:url(C:/Users/me/project/.vinext/fonts/geist-abc/geist-123.woff2)}
// expected: @font-face{src:url(/assets/_vinext_fonts/geist-abc/geist-123.woff2)}
```

For comparison, the POSIX shape rewrites correctly:

```js
const posix = "/home/me/project/.vinext/fonts";
rewrite(`@font-face{src:url(${posix}/geist-abc/geist-123.woff2)}`, posix, "assets");
// → @font-face{src:url(/assets/_vinext_fonts/geist-abc/geist-123.woff2)}
```

## Relationship to #1605

This is the same class as #1605 ("Normalize internal paths to forward slashes for Windows support"), but the fonts plugin is not in that issue's *Known affected areas* list, and the failure is user-visible in production rather than a build-time break — the page renders, every check passes, and the fonts are simply gone.

`cacheDir` here comes from `path.join(config.root, ".vinext", "fonts")` in `configResolved`, which is precisely the "normalize at filesystem boundaries" case #1605 proposes, and #1604's `toSlash()` is already available for it.

## Suggested fix

Normalise `cacheDir` at the boundary with `toSlash()`, and normalise the separators in the tail that follows — the second half matters for the `1.0.0-beta.6` shape, where the path is substituted but `\geist-abc\geist-123.woff2` remains:

```js
// configResolved
cacheDir = toSlash(path.join(config.root, ".vinext", "fonts"));

function _rewriteCachedFontCssToServedUrls(css, cacheDir, assetsDir = DEFAULT_ASSETS_DIR) {
  if (!cacheDir || !css.includes(cacheDir)) return css;
  const prefix = assetsDir || DEFAULT_ASSETS_DIR;
  const served = `/${prefix}/${VINEXT_FONT_URL_NAMESPACE}`;
  const out = css.split(cacheDir).join(served);
  // A backslash is not a path separator in a URL. Browsers normalise it for
  // http(s); the preload Link header and other consumers need not.
  const servedUrl = new RegExp(`${escapeForRegExp(served)}[^\\s"'()]*`, "g");
  return out.replace(servedUrl, (url) => url.replace(/\\+/g, "/"));
}
```

With `cacheDir` normalised at the boundary, the writer's existing `filePath.replaceAll("\\", "/")` lines up with it and the comparison succeeds. If the writer is instead left emitting `path.join` output (as on `1.0.0-beta.6`), the tail normalisation is what keeps the resulting URL valid.

## Environment

- vinext 0.0.50 (also verified against 0.2.1 and 1.0.0-beta.6 from npm)
- Build machine: Windows; the same source builds correctly on macOS and Linux
- Deploy target: Cloudflare Workers via `@cloudflare/vite-plugin`

james-elicx:
## Summary

- normalize the Vite project root before deriving the Google Fonts cache directory
- canonicalize the cache-directory argument before matching writer-normalized CSS
- add a Windows regression for the 0.0.50/0.2.1 prefix mismatch reported in #2933

The fix is confined to build-time path normalization. It adds no request-time work and does not migrate pre-existing `.vinext` font caches.

Next.js emits origin-served font URLs under `/_next/static/media`; this keeps vinext self-hosted font CSS, preload tags, and `Link` headers on the equivalent URL-space contract.

Fixes #2933

## Testing

- `vp test run tests/font-google.test.ts tests/app-router-font-google-prod.test.ts tests/pages-router-font-google-prod.test.ts`
- `vp check tests/font-google.test.ts packages/vinext/src/plugins/fonts.ts`
- `vp run vinext#build`

### DIFF
--- a/packages/vinext/src/plugins/fonts.ts
+++ b/packages/vinext/src/plugins/fonts.ts
@@ -26,7 +26,7 @@
 
 import type { Plugin } from "vite";
 import { parseAst } from "vite";
-import path from "pathslash";
+import path, { toSlash } from "pathslash";
 import fs from "node:fs";
 import { escapeRegExp } from "../utils/regex.js";
 import { lastSignificantChar } from "../utils/has-trailing-comma.js";
@@ -138,9 +138,10 @@ export function _rewriteCachedFontCssToServedUrls(
   cacheDir: string,
   assetsDir: string = DEFAULT_ASSETS_DIR,
 ): string {
-  if (!cacheDir || !css.includes(cacheDir)) return css;
+  const normalizedCacheDir = toSlash(cacheDir);
+  if (!normalizedCacheDir || !css.includes(normalizedCacheDir)) return css;
   const prefix = assetsDir || DEFAULT_ASSETS_DIR;
-  return css.split(cacheDir).join(`/${prefix}/${VINEXT_FONT_URL_NAMESPACE}`);
+  return css.split(normalizedCacheDir).join(`/${prefix}/${VINEXT_FONT_URL_NAMESPACE}`);
 }
 
 /**
@@ -661,7 +662,7 @@ export function createGoogleFontsPlugin(fontGoogleShimPath: string, shimsDir: st
     enforce: "pre",
 
     configResolved(config) {
-      cacheDir = path.join(config.root, ".vinext", "fonts");
+      cacheDir = path.join(toSlash(config.root), ".vinext", "fonts");
     },
 
     // Dev-mode equivalent of the production `writeBundle` copy step. In dev
--- a/tests/font-google.test.ts
+++ b/tests/font-google.test.ts
@@ -1464,6 +1464,27 @@ describe("_rewriteCachedFontCssToServedUrls", () => {
     expect(out).toContain("format('woff2')");
   });
 
+  it.runIf(process.platform === "win32")(
+    "matches writer-normalized CSS when the cache directory uses native Windows separators",
+    () => {
+      // Regression for the 0.0.50 / 0.2.1 shape reported in #2933: the CSS
+      // writer used forward slashes while configResolved retained backslashes.
+      // Next.js emits font URLs under /_next/static/media rather than leaking
+      // the build machine's filesystem path:
+      // https://github.com/vercel/next.js/blob/canary/test/e2e/next-font/index.test.ts
+      const cacheDir = path.join("C:\\", "Users", "me", "project", ".vinext", "fonts");
+      const canonicalCacheDir = cacheDir.replaceAll("\\", "/");
+      const css = `src: url(${canonicalCacheDir}/geist-abc/geist-def.woff2) format('woff2');`;
+
+      const out = rewriteCachedFontCssToServedUrls(css, cacheDir);
+
+      expect(out).toBe(
+        "src: url(/_next/static/_vinext_fonts/geist-abc/geist-def.woff2) format('woff2');",
+      );
+      expect(out).not.toContain("C:/Users/me/project/.vinext/fonts");
+    },
+  );
+
   it("rewrites every occurrence when the same path appears multiple times", () => {
     // The broken path can appear in the same cached CSS via the cyrillic /
     // latin-ext / latin @font-face blocks Google Fonts returns per family,


james-elicx:
This is not a blocker for the requested scope.

The maintainer explicitly does not want compatibility for reusing pre-existing `.vinext/fonts/style.css` files written by an older version. The PR body already states that those caches are outside this fix and must be regenerated.

For a fresh build, `config.root` is normalized at the external boundary and every downstream `fontDir` / `filePath` join uses `pathslash`. The published beta.6 artifact also imports `pathslash` and uses it for both `cacheDir` and `filePath`; its bundled win32 `join` emits forward slashes. Current code therefore does not generate the mixed-separator tail this review wants to repair. That tail is only reachable by reading an existing on-disk cache, which is the excluded migration case.

Also, `_rewriteCachedFontCssToServedUrls` runs while transforming the font call during Vite dev/build. It is not request-path work. We nevertheless removed the regex because there is no reason to carry cache-migration logic when that compatibility is explicitly out of scope.

The intended contract is now narrow:

- normalize the external `config.root` / `cacheDir` boundary;
- keep the current writer-normalized CSS path matching on Windows;
- require `.vinext` cache regeneration when upgrading from an older cache format.

Please evaluate the diff against that stated fresh-build contract rather than requesting legacy cache migration support.

PR #2876, opened 2026-08-11
fix(app-router): match current navigation identity

james-elicx:
## Summary

Fix the three non-cache `test/e2e/app-dir/app/index.test.ts` failures recorded by [deploy-suite run 31439707085](https://github.com/cloudflare/vinext/actions/runs/31439707085) / [test report job 93624401572](https://github.com/cloudflare/vinext/actions/runs/31439707085/job/93624401572). The archived report recorded 100 passed, 3 failed, and 5 skipped for this suite.

Failure mapping:

- `<Link /> > should soft push`: a new push after back could replay the departed visited-response snapshot instead of fetching and committing the current server output. Departed snapshots are now demoted after history restoration.
- `<Link /> > should soft replace`: an exact-current Link could reuse the response that produced the page already on screen. Exact raw URL identity now refreshes page segments without adding history, including basePath, hash, and encoded-query cases.
- `template component > should render the template that is a server component and rerender on navigation`: server template output followed child remount identity, replacing the template seed during child navigation. Template seed identity is now derived from parameters owned by the template boundary while child template state still remounts.

The navigation planner now makes browser-space versus already app-relative snapshot URLs explicit, preserves raw query spelling (`%20` versus `+`), handles basePath prefix collisions, refreshes an unchanged hash, and treats a changed or removed hash as same-document navigation.

## Validation

- Next.js `v16.2.6` (`ee6e79b1792a4d401ddf2480f40a83549fe8e722`):
  - `REPO="$(pwd)" NEXTJS_DIR="/Users/jamesanderson/Developer/vinext/.nextjs-ref" ./scripts/run-targeted-nextjs-e2e.sh test/e2e/app-dir/app/index.test.ts`
  - **103 passed, 5 skipped, 0 failed**
- 1,538 core unit assertions passed across the navigation shim, planner intent, and browser entry suites; 122 additional planner/prefetch assertions passed.
- 5 focused App Router browser regressions passed.
- 5 fresh-production basePath/hash regressions passed, including zero RSC requests on hash removal and exact `%20` identity.
- Focused format, lint, type checks, and `git diff --check` passed.
- Four independent reviews completed; the final review reported **NO FINDINGS**.

This PR intentionally targets only the three non-cache failures above.

Issue #1523, opened 2026-05-22
App Router: client `<Link>` soft navigation triggers full server re-render


james-elicx:
> *This issue was created by an agent analysing CI failures from the [Next.js Deploy Suite](https://github.com/cloudflare/vinext/actions/runs/26282973805) (vinext `main` vs Next.js `v16.2.6`, 2026-05-22).*

## Problem

Soft client-side navigations via `<Link>` (push, back/forward) trigger a fresh server render instead of reusing the cached client tree. The deployment build ID changes between the initial render and after navigation, indicating the client re-fetches rather than soft-navigates. Layout is re-rendered when it should be reused.

```
Soft <Link> navigation rerenders layout (build IDs differ)
```

## Estimated Impact

~4 test failures across the deploy suite.

## Affected Test Suites

- `test/e2e/app-dir/app/index.test.ts` (4 failures)

## Recommendation

1. **Reproduce first in vinext's own test suite.** Add a test that captures the deployment ID once, navigates via `<Link>` to a sibling route, and asserts the build ID is the same (i.e. no full server re-render).

2. **Make `<Link>` navigations use the cached client tree.** On click, perform an RSC payload fetch and reuse the existing layout/cache subtree instead of dispatching a full server render.

3. **Cover back/forward.** Same expectation applies to history traversal.

---

Part of #1328.

PR #2939, opened 2026-08-16
test(app-router): reproduce shared layout teardown on soft navigation

james-elicx:
> **Draft: reproduction only.** This PR contains no fix — the added e2es are expected to fail on CI. It exists to pin down the failure described in #1523 with a runnable fixture.

## Problem

A shared layout is discarded during a soft navigation that only changes segments *below* it. The layout's DOM node is replaced and any client state it owns is lost, even though its dynamic segment never changes value.

`/layout-identity/[slug]` and everything beneath it resolve the same `slug`, so navigating from the parent into a child changes only the segments under the `[slug]` layout. That layout should be reconciled in place.

This is the client-side half of #1523 ("Soft `<Link>` navigation triggers full server re-render… Layout is re-rendered when it should be reused"). #2876 fixes the three non-cache failures in that suite and deliberately leaves this one.

## What was added

A `/layout-identity` fixture in `examples/app-router-cloudflare`, deliberately shaped like a real app rather than reduced to the smallest possible case:

- a shared `[slug]/layout.tsx` holding a client component with state
- a **`loading.tsx` boundary above** the shared layout
- a **parallel slot** (`@aside`) beside `children`, whose content changes on every navigation
- a **static** child segment (`[slug]/child`)
- a **nested dynamic** segment with its own layout (`[slug]/items/[item]`), so a navigation can cross two layout levels at once

Plus `tests/e2e/cloudflare-workers/layout-identity.spec.ts` covering each navigation shape.

## Results on this branch

```
✓ navigates without a full document load
✘ keeps the shared layout element mounted across a static child segment
✘ keeps the shared layout element mounted across a nested dynamic segment
✘ preserves client state owned by the shared layout
✘ keeps the shared layout mounted when navigating back to the parent
✘ updates the parallel slot alongside the retained layout

5 failed, 1 passed
```

Run with:

```sh
cd examples/app-router-cloudflare && npx vp build && npx wrangler dev --config dist/server/wrangler.json --port 4176
npx playwright test --project=cloudflare-workers layout-identity
```

The passing test is the control: a `window` marker set before the click survives it, so this is a client-side navigation, not a document load. The teardown is the router discarding a layout it should have retained.

The state failure is the user-visible one — the layout's counter is incremented to 1, and after navigating reads `Layout count: 0`.

## Isolating the trigger

The `loading.tsx` boundary above the shared layout is what causes it. Measured on this fixture:

| fixture | result |
| --- | --- |
| as committed | 5 failed, 1 passed |
| identical tree, `layout-identity/loading.tsx` deleted | **6 passed** |

I also built the fixture up incrementally, and neither the parallel slot nor the nested dynamic segment triggers it on their own — with no `loading.tsx` present, every combination of those passed. Only introducing the boundary above the shared layout produces the teardown.

## Where I think the gap is

`preserveElementIds` in the visible-commit path never admits layout-kind elements. The only element kind preserved across a navigation lane is `template` (`app-browser-visible-commit.ts`, via the `preserveTemplateIds` logic that #2876 carries). Layout-kind machinery exists elsewhere (`app-elements-wire.ts`, `app-history-state.ts`) but is not wired into that decision, so a layout whose segment value is unchanged is not retained when the boundary above it re-suspends.

I have not attempted a fix here — happy to follow up with one, or hand this to whoever picks up #1523.

## Environment

Reproduced on `main` (`vinext@1.0.0-beta.6`, 6 commits behind HEAD at the time of writing, none touching App Router navigation), Next.js 16.2.2, React 19.2.5, production build served with `wrangler dev`.

Separately confirmed against the pkg.pr.new prerelease of #2876 (`5719d11`) on a real application: the shared layout is still torn down there, consistent with that PR scoping itself away from this failure.

### DIFF
--- a/examples/app-router-cloudflare/app/layout-identity/@aside/[slug]/child/page.tsx
+++ b/examples/app-router-cloudflare/app/layout-identity/@aside/[slug]/child/page.tsx
@@ -1,0 +1,4 @@
+export default async function AsideChild({ params }: { params: Promise<{ slug: string }> }) {
+  const { slug } = await params;
+  return <p data-testid="aside-content">aside: {slug} child</p>;
+}
--- a/examples/app-router-cloudflare/app/layout-identity/@aside/[slug]/items/[item]/page.tsx
+++ b/examples/app-router-cloudflare/app/layout-identity/@aside/[slug]/items/[item]/page.tsx
@@ -1,0 +1,12 @@
+export default async function AsideItem({
+  params,
+}: {
+  params: Promise<{ slug: string; item: string }>;
+}) {
+  const { slug, item } = await params;
+  return (
+    <p data-testid="aside-content">
+      aside: {slug} / {item}
+    </p>
+  );
+}
--- a/examples/app-router-cloudflare/app/layout-identity/@aside/[slug]/page.tsx
+++ b/examples/app-router-cloudflare/app/layout-identity/@aside/[slug]/page.tsx
@@ -1,0 +1,4 @@
+export default async function AsideParent({ params }: { params: Promise<{ slug: string }> }) {
+  const { slug } = await params;
+  return <p data-testid="aside-content">aside: {slug} parent</p>;
+}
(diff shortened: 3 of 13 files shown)


james-elicx:
Folded into #2940 so the production Worker regression E2E and parity fix can land together in one PR targeting main.

Issue #2865, opened 2026-08-10
App Router search-only navigation can fail to commit when removing the last query parameter


NathanDrake2406:
## Summary

Under vinext 1.0.0-beta.5, an App Router transition from a filtered URL such as `/top?provider=8` back to the same pathname without search parameters can fail to commit. The RSC request for `/top` reaches the Worker and returns 200, but the client remains in the provider-filtered state or falls into vinext's default client error boundary.

This looks related to the stale search-param cache behavior tracked in #1621, but it is occurring on beta.5 after that issue was closed.

## Reproduction shape

The page is dynamic and reads `searchParams`. A client filter uses `useSearchParams()`, optimistic state, and `router.push()`:

```tsx
"use client";

import { useOptimistic, useTransition } from "react";
import { useRouter, useSearchParams } from "next/navigation";

export function ProviderToggle() {
  const router = useRouter();
  const routeSearchParams = useSearchParams();
  const [isPending, startTransition] = useTransition();
  const [optimisticQuery, setOptimisticQuery] = useOptimistic(
    routeSearchParams.toString(),
    (_current, next: string) => next,
  );

  const active = new URLSearchParams(optimisticQuery).get("provider") === "8";

  return (
    <button
      aria-busy={isPending}
      onClick={() => {
        const params = new URLSearchParams(optimisticQuery);
        if (active) params.delete("provider");
        else params.set("provider", "8");

        const query = params.toString();
        startTransition(() => {
          setOptimisticQuery(query);
          router.push(query ? `/top?${query}` : "/top", { scroll: false });
        });
      }}
    >
      Netflix
    </button>
  );
}
```

## Steps

1. Load `/top`.
2. Click the provider toggle. The navigation to `/top?provider=8` commits and filtered server content renders.
3. Click the active provider toggle again. This calls `router.push("/top", { scroll: false })`.
4. Observe the transition.

## Actual behavior

The request sequence includes a successful `GET /top?_rsc=...` response, but the navigation can fail to commit on the client. The provider remains active / the filtered tree remains mounted. During the same production session, the client also reached vinext's default `This page couldn’t load` boundary without a server digest.

## Expected behavior

The transition commits to `/top`, `useSearchParams()` becomes empty for `provider`, and the unfiltered server tree replaces the filtered tree.

## Notes

- This was observed in a real Cloudflare Workers deployment immediately after migrating an existing Next.js App Router application to vinext.
[...]

PR #2948, opened 2026-08-17
test(app-router): reproduce deployed optimistic search navigation stall

NathanDrake2406:
> **Draft: reproduction only.** This PR contains no Vinext fix. It adds a deployed-Workers regression target for #2865 so the navigation path can be investigated in the environment where it fails.

## Problem

A dynamic App Router page uses `useSearchParams()` as its committed state, mirrors the query with `useOptimistic()`, and calls `router.push()` inside the same transition. On a deployed Cloudflare Worker, a same-path provider change can receive a successful RSC response but never commit the new router state.

The visible result is intermittent: Netflix or Prime may select and clear normally, then Disney can remain stuck in its optimistic state. Removing the active provider can fail in the same way, leaving the filtered tree mounted at the old URL state.

This is the deployed-runtime follow-up to #2865. The same application and reduced interaction pass under `wrangler dev`, so a local preview is not a sufficient control for this bug.

## What was added

A production-shaped `/optimistic-search-navigation` fixture in `examples/app-router-cloudflare`:

- a dynamic server page that reads `searchParams.provider`
- a provider filter using `useSearchParams()`, `useOptimistic()`, `useTransition()`, and `router.push()`
- a streamed provider-specific result list behind `Suspense`
- a parallel modal slot matching the affected list route
- Netflix (`8`), Prime (`9`), and Disney (`337`) provider transitions

The Cloudflare Workers E2E covers:

- plain `router.push()` as a control
- selecting a provider inside the optimistic transition
- removing the only search parameter
- overlapping provider transitions
- document identity, proving each successful result is a soft navigation
- five deployed-only cycles across all three providers (30 commits total)

The deployed-example workflow now includes this spec, so an internal preview deployment exercises the race on `workers.dev` rather than Miniflare.

## Results on this branch

Local production Workers build:

```text
✓ router.push commits the server response without useOptimistic
✓ commits a provider selected inside a useOptimistic transition
✓ commits the final state when the active provider is toggled off
✓ commits the last provider when transitions overlap
- deployed-only provider cycle skipped

4 passed, 1 skipped
```

The selected-provider control also passed 40 fresh local browser runs on current `main`. The same reduced fixture passed locally on the `vinext@1.0.0-beta.6` tag. That is expected and preserves the important finding from the affected application: both beta.6 and current `main` pass under `wrangler dev`, while the beta.6 deployed Worker fails intermittently.

Run the local controls with:

```sh
cd examples/app-router-cloudflare
npx vp build
npx wrangler dev --config dist/server/wrangler.json --port 4176

PLAYWRIGHT_PROJECT=cloudflare-workers \
  npx playwright test tests/e2e/cloudflare-workers/optimistic-search-navigation.spec.ts
```

To run the deployed stress case, point Playwright at an HTTPS deployment:

```sh
PLAYWRIGHT_PROJECT=cloudflare-workers \
VINEXT_E2E_BASE_URL=https://<preview>-app-router-cloudflare.vinext.workers.dev \
  npx playwright test tests/e2e/cloudflare-workers/optimistic-search-navigation.spec.ts
```

Fork PRs do not receive the repository deploy secrets, so the deployed lane will remain skipped until a maintainer runs the branch with repository deployment access.

## Isolating the trigger

Captured on the affected deployment:

| interaction | result |
| --- | --- |
| direct `router.push()` | commits |
| outer transition plus ordinary state | commits |
| `useOptimistic()` plus `router.push()` | can remain pending |
| same navigation with Vinext synchronous visible commits forced | 20 consecutive Disney toggles committed |

During a failed navigation, the provider API and RSC request both completed with status 200. The client navigation promise remained unresolved and Vinext retained one active navigation snapshot. This points to the transition visible-commit path, not provider data or network failure.

## Where I think the gap is

`dispatchApprovedVisibleCommit()` resolves `pendingRouterState` and normally relies on React mounting the new `NavigationCommitSignal` to publish URL and scroll effects. The failed deployed sessions never settled that signal: the optimistic transition stayed pending and the navigation snapshot remained active. Forcing the existing synchronous visible-commit mode removed the failure in the affected app.

I have not attempted a fix here. The fixture and deployed stress lane are intended to give the maintainer a focused red/green target for that commit lifecycle.

## Validation

- `npx vp build` in `examples/app-router-cloudflare`
- targeted Cloudflare Workers E2E: 4 passed, 1 skipped
- selected-provider E2E repeated 40 times: 40 passed locally
- `npx vp check`
- full pre-commit checks and Knip

Tracks #2865. The issue should remain open until the deployed stress case is confirmed green with a fix.

### DIFF
--- a/.github/workflows/deploy-examples.yml
+++ b/.github/workflows/deploy-examples.yml
@@ -168,7 +168,10 @@ jobs:
         env:
           PLAYWRIGHT_PROJECT: cloudflare-workers
           VINEXT_E2E_BASE_URL: ${{ github.event_name == 'pull_request' && format('https://pr-{0}-app-router-cloudflare.vinext.workers.dev', github.event.pull_request.number) || 'https://app-router-cloudflare.vinext.workers.dev' }}
-        run: vp exec playwright test tests/e2e/cloudflare-workers/web-worker.spec.ts
+        run: >-
+          vp exec playwright test
+          tests/e2e/cloudflare-workers/web-worker.spec.ts
+          tests/e2e/cloudflare-workers/optimistic-search-navigation.spec.ts
 
   comment:
     name: Comment Preview URLs
--- a/examples/app-router-cloudflare/app/optimistic-search-navigation/@modal/default.tsx
+++ b/examples/app-router-cloudflare/app/optimistic-search-navigation/@modal/default.tsx
@@ -1,0 +1,3 @@
+export default function ModalDefault() {
+  return null;
+}
--- a/examples/app-router-cloudflare/app/optimistic-search-navigation/layout.tsx
+++ b/examples/app-router-cloudflare/app/optimistic-search-navigation/layout.tsx
@@ -1,0 +1,16 @@
+import type { ReactNode } from "react";
+
+export default function OptimisticSearchNavigationLayout({
+  children,
+  modal,
+}: {
+  children: ReactNode;
+  modal?: ReactNode;
+}) {
+  return (
+    <div data-testid="provider-list-layout">
+      {children}
+      {modal}
+    </div>
+  );
+}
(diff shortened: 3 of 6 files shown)


NathanDrake2406:
@james-elicx enjoy. Only happens on Workers (tested on both prod and and isolated prod)

PR #2952, opened 2026-08-17
fix(app-router): settle optimistic search navigations

james-elicx:
## Summary

- carry forward the deployed-Workers reproduction from #2948
- add a deterministic local production-Worker stress case for the same stuck optimistic transition
- keep Flight decoding streaming from the first response bytes
- recover only a still-uncommitted programmatic same-path search navigation when its complete response is available
- leave links, traversals, refreshes, different-path navigations, and unchanged-search/hash navigations on their existing paths

## Root cause

Vinext resolves the deferred browser-router state as soon as the first Flight model decodes, while nested chunks for the retained same-route tree can still be arriving. When that retained client subtree owns a `useOptimistic()` update in the same transition, React can render the approved router state but leave that concurrent render uncommitted indefinitely. URL publication, snapshot release, and `useTransition()` settlement are commit-gated, so all three remain stuck even though the RSC request and router-state promise have resolved.

The fix starts `createFromFetch()` immediately, preserving progressive Flight decoding and the existing streamed render. For the narrow programmatic same-path search case, it also observes completion of the already-existing cache tee. If the normal React commit has not removed that render from the pending-commit map by response completion, and the navigation is still current with an attached browser root, it synchronously republishes the fully received candidate as a terminal recovery. Normal commits suppress recovery; stale navigations and detached roots cannot recover; the existing commit signal remains responsible for history, scroll, and settlement exactly once.

## Next.js parity

An equivalent Next.js 16.2.7 production app with the same 1,000-row streamed response and the same CDP CPU/network constraints completed 3/3 five-cycle runs. It retained the old provider content while pending and committed the URL and new content together without showing the Suspense fallback.

The lifecycle matches the current Next.js action queue and uncached Segment Cache path:

- [`use-action-queue.ts`](https://github.com/vercel/next.js/blob/ee6e79b1792a4d401ddf2480f40a83549fe8e722/packages/next/src/client/components/use-action-queue.ts) resolves the deferred router action with the completed navigation state.
- [`segment-cache/navigation.ts`](https://github.com/vercel/next.js/blob/ee6e79b1792a4d401ddf2480f40a83549fe8e722/packages/next/src/client/components/segment-cache/navigation.ts) awaits `fetchServerResponse()` for an uncached navigation before constructing that state.
- [`app-router.tsx`](https://github.com/vercel/next.js/blob/ee6e79b1792a4d401ddf2480f40a83549fe8e722/packages/next/src/client/components/app-router.tsx) publishes history from the committed state in an insertion effect.

## Validation

Baseline on the reproduction commit:

- constrained large streamed tree: 3/3 failed
- large tree without constraints: 3/3 passed
- CPU constraints only: 3/3 passed
- network constraints only: 3/3 passed
- original local controls: 4/4 passed

Fixed head `ea796ebe8`:

- hardened constrained Worker stress: 3/3 passed
- each stress run completed 32 pushes with exact history growth
- old content remained visible while pending, with no Suspense fallback flash
- original four Worker controls: 4/4 passed
- `vp test run tests/app-browser-entry.test.ts`: 241/241 passed
- focused `vp check`: passed
- independent exact-head review: no actionable findings

## Status

Ready for review on `ea796ebe8f72fac426522711bed9087c20815e79`:

- all 70 exact-head checks green
- [Big Bonk](https://github.com/cloudflare/vinext/actions/runs/32031482495): LGTM, no actionable defects
- [Next.js deploy suite](https://github.com/cloudflare/vinext/actions/runs/32031865120): 2,680 passed / 124 failed / 631 skipped
- matching baseline: 2,678 passed / 126 failed / 631 skipped; zero new failure identities and two resolved

Refs #2865. Supersedes the reproduction-only #2948.

### DIFF
--- a/.github/workflows/deploy-examples.yml
+++ b/.github/workflows/deploy-examples.yml
@@ -168,7 +168,10 @@ jobs:
         env:
           PLAYWRIGHT_PROJECT: cloudflare-workers
           VINEXT_E2E_BASE_URL: ${{ github.event_name == 'pull_request' && format('https://pr-{0}-app-router-cloudflare.vinext.workers.dev', github.event.pull_request.number) || 'https://app-router-cloudflare.vinext.workers.dev' }}
-        run: vp exec playwright test tests/e2e/cloudflare-workers/web-worker.spec.ts
+        run: >-
+          vp exec playwright test
+          tests/e2e/cloudflare-workers/web-worker.spec.ts
+          tests/e2e/cloudflare-workers/optimistic-search-navigation.spec.ts
 
   comment:
     name: Comment Preview URLs
--- a/examples/app-router-cloudflare/app/optimistic-search-navigation/@modal/default.tsx
+++ b/examples/app-router-cloudflare/app/optimistic-search-navigation/@modal/default.tsx
@@ -1,0 +1,3 @@
+export default function ModalDefault() {
+  return null;
+}
--- a/examples/app-router-cloudflare/app/optimistic-search-navigation/layout.tsx
+++ b/examples/app-router-cloudflare/app/optimistic-search-navigation/layout.tsx
@@ -1,0 +1,16 @@
+import type { ReactNode } from "react";
+
+export default function OptimisticSearchNavigationLayout({
+  children,
+  modal,
+}: {
+  children: ReactNode;
+  modal?: ReactNode;
+}) {
+  return (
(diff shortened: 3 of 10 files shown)


NathanDrake2406:
nice one! working
I'll put up a repro for the regional headers bug tomorrow

PR #2958, opened 2026-08-17
test(client-assets-build): root fixtures on the repo volume to avoid EXDEV

shulaoda:
## Problem

`tests/client-assets-build.test.ts` fails on Windows with `EXDEV: cross-device link not permitted`. The fixture is created under `os.tmpdir()` (drive C: on this machine) but junctions its `node_modules` to the repo's `node_modules` (drive E:), so plugin-rsc's `.vite-rsc-temp` staging directory lands on E:. The build output dirs are also created under `os.tmpdir()` (C:), and plugin-rsc renames between the two — a cross-drive rename, which fails on Windows. On Linux `os.tmpdir()` and the repo share one filesystem, so the rename stays intra-volume and passes.

Fix: root the fixture and its output dirs under `import.meta.dirname` (same volume as the repo `node_modules`), matching the pattern already used by `externals-transitive.test.ts` and `static-image-emission.test.ts`. All build inputs, staging, and outputs then live on one volume.

## Affected tests

- `tests/client-assets-build.test.ts > client asset sidecar builds > keeps App Router metadata at the stable root of nested custom server outputs`

### DIFF
--- a/tests/client-assets-build.test.ts
+++ b/tests/client-assets-build.test.ts
@@ -1,5 +1,4 @@
 import fs from "node:fs/promises";
-import os from "node:os";
 import path from "node:path";
 import { build, createBuilder } from "vite";
 import { describe, expect, it } from "vite-plus/test";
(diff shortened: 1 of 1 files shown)


james-elicx:
I think a lot of our tests work like this (unfortunately), so I imagine you'll get it for more than just this?

shulaoda:
> I think a lot of our tests work like this (unfortunately), so I imagine you'll get it for more than just this?

I will check it

PR #2987, opened 2026-08-18
Use GPT-5.6 Terra and Sol for Bonk

southpolesteve:
## Summary

- use GPT-5.6 Terra with `high` reasoning for `/bonk` and the Next.js tracker
- use GPT-5.6 Sol with `max` reasoning for `/bigbonk`
- update the pinned Bonk action to `d00cbcf581a5463f7adc2d81e470234b1727f8dd`
- temporarily install OpenCode from its `dev` channel
- make OpenCode permissions non-interactive in GitHub Actions

## Root cause

The existing OpenCode `1.15.13` pin cannot use GPT-5.6 with reasoning and tools through the legacy Chat Completions route.

Upgrading directly to the latest stable OpenCode release, `1.18.18`, also fails: it sends the Cloudflare AI Gateway token as an OpenAI API key and receives a 401.

OpenCode's current dev build contains the upstream AI Gateway fix. Bonk cannot pin that exact npm version because its version validator rejects `0.0.0-dev-202608181849`, so this uses Bonk's explicit `opencode_dev: true` path for now.

OpenCode defaults `external_directory`, `question`, and `doom_loop` permissions to interactive prompts. An unattended GitHub Actions run can therefore wait forever if an agent or subagent reads outside the checkout. These workflows allow the trusted runner workspace and deny interactive question/loop prompts. The automated Next.js tracker also denies file edits.

## Validation

The temporary no-push smoke workflow exercised both production model configurations through Cloudflare AI Gateway:

- Terra / `high`: passed
- Sol / `max`: passed
- OpenCode installed: `0.0.0-dev-202608181849`

Passing run: https://github.com/cloudflare/vinext/actions/runs/32176160708

The smoke workflow was removed after validation.

Workerd's automatic Bonk reviewer independently exercised the full repository-review path. Its first run exposed the unattended `external_directory` prompt; after applying the permission policy used here, the next run completed in 2m59s and posted a real review:

- timed-out permission-prompt run: https://github.com/cloudflare/workerd/actions/runs/32176642011
- passing review run: https://github.com/cloudflare/workerd/actions/runs/32179760200

## Follow-up

Once OpenCode publishes a stable release containing the gateway fix, replace `opencode_dev: true` with a stable version pin.

GitHub loads `issue_comment` workflows from the default branch, so `/bonk` on this PR would still use Vinext's old configuration. The full slash-command path can be tested after this PR merges.


NathanDrake2406:
Assuming @james-elicx want this, I think it's better in terms of price/performance/speed to go for Sol high/xhigh and Terra max instead, according to: 

https://deepswe.datacurve.ai/

james-elicx:
I kind of liked having Opus in CI and GPT 5.6 locally - made it nice for getting different types of reviews. I think it would be quite cool to end up with 2 or 3 different models that all do reviews in CI and aggregate their findings and weigh them against each other, but i imagine bonk doesnt support that

NathanDrake2406:
Time to build one :)

PR #2994, opened 2026-08-19
fix: normalize path.relative results to forward slashes on Windows

AiAlchemist0:
## Summary

- Add `toPosixRelative()` so `path.relative()` results used internally are always forward-slash (Node emits `\` on Windows).
- Use it for `isPathInside` / `isPathInsideOrEqual` (`../` checks), Vite glob `toRelativeFileEntry`, hybrid-route `sourcePath`, and stylesheet manifest relatives.
- `scanWithExtensions` already yields `toSlash()` on main (#1604). This is the leftover `path.relative` slice of #1605.
- Windows CI already runs on `main`; this does not add a second job.

Draft #1745 also claimed #1605 but only touched routing glob discovery and has been stale since June. Happy to rebase or drop if that lands first.

## Testing

- `pnpm exec vp test run --project unit tests/plugin-utils.test.ts` (24 passed, on Windows)

## Notes

Does not close all of #1605 (build packaging + NTFS-illegal fixtures remain). Closes the `path.relative` / `isPathInside` gap.

Related: #1605, #1604, #1578

### DIFF
--- a/packages/vinext/src/build/css-url-assets.ts
+++ b/packages/vinext/src/build/css-url-assets.ts
@@ -180,7 +180,7 @@ export function rebaseCssUrlAssetReferences(
     if (!CSS_ASSET_EXT_RE.test(lower) || lower.endsWith(".css")) return null;
 
     const absolutePath = path.resolve(sourceDirectory, parts.path);
-    let rebasedPath = toSlash(path.relative(targetDirectory, absolutePath));
+    let rebasedPath = path.relative(targetDirectory, absolutePath);
     if (!rebasedPath.startsWith(".")) rebasedPath = `./${rebasedPath}`;
     return joinUrl({ ...parts, path: rebasedPath });
   });
--- a/packages/vinext/src/build/css-url-assets.ts
+++ b/packages/vinext/src/build/css-url-assets.ts
@@ -180,7 +180,7 @@ export function rebaseCssUrlAssetReferences(
     if (!CSS_ASSET_EXT_RE.test(lower) || lower.endsWith(".css")) return null;
 
     const absolutePath = path.resolve(sourceDirectory, parts.path);
-    let rebasedPath = path.relative(targetDirectory, absolutePath);
+    let rebasedPath = toSlash(path.relative(targetDirectory, absolutePath));
     if (!rebasedPath.startsWith(".")) rebasedPath = `./${rebasedPath}`;
     return joinUrl({ ...parts, path: rebasedPath });
   });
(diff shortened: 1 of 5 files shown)


packages/vinext/src/build/css-url-assets.ts
```diff
@@ -180,7 +180,7 @@ export function rebaseCssUrlAssetReferences(
     if (!CSS_ASSET_EXT_RE.test(lower) || lower.endsWith(".css")) return null;
 
     const absolutePath = path.resolve(sourceDirectory, parts.path);
-    let rebasedPath = path.relative(targetDirectory, absolutePath);
+    let rebasedPath = toSlash(path.relative(targetDirectory, absolutePath));
```
shulaoda: Why do we need the `toSlash` wrapper here? Shouldn’t `path.relative` already return a normalized path, given that `path` comes from `pathslash`?

shulaoda: See https://github.com/cloudflare/vinext/pull/2502

AiAlchemist0: You're right — `path` here is already pathslash, and `pathslash.relative` normalizes separators (https://github.com/cloudflare/vinext/pull/2502). The extra `toSlash` was leftover from treating `path.relative` as Node's. Dropped it in aeb40a8, and the same wrap inside `toPosixRelative`.

AiAlchemist0: Agreed — #2502 is the contract. `toSlash` only at the external-origin boundary, not around `path.relative`. Addressed in aeb40a8.

packages/vinext/src/utils/path.ts
```diff
@@ -18,9 +18,17 @@ export function canonicalizeFilePath(filePath: string): string {
   }
 }
 
+/**
+ * `path.relative` for internal vinext use. `path` is pathslash, so the
+ * result is already forward-slash on Windows — no extra `toSlash`.
+ */
+export function toPosixRelative(from: string, to: string): string {
```
shulaoda, line 25: So, is it necessary right now?

james-elicx:
It doesn't look like this pr is doing anything so i'm going to close it.

packages/vinext/src/utils/path.ts
AiAlchemist0: No. After #2502 `path.relative` from pathslash already returns `/`. `toPosixRelative` is just an alias now. I'll close this.

AiAlchemist0:
Closing — leftover Windows `toSlash` around `path.relative` is already handled by pathslash (#2502). Nothing left here that isn't a no-op wrapper.

PR #3007, opened 2026-08-20
fix(build): avoid workerd cwd in module identity

### LINKED ISSUE #3006: 1.0.0-beta.7: emitted server chunk calls process.cwd() at module init, throwing on workerd (500 on every SSR page)
## Summary

On `vinext@1.0.0-beta.7`, the emitted server chunks carry a runtime preamble whose fallback path calls `process.cwd()`. `workerd` has no `process.cwd`, so the preamble throws at **module-evaluation time** and the SSR bundle never initialises. Every App Router HTML/RSC render returns HTTP 500 (Cloudflare error 1101).

`vinext@0.2.1` is unaffected on the same app, same config, same build command.

The failure is invisible in the dev loop: `vite build` succeeds, `vite dev` serves every page correctly, unit and E2E suites pass. It only appears once the built Worker runs, because in `vite dev` `import.meta.url` **is** a `file:` URL and the fallback branch is never taken.

## Emitted code

Head of `dist/server/ssr/index.js` (formatted):

```js
import * as e from "node:process";
import * as t from "node:url";
const n = (() => {
  let n = import.meta.url;
  return {
    url: typeof n === `string` && n.startsWith(`file:`)
      ? n
      : t.pathToFileURL(e.cwd().replace(/[\\/]$/, ``) + `/ssr/index.js`).href
  };
})();
```

In `workerd`, `import.meta.url` is not a `file:` string, so the ternary takes the fallback and calls `e.cwd()` — `process.cwd`, which is not implemented (with `nodejs_compat` enabled).

## Error

```
Uncaught TypeError: e.cwd is not a function
    at dist/server/ssr/index.js:1:168
    at dist/server/ssr/index.js:1:219
```

Nothing is logged for the failing requests via `wrangler tail`, because the throw happens before any handler is reached. Route handlers that don't pull in the SSR bundle keep working, which makes the Worker look healthy: `/api/*` and `/sitemap.xml` returned 200 while every HTML page returned 500.

## Where it comes from

`src/plugins/import-meta-url.ts` → `finalizeEmittedModuleIdentity()`:

```js
const processCwd = `${processNamespaceBinding}.cwd()`;
const emittedPathFallback = emittedFileName
  ? `(${processCwd}.replace(/[\\\\/]$/, "") + ${JSON.stringify(`/${emittedFileName}`)})`
  : processCwd;
```

…used in the `url` branch of the preamble:

```js
`url: typeof runtimeUrl === "string" && runtimeUrl.startsWith("file:") ? runtimeUrl : ${urlNamespaceBinding}.pathToFileURL(${emittedPathFallback}).href,`
```

The `renderChunk` hook applies this to any **server** chunk (`consumer === "server"`, `format === "es"`) that contains an `import.meta.url` sentinel, so it isn't specific to one entry.

## Reproduction

App Router project, Cloudflare Workers target, custom Worker entry (`"main": "src/worker-entry.ts"`).

```bash
npm run build                     # vite build — succeeds
npx wrangler dev --config dist/server/wrangler.json --local --port 8788
curl -o /dev/null -w '%{http_code}' http://localhost:8788/about
```

Tested on pages that touch **no** bindings and no data (`/about`, `/terms`, `/privacy`, `/pro`) so nothing app-specific is involved:

| vinext | `/about` | `/terms` | `/privacy` | `/pro` |
|---|---|---|---|---|
| `1.0.0-beta.7` | 500 | 500 | 500 | 500 |
| `0.2.1` | 200 | 200 | 200 | 200 |

Downgrading vinext and rebuilding is the only change between rows.

Confirmed in a real deployment too: merging the upgrade took our dev environment down with 500s on every HTML route until it was reverted.

## Environment

- `vinext` `1.0.0-beta.7` (broken) vs `0.2.1` (works)
- `@vitejs/plugin-rsc` `0.5.34`
- `vite` `8.1.3`
- `next` `16.2.3`
- `wrangler` `4.107.1`
- `@cloudflare/vite-plugin` `1.43.1`
- Node 22.16.0, macOS (darwin 25.6.0)
- `compatibility_flags: ["nodejs_compat"]`

Note: beta.7 refuses to build against `@vitejs/plugin-rsc@0.5.27` with `vinext: callable use cache requires @vitejs/plugin-rsc 0.5.34 or newer`, so the plugin bump is required to reach this point.

## Suggested fix

`emittedFileName` is already known at `renderChunk` time, so the fallback does not need a runtime `cwd()` at all — it could be emitted as a literal. Failing that, guarding the call would be enough:

```js
typeof process.cwd === "function" ? process.cwd() : ""
```

Either way, deriving a module URL from `process.cwd()` cannot work on a runtime with no filesystem and no working directory, which is the primary deployment target.

## One caveat

Our project uses a custom Worker entry rather than vinext's default. I have not verified whether a stock scaffolded app also trips this — the mechanism looks generic (any server chunk containing `import.meta.url` gets the preamble), but whether a default app produces such a chunk is worth checking on your side.

james-elicx:
## Summary

- resolve URL-only emitted module identity from `import.meta.filename` before using the emitted absolute fallback
- stop importing `node:process` for URL-only preambles and guard the remaining CJS `cwd()` fallback
- cover the reported runtime contract in the checked-in `cf-app-basic` Cloudflare fixture and run that Playwright project in CI

## Root cause

The emitted URL-only preamble treated every non-`file:` `import.meta.url` as a filesystem-relative Node path and called `process.cwd()` during module evaluation. The Workerd runtime from the report exposed `node:process` without `cwd()`, so SSR chunks could throw before the Worker handler initialized.

Workerd may instead supply emitted identity through `import.meta.filename`. When neither a file URL nor a filename is available, the emitted chunk filename itself is a safe absolute fallback.

The custom Worker entry is not the cause: the same checked-in reproduction returns 500 on the pre-fix base with the stock `vinext/server/fetch-handler` entry. The trigger is any emitted server chunk that reads module identity under the affected Workerd contract.

## Regression coverage

`tests/fixtures/cf-app-basic` now contains:

- a custom Worker entry matching the report
- a dynamic App Router page so the request cannot pass through prerendered output
- a checked-in URL-only ESM dependency that becomes a dedicated SSR chunk
- a final-bundle-only simulation of Workerd's non-file URL, absent filename, and missing `cwd()` behavior

Applying the exact fixture commit to pre-fix `main` makes the Playwright spec fail with HTTP 500 and emits the reported `cwd()` call. The fixed branch returns 200 and `file:///ssr/_next/static/module-url-*.js`.

The `cloudflare-encoded-paths` project is now included in the CI Playwright matrix, so this regression runs on every PR.

## Validation

- pre-fix exact fixture commit: focused Cloudflare spec fails with HTTP 500
- fixed focused Cloudflare spec: 1 passed
- full `cloudflare-encoded-paths` project locally: 9 passed
- exact-head CI `cloudflare-encoded-paths`: 9 passed
- `tests/prerender.test.ts`: 100 passed
- `tests/nextjs-compat/hybrid-pages-api.test.ts`: 5 passed
- `vp test run tests/import-meta-url.test.ts tests/cjs-globals-runtime.test.ts`: 88 passed
- targeted formatting/checks, actionlint, and zizmor: clean
- three fresh independent reviews: no findings

Closes #3006

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -521,6 +521,10 @@ jobs:
             label: cloudflare-workers
             shardIndex: 1
             shardTotal: 1
+          - project: cloudflare-encoded-paths
+            label: cloudflare-encoded-paths
+            shardIndex: 1
+            shardTotal: 1
           - project: cloudflare-sentry-app
             label: cloudflare-sentry-app
             shardIndex: 1
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,10 @@ tests/e2e/cloudflare-workers/fixtures/client-module-with-package-type/node_modul
 !tests/e2e/cloudflare-workers/fixtures/client-module-with-package-type/node_modules/lib-cjs/**
 !tests/e2e/cloudflare-workers/fixtures/client-module-with-package-type/node_modules/lib-esm/
 !tests/e2e/cloudflare-workers/fixtures/client-module-with-package-type/node_modules/lib-esm/**
+!tests/fixtures/cf-app-basic/node_modules/
+tests/fixtures/cf-app-basic/node_modules/*
+!tests/fixtures/cf-app-basic/node_modules/module-identity-dependency/
+!tests/fixtures/cf-app-basic/node_modules/module-identity-dependency/**
 dist/
 !packages/types/next/upstream/dist/
 !packages/types/next/upstream/dist/**
--- a/packages/vinext/src/plugins/import-meta-url.ts
+++ b/packages/vinext/src/plugins/import-meta-url.ts
@@ -443,15 +443,16 @@ function finalizeEmittedModuleIdentity(
   const urlNamespaceBinding = selectRuntimeBinding("__vinext_module_url");
   const identityBinding = selectRuntimeBinding("__vinext_module_identity");
   const emittedFileName = toSlash(fileName).replace(/^\.\//, "").replace(/^\/+/, "");
-  const processCwd = `${processNamespaceBinding}.cwd()`;
+  const processCwd = `(typeof ${processNamespaceBinding}.cwd === "function" ? ${processNamespaceBinding}.cwd() : "")`;
   const emittedPathFallback = emittedFileName
     ? `(${processCwd}.replace(/[\\\\/]$/, "") + ${JSON.stringify(`/${emittedFileName}`)})`
-    : processCwd;
+    : `(${processCwd} || "/")`;
   const emittedDirName = path.dirname(emittedFileName);
   const emittedDirFallback =
     emittedDirName === "."
-      ? processCwd
+      ? `(${processCwd} || "/")`
       : `(${processCwd}.replace(/[\\\\/]$/, "") + ${JSON.stringify(`/${emittedDirName}`)})`;
+  const absoluteEmittedPath = JSON.stringify(`/${emittedFileName}`.replace(/\/$/, "") || "/");
   const replacements: Record<EmittedModuleIdentityField, string> = {
     __filename: `${identityBinding}.filename`,
     __dirname: `${identityBinding}.dirname`,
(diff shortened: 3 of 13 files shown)


james-elicx:
@ask-bonk This review workflow has a hard 10-minute execution limit. On future runs, please prioritize completing the code review and posting an actionable verdict within that limit; avoid spending the full budget on additional local reproduction once the branch evidence is sufficient.

james-elicx:
/bigbonk review for issues

Hard time limit: this workflow is canceled after 10 minutes. Prioritize completing the review and posting an actionable verdict within that limit; do not spend the full budget on additional local reproduction once the branch evidence is sufficient.

james-elicx:
/bigbonk review for issues

Hard time limit: this workflow is canceled after 10 minutes. Prioritize completing the review and posting an actionable verdict within that limit.

PR #3023, opened 2026-08-21
fix(build): filter Pages-owned RSC warm paths

james-elicx:
## Summary

- resolve each concrete App-discovered warm path through the same cross-router ownership decision used at runtime
- exclude Pages page and Pages API-owned collisions only from full/loading-shell RSC warm targets
- retain generic HTML warm discovery so the deployed Worker still decides response cacheability
- compare against the best concrete App match, preserving more-specific App routes over broad Pages matches

This remains independent of local prerender output and the prerender flag.

## Next.js references

- [`test/e2e/app-dir/use-params/use-params.test.ts`](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/use-params/use-params.test.ts)
- [`test/e2e/app-dir/pages-to-app-routing/pages-to-app-routing.test.ts`](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/pages-to-app-routing/pages-to-app-routing.test.ts)
- [`packages/next/src/server/base-server.ts`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/base-server.ts)

## Testing

- `vp test run tests/prerender-paths.test.ts tests/hybrid-route-priority.test.ts` (28 passed)
- `vp check packages/vinext/src/build/prerender-paths.ts tests/prerender-paths.test.ts`

Stacked on #3022.

PR #3021, opened 2026-08-21
fix(cloudflare): validate reusable warm responses

james-elicx:
## Summary

- validate the effective Cloudflare cache policy using `Cloudflare-CDN-Cache-Control` / `CDN-Cache-Control` / `Cache-Control` precedence
- reject `CF-Cache-Status: STALE` because it does not prove a fresh fill completed
- reject zero or negative shared-cache freshness instead of reporting it as warmed
- preserve coherent non-cacheable responses as skipped
- prevent named Wrangler environments from inheriting the top-level production custom domain

## Test plan

- `vp test run tests/cloudflare-cdn-warm.test.ts tests/cloudflare-cdn-warm-deploy.test.ts`
- `vp check packages/cloudflare/src/cdn-warm.ts packages/cloudflare/src/deploy.ts tests/cloudflare-cdn-warm.test.ts tests/cloudflare-cdn-warm-deploy.test.ts`

Stacked on #3020.

PR #3024, opened 2026-08-22
fix(cloudflare): keep CDN warmup correct with custom adapters and i18n

james-elicx:
## Summary

Keeps deploy-time CDN warming correct when applications use custom cache adapters or hybrid App/Pages Router i18n routes.

- expose build-identity response headers as a generic CDN adapter capability; the built-in Cloudflare adapter now guarantees `X-Vinext-Build-Id` on every page response
- verify staged HTML only when the adapter declares that capability; strict warmup fails before upload when exact-build HTML warming cannot be proven, while non-strict deploys defer that HTML work until after promotion and label it best-effort
- validate response build identity before accepting cache admission or a non-cacheable skip, so an old deployment's `BYPASS` response cannot suppress warming for a route that became cacheable
- treat Cloudflare cache admission statuses as authoritative while rejecting unsafe plain `Set-Cookie` `MISS` responses and accepting explicit field-qualified cookie policy
- preserve hybrid route ownership under Pages Router i18n by normalizing locale prefixes for page matching without reclassifying locale-prefixed App routes as Pages API routes

This is stacked on #3023.

## Validation

- 88 focused tests across CDN response validation, staged deploy orchestration, adapter capabilities, prerender path discovery, deploy config wiring, and hybrid i18n handoff
- targeted formatting, lint, and type checks for all 11 changed files
- `vp run vinext#build`

Cloudflare admission behavior was checked against the current [cache response](https://developers.cloudflare.com/cache/concepts/cache-responses/) and [CDN-Cache-Control](https://developers.cloudflare.com/cache/concepts/cdn-cache-control/) documentation. The hybrid locale/API behavior follows Next.js' `test/e2e/i18n-api-support/index.test.ts` coverage.

### DIFF
--- a/packages/cloudflare/src/cache/cdn-adapter.ts
+++ b/packages/cloudflare/src/cache/cdn-adapter.ts
@@ -23,6 +23,7 @@ export function cdnAdapter(options?: Record<string, never>) {
     adapter: fileURLToPath(import.meta.resolve("./cdn-adapter.runtime.js")),
     options,
     capabilities: {
+      buildIdentity: "response-header" as const,
       responseVary: "verbatim" as const,
     },
   };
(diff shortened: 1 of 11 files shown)


james-elicx:

PR #3032, opened 2026-08-22
fix(rsc): align canonical warm ownership and request variants

james-elicx:
## Summary

Consolidates the client-identity, route-ownership, and deployed-proof work previously split across #3027–#3032.

- align locale-prefixed hybrid App/Pages ownership between browser navigation, runtime routing, and warm discovery
- verify warmed HTML and RSC responses against the exact immutable build and supported variant
- conservatively keep rewrite-affected requests contextual instead of collapsing them into canonical cache entries
- prove Link, `router.prefetch()`, `router.push()`, explicit full prefetch, and soft navigation reuse on a real deployed Worker
- isolate every deployed proof attempt behind a fresh Worker/cache namespace and bind observed hits to the expected build
- keep parallel/interception-only loading shells contextual and hashed while canonicalizing ordinary main-tree `loading.tsx` shells

## Scope

Only the request identity shared by browser RSC navigation and deploy-time CDN warming, plus its deployed proof. Special interception, mounted-slot, route-tree, segment, and other contextual variants remain deferred.

## Validation

The retained commits keep their focused client, ownership, discovery, cache-validation, workflow, and deployed Playwright coverage. The cumulative exact head has passed the complete CI suite and deployed browser-reuse proof.

Consolidates #3027, #3028, #3029, #3030, #3031, and #3032. Stacked on #3026.

PR #3033, opened 2026-08-22
fix(build): keep CDN warm discovery route-correct

james-elicx:
Stacked on #3032.

## Summary

- discover normal static child routes whose params are generated by a parent layout
- exclude configured-rewrite public paths from both HTML and canonical RSC warming
- preserve that exclusion when optional locally-prerendered fallback shells are included
- evaluate rewrite sources against the actual trailing-slash warm URL and every configured domain default locale

## Scope

Only changes the deploy warm plan. It does not execute rewrites during discovery, locally render pages, alter runtime routing, or expand the deferred special RSC variants.

## Validation

- 149 focused discovery, prerender, and Cloudflare warm-plan tests
- targeted `vp check` on all 6 touched files
- `vp run vinext#build`


james-elicx:

PR #3034, opened 2026-08-22
fix(cache): preserve valid CDN policy during RSC warmup

james-elicx:
Stacked on #3033.

## Summary

- treat field-qualified `private` / `no-cache` directives as field policy rather than whole-response cache opt-outs
- preserve cacheable adapter policy for responses that strip named fields such as `Set-Cookie`
- stop staged propagation retries once either immutable build identity proves the current Worker served a deterministic invalid response

## Scope

Only cache-policy interpretation and warmup retry classification used by the RSC CDN prewarming feature. No general cache key or runtime routing changes.

## Validation

- 106 focused cache policy, RSC finalization, Cloudflare adapter, and warmup tests
- targeted `vp check` on all 5 touched files
- `vp run vinext#build`


james-elicx:

PR #3038, opened 2026-08-22
fix(cloudflare): stabilize staged Worker before CDN warmup

james-elicx:
## Summary

- wait for version-override routing to remain on the uploaded build before sending any real CDN warm key
- use unique no-cache readiness queries so a probe can never satisfy a later probe from cache
- keep each canonical full/loading/HTML warm request single-shot after readiness, with existing failed-key retry behavior unchanged
- fail strict deployment before promotion when readiness cannot be established; non-strict deployment promotes and warms the untouched plan afterward

## Root cause

The deployed proof failed twice while staging was still visibly routing a mix of the seed and uploaded Workers. In the second attempt, the loading-shell key accepted immediately during that mixed phase later missed, while the full key that reached the uploaded build only after six attempts later hit. URL, query, Vary fields, build ID, version UUID, host, and TTL matched. The existing 15-second post-fill delay occurred too late to make an early transitional fill reusable.

## Scope

This changes only Cloudflare CDN warmup orchestration and its focused tests. It does not change client RSC identity, ISR eligibility/discovery, middleware, dev mode, TPR, cache adapters, or general request handling.

## Validation

- 77 focused Cloudflare warm/deploy/version tests passed
- vp check passed across 3,104 formatted files and 1,315 lint/typechecked files
- vinext build passed
- @vinext/cloudflare build passed

## References

- https://developers.cloudflare.com/workers/cache/cache-keys/
- https://developers.cloudflare.com/workers/cache/
- https://developers.cloudflare.com/workers/versions-and-deployments/version-overrides/

### DIFF
--- a/packages/cloudflare/src/cdn-warm.ts
+++ b/packages/cloudflare/src/cdn-warm.ts
@@ -1,5 +1,7 @@
 import path from "node:path";
 import fs from "node:fs";
+import { randomUUID } from "node:crypto";
+import { setTimeout as delay } from "node:timers/promises";
 import {
   PRERENDER_PATHS_MANIFEST,
   type PrerenderPathManifest,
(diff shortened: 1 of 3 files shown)


james-elicx:

PR #3039, opened 2026-08-22
fix(cloudflare): preserve strict CDN warmup invariants

james-elicx:
## Summary

- abort strict warmup before promotion when the current traffic split cannot stage the upload at 0%
- fail no-promote mode when staged readiness fails instead of returning an unwarmed 0% version as success
- apply the configured warmup retry budget to readiness failures while retaining the fixed six-probe stability window

## Scope

This is a focused follow-up from the cumulative RSC prewarming review. It changes only staged deploy failure semantics and focused tests.

## Validation

- 78 focused Cloudflare warm/deploy/version tests passed
- vp check passed
- @vinext/cloudflare build passed


james-elicx:

PR #3041, opened 2026-08-22
test(cloudflare): prove deployed Pages CDN prewarming

james-elicx:
## Summary

- add one non-conflicting Pages ISR route to the existing hybrid workers-cache fixture
- require the staged deploy warmup to discover and fill that HTML key
- assert the first post-promotion Pages request is an untouched CDN HIT from the exact uploaded build before running the existing RSC browser proof

## Scope

This reuses the existing Worker, seed deployment, strict warm command, production hostname, promotion readiness check, and Playwright job. It does not add a second deploy flow or change production behavior.

## Validation

- workers-cache hybrid App/Pages production build passed locally
- 20 focused discovery tests passed
- vp check passed
- deployed strict warm/HIT proof runs in this PR CI

### DIFF
--- a/examples/workers-cache/pages/pages-prewarm.tsx
+++ b/examples/workers-cache/pages/pages-prewarm.tsx
@@ -1,0 +1,7 @@
+export async function getStaticProps() {
+  return { props: {}, revalidate: 300 };
+}
+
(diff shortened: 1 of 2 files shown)


james-elicx:

PR #3042, opened 2026-08-22
fix(cloudflare): reject incomplete no-promote CDN warmup

james-elicx:
## Summary

- reject `--warm-cdn-no-promote` when any staged warm request remains unsuccessful
- keep the uploaded version staged at 0% and report the existing cleanup guidance instead of returning false success
- cover readiness-success plus real-request-failure without promoting

## Scope

This is a focused follow-up from the cumulative RSC/ISR CDN prewarming review. It changes only no-promote completion semantics and its focused deploy test.

## Validation

- 26 Cloudflare warm-deploy tests passed

### DIFF
--- a/packages/cloudflare/src/deploy.ts
+++ b/packages/cloudflare/src/deploy.ts
@@ -727,11 +727,23 @@ export async function deployWithCdnWarmup(
     );
   }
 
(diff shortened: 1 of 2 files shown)


james-elicx:

PR #3043, opened 2026-08-22
fix(cloudflare): handle non-cacheable warm paths safely

james-elicx:
## Summary

- skip same-build 307/404 responses only when Cloudflare and response policy explicitly prove they are non-cacheable
- continue failing non-success responses from an unexpected build or without cache opt-out evidence
- validate the complete Pages `getStaticPaths` result instead of silently emitting a partial warm manifest

## Next.js reference

Pages result validation follows `packages/next/src/build/static-paths/pages.ts`: the result must be an object containing only an array `paths` and a boolean or `blocking` `fallback`.

## Scope

This is a focused cumulative-review follow-up for RSC/ISR CDN warm discovery and validation. It does not alter route rendering, middleware, dev mode, or general cache behavior.

## Validation

- 60 focused warm/discovery tests passed

### DIFF
--- a/packages/cloudflare/src/cdn-warm.ts
+++ b/packages/cloudflare/src/cdn-warm.ts
@@ -516,15 +516,6 @@ function validateRscWarmResponse(
   expectedBuildId?: string,
   expectedRscBuildId?: string,
 ): WarmValidation {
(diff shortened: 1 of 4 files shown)


james-elicx:

PR #3044, opened 2026-08-22
test(cloudflare): prove prewarmed App HTML cache reuse

james-elicx:
## Summary

- require the first post-promotion App Router HTML request to be a CDN `HIT`
- bind the response to the exact uploaded build and expected page body
- keep the assertion ahead of all RSC browser traffic so the test cannot fill the key itself

## Scope

This is the final narrow deployed-proof follow-up from the cumulative RSC/ISR CDN prewarming review. It changes no production behavior.

## Validation

- the deployed Worker proof runs in this PR CI


james-elicx:

PR #3045, opened 2026-08-22
fix(cloudflare): enforce no-promote CDN warmup contract

james-elicx:
## Summary

- reject `--warm-cdn-no-promote` when discovery finds no requests instead of silently running an ordinary deployment
- reject no-promote HTML warming before upload when the adapter cannot verify response build identity
- preserve normal best-effort promotion behavior when no-promote is not requested

## Scope

This is a narrow exact-head review follow-up for the CDN prewarming promotion-control flag. It does not change cache admission, routing, middleware, or general deployment behavior.

## Validation

- 39 focused deploy tests passed


james-elicx:

PR #3047, opened 2026-08-22
fix(build): validate generated CDN warm path params

james-elicx:
## Summary

- reject malformed `getStaticPaths` entries before they can invent or collapse CDN warm keys
- validate App `generateStaticParams` string-versus-array parameter shapes against the route pattern
- surface invalid generator return shapes during CDN path discovery instead of silently treating them as an empty result

## Next.js reference

The validation mirrors `packages/next/src/build/static-paths/pages.ts`: entry keys are limited to `params`/`locale`, ordinary params are strings, catch-all params are string arrays, and query-bearing string entries cannot be silently normalized into another path.

## Scope

This validation is used by build-discovered CDN warm paths. The existing non-discovery prerender fallback for malformed App generator values is preserved.

## Validation

- 54 focused discovery/endpoint tests passed

### DIFF
--- a/packages/vinext/src/build/prerender-paths.ts
+++ b/packages/vinext/src/build/prerender-paths.ts
@@ -137,6 +137,78 @@ function validatePagesStaticPathsResult(
   };
 }
 
+type DynamicPatternParam = { name: string; optional: boolean; repeat: boolean };
+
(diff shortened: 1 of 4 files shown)


james-elicx:

PR #3048, opened 2026-08-22
fix(build): exclude route handlers from CDN warm paths

james-elicx:
## Summary

- apply runtime App/Pages ownership resolution to HTML warm paths as well as RSC paths
- exclude concrete App candidates shadowed by App route handlers or Pages API routes
- retain independently discovered Pages page keys and App page keys in original discovery order

## Scope

This changes only which build-discovered page URLs are sent to CDN warmup. It does not change runtime route priority or handler behavior.

## Validation

- 45 focused discovery/deploy-wiring tests passed


james-elicx:

PR #3049, opened 2026-08-22
fix(build): resolve all CDN warm path owners

james-elicx:
## Summary

- resolve runtime ownership across the complete App+Pages discovered path union
- exclude a Pages-discovered key when its runtime winner is an App route handler
- add full/loading RSC variants when a Pages-discovered key resolves to an App page
- retain HTML warming when the runtime winner is a Pages page

## Scope

This only filters and classifies build-discovered CDN warm keys. Runtime routing and handler behavior are unchanged.

## Validation

- 34 focused discovery tests passed


james-elicx:

PR #3050, opened 2026-08-22
fix(build): preserve exact CDN warm path identity

james-elicx:
## Summary

- reject relative, double-slash, query-bearing, and fragment-bearing `getStaticPaths` strings instead of normalizing them into a different warm key
- preserve the exact spelling of explicit locale prefixes because Workers cache keys are case-sensitive
- require App `generateStaticParams` entries to be plain parameter objects before combining them with parent params

## Next.js reference

This follows the string route-regex and plain-object checks in Next.js `packages/next/src/build/static-paths/pages.ts` and `app.ts`.

## Scope

This only validates and preserves build-discovered CDN warm request paths. Runtime routing and ordinary application requests are unchanged.

## Validation

- 57 focused discovery/endpoint tests passed


james-elicx:

PR #3051, opened 2026-08-22
fix(build): preserve encoded Pages CDN warm keys

james-elicx:
## Summary

- preserve the validated original spelling of string entries returned by Pages Router getStaticPaths
- continue matching those entries against the route and locale before adding them to the warm plan
- keep object-form params on the existing canonical encoding path
- prevent percent-encoding case and unreserved-character normalization from warming a different Workers cache key

## Tests

- vp test run tests/prerender-paths.test.ts tests/cloudflare-cdn-warm.test.ts
- vp check packages/vinext/src/build/prerender-paths.ts tests/prerender-paths.test.ts

Stacked on #3050.


james-elicx:

PR #3052, opened 2026-08-22
fix(build): exclude unsafe Pages CDN warm paths

james-elicx:
## Summary

- exclude concrete Pages-only warm paths when a real Pages API route owns the request
- reject malformed percent-encoding from string getStaticPaths entries during discovery
- prevent deploy warming from invoking API handlers or requesting paths the runtime rejects

## Tests

- vp test run tests/prerender-paths.test.ts
- vp check packages/vinext/src/build/prerender-paths.ts tests/prerender-paths.test.ts

Stacked on #3051.


james-elicx:

PR #3063, opened 2026-08-23
perf(server): avoid base64 fallback for split UTF-8 Flight chunks

sunflower0305:
## Summary

RSC streams can split a valid UTF-8 code point across upstream chunks. `createRscEmbedTransform` currently decodes each chunk independently with a fatal decoder, so split CJK or emoji bytes are classified as binary and embedded as Base64.

This change keeps one streaming fatal decoder for the embed stream and tracks the raw bytes that the decoder has not released yet. Valid split text stays in text chunks, while invalid binary data and incomplete UTF-8 at EOF still use the existing `[3, base64]` representation without losing bytes. The decoder is reset after binary fallback so later text can be decoded normally.

This follows the streaming decode used by the [Next.js Web Flight transport](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/app-render/use-flight-response.tsx), with explicit raw-byte retention for lossless error and EOF fallback.

## Coverage

- CJK and emoji split at every UTF-8 byte boundary
- invalid binary followed by valid text
- incomplete UTF-8 followed by binary data
- incomplete UTF-8 at EOF
- existing BOM preservation

## Application result

Measured on a blog page with 108 RSC chunks:

- Base64 chunks: 54 → 0
- raw RSC script size: -36,108 bytes (-9.1%)
- separately gzipped size: about 36.9 KB smaller

## Verification

- `vp test run tests/app-ssr-stream.test.ts -t "createRscEmbedTransform raw buffer"`
- `vp check`
- `vp run vinext#build`

### DIFF
--- a/packages/vinext/src/server/app-ssr-stream.ts
+++ b/packages/vinext/src/server/app-ssr-stream.ts
@@ -119,6 +119,39 @@ function createNextFlightCleanupScript(): string {
   return 'self.addEventListener("DOMContentLoaded",()=>{if(self.__next_f?.push===Array.prototype.push)self.__next_f.length=0},{once:true})';
 }
 
+const EMPTY_BYTES = new Uint8Array();
+
+function getIncompleteUtf8Tail(previousTail: Uint8Array, chunk: Uint8Array): Uint8Array {
+  const previousTailLength = previousTail.byteLength;
+  const totalLength = previousTailLength + chunk.byteLength;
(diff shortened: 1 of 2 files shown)


sunflower0305:
For reference, the measurements in the PR description were collected from this [blog page](https://blog.zhangleyang.com/2026-08-14-tprl-4), which produced 108 RSC chunks. The source is available in the [blog repository](https://github.com/sunflower0305/blog).

PR #3070, opened 2026-08-24
fix(metadata): encode metadata social image URLs

### LINKED ISSUE #3066: Metadata: og:image / twitter:image URLs are not percent-encoded
### What happened

vinext emits `og:image` / `twitter:image` URLs without percent-encoding, where Next.js encodes them. A URL containing a space is passed through verbatim.

```
Next.js 16.3.2
https://file.devfive.kr/file/2025_06_10/1749539238715_1732690349249_1728972833495_image%2016.webp

vinext
https://file.devfive.kr/file/2025_06_10/1749539238715_1732690349249_1728972833495_image 16.webp
```

Note `image%2016.webp` vs `image 16.webp`.

### Reproduction

Return a metadata image whose URL contains a character that must be percent-encoded:

```ts
export async function generateMetadata(): Promise<Metadata> {
  return {
    openGraph: {
      images: ['https://example.com/path/my image.webp'],
    },
  }
}
```

- Next.js renders `content="https://example.com/path/my%20image.webp"`.
- vinext renders `content="https://example.com/path/my image.webp"`.

In our case the URL comes from a CMS field, so we do not control whether it contains spaces.

### Impact

A raw space inside a `content` URL is not a valid URI. Crawlers and preview generators (Open Graph consumers, chat clients) may fail to fetch the image, so link previews can silently break. Affected 2 of 25 captured URLs in our app — only those whose CMS image name contains a space.

### Expected

Match Next.js and normalize metadata image URLs (e.g. via `new URL(...).href` or equivalent encoding) before writing them into `og:image` / `twitter:image`.

### Environment

- vinext 1.0.0-beta.8, Vite 8.2.2
- Next.js 16.3.2 API surface, App Router, `generateMetadata`
- `--platform=node`, `output: 'standalone'`

Found by a golden-master diff of SSR HTML between the two runtimes on the same source and same API fixtures.

NriotHrreion:
Closes #3066

## Overview

Next.js percent-encodes og and twitter image URLs during metadata resolution, while vinext could emit absolute string URLs unchanged when they contained spaces or other characters requiring encoding. This produced invalid metadata URLs and could cause social crawlers or link-preview generators to fail when fetching CMS-provided images.

This PR restores Next.js-compatible URL serialization for social image metadata.

## What changed

In `packages/vinext/src/shims/metadata.tsx`, `resolveSocialImageUrl()` uses `resolveMetadataUrl()` to resolve social image URLs.

https://github.com/cloudflare/vinext/blob/20fdac4cec59fd0f8dcdf490693e5205ccc33dff/packages/vinext/src/shims/metadata.tsx#L835-L842

However, `resolveMetadataUrl()` will return the original URL if there is no `metadataBase` provided.

https://github.com/cloudflare/vinext/blob/20fdac4cec59fd0f8dcdf490693e5205ccc33dff/packages/vinext/src/shims/metadata.tsx#L751-L759

So just pass the social image URL into `new URL()`, and the URL will be encoded automatically. If `new URL()` fails, this means it may be a relative URL, then just pass it to the `resolveMetadataUrl()`.

```ts
try {
  return new URL(imageUrl).href;
} catch {
  // Relative social image URLs are composed with metadataBase below.
}
// ...
```

## Testing

- `pnpm test tests/features.test.ts`

### DIFF
--- a/packages/vinext/src/shims/metadata.tsx
+++ b/packages/vinext/src/shims/metadata.tsx
@@ -831,12 +831,21 @@ function resolveSocialImageUrl(
   metadataBase: URL | null | undefined,
 ): string {
   const imageUrl = isSocialImageDescriptor(image) ? image.url : image;
+
+  // Next.js resolves social image URLs through resolveUrl(), which parses
+  // absolute strings before considering metadataBase. Returning the serialized
+  // URL also percent-encodes characters such as spaces in image paths.
+  if (imageUrl instanceof URL) {
+    return imageUrl.href;
+  }
+  try {
+    return new URL(imageUrl).href;
+  } catch {
+    // Relative social image URLs are composed with metadataBase below.
+  }
+
   const metadataRoute = isSocialImageDescriptor(image) && isMetadataRouteSocialImage(image);
-  if (
-    typeof imageUrl === "string" &&
-    !isAbsoluteOrProtocolRelativeUrl(imageUrl) &&
-    (!metadataBase || metadataRoute)
-  ) {
+  if (!isAbsoluteOrProtocolRelativeUrl(imageUrl) && (!metadataBase || metadataRoute)) {
     return resolveMetadataUrl(imageUrl, getSocialImageMetadataBaseFallback(metadataBase));
   }
   return resolveMetadataUrl(imageUrl, metadataBase);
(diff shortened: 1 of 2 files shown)


james-elicx:
/bigbonk review for issues — hard time limit: 10 minutes; review exact head 4febde5f596c711fff49db15d2cfeb1829856a44. Be verdict-first and inspect only social-image URL resolution/serialization against Next.js 16.2.7, including absolute, relative, metadata-route, cross-origin, and protocol-relative inputs. Post explicit LGTM/no findings or concise actionable in-scope findings before the deadline.

PR #3092, opened 2026-08-26
fix(cache): gate CDN admission on probed routes

james-elicx:
Capability stack **4/9**. Exact head: `f50e5a1d7526ec067df7306085c5f05d253e457d`. Base: #3091.

## Summary

- emit an inert cacheability-manifest ESM module in App Router Worker builds
- admit exact manifest identities only after the ordinary response completes cleanly
- preserve a safe no-manifest fallback by buffering one bounded cold response to EOF before applying its proven static policy
- apply the same completed-response boundary to potentially cacheable App Route Handlers, matching Next.js response draining before static completion
- convert late Route Handler stream errors into uncached 500 responses and observe dynamic request reads made during stream consumption
- keep the Cloudflare adapter's provisional responses private until the outer admission boundary completes
- preserve independently classified Pages Router responses when hybrid routing hands off outside the App-only admission layer
- preserve every explicit handler-owned public or private cache policy on ordinary origin-managed requests, matching Next.js, while retaining adapter-owned fail-closed policy during CDN probe/admission
- cap capture at 4 MiB per response and 16 MiB across concurrent isolate requests
- replay captured chunks without allocating a second contiguous body

## Fail-closed behavior

- an identity absent from a present manifest is `no-store`
- malformed, wrong-build, dynamic, contextual RSC, slow, oversized, incomplete, or middleware-eligible requests remain private
- middleware `has`/`missing` conditions are evaluated by pathname eligibility, not only by the probe request's current headers
- conditional redirects, rewrites, and response headers cannot become shared responses
- framework-owned provisional `no-store` may be replaced after proof, while application/config/private policy remains an absolute veto
- a final ordinary response is revalidated independently of the earlier probe
- CDN purge is safe because every refill repeats completed-response admission

The manifest is a Worker module asset containing classification only—never response bodies. Origin-managed/KV adapters retain their existing streaming path.

## Next.js reference

Next.js drains cacheable App Route Handler responses before resolving static generation (`packages/next/src/server/route-modules/app-route/module.ts`). The built-workerd regression here covers late stream request reads and late stream failure at that same completion boundary.

## Review guide

1. `cacheability-manifest.ts` validates the embedded artifact.
2. `cacheability-request.ts` owns exact matching, capture limits, and admission.
3. `app-route-handler-execution.ts` completes potentially cacheable Route Handler bodies before public admission.
4. `app-router-entry.ts` gates obvious non-page requests before loading admission code.
5. `cdn-adapter.runtime.ts` fails closed while classification is pending.
6. The built-workerd suites prove App Page and Route Handler completion behavior.

This PR does not alter cache keys, response `Vary`, Worker entrypoints, or request routing. Existing framework and adapter `Vary` replacement semantics are unchanged.

## Review size

Layer-only diff against this PR's base: **47 files, +2,723/-76**.

## Validation

### DIFF
--- a/packages/cloudflare/src/cache/cdn-adapter.runtime.ts
+++ b/packages/cloudflare/src/cache/cdn-adapter.runtime.ts
@@ -267,6 +267,11 @@ export class CloudflareCdnCacheAdapter implements CdnCacheAdapter {
       "CDN-Cache-Control": toEdgeCacheControl(input.cacheControl),
       "Cloudflare-CDN-Cache-Control": null,
       "Cache-Tag": input.tags ? formatCacheTag(input.tags) : null,
+      // Workers Cache is consulted before the Worker runs. Vary the reusable
+      // anonymous response by Cookie so a draft/preview request cannot consume
+      // that HIT before vinext validates its signed bypass cookie. The draft
+      // response itself remains no-store once the Worker handles the miss.
+      Vary: "Cookie",
     };
 
     return headers;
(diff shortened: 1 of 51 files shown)


james-elicx:

PR #3093, opened 2026-08-26
feat(cache): deploy probed manifests in two stages

james-elicx:
Capability stack **6/9**. Exact head: `03d5e5edf7fa831c68e110ec96a00c1af21f12e6`. Base: #3103.

## Summary

Implement the two-upload deployment protocol:

1. upload the built Worker with its inert cacheability module
2. stage that probe version at 0%
3. discover concrete requests against that exact staged version
4. probe exact HTML, full-RSC, and loading-shell identities
5. abort if classification or deployment-state validation fails
6. omit dynamic identities and deterministically bound the static-only manifest
7. verify the generated Wrangler main reaches the manifest module
8. upload the isolated manifest-bearing artifact
9. stage, warm, and promote only static identities

A valid empty manifest is also deployed: zero-discovered and all-dynamic applications upload the final fail-closed Worker, skip warming, and either promote it or leave it staged when `--warm-cdn-no-promote` is set.

The normal flow makes one final cache-fill request per admitted identity. It does not perform certification or a second fill request; #3094 adds that only behind `--warm-cdn-certify`.

## Deployment-state safety

- long-running probe/upload/warm windows compare both Wrangler's deployment ID and the exact version percentages
- deployment state is checked after probe staging, before and after the final upload, after final staging, and before promotion
- if `wrangler versions deploy` throws after Cloudflare created the deployment but failed while synchronizing settings, the CLI immediately re-reads state and reports whether the requested stage/promotion happened, the prior state remains, or the outcome is unknown
- successful `--warm-cdn-no-promote` and post-stage failures apply production triggers exactly once; the help text explicitly states that no-promote leaves those triggers applied while the old version remains at 100%

### Remaining concurrency boundary

Cloudflare's create-deployment API and Wrangler expose no ETag, `If-Match`, expected deployment ID, or other compare-and-swap precondition. Deployment-ID checks close observable long windows, but cannot make the final status-read → deployment-POST pair atomic. Deploy pipelines for the same Worker must therefore be serialized; a separate writer racing inside that final API-call window can still overwrite state, as it can with ordinary `wrangler versions deploy`.

## Artifact model

The manifest is an ESM module included in the second Worker upload. It is **not** an environment, service, KV, or static-assets binding. The original build directory is not mutated; an isolated copy replaces only `__vinext_cacheability_manifest.js`.

Application code is not rebuilt between uploads. Dynamic identities are represented by absence and are neither embedded nor rendered again during final warming.

## Review guide

1. `packages/cloudflare/src/cacheability-probe.ts` sends authenticated version-pinned probes, cancels stale bodies, and emits a deterministic static-only result.
2. `packages/cloudflare/src/cacheability-artifact.ts` validates module reachability and enforces route/byte bounds.
3. `packages/cloudflare/src/deploy.ts` owns probe upload → final upload → static-only warm → promotion, deployment identity checks, and post-command reconciliation.
4. `packages/cloudflare/src/version-deploy.ts` parses Wrangler's deployment ID and traffic state.
5. `tests/cloudflare-cdn-warm-deploy.test.ts` proves mixed filtering, fail-closed orchestration, no-promote behavior, concurrent-state rejection, and all reconciled failure outcomes.

## Review size

Layer-only diff against this PR's base: **28 files, +2,888/-119**.

## Validation

- exact-layer CDN deploy orchestration — **57/57** before the final Route Handler extension
- valid zero-discovered, all-dynamic, and `--warm-cdn-no-promote` empty-manifest flows are covered


james-elicx:

PR #3095, opened 2026-08-26
fix(cache): preserve Cache Components ownership during probing

james-elicx:
Stacked on #3094.

## What this fixes

This follow-up aligns staged cacheability probes with Next.js Cache Components ownership semantics:

- a public `use cache` boundary owns an inner `fetch(..., { cache: "no-store" })`, so that fetch does not demote the enclosing route
- `use cache: private` marks the route dynamic and suspends before private user code or cache-key construction
- the outer prerender owns that suspension, so user `try/catch` and Suspense fallbacks cannot turn it into a static certification
- `use cache: private` inside public `use cache` or `unstable_cache()` fails synchronously, matching Next.js, and cannot persist private output in the outer shared cache
- a genuine application 5xx remains a probe failure rather than being hidden by earlier dynamic usage
- the ordinary App Router request path avoids probe-only ownership allocation

The unsafe legacy-payload upgrade and bounded-fill coordinator have been split into #3099 and #3100. They are not part of this diff.

## Next.js parity

The workerd E2E ports the relevant ownership and invalid-nesting behavior from:

- https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/use-cache/use-cache.test.ts
- https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/cache-components-errors/use-cache-private.util.ts
- https://github.com/vercel/next.js/blob/canary/packages/next/src/server/use-cache/use-cache-wrapper.ts

### DIFF
--- a/tests/e2e/cacheability-components/fixture/next.config.ts
+++ b/tests/e2e/cacheability-components/fixture/next.config.ts
@@ -1,3 +1,6 @@
+// @ts-nocheck
+// This isolated non-workspace fixture receives its dependencies from the
+// Playwright web-server setup immediately before the production build.
 import type { NextConfig } from "vinext";
 
 export default {
--- a/tests/e2e/cacheability-components/fixture/tsconfig.json
+++ b/tests/e2e/cacheability-components/fixture/tsconfig.json
@@ -5,8 +5,7 @@
     "moduleResolution": "bundler",
     "jsx": "react-jsx",
     "strict": true,
-    "skipLibCheck": true,
-    "types": ["@cloudflare/workers-types"]
+    "skipLibCheck": true
   },
-  "include": ["app", "*.ts"]
+  "include": ["app"]
 }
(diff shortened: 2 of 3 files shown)


james-elicx:
/bigbonk review for issues

Hard time limit: 20 minutes. Return all findings available before the deadline; do not wait indefinitely for additional analysis.

james-elicx:
/bigbonk review for issues

Hard time limit: 20 minutes. Review exact head 62917048e5688b12508bb95fcdd81395a102268f and return all findings available before the deadline; do not wait indefinitely for additional analysis.

james-elicx:
/bigbonk review for issues

Hard time limit: 20 minutes. Review exact head 7f4001a573f6b153fafb68d61a4b6023aef4d885 and return all findings available before the deadline; do not wait indefinitely.

james-elicx:
/bigbonk review for issues

Hard time limit: 20 minutes. Review exact head 7f4001a573f6b153fafb68d61a4b6023aef4d885 and return every finding available before the deadline. If the full review cannot finish, post the partial verdict instead of waiting or cancelling without a result.

PR #3098, opened 2026-08-26
feat(cache): probe Pages Router cacheability

james-elicx:
Capability stack **8/9**. Exact head: `b40e1a98c03334c1aaf5c7eaf9930acd47a8d66c`. Base: #3094.

This is the Pages Router capability layer. Without it, App Router probing and two-stage warming work, while Pages routes retain safe route-owned caching but are not classified or warmed by the two-stage flow. #3113 completes the cumulative capability by adding statically eligible Route Handlers.

## What this adds

- classify Pages Router HTML and `/_next/data/<buildId>/*.json` request identities inside the staged Worker
- treat automatic static optimization and `getStaticProps`/ISR HTML as cache candidates
- discover and warm Pages data identities only for real `getStaticProps`/ISR routes; automatic-static pages do not invent data endpoints
- treat `getServerSideProps` and custom page `getInitialProps` as dynamic without executing request-time data functions during probing
- match Next.js custom `_app.getInitialProps` behavior: automatic-static pages become dynamic, while `getStaticProps` pages remain cache candidates
- preserve base paths and the runtime build ID when constructing Pages data paths
- include the locale in Pages data identities even for the default locale, matching Next.js client requests
- apply hybrid App/Pages runtime ownership before generating Pages data targets
- validate Pages data fills as JSON rather than applying RSC response rules
- emit the cacheability-manifest module asset in Pages-only Cloudflare Worker builds
- register the CDN adapter before deciding admission so Pages-only/manual deployments still honor completed-response admission when the embedded manifest is `null`
- carry discovered Pages HTML/data paths through probing, manifest generation, readiness, warming, retry, and admission
- transfer hybrid App/Pages admission ownership explicitly so the App layer cannot overwrite Pages policy
- keep unlisted paths, query variants, middleware/config-header variants, preview-cookie cleanup, unknown data paths, and failed responses private
- fail closed when Pages redirects or rewrites depend on headers, cookies, or hosts in every routing phase
- preserve explicit public `next.config` cache policy for App Pages, including `force-dynamic`, while retaining middleware-over-config precedence

## Warming and certification behavior

The default flow probes each eligible App or Pages identity once and requests it once again to fill the final Worker's CDN cache. It does **not** make a second fill/certification request.

If `--warm-cdn-certify` is explicitly supplied, the optional header-only pass from #3094 re-requests warmed identities before promotion. Certification is disabled by default.

## Next.js parity

The classification follows the Pages data-contract split used by Next.js: automatic static optimization and `getStaticProps` HTML are reusable unless a custom `_app.getInitialProps` opts automatic-static pages out; `getStaticProps` pages remain static. `getStaticProps`/ISR data endpoints receive the same static/ISR cache policy; `getServerSideProps` and page `getInitialProps` remain request-time. Default-locale data URLs and hybrid route ownership match the upstream client/build behavior. The tests link to the upstream prerender, i18n data-route, preview, middleware matcher, and getServerSideProps suites.

### Known safe parity gap

Discoverable exact identities match the staged probe. Runtime-generated `fallback: true`/`blocking` paths that were not discovered remain `no-store` at the CDN, while Next.js can cache them after successful runtime generation. This is fail-closed and prevents disclosure, but is a CDN-performance parity gap for a separately reviewable pattern-contract/runtime-admission follow-up.

## External gateway boundary

A preview/draft cookie request can still consume an already-cached anonymous edge HIT before this single Worker entrypoint runs. This stack deliberately does not add `Vary: Cookie` or multi-entrypoint behavior; the separate uncached-gateway stack owns that edge-before-Worker boundary.

## Rollout requirement

Deploying the fixed Worker cannot remove objects that a vulnerable Worker version already admitted to the shared CDN cache. The security rollout must purge/invalidate existing cache entries when customers upgrade and redeploy. After that purge, this stack prevents unsafe refills; the embedded classification manifest itself survives later content-cache purges.

## Review guide

1. `packages/vinext/src/server/pages-router-entry.ts` exposes authenticated Pages probe handling, registers adapters before admission, and enables capability-based no-manifest completion handling.
2. `packages/vinext/src/server/pages-page-handler.ts` classifies Pages HTML and data contracts.
3. `packages/vinext/src/build/prerender-paths.ts` reports concrete Pages data identities for real `getStaticProps`/ISR routes.
4. `packages/vinext/src/server/app-pages-bridge.ts` transfers hybrid Pages admission ownership.
5. `packages/vinext/src/server/cacheability-request.ts` and `cacheability-manifest.ts` enforce exact HTML/data identity and fail-closed admission.
6. `packages/cloudflare/src/cdn-warm.ts` sends browser-equivalent HTML, RSC, and Pages data requests and validates each representation.
7. The Pages cacheability and Route Handler Playwright specs exercise hybrid and Pages-only built Workers.

## Review size

Layer-only diff against this PR's base: **49 files, +1,754/-165**.

## Validation

- Pages/hybrid built-workerd regression — **5/5**

### DIFF
--- a/packages/cloudflare/src/cacheability-probe.ts
+++ b/packages/cloudflare/src/cacheability-probe.ts
@@ -314,7 +314,7 @@ export async function probeStagedWorkerCacheability(options: {
       }
       if (
         result.version !== 1 ||
-        result.kind !== "app-page" ||
+        (result.kind !== "app-page" && result.kind !== "pages-page") ||
         typeof result.pattern !== "string" ||
         !result.pattern.startsWith("/") ||
         !isProbeRouteState(result.state) ||
@@ -335,7 +335,7 @@ export async function probeStagedWorkerCacheability(options: {
       if (result.state !== "static-candidate") continue;
 
       const route: CacheabilityManifestRoute = {
-        kind: "app-page",
+        kind: result.kind,
         pattern: result.pattern,
         representation: identity.representation,
         requestKey: identity.requestKey,
--- a/packages/cloudflare/src/cdn-warm.ts
+++ b/packages/cloudflare/src/cdn-warm.ts
@@ -91,6 +91,7 @@ export type PrerenderWarmPlan = {
   buildIdentity?: string;
   deploymentId?: string;
   loadingShellPaths: string[];
+  pagesPaths?: string[];
   paths: string[];
   rscBuildId?: string;
   rscPaths: string[];
@@ -241,6 +242,7 @@ export function readPrerenderWarmPlan(
     loadingShellPaths: supportsCanonicalRsc
       ? (manifest.loadingShellPaths ?? []).map(applyConfig)
       : [],
+    ...(manifest.pagesPaths ? { pagesPaths: manifest.pagesPaths.map(applyConfig) } : {}),
     paths: htmlPaths,
     ...(supportsCanonicalRsc ? { rscBuildId: manifest.rscBuildId } : {}),
     rscPaths: supportsCanonicalRsc ? manifest.rscPaths!.map(applyConfig) : [],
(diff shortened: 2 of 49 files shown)


james-elicx:
/bigbonk review for issues

Hard time limit: 10 minutes. Review the exact current head only.

james-elicx:
/bigbonk review for issues

Please return all findings within a hard 10-minute limit.

james-elicx:

PR #3100, opened 2026-08-26
fix(cache): bound unstable cache fill ownership

james-elicx:
Follow-up stack 2/3. Exact head: `5b5cbf23219c039deb7ad0c384d419a3fda437f8`. Base: #3099.

This layer hardens the same-isolate `unstable_cache()` replacement-fill coordinator introduced by the parent:

- assigns publication ownership to the active fill
- prevents an expired or cancelled owner from overwriting a replacement result
- uses a bounded lease so a stalled computation cannot block the key forever
- distinguishes a computing owner from one already publishing to the adapter
- bounds retained pending-fill state by entry count and total key characters
- stores fixed-size key tokens in temporary bypass state

The focused tests cover owner expiry, replacement publication, stalled publication, waiter replay, and state bounds.

Validation before the split:

- targeted cache/runtime tests green
- ppr-impact-demo workerd E2E green


james-elicx:
Closing this layer because the reviewed cross-request cold-miss coordinator diverged from current Next.js behavior. #3099 now retains only the safe envelope migration and request-owned invalid-publication checks; with coalescing removed, this stalled-fill ownership layer has no valid role.

PR #3109, opened 2026-08-26
test(stream): drain transforms before releasing readers

james-elicx:
## Summary

- drain the transform to completion in the single-chunk test helper instead of cancelling while asynchronous `flush()` is still running
- avoid the deterministic `ERR_INVALID_STATE: Unable to enqueue` failure introduced by Node 24.20 stream cancellation timing

This is a test-harness compatibility prerequisite and intentionally contains no production cacheability changes.

## Validation

- failing test reproduced under Node 24.20.0 before this change
- `PATH=/tmp/node-v24.20.0/bin:$PATH vp test run tests/app-ssr-stream.test.ts` (47/47)
- `vp run check`
- `git diff --check`

### DIFF
--- a/tests/app-ssr-stream.test.ts
+++ b/tests/app-ssr-stream.test.ts
@@ -430,8 +430,13 @@ async function readSingleTransformChunk(
 
   await writer.write(new TextEncoder().encode(chunk));
   const result = await reader.read();
-  await writer.close();
-  await reader.cancel();
+  const close = writer.close();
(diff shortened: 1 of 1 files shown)


james-elicx:
Superseded by #3108. Both independent CI prerequisites must be present together for the bottom PR to be green under Node 24.20, so #3108 retains them as separate commits in one prerequisite PR.

PR #3138, opened 2026-09-01
fix(cloudflare): support explicit CDN warm targets

james-elicx:
## Summary

- add `--warm-cdn-target <origin>` to explicitly select the hostname used for staged discovery, cacheability probing, readiness checks, cache fills, and optional certification
- validate that the target is an HTTPS origin without a path, query, credentials, or non-default port
- use the canonical `https://vinext.dev` target for the dashboard-managed web custom domain

## Why

Wrangler's version commands only expose version/deployment metadata and `wrangler triggers deploy` only prints domains declared in the local configuration. A dashboard-managed custom domain is therefore invisible to the existing inference and falls back to `workers.dev`.

CDN cache identity is hostname-specific, so warming the `workers.dev` hostname does not warm the application's custom domain. This was exposed by the production web Worker's intentional redirect from `vinext-web.vinext.workers.dev` to `vinext.dev`.

The explicit target continues to use the Worker version-override header and the existing build/version checks, so a target that does not route to the staged version fails readiness rather than being promoted.

## Next.js parity

This does not change route classification or cache eligibility. It only selects the Cloudflare hostname through which the existing Next.js-compatible classification and warming flow runs.

## Validation

- `vp check`
- `vp test run tests/deploy.test.ts tests/cloudflare-cdn-warm-deploy.test.ts` (409 tests)

### DIFF
--- a/.github/workflows/deploy-examples.yml
+++ b/.github/workflows/deploy-examples.yml
@@ -67,6 +67,7 @@ jobs:
             working_directory: apps/web
             wrangler_config: dist/server/wrangler.json
             warm_cdn_cache: true
+            warm_cdn_target: https://vinext.dev
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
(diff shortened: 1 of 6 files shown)


james-elicx:
/bigbonk review for issues\n\nPlease review exact head 5a2c72a4fe8f636a7d33e1603f648878ce6b2877. Hard time limit: 8 minutes; return your findings within that limit even if the review is incomplete.

PR #3183, opened 2026-09-04
test(link): ensure Enter activates focused links

pascoa-lab:
## Summary
- preserve consumer onKeyDown behavior
- explicitly activate focused Link anchors on unmodified Enter
- reuse the existing click path so onClick, onNavigate, routing, and safety checks remain consistent
- add an App Router keyboard-navigation regression test

## Validation
- targeted vp check passes
- focused Playwright App Router test passes


james-elicx:
Updated this PR to test existing behavior only. The Link implementation change was removed because the new Playwright case already passes on origin/main (2bc9d542b): 1 passed (6.3s). Chromium synthesizes a click when Enter is pressed on a focused anchor, and the test verifies that navigation remains client-side.

james-elicx:
/bigbonk review for issues — hard time limit: 10 minutes; review the cumulative diff at the current exact head and return an explicit LGTM/no-findings verdict or concise actionable in-scope findings.

PR #3186, opened 2026-09-07
fix(app-router): dedupe next/dynamic module preloads

### LINKED ISSUE #3120: next/dynamic preload hints download shared chunks twice (rel=preload as=script vs modulepreload)
`DynamicPreloadChunks` calls `ReactDOM.preload(href, { as: "script", … })` for each JS file of a rendered `next/dynamic()` boundary. Chunks the page entry also loads already carry a `modulepreload` hint from `@vitejs/plugin-rsc`. Classic-script and module are different fetch modes, so the browser reuses neither hint for the other and downloads the file twice.

`utils/lazy-chunks.ts` documents the assumption this rests on — "the browser dedupes preloads by URL". Dedupe is per destination/mode, not per URL.

Measured on one route (vinext 1.0.0-beta.8, Cloudflare Workers): 65 chunks carry both hints, and all 65 are fetched twice — 223 script requests for 158 distinct URLs, 212 KB of extra transfer (679 KB instead of 467 KB).

Fix, in `src/shims/dynamic-preload-chunks.tsx` — emit the same hint the eager chunks use:

```diff
-const preloadOptions = { as: "script", fetchPriority: "low", nonce };
-ReactDOM.preload(assetHref, preloadOptions);
+const preloadOptions = { as: "script", crossOrigin: "", fetchPriority: "low", nonce };
+ReactDOM.preloadModule(assetHref, preloadOptions);
```

Verified on a live deploy: 158 script requests, 0 duplicates, `rel="preload" as="script"` count 65 → 0.

keeltrace:
## Summary

- emit rendered `next/dynamic` JavaScript hints as `modulepreload` so they use the same fetch mode as Vite's eager ES-module hints
- render the module preload resource directly so React preserves the existing CSP nonce and low fetch priority
- update production and Cloudflare Worker coverage, including an assertion that the old low-priority classic-script preload is no longer emitted

## Why

`DynamicPreloadChunks` currently uses `ReactDOM.preload(..., { as: "script" })`, while the same Vite chunks may already be hinted with `modulepreload`. Browsers treat classic-script and module fetches as different modes, so the two hints do not dedupe and the chunk can be downloaded twice.

Issue #3120 measured 65 duplicate chunk fetches and 212 KB of avoidable transfer on one route.

The issue suggested `ReactDOM.preloadModule()`, but the React 19 API used here does not preserve the full existing hint contract (notably the low fetch priority through the public wrapper). Rendering `<link rel="modulepreload">` directly keeps `nonce`, `fetchPriority`, and `crossOrigin` while React still hoists the resource into `<head>` during SSR.

This is an intentional vinext/Vite divergence from Next.js's webpack preload implementation: these emitted client chunks are native ES modules.

## Validation

- `vp check packages/vinext/src/shims/dynamic-preload-chunks.tsx tests/app-router-production-server.test.ts tests/e2e/cloudflare-workers/dynamic-preload.spec.ts` — passed on final commit
- focused production regression on final commit — passed
- red/green proof: the new production assertion fails against the previous source (`0` matching dynamic modulepreloads) and passes with this fix
- `vp run vinext#build` — passed
- targeted production `next/dynamic` matrix — 5 passed directly; the two asset-prefix build cases exceeded the loaded host's existing 60s test timeout and both passed when rerun with a timeout-only 180s harness change
- focused Cloudflare Workers Playwright regression — passed, including nonce preservation and client interaction

Fixes #3120

## Next.js parity reference

Next.js emits classic-script preloads for its webpack chunks in [`preload-chunks.tsx`](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/lazy-dynamic/preload-chunks.tsx). Vinext deliberately uses `modulepreload` instead because Vite emits native ES modules; the request mode must match Vite’s eager hints for browser fetch reuse.


james-elicx:
/bigbonk review for issues — time limit: 10 minutes; please return a concise clean verdict or actionable in-scope findings.

PR #3210, opened 2026-09-10
refactor(cache): reduce response store metadata calls

james-elicx:
## Summary

- collapse response-body writes into one reservation RPC and one publication RPC
- retain pending-object rows as the durable R2 object registry
- move orphan cleanup from every put to the metadata Durable Object alarm
- return the published entry from the publication transaction

## Validation

- `vp check packages/workers-response-store/src/binding.ts packages/workers-response-store/src/metadata-do.ts packages/workers-response-store/tests/e2e.test.mjs`


james-elicx:

PR #3211, opened 2026-09-10
perf(cache): remove eager response store tag index

james-elicx:
## Summary

- stop writing one SQLite row for every cache tag on every publication
- select tagged entries from stored metadata only during explicit refresh and purge mutations
- retain tag invalidation timestamps for soft-tag correctness
- preserve the old tables for safe rollback compatibility without backfilling or writing them

## Validation

- `vp check packages/workers-response-store/src/metadata-do.ts packages/workers-response-store/tests/e2e.test.mjs`


james-elicx:

PR #3212, opened 2026-09-10
perf(cache): cache response store tag expirations

james-elicx:
## Summary

- resolve framework soft-tag state through cacheable internal marker responses
- add a version-wide invalidation watermark that skips tag-set lookups for newer entries
- evict watermark and tag-set markers whenever a tag is purged
- preserve the existing service-binding RPC signature order for rolling deployment compatibility

## Validation

- `pnpm --filter @vinext/cloudflare run test:response-store` (29 tests)


james-elicx:

PR #3233, opened 2026-09-11
fix(server): fall back to a passthrough when console.createTask is inert

### LINKED ISSUE #3234: Dev server crashes on Workers: console.createTask exists but throws; React dev builds call it in every environment
**Context**

Running an App Router app in dev on the Cloudflare runtime (`vinext dev` with the workerd runner).

**Observed**

Every route 500s. Two distinct crashes, both rooted in `console.createTask`:

1. The RSC environment dies at module initialization — React's dev build calls `console.createTask(...)` at module init (scheduler task tracing)
2. The SSR child environment dies again during HTML rendering for the same reason

workerd's console **exposes** `createTask` but **throws** "not implemented" when it is actually called, so React's feature detection happily picks it up and then crashes on first use. A stub imported before the vinext handler fixes the RSC environment, but the SSR environment needs the polyfill too — the fix has to reach every vite environment.

**Expected**

Dev rendering should work on Workers out of the box; the runtime gap should be shimmed once, upstream, for all server environments.

**Workaround (app-side)**

```ts
// imported before vinext/server/fetch-handler
(console as any).createTask ??= (name: string) => ({ name, run: (fn: any) => fn() });
```

Fixed by #3233.

JamesbbBriz:
## Summary

workerd installs a truthy `console.createTask` through `node:console`, but calling it throws `ERR_METHOD_NOT_IMPLEMENTED`. React development builds use a truthiness check and invoke the method during module initialization and rendering, which crashes Worker RSC and SSR environments.

The shared server-globals module now deliberately loads `node:console` before probing the API, removes an unusable implementation so React uses its own no-task fallback, and preserves a working implementation such as the Node task tracer. The probe verifies that `task.run()` synchronously executes and returns its callback result. Generated App Router and Pages Router entries already load this module before user modules.

The cleanup handles throwing accessors, truthy non-functions, incomplete task implementations, configurable and writable descriptors, and repeated installation. Tests restore the original console descriptor even on Node versions where `createTask` starts absent.

## Parity and runtime evidence

- Next.js vendored React and workspace React 19.2.7 both select `console.createTask` by truthiness and otherwise use a no-task fallback.
- A real workerd probe confirms importing `node:console` installs a writable/configurable `createTask` that throws `The Console.createTask method is not implemented`.
- Node 24.15 exposes a working writable/configurable implementation whose task has a synchronous `run` method; it is left untouched.
- No dedicated Next.js test covers the workerd-specific runtime gap.

## Validation

- `vp test run tests/edge-globals.test.ts` — 11 passed
- `vp test run tests/entry-templates.test.ts -t "installs server globals"` — 2 passed
- `vp check packages/vinext/src/server/server-globals.ts tests/edge-globals.test.ts`
- `vp run vinext#build`
- `PLAYWRIGHT_PROJECT=cloudflare-workers vp run test:e2e -- tests/e2e/cloudflare-workers/ssr.spec.ts --workers=1` — 11 passed on real workerd
- Inspected the generated Worker RSC and SSR bundles to confirm both retain the `node:console` import and shared probe

Fixes #3234

### DIFF
--- a/packages/vinext/src/server/server-globals.ts
+++ b/packages/vinext/src/server/server-globals.ts
@@ -60,17 +60,9 @@ function installCreateTaskFallback(): void {
   const existing = (console as ConsoleWithCreateTask).createTask;
   if (typeof existing !== "function") return;
 
-  let usable = true;
   try {
-    const probe = (existing as (name: string) => { run: (fn: () => void) => void })(
-      "vinext:createTask-probe",
-    );
-    probe.run(() => {});
+    (existing as (name: string) => ConsoleTaskLike)("vinext:createTask-probe").run(() => {});
   } catch {
-    usable = false;
-  }
-
-  if (!usable) {
     (console as ConsoleWithCreateTask).createTask = (name: string): ConsoleTaskLike => ({
       name,
       run: (fn) => fn(),
--- a/packages/vinext/src/server/server-globals.ts
+++ b/packages/vinext/src/server/server-globals.ts
@@ -15,6 +15,12 @@ const PHASE_PRODUCTION_BUILD = "phase-production-build";
 
 type BrowserGlobalName = "window" | "document";
 
+type ConsoleTaskLike = { name: string; run: <T>(fn: () => T) => T };
+
+type ConsoleWithCreateTask = typeof console & {
+  createTask?: unknown;
+};
+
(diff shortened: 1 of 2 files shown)


james-elicx:
/bigbonk review for issues

PR #3237, opened 2026-09-11
fix(image): stop forcing object-fit on fill images

### LINKED ISSUE #3238: next/image with fill crops images that should use object-fit: contain — inline cover overrides caller intent
**Context**

`next/image` with the `fill` prop, where the image needs `object-fit: contain` (logos, diagrams, wide banners with text).

**Observed**

vinext renders fill images with an inline `object-fit: cover`, cropping the image from the top and bottom. Passing `objectFit` through the `style` prop does override it, but nothing else does — no stylesheet rule can win against the inline style, so the documented "style your own object-fit" approach only works if you know about the hardcoded value.

**Expected — Next.js parity**

Next.js builds its fill `imgStyle` from the legacy `objectFit` prop (usually `undefined` for modern usage) and assigns the caller's `style` last, so there is **no inline object-fit at all** unless the caller sets one (`next/dist/shared/lib/get-img-props.js`):

```js
const imgStyle = Object.assign(fill ? {
    position: 'absolute', height: '100%', width: '100%',
    left: 0, top: 0, right: 0, bottom: 0,
    objectFit, objectPosition
} : {}, ..., style);
```

vinext should leave the value to the caller the same way.

Fixed by #3237.

JamesbbBriz:
Next.js builds its fill `imgStyle` from the legacy `objectFit` prop (usually undefined) and never sets an inline `object-fit`, leaving the value to the caller's style or stylesheet (`get-img-props.js` assigns `style` last). vinext hardcoded `objectFit: "cover"`, so fill images that need `contain` (logos, diagrams) rendered cropped even though passing `objectFit` through `style` is the documented pattern.

Drops the hardcoded value so fill parity holds: the legacy `objectFit` prop still flows through style, explicit styles still win, and blur placeholder backgrounds keep their own `cover` sizing independent of this path.

Fixes #3238

### DIFF
--- a/packages/vinext/src/shims/image.tsx
+++ b/packages/vinext/src/shims/image.tsx
@@ -222,12 +222,14 @@ function getFillStyle(
   style?: React.CSSProperties,
   backgroundStyle?: React.CSSProperties,
 ): React.CSSProperties {
+  // Next.js does not set an inline object-fit for fill images: the style is
+  // whatever the caller passes (or their CSS), so the legacy objectFit prop
+  // flows through style and anything else stays under stylesheet control.
   return {
     position: "absolute",
     inset: 0,
     width: "100%",
     height: "100%",
-    objectFit: "cover",
     ...backgroundStyle,
     ...style,
   };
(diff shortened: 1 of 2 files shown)


james-elicx:
/bigbonk review for issues

james-elicx:
/bigbonk review for issues

PR #3241, opened 2026-09-11
fix(worker): inline NEXT_DEPLOYMENT_ID into worker bundles

AhmedElBanna80:
## Problem

`process.env.NEXT_DEPLOYMENT_ID` and `process.env.__VINEXT_DEPLOYMENT_ID` are set as top-level Vite `define`s, and the comment there explicitly intends worker code to read `process.env.NEXT_DEPLOYMENT_ID`. But worker builds run in their **own plugin container** that the top-level defines don't reach — as the sibling `createWorkerImageImportsPlugin` already notes ("Worker builds use their own plugin container, so the top-level … plugin cannot perform this transform"). So a worker that reads `process.env.NEXT_DEPLOYMENT_ID` sees an un-replaced member access → `null`/`undefined`.

Surfaced by the Next.js e2e fixture `test/e2e/app-dir/worker` (`NEXT_DEPLOYMENT_ID` null inside the worker).

## Fix

Add a worker-scoped define plugin (`createWorkerDeploymentIdDefinePlugin`) registered in `worker.plugins` alongside the existing worker-image plugin. It inlines both identifiers into worker scripts, mirroring the top-level define values for parity (`NEXT_DEPLOYMENT_ID` → the string or `false`; `__VINEXT_DEPLOYMENT_ID` → the string or `""`). It is a **no-op** where the identifiers are absent, and harmless if the top-level define already reaches workers (the filter simply won't match a value that's already replaced).

`packages/vinext/src/plugins/worker-image-imports.ts` (the plugin, beside its sibling) + registration in `packages/vinext/src/index.ts`.

## Test

`tests/worker-deployment-id-define.test.ts` exercises the transform directly: inlines a configured id, inlines `false` when unset (Next.js parity), handles the internal identifier, and is a no-op when the id is never read. Mutation-proved (neutering the replacement reds the positive cases).

## Scope

This fixes the **`NEXT_DEPLOYMENT_ID`-is-null** half of the fixture. The `?dpl=` skew query on the worker *script URL* is a separate asset-URL concern (`renderBuiltUrl`/`assetPrefix`) and is not addressed here.

### DIFF
--- a/packages/vinext/src/index.ts
+++ b/packages/vinext/src/index.ts
@@ -276,7 +276,10 @@ import {
   createImportMetaUrlPlugin,
   type EmittedModuleFileNameResolver,
 } from "./plugins/import-meta-url.js";
-import { createWorkerImageImportsPlugin } from "./plugins/worker-image-imports.js";
+import {
+  createWorkerDeploymentIdDefinePlugin,
+  createWorkerImageImportsPlugin,
+} from "./plugins/worker-image-imports.js";
 import { createRequireContextPlugin } from "./plugins/require-context.js";
 import {
   createRequireConditionResolutionPlugin,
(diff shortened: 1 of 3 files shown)


AhmedElBanna80:
Rebased on the current `main` (merge, no force-push) — no conflicts. `tests/worker-deployment-id-define.test.ts` passes (4/4) against the merged tree. Ready for review.

AhmedElBanna80:
Status check on `1.0.0-beta.11` / current `main`: this branch is 1 commit behind `main`, GitHub reports it as mergeable with no conflicts, and all CI jobs on the PR are green (unit, integration, E2E, create-next-app, Workers). No rebase needed at the moment; I'll merge `main` in if that changes.

PR #3242, opened 2026-09-11
fix(cloudflare): return redirects from cache-facing fetches

### LINKED ISSUE #3243: Shared-cache renders that redirect return 500 "Too many redirects" since #3168
## Description

Any route whose shared-cache render responds with a redirect (e.g. an auth gate that redirects unauthenticated visitors to a login page) returns `500` with `TypeError: Too many redirects` since #3168 (35af7910).

A stock Payload CMS app is a reliable reproducer: `/admin` redirects unauthenticated visitors to `/admin/login`.

## Repro

1. Payload CMS 3 app on vinext (current `main`), Cloudflare platform, `cdnAdapter` workers cache, D1/KV/R2 bindings, no admin users created yet.
2. `GET /admin` → `500` (plain `Internal Server Error` body).
3. Worker logs show the same render repeated ~20 times, then:
   ```
   TypeError: Too many redirects.<original url>?...__vinext_cache_key=..., /admin/login?redirect=..., /admin/login?redirect=..., ... (20 hops)
   ```

On 80c5f590 (commit before 35af7910) the same request correctly returns `307 -> /admin/login` and the login page renders.

## Analysis (from instrumentation on a deployed repro)

- Public traffic enters the cache-fronted response entrypoint directly.
- On a cache miss the render returns `307 -> /admin/login` (`Cache-Control: no-store`, nothing cached).
- A fetch through the cache-enabled entrypoint **follows redirects by default**; the follow-up request re-enters the entrypoint with the redirect target URL — but the response-stage invocation/props still describe the **original** URL (verified: every hop rendered with the original `requestUrl` while the entrypoint received the login URL).
- The redirect source re-renders and redirects again until the runtime's redirect budget is exhausted.

Two contributing defects: the cache-facing fetch follows redirects, and the invocation props appear stale across those internal redirect hops. The first is enough to restore correct behavior: serve the redirect to the client instead of following it.

## Fix

Proposed in #3242 (`redirect: "manual"` on the cache-facing request). Happy to dig into the stale-props transport as a follow-up if maintainers want that fixed too — happy to be assigned this issue.

JamesbbBriz:
## Problem

Since #3168 (`35af7910`), any shared-cache App or Pages render that responds with a redirect returns `500 TypeError: Too many redirects` instead of surfacing the redirect to the client.

A Payload CMS `/admin` route exposed this: unauthenticated visitors should receive `307 -> /admin/login`, but the request instead exhausted the Workers redirect budget.

Fixes #3243.

## Root cause

`createCacheFacingRequest` intentionally constructs a fresh `Request` so browser cache-bypass state does not leak into the cache-enabled entrypoint. A fresh Workers `Request` defaults to `redirect: "follow"`, unlike the client-originated Worker request whose mode is `manual`.

When a response-stage render returns a redirect, following it re-enters the cache entrypoint for the destination while the configurable-entrypoint invocation still describes the original route. The original route renders and redirects again until workerd exhausts its redirect limit.

Cloudflare also documents that followed redirects forward request headers to the destination, including cross-origin destinations, so the cache-stage transport must treat a render redirect as its terminal response.

## Fix

Set `redirect: "manual"` on the single cache-facing request factory. This returns the 3xx response to the uncached gateway, preserves its status and `Location`, and matches the pre-#3168 gateway behavior.

The same factory serves both App and Pages shared response stages. Bypass/probe stages keep their direct request path.

## Parity

Next.js document requests surface `redirect()` as a 307. The runtime regression ports that invariant from:

https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/rsc-redirect/rsc-redirect.test.ts

## Validation

- Added focused unit coverage that asserts the cache-facing request uses manual redirect handling.
- Added a real Wrangler/workerd regression for an App Router `redirect()` and a Pages Router `getStaticProps` redirect, with client redirect following disabled.
- Baseline without the fix: `500 TypeError: Too many redirects`, with 20 repeated destination hops.
- With the fix: both App and Pages routes return 307 with the expected `Location`.
- `vp test run tests/cloudflare-cdn-worker.test.ts`: 60/60 passed.
- Targeted `vp check`: formatting, lint, and types passed.
- `vp run @vinext/cloudflare#build`: passed.

### DIFF
--- a/packages/cloudflare/src/cache/cdn-adapter.worker.ts
+++ b/packages/cloudflare/src/cache/cdn-adapter.worker.ts
@@ -305,6 +305,11 @@ async function createCacheFacingRequest(
   return new Request(url, {
     ...init,
     method: request.method,
+    // A cache-fronted fetch follows redirects by default; following would
+    // re-enter the entrypoint for the redirect target while the invocation
+    // props still describe the original URL, re-rendering the redirect
+    // source forever. Return redirects to the caller instead.
+    redirect: "manual",
   });
 }
 
--- a/tests/cloudflare-cdn-worker.test.ts
+++ b/tests/cloudflare-cdn-worker.test.ts
@@ -667,6 +667,29 @@ describe("Cloudflare CDN multi-stage Worker facade", () => {
     expect(stages.response).not.toHaveBeenCalled();
   });
 
+  it("returns redirects from shared dispatch instead of following them", async () => {
+    // A cache-fronted fetch follows redirects by default. When the render
+    // responds with a redirect, following it re-enters the entrypoint for the
+    // redirect target while the invocation still describes the original URL,
+    // re-rendering the redirect source until the runtime's redirect budget is
+    // exhausted (observed as `TypeError: Too many redirects` on any route that
+    // redirects unauthenticated visitors, e.g. a CMS admin login gate).
+    const fetch = vi.fn().mockResolvedValue(new Response(null, { status: 307 }));
+    const binding = vi.fn(() => ({ fetch }));
+    stages.request.mockImplementation((_request, _env, _ctx, dispatch) =>
+      dispatch(new Request("https://example.com/admin"), { kind: "app-page" }, { cache: "shared" }),
+    );
+
+    const result = await worker.fetch(new Request("https://example.com/admin"), {}, {
+      exports: { VinextCachedResponse: binding },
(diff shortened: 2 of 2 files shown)


james-elicx:
/bigbonk review for issues

james-elicx:
/bigbonk review for issues

PR #3299, opened 2026-09-15
test(cache): add App Router Response Store load harness

james-elicx:
## Scenario

This adds an isolated vinext App Router load-test app for the production Workers Response Store path. It intentionally uses unique route IDs so each lifecycle starts from a cold key rather than measuring an already-warm Workers Cache hit.

### Cacheable lifecycle

- Requests `/cacheable/:id`, an App Router page forced to render at runtime.
- `next.config.mjs` applies `Cache-Control: public, s-maxage=15, stale-while-revalidate=15` only to this route.
- Verifies `MISS` on the first request, then the same render token on a stored `HIT`.
- Waits for freshness to expire, verifies stale `UPDATING` with the old token, then polls until a recovered `HIT` has a new token.
- This exercises cold metadata/body admission, Workers Cache population, SWR claim/regeneration, publication, and return to a hit.

### Dynamic lifecycle

- Requests `/dynamic/:id` twice at the same URL.
- The page remains `force-dynamic` with no public cache-policy override.
- Verifies both responses are `no-store`, neither reports `HIT`/`UPDATING`, and each render has a different token.
- This exercises repeated Response Store misses that must never be admitted.

## Load shape and output

- Defaults to 5,000 cacheable and 5,000 dynamic lifecycles.
- Uses a closed-loop concurrency ladder of 50, 100, 200, 400, and 800; this is not a fixed-arrival-rate RPS benchmark.
- Stops a scenario after its cumulative lifecycle failure rate exceeds 1%.
- Emits ignored, chart-ready JSON with per-second RPS, p50/p95/p99/max latency, statuses, cache states, errors, lifecycle totals, and admission/SWR transition timings.
- Enables logs, invocation logs, traces, and 1% head sampling on both the app and cache Workers.
- Keeps the manually provisioned Workers, Durable Object namespace, and R2 bucket after runs.

The in-directory README documents prerequisites, one-time bucket creation, build/deploy ordering, smoke and high-load commands, every runner option, JSON fields, and CSV export for charting.

## Observed App Router experiment

- Cacheable lifecycles were clean through concurrency 200.
- At concurrency 400, 107 of 1,333 cacheable lifecycles failed (8.03%).
- A dedicated concurrency-800 cacheable run failed 871 of 1,000 lifecycles (87.1%).
- Dynamic lifecycles reached concurrency 800 with 37 of 1,668 failures at that level (2.22%).
- Failures included `Durable Object is overloaded. Requests queued for too long.`
- The dashboard showed a 249.42 Durable Object invocations/s minute-average peak. This is experiment context, not a precise single-instance capacity limit.

The generated experiment JSON and graph artifacts are not committed.

## Validation

- `node --check examples/response-store-demo/load-test/load.mjs`
- App Router production build completed through all five vinext build stages.
- Live smoke: 1 cacheable and 1 dynamic lifecycle passed, 9 requests total, zero failures.
- `git diff --check`


james-elicx:
Closing because this load experiment was not requested as a pull request. The harness and result work will continue locally.

PR #3300, opened 2026-09-15
feat(cloudflare): support Vite plugin v2

dario-piotrowicz:
Add an experimental initializer flag that installs the plugin v2 preview package and generates cloudflare.config.ts for new projects.

Support the v2 Build Output layout during CDN cache finalization, while preserving compatibility with the legacy Wrangler output.

Document the opt-in flow and add coverage for initialization, configuration generation, dependency installation, and builds.


packages/vinext/src/init.ts
```diff
@@ -225,6 +225,10 @@ export type InitDependencyGroups = {
   devDependencies: string[];
 };
 
+// TODO: replace this version specifier with the proper npm beta one when available
+export const CLOUDFLARE_VITE_PLUGIN_V2_SPECIFIER =
+  "https://pkg.pr.new/@cloudflare/vite-plugin@15508";
```
dario-piotrowicz, line 230: obviously this is temporary and ideally to be addressed before this PR is merged we'll need to use the proper npm specifier as soon as the v2 plugin is published to npm

dario-piotrowicz:
Closing in favour of https://github.com/cloudflare/vinext/pull/3230

PR #3303, opened 2026-09-16
fix(response-store): index orphan cleanup lookups

james-elicx:
## Summary

- add a persistent partial index for active response-body object keys
- avoid relying on SQLite `rowsWritten` for publication after index maintenance
- verify the orphan-cleanup join uses the covering index without an automatic index

## Validation

- `npm run check` in `packages/workers-response-store`
- `npx vitest run tests/e2e.test.ts tests/service-binding-e2e.test.ts` (48 tests)
- `npm run build:examples` in `packages/workers-response-store`


james-elicx:
/bigbonk review for issues

PR #3304, opened 2026-09-16
perf(response-store): coalesce binding metadata misses

james-elicx:
## Summary

- coalesce concurrent metadata lookups for the same version-scoped cache key inside a binding isolate
- retain metadata misses for one second in a bounded 1,024-entry LRU
- invalidate local misses before and after writes, including concurrent cold-write races

The coalescing shares metadata only; response bodies and streams remain request-local.

## Validation

- `npm run check` in `packages/workers-response-store`
- `npm run build:examples` in `packages/workers-response-store`
- `npx vitest run tests/e2e.test.ts tests/service-binding-e2e.test.ts` (51 tests)


james-elicx:
/bigbonk review for issues

PR #3306, opened 2026-09-17
fix(cache): expose response store entrypoints in dev

james-elicx:
## Summary

- run adapter-owned host-entry transforms during local development
- keep adapter-owned multi-stage request routing build-only
- exercise the self-contained Response Store in the existing Cloudflare app under both Vite dev and built Wrangler E2Es

Self-contained Workers Response Store configuration declares Durable Object and Worker entrypoints in Wrangler. Vinext previously applied the adapter host-entry transform only during builds, so the Cloudflare development Worker did not export CacheMetadata and Miniflare attempted to wrap an undefined Durable Object class.

The E2E fixture is opt-in through VINEXT_RESPONSE_STORE_E2E, so the normal app-router-cloudflare deployment remains unchanged.

## Red/green proof

With the E2E Response Store configuration present and the vinext fix removed, Vite dev fails during startup with createDurableObjectWrapper receiving an undefined CacheMetadata export. Restoring the fix starts successfully and serves the shared smoke test with HTTP 200. The same test also passes against the production build served by Wrangler.

## Testing

- `vp test run tests/fetch-handler.test.ts tests/cache-adapters-config.test.ts tests/cdn-adapter-config.test.ts`
- `vp test run tests/build-optimization.test.ts -t "keeps transform-only response-stage outputs self-contained across output arrays"`
- `vp check packages/vinext/src/index.ts tests/fetch-handler.test.ts examples/app-router-cloudflare/vite.config.ts playwright.config.ts tests/e2e/cloudflare-response-store.spec.ts`
- `pnpm install --frozen-lockfile`
- `VINEXT_RESPONSE_STORE_E2E=1 npx vp build` in examples/app-router-cloudflare
- shared Playwright smoke test against Vite dev and Wrangler/Miniflare
- `vp run vinext#build`


james-elicx:
/bigbonk review for issues

james-elicx:
/bigbonk review for issues

PR #3308, opened 2026-09-17
fix(response-store): stabilize cached variant selection

james-elicx:
## Summary

- mark every framework selector as present on opaque Response Store lookup requests
- keep original selector values encoded only in the opaque lookup key
- preserve cached response status, body, and public headers

## Why

The Response Store lookup URL already partitions every framework request selector. Giving each selector a fixed marker keeps cache variant selection consistent without forwarding large request values.

## Testing

- `vp check packages/cloudflare/src/cache/response-store-adapter.worker.ts tests/cloudflare-response-store-worker.test.ts`
- `vp test run tests/cloudflare-response-store-worker.test.ts`
- `pnpm --filter @vinext/cloudflare exec vitest run tests/response-store-adapter.e2e.test.ts` (17 tests)


james-elicx:
/bigbonk review for issues

You have an eight-minute time limit. Finish the review and post all findings within eight minutes.

PR #3315, opened 2026-09-17
fix(cache): round-trip Infinity revalidate in kvDataAdapter

### LINKED ISSUE #3311: kvDataAdapter rejects its own entries for static App Router pages (revalidate: Infinity is stored as null) since 1.0.0-beta.9
## Summary

Since `vinext@1.0.0-beta.9`, a statically generated App Router page is written to the data cache with `cacheControl.revalidate: Infinity`. `kvDataAdapter` stores the entry with `JSON.stringify`, which turns `Infinity` into `null`. On the next read, `validateCacheEntry` requires `cacheControl.revalidate` to be a number, so the adapter rejects its own entry, logs `Invalid cache entry shape`, deletes it, and returns a miss.

The page still renders, but every request to it:

1. reads the KV entry, rejects it and logs `console.error`
2. deletes the entry in the background
3. re-renders and writes the full HTML back to KV

The page is never served from the data cache, and every page view becomes a KV write.

## Versions

| Package | Version |
| --- | --- |
| `vinext` | 1.0.0-beta.10 (first bad: 1.0.0-beta.9) |
| `@vinext/cloudflare` | 1.0.0-beta.8 |
| `@cloudflare/vite-plugin` | 1.54.11 |
| `vite` | 8.3.0 |
| `@vitejs/plugin-rsc` | 0.5.35 |
| `react` / `react-dom` | 19.3.0 |
| `wrangler` | 4.133.0 |

## Reproduction

A minimal app: one dynamic segment with `generateStaticParams`, and the KV data adapter.

<details>
<summary>Files</summary>

`package.json`

```json
{
  "name": "vinext-kv-infinity-repro",
  "private": true,
  "type": "module",
  "scripts": {
    "build": "vinext build",
    "preview": "wrangler dev"
  },
  "dependencies": {
    "react": "^19.2.8",
    "react-dom": "^19.2.8",
    "vinext": "1.0.0-beta.10"
  },
  "devDependencies": {
    "@cloudflare/vite-plugin": "^1.52.1",
    "@vinext/cloudflare": "1.0.0-beta.8",
    "@vitejs/plugin-rsc": "^0.5.34",
    "react-server-dom-webpack": "^19.2.8",
    "vite": "^8.2.1",
    "wrangler": "^4.123.0"
  }
}
```

`vite.config.ts`

```ts
import { defineConfig } from "vite";
import vinext from "vinext";
import { cloudflare } from "@cloudflare/vite-plugin";
import { kvDataAdapter } from "@vinext/cloudflare/cache/kv-data-adapter";

export default defineConfig({
  plugins: [
    vinext({ cache: { data: kvDataAdapter() } }),
    cloudflare({ viteEnvironment: { name: "rsc", childEnvironments: ["ssr"] } }),
  ],
});
```

`wrangler.jsonc`

```jsonc
{
  "name": "vinext-kv-infinity-repro",
  "main": "vinext/server/app-router-entry",
  "compatibility_date": "2026-08-03",
  "compatibility_flags": ["nodejs_compat"],
  "assets": { "binding": "ASSETS", "not_found_handling": "none" },
  "kv_namespaces": [{ "binding": "VINEXT_KV_CACHE", "id": "00000000000000000000000000000000" }]
}
```

`app/layout.tsx`

```tsx
export default function RootLayout({ children }: { children: React.ReactNode }) {
  return (
    <html lang="en">
      <body>{children}</body>
    </html>
  );
}
```

`app/[lang]/page.tsx`

```tsx
export function generateStaticParams() {
  return [{ lang: "en" }, { lang: "fr" }];
}

export default async function Page({ params }: { params: Promise<{ lang: string }> }) {
  const { lang } = await params;
  return <h1>Hello ({lang})</h1>;
}
```

</details>

Steps:

```sh
bun install && bun run build
wrangler dev
# in another shell, request the same page a few times
curl -s -o /dev/null -D - http://localhost:8787/en   # repeat
```

## Observed

Every request is a miss:

```
req 1: Cache-Control: no-store, must-revalidate X-Vinext-Cache: MISS
req 2: Cache-Control: no-store, must-revalidate X-Vinext-Cache: MISS
req 3: Cache-Control: no-store, must-revalidate X-Vinext-Cache: MISS
req 4: Cache-Control: no-store, must-revalidate X-Vinext-Cache: MISS
```

The `wrangler dev` log shows one error per read after the first write:

```
✘ [ERROR] [vinext] Invalid cache entry shape for key: app:<build-id>:/en:html
```

The stored entry, with `value` omitted:

```json
{
  "tags": ["/en", "_N_T_/en", "_N_T_/layout", "_N_T_/[lang]/layout", "_N_T_/[lang]/page"],
  "lastModified": 1789640334484,
  "revalidateAt": null,
  "expireAt": 1821176334484,
  "cacheControl": { "revalidate": null, "expire": 31536000 }
}
```

`set()` builds `cacheControl` only when `typeof effectiveRevalidate === "number"`, so a `revalidate: null` inside it can only come from a non-finite number serialised by `JSON.stringify`. `revalidateAt` is `null` for the same reason (`now + Infinity * 1000`).

## Expected

The entry round-trips, and the second request is served from the data cache without an error. Before beta.9 these pages were not written to the data cache at all, so the mismatch never surfaced.

## Bisect

Same app, `wrangler dev`, the same page requested four times:

| `vinext` | `@vinext/cloudflare` | Page entry written to KV | `Invalid cache entry shape` errors |
| --- | --- | --- | --- |
| 1.0.0-beta.6 | 1.0.0-beta.6 | no | 0 |
| 1.0.0-beta.8 | 1.0.0-beta.6 | no | 0 |
| 1.0.0-beta.9 | 1.0.0-beta.7 | yes | 3 |
| 1.0.0-beta.10 | 1.0.0-beta.6 | yes | 3 |
| 1.0.0-beta.10 | 1.0.0-beta.8 | yes | 3 |

The adapter's `set()` and `validateCacheEntry` logic is the same in `@vinext/cloudflare` beta.6 and beta.8. The change in beta.9 is that core now writes these static pages to the data cache, which exposes the round-trip bug. (All rows were bisected on a larger app with the same route shape. Rows 2 and 5 were also confirmed on the minimal app above.)

## Where

On current `main` (`ebbebfc`):

- `set()` stores `revalidate: effectiveRevalidate` as-is, including `Infinity`:
  https://github.com/cloudflare/vinext/blob/ebbebfc1c70c96e277cf6f351c5cfb52b33e68d0/packages/cloudflare/src/cache/kv-data-adapter.runtime.ts#L482-L491
- `validateCacheEntry` rejects a non-number `cacheControl.revalidate`:
  https://github.com/cloudflare/vinext/blob/ebbebfc1c70c96e277cf6f351c5cfb52b33e68d0/packages/cloudflare/src/cache/kv-data-adapter.runtime.ts#L687

## Possible fix

JSON cannot carry `Infinity`, so the adapter needs to encode it explicitly. For example: write a non-finite `revalidate` as `null` (or a sentinel), accept that value in `validateCacheEntry`, and restore it to `Infinity` in `get()` before returning `cacheControl`. A round-trip test that runs `set()` with `cacheControl: { revalidate: Infinity }` and then `get()` would catch a regression.

## Impact

Besides the log noise, each request to such a page does a KV read, a delete and a put of the full HTML. KV writes are billed, and a key accepts at most one write per second, so a popular page competes with itself.

snhsish:
## Summary

- Round-trip the `Infinity` revalidate value used by runtime-rendered static App Router pages through Cloudflare KV.
- Accept JSON's existing `null` representation at the cache-entry validation boundary and restore it to `Infinity` before callers see the entry.
- Reject `NaN` and `-Infinity` writes so invalid values cannot be confused with the static-page sentinel.

## Why

`JSON.stringify(Infinity)` produces `null`. The KV adapter previously wrote that value for a static page, then rejected the same row on its next read. Each request therefore became a miss followed by a delete, render, and full KV rewrite.

This also repairs beta.9/beta.10 rows already stored with `cacheControl.revalidate: null`.

Fixes #3311.

## Simplification

The original patch added 91 production lines and 57 test lines. The final cumulative diff adds 12 production lines (and removes 1) plus 38 test lines.

The implementation uses native JSON serialization and normalizes once at the existing schema boundary. This removes the custom serialized metadata type, four number codec helpers, duplicate return-path decoding, raw/encoded timestamp pairs, and a redundant legacy-row test. The retained production code is only:

1. a guard that permits positive `Infinity` but rejects other non-finite revalidate values;
2. validator acceptance of the persisted `null` representation; and
3. normalization back to `Infinity` before returning a validated entry.

## Next.js parity and runtime scope

Next.js represents non-revalidating cache policy as `false` or its finite `INFINITE_CACHE` sentinel internally. Vinext's existing runtime App Router path intentionally uses positive `Infinity`, and its response-header logic already treats that value as the static policy. This PR preserves that current vinext contract rather than changing cache semantics.

The Cloudflare KV runtime adapter is shared by the App/Pages dev and production registrations. Deploy-time prerender KV seeding continues to represent static routes as an absent cache-control claim, while TPR only writes finite revalidate values; neither path needs a change.

## Red / green proof

- Base `e97de291c`: the focused regression failed with `Invalid cache entry shape` and returned no cache hit.
- Head: the same set -> JSON KV storage -> parse -> validate -> get round trip passes, preserves the 30-day KV TTL, and does not delete the entry.

## Testing

- `vp test run tests/kv-cache-handler.test.ts tests/isr-cache.test.ts tests/fetch-cache.test.ts` (314 passed)
- `vp test run tests/app-page-cache.test.ts -t "indefinite|non-finite"` (2 passed)
- `vp check packages/cloudflare/src/cache/kv-data-adapter.runtime.ts tests/kv-cache-handler.test.ts`

The focused test exercises the complete adapter serialization boundary directly; there is no separate worker-specific branch between the KV string and this parser.


james-elicx:
/bigbonk review for issues

PR #3347, opened 2026-09-21
fix(config): prioritize explicit aliases over tsconfig paths

hyoban:
## Summary

Explicit `resolve.alias` entries currently lose to Vinext's generated tsconfig aliases. Because those generated entries skip stylesheet resolution, CSS such as `url("~@/icon.svg")` can remain unresolved even when the app explicitly configures `~@`. Dify worked around this with a post-order config plugin: https://github.com/langgenius/dify/pull/42654.

Keep explicit Vite aliases ahead of generated aliases while preserving their order, regular expressions, and custom resolvers. Identify explicit Next config aliases by their source instead of comparing replacement values, so an explicit alias matching the tsconfig target still applies to CSS. Copy the resolve configuration before rearranging aliases so Vite can reuse the original inline aliases on server restart. Inferred tsconfig aliases continue to skip stylesheet resolution, preserving package CSS exports.

Next.js reference: its [CSS loader](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack/loaders/css-loader/src/index.ts) resolves CSS imports and asset URLs through the bundler resolver. This change makes explicit Vite configuration usable without broadening inferred tsconfig aliases to CSS.

## Validation

- Added dev and production-build regressions for object aliases, regular expression aliases, custom resolvers, and explicit Next aliases matching tsconfig targets. Build checks verify that CSS URLs resolve to emitted files; dev checks repeat resolution after `server.restart()` to catch inline-config mutation.
- 199 targeted tests passed; 2 platform-specific tests skipped across alias resolution, Vite 8 config, alias builds, build optimization, and Sass coverage.
- `pnpm run check` passed, including repository-wide format, lint, type, Next.js type-sync, and shim checks.
- `PLAYWRIGHT_PROJECT=cloudflare-workers pnpm run test:e2e`: 87 passed, 3 skipped after the restart fix.
- Full test suite and other browser E2E projects were not run locally.


james-elicx:
/bigbonk review for issues

